PR #33, opened 2016-10-03
Lockfile

### LINKED ISSUE #9: Make sure IndexWriter "locks" the directory
There should be only one IndexWriter working at a time (within our process, or within another process). We should probably add a lockfile system to prevent people from making a mistake.

currymj:
IndexWriter will attempt to create a lockfile (with a specific constant filename) when it is opened. It returns an error if the file exists. When IndexWriter is dropped it deletes the file. This should prevent multiple IndexWriters from simultaneously existing for the same index.

should fix #9

### DIFF
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -82,8 +82,7 @@ impl Drop for IndexWriter {
     fn drop(&mut self) {
         let lockfile_path = Path::new(LOCKFILE_NAME);
         match self.index.directory_mut().delete(lockfile_path) {
-            Ok(_) => (),
-            Err(_) => (),
+            Ok(_) | Err(_) => ()
         }
     }
 }
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -121,10 +121,14 @@ impl IndexWriter {
         Ok(())
     }
 
-    /// Open a new index writer
+    /// Open a new index writer. Attempts to acquire a lockfile.
     ///
-    /// num_threads tells the number of indexing worker that
+    /// num_threads specifies the number of indexing workers that
     /// should work at the same time.
+    /// # Errors
+    /// If the lockfile already exists, returns `Error::FileAlreadyExists`.
+    /// # Panics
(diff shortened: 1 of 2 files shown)


fulmicoton:
You relied solely on the current directory API (as opposed to using OS-specific locking API). I think this is fine. Recover from panics will require actions from the user, but this sounds ok. 

We need to document the lock behavior in `Index.writer.*`.

Considering how important this API is, we might consider having a specific Error for opening a writer.

Either way, people should understand exactly what is happening if they have trouble acquiring the lock. (e.g. If the process panicked or was killed before, and the file is still here). Users should know that it is ok to just remove the lock file if he knows the process which created has terminated. 

He should also know that in that case, the index will rollback in the state of the last commit.
(I let you judge if this should be logged, in the doc, in the error or any combinations of those.)

src/indexer/index_writer.rs
```diff
@@ -82,8 +82,7 @@ impl Drop for IndexWriter {
     fn drop(&mut self) {
         let lockfile_path = Path::new(LOCKFILE_NAME);
         match self.index.directory_mut().delete(lockfile_path) {
-            Ok(_) => (),
-            Err(_) => (),
+            Ok(_) | Err(_) => ()
```
fulmicoton, line 85: oh you use clippy ? cool. I should probably integrate it in Travis.

src/indexer/index_writer.rs
```diff
@@ -309,16 +309,20 @@ mod tests {
...
     fn test_lockfile_stops_duplicates() {
         
 		    let mut schema_builder = schema::SchemaBuilder::default();
 		    let text_field = schema_builder.add_text_field("text", schema::TEXT);
 		    let index = Index::create_in_ram(schema_builder.build());
 			  let index_writer = index.writer(40_000_000).unwrap();
-        let index_writer_two = index.writer(40_000_000).unwrap();
+        match index.writer(40_000_000) {
+            Err(Error::FileAlreadyExists(_)) => {},
+            _ => panic!("Expected FileAlreadyExists error")
```
fulmicoton: Great

PR #34, opened 2016-10-12
Added a macro to define document simply

### LINKED ISSUE #6: Add a macro to define document simply.
This would be handy for unit tests for instance.

Maybe something like :

`doc!(
  field1=>"toto",
  field2=>5,
)`

Value implements `From<String>` and `From<u32>` so that should be feasible.

White-Oak:
Added a test to cover that.
Added `impl From<&str> for Value` for easier doc macro usage.

Closes #6.


fulmicoton:
Thanks @White-Oak !

Issue #46, opened 2016-10-18
LogMergePolicy


currymj:
Implement a merge policy inspired by [LogMergePolicy](https://lucene.apache.org/core/4_0_0/core/org/apache/lucene/index/LogMergePolicy.html) from Lucene.

Segment size will be determined by the number of documents.

Unless I am mistaken, this merge policy should not need to store any state.

fulmicoton:
Merged your branch after internalizing the parameters in the struct, adding setters for them.

Tested it over wikipedia... It worked out fine.

PR #49, opened 2016-10-27
WIP: Issue/46 LogMergePolicy

currymj:
Implements a LogMergePolicy as mentioned in #46.

I would like to add some tests. However, I'm not totally clear on what might need to be tested. Thoughts?

### DIFF
--- a/src/indexer/log_merge_policy.rs
+++ b/src/indexer/log_merge_policy.rs
@@ -87,9 +87,14 @@ mod tests {
     }
 
     #[test]
-    fn test_log_merge_policy_small() {
-        // deal correctly with tiny segments?
+    fn test_log_merge_policy_within_levels() {
+        // multiple levels all get merged correctly
+        let test_input = vec![SegmentMeta::new(SegmentId::generate_random(), 10),
+                              SegmentMeta::new(SegmentId::generate_random(), 11),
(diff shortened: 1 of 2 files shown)


fulmicoton:
I put a bunch of comments.

About unit test, you could make calls with very specific values. (no segment, X segment on the same level, X segment on level 0, X segments on level 2, segments below the lowest level threshold)

fulmicoton:
@currymj Sorry I am new to the code review functionality. I wrote those comments weeks ago and apparently did not finalize the review. Let me know if those make sense.

currymj:
This is all helpful advice! I was just digging through the Lucene code to try to reverse engineer how/why they use the lowest level setting, but your explanation makes a lot of sense. I don't think I will find it too difficult to add this and to just copy the logic from Lucene for adding a level cutoff whenever the floor is reached, etc.

fulmicoton:
Cool. I'll test that at home on indexing wikipedia. It will probably shave off a lot of time over the current merge policy. If I have time, I'll make a visualization of how it behaves.

currymj:
Right now the values for min merge size and segment size are very small hardcoded constants. They should probably be configurable somewhere -- maybe a LogMergePolicy instance should contain these parameters?

Issue #24, opened 2016-09-17
Compilation error: missing simdcomp


Fiedzia:
running: "c++" "-O0" "-ffunction-sections" "-fdata-sections" "-g" "-m64" "-fPIC" "-I" "./cpp/simdcomp/include" "-std=c++11" "-O3" "-mssse3" "-o" "/home/maciej/git/tantivy/target/debug/build/tantivy-c843e95d6308187a/out/cpp/simdcomp_wrapper.o" "-c" "cpp/simdcomp_wrapper.cpp"
cargo:warning=cpp/simdcomp_wrapper.cpp:4:22: fatal error: simdcomp.h: No such file or directory
cargo:warning=compilation terminated.
ExitStatus(ExitStatus(256))

How should I install it? Documentation doesn't mention it.

fulmicoton:
simdcomp is a git submodule

You need to run

```
git submodule init
git submodule update
```

to download it

Fiedzia:
Thanks, works now.

fulmicoton:
Reopening. I really need to write something in the README.

manuel-woelker:
Ran into this today, would you consider a PR to use git subtree instead to make it easier for new contributors?

fulmicoton:
To be honest I didn't know about git subtree and yes it sounds like a good idea.

I am looking forward to your PR.

PR #73, opened 2017-01-26
Use git subtree mechanism for simdcomp to simplify build (cf. #24)

manuel-woelker:
This should make the initial checkout and build more robust

The subtree was created with

```
git remote add simdcomp https://github.com/lemire/simdcomp.git
git fetch simdcomp
git subtree add --prefix cpp/simdcomp simdcomp 0dca28668f1fb6d343dc3c62fa7750a00f1d7201 --squash
```


fulmicoton:
Awesome thank you!

PR #76, opened 2017-02-17
Updated dependency versions and implementations

Ameobea:
- Updated `byteorder` error usage (now returns straight `Error`s)
 - Updated `Uuid` implementation (`to_simple_string` now `.simple().to_string()`)

All of your beautiful, extensive tests pass and the crate compiles as expected!  Should be no issues here at all; almost all of the updated crates had no breaking changes.


src/common/serialize.rs
```diff
@@ -62,12 +53,10 @@ impl BinarySerializable for u32 {
     fn serialize(&self, writer: &mut Write) -> io::Result<usize> {
         writer.write_u32::<Endianness>(*self)
               .map(|_| 4)
-              .map_err(convert_byte_order_error)
```
fulmicoton, line 65: Awesome! So byteorder now uses io::Error.

Ameobea: Indeed!  Always nice to have stuff do more on its own rather than having to do it yourself :+1:

fulmicoton: Negative LOC commit are the bests !

fulmicoton:
kudos @Ameobea !

PR #98, opened 2017-03-08
Convert simd wrapper to C

KodrAus:
@fulmicoton conversation starter about using a C wrapper instead of C++ for the simdcomp stuff. I'm not sure if there was a specific reason we're using C++, but not needing it makes the `musl` target much easier.

~~**EDIT:** You can ignore the `Cross.toml` file, it's an artifact from me testing the build with `musl`. It really belongs to `tantivy-cli`.~~


KodrAus:
@fulmicoton Ok this should be ready for review now.

src/compression/compression_simd.rs
```diff
@@ -1,44 +1,46 @@
...
-        offset: u32) -> size_t;
-        
-    fn compress_unsorted_cpp(
-        data: *const u32,
-        output: *mut u8) -> size_t;
-
-    fn uncompress_unsorted_cpp(
-        compressed_data: *const u8,
-        output: *mut u32) -> size_t;
+mod simdcomp {
```
fulmicoton, line 5: for info, what is the benefit of putting this in an extra module?

KodrAus: There isn't really one, but since I removed the `_cpp` suffix on the wrapper functions putting the C functions in a module was the smallest change so they don't conflict with the other functions in there.

fulmicoton: oh i see. that makes sense. LGTM

PR #102, opened 2017-03-18
Make directory syncing work on Windows

lnicola:


KodrAus:
Build is failing because nightly cargo is busted.

So on Windows we require write access, but on Unix we only need read access. Could this be a problem if there's already an open handle somewhere else? Or is this only called when the file is first created?

lnicola:
I've never used `tantivy`, but it doesn't look like it's keeping open handles for the directory. The function is called whenever files are created or deleted.

About Unix, it's strange. I tried not requesting for read access and it failed with `EINVAL`, and I tried requesting write access and it failed with `EISDIR`:

```
       EISDIR pathname refers to a directory and the access requested involved writing (that is, O_WRONLY or O_RDWR is set).
```

On the other hand, the `fsync()` manpage says:

```
       On some UNIX systems (but not Linux), fd must be a writable file descriptor.
```

I have no idea what Mac OS does, but the old code used `File::open()` which only gives read access, so I figured it's safe to keep that behaviour.

lnicola:
Closing in favour of the branch.

PR #113, opened 2017-03-30
Mark "cpp" folder as linguist-vendored in .gitattributes

kaedroho:
This repo is currently being detected as a C project because of some vendored libraries in the "cpp" folder.

According to https://github.com/github/linguist#using-gitattributes you can use ``.gitattributes`` tell GitHub to not count this folder when detecting the language.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+cpp/* linguist-vendored


lnicola:
Looks good to me.

fulmicoton:
@kaedroho Thanks!

PR #129, opened 2017-05-02
Replace rustc_serialize with serde

### LINKED ISSUE #123: Drop rustc-serialize dependency
It has been deprecated: [announcement](https://users.rust-lang.org/t/deprecation-of-rustc-serialize/10485?u=dtolnay).

Is there any functionality you would require from Serde or another library before this would be possible?

KodrAus:
Closes #123 

This is at a point where previous tests pass. I haven't tested against an actual tantivy index yet though, I'll give that a go and make sure the serialised format hasn't changed.

I also haven't tried too hard to optimise this yet, but I think it's a decent start.

How do you feel about me deriving `PartialEq` on `Schema` and its children so the tests can verify all fields are deserialised correctly?


src/schema/named_field_document.rs
```diff
@@ -11,33 +9,5 @@ use rustc_serialize::Encoder;
 /// A `NamedFieldDocument` is a simple representation of a document
 /// as a `BTreeMap<String, Vec<Value>>`.
 ///
+#[derive(Serialize)]
```
fulmicoton, line 12: cool

PR #134, opened 2017-05-07
Replace rustc_serialize with serde (updated)

### LINKED ISSUE #123: Drop rustc-serialize dependency

lnicola:
This is a the same as https://github.com/tantivy-search/tantivy/pull/129 but rebased over master.

Closes #123


src/schema/schema.rs
```diff
@@ -516,12 +538,13 @@ mod tests {
             let json_err = schema.parse_document(r#"{
                 "title": "my title",
                 "author": "fulmicoton",
-                "count": 50000000000000000000
+                "count": 50,
+                "popularity": 9223372036854775808,
             }"#);
             match json_err {
-                Err(NotJSON(_)) => {
+                Err(NotJSON(_)) | Err(NotJSONObject(_)) => {
```
lnicola: Not sure how I feel about this.

KodrAus: Hmm yeh that's not great, is it. Maybe those errors just need better names, or could be combined into 1

src/schema/field_type.rs
```diff
@@ -63,31 +62,28 @@ impl FieldType {
                     }
                 }
             }
-            Json::U64(ref field_val_u64) => {
+            JsonValue::Number(ref field_val_num) => {
                 match *self {
                     FieldType::I64(_) => {
-                        if *field_val_u64 > (i64::max_value() as u64) {
-                            Err(ValueParsingError::OverflowError(format!("Value {:?} is too high for a i64.", field_val_u64)))
+                        if let Some(field_val_u64) = field_val_num.as_u64() {
```
KodrAus: This section needs to be fixed up now too. It doesn't make sense

KodrAus: This can use [`field_val_num.as_i64()`](https://docs.serde.rs/serde_json/struct.Number.html#method.as_i64) which will do the bounds check and return `None` if it's not an `i64`.

src/schema/field_type.rs
```diff
@@ -63,31 +62,23 @@ impl FieldType {
...
-                    FieldType::I64(_) => {
-                        Ok(Value::I64(* field_val_i64))
-                    }
-                    FieldType::U64(_) => {
-                        Err(ValueParsingError::TypeError(format!("Expected a positive integer, got {:?}", json)))
+                        if let Some(field_val_u64) = field_val_num.as_u64() {
+                            Ok(Value::U64(field_val_u64))
+                        }
+                        else {
+                            Err(ValueParsingError::TypeError(format!("Expected an u64 int, got {:?}", json)))
```
KodrAus: Should these two arms return the same kind of error? Since they'll fail for the same reasons?

src/schema/named_field_document.rs
```diff
@@ -11,9 +9,11 @@ use rustc_serialize::Encoder;
 /// A `NamedFieldDocument` is a simple representation of a document
 /// as a `BTreeMap<String, Vec<Value>>`.
 ///
+#[derive(Serialize)]
 pub struct NamedFieldDocument(pub BTreeMap<String, Vec<Value>>);
 
-
+// TODO: Remove
```
KodrAus: I think we can remove this now :)

src/schema/field_type.rs
lnicola: Sigh. I spotted that, but it slipped out of my mind :(.

KodrAus:
Do you have any feedback on this @fulmicoton?

PR #144, opened 2017-05-11
FIXED inconsistent Term's field serialization.

### LINKED ISSUE #138: Term::text() doesn't panic if the term isn't valid UTF-8
There's a problem with the implementation of the following method:

```rust
    /// Returns the text associated with the term.
    ///
    /// # Panics
    /// If the value is not valid utf-8. This may happen
    /// if the index is corrupted or if you try to 
    /// call this method on a non-string type.
    pub unsafe fn text(&self) -> &str {
        str::from_utf8_unchecked(self.value())
    }
```
(https://github.com/tantivy-search/tantivy/blob/0dad02791c74e572ec005abb1f75e3601bd25db5/src/schema/term.rs#L110-L118)

Unlike the comment suggests, ``str::from_utf8_unchecked`` does not panic if the passed value isn't valid UTF-8 (If it did, the function wouldn't be ``unsafe``). This is pretty bad as if it's invalid, the returned string would be invalid too and could cause a confusing error later on.

The following implementation should make it behave correctly:

```rust
    /// Returns the text associated with the term.
    ///
    /// # Panics
    /// If the value is not valid utf-8. This may happen
    /// if the index is corrupted or if you try to 
    /// call this method on a non-string type.
    pub fn text(&self) -> &str {
        str::from_utf8(self.value()).expect("Term is not a valid UTF-8 string")
    }
```

### LINKED ISSUE #139: allocate_vec returns uninitialised memory
https://github.com/tantivy-search/tantivy/blob/0dad02791c74e572ec005abb1f75e3601bd25db5/src/common/mod.rs#L31-L41

This could lead to a number of problems if not used carefully but most rust developers would not expect this behaviour from a "safe" function.

I really think this should be replaced with something safer, or at least marked as ``unsafe``

### LINKED ISSUE #141: Segfault during commit
As reported by @aabstain .

`
akss:resources akss$ RUST_BACKTRACE=1 cargo run
   Compiling tantivy_sandbox v0.1.0 (file:///Users/akss/programming/rust/tantivitySandbox)
    Finished dev [unoptimized + debuginfo] target(s) in 3.70 secs
     Running `/Users/akss/programming/rust/tantivitySandbox/target/debug/tantivy_sandbox`
10000
thread 'indexing thread 0 for gen 0' panicked at 'index out of bounds: the len is 3 but the index is 16777216', /Users/travis/build/rust-lang/rust/src/libcollections/vec.rs:1490
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::default_hook
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::begin_panic
   5: std::panicking::begin_panic_fmt
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::panicking::panic_bounds_check
   9: <collections::vec::Vec<T> as core::ops::Index<usize>>::index
  10: tantivy::schema::schema::Schema::get_field_entry
  11: tantivy::postings::serializer::PostingsSerializer::load_indexing_options
  12: tantivy::postings::serializer::PostingsSerializer::new_term
  13: <tantivy::postings::postings_writer::SpecializedPostingsWriter<'a, Rec> as tantivy::postings::postings_writer::PostingsWriter>::serialize
  14: tantivy::indexer::segment_writer::write
  15: tantivy::indexer::segment_writer::SegmentWriter::finalize
  16: tantivy::indexer::index_writer::index_documents
  17: tantivy::indexer::index_writer::IndexWriter::add_indexing_worker::{{closure}}
  18: <std::panic::AssertUnwindSafe<F> as core::ops::FnOnce<()>>::call_once
  19: std::panicking::try::do_call
  20: __rust_maybe_catch_panic
  21: std::panicking::try
  22: std::panic::catch_unwind
  23: std::thread::Builder::spawn::{{closure}}
  24: <F as alloc::boxed::FnBox<A>>::call_box
  25: std::sys::imp::thread::Thread::new::thread_start
  26: _pthread_body
  27: _pthread_start`

### LINKED ISSUE #142: Remove misleading code
Get rid of `Term::allocate(Field(0), ...)` in 

https://github.com/tantivy-search/tantivy/blob/master/src/postings/postings_writer.rs#L121

The field argument is not useful

fulmicoton:
Also.

Cleaned up the code to make sure that the logic
is only in one place.
Removed allocate_vec

Closes #141
Closes #139
Closes #142
Closes #138

### DIFF
--- a/src/schema/term.rs
+++ b/src/schema/term.rs
@@ -1,12 +1,13 @@
 use std::fmt;
 
 use common;
-use common::BinarySerializable;
-use common::allocate_vec;
 use byteorder::{BigEndian, ByteOrder};
 use super::Field;
 use std::str;
 
+
+/// Size (in bytes) of the buffer of a int field.
+const INT_TERM_LEN: usize = 4 + 8;
 
 /// Term represents the value that the token can take.
 ///
@@ -15,13 +16,6 @@ use std::str;
 pub struct Term(Vec<u8>);
 
 impl Term {
-    
-    /// Pre-allocate a term buffer. 
-    pub fn allocate(field: Field, num_bytes: usize) -> Term {
-        let mut term = Term(Vec::with_capacity(num_bytes));
-        field.serialize(&mut term.0).expect("Serializing term in a Vec should never fail");
-        term
-    }
 
     /// Set the content of the term.
     pub fn set_content(&mut self, content: &[u8]) {
@@ -40,6 +34,14 @@ impl Term {
         Field(self.field_id())
     }
 
+    /// Returns the field.
+    pub fn set_field(&mut self, field: Field) {
+        if self.0.len() < 4 {
+            self.0.resize(4, 0u8);
+        }
+        BigEndian::write_u32(&mut self.0[0..4], field.0);
+    }
+
     /// Builds a term given a field, and a u64-value
     ///
     /// Assuming the term has a field id of 1, and a u64 value of 3234,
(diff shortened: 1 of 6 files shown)


src/schema/term.rs
```diff
@@ -76,10 +86,22 @@ impl Term {
...
+    }
+
+    /// Creates a new Term with an empty buffer, 
+    /// but with a given capacity.
+    ///
+    /// It is declared unsafe, as the term content
+    /// is not initialized, and a call to `.field()`
+    /// would panic.
+    #[doc(hidden)]
+    pub unsafe fn with_capacity(num_bytes: usize) -> Term {
```
KodrAus: I think we could use `pub(crate)` here instead of `pub` because we're on `nightly`. So the function is internal to `tantivy` and won't need `[doc(hidden)]` because consumers won't be able to call it.

I'm also wondering if there's a better name that describes the `Term` we receive. Maybe `Term::uninitialized`?

KodrAus (approved):
This is great! Just a small, non-critical comment about `Term::with_capacity`

src/schema/term.rs
fulmicoton: Oh this is the first time I hear about `pub(crate)` ! I have been wishing for something like for so long. Awesome.

Wouldn't `Term::uninitialized` suggest that the buffer is unitialized and with a len of  `num_bytes`?

KodrAus: Hmm, that's true. It's probably not worth bikeshedding the name then. At least it's `unsafe` and the docs tell us why.

PR #146, opened 2017-05-11
Issue/136

fulmicoton:
This is quite a bit of a redesign of the indexer, mostly for performance reasons.

- The Hashmap is now shared for all of the fields
- The hash table now uses quadratic probing. Even at low capacity, linear probing was very pathological when there was a lot of distinct terms.
- We do not call `.close_doc()` for all of the terms recorder. When there was a lot of terms, this would create a huge raise in the size of the heap, which would exceed its limit and force reallocation.
- We avoid copying terms during postings serialization


src/datastruct/stacker/hashmap.rs
```diff
@@ -57,34 +53,58 @@ pub enum Entry {
...
-    #[inline]
-    fn bucket(&self, key: &[u8]) -> usize {
-        let hash: u64 = djb2(key);
-        (hash as usize) & self.mask
+    fn probe(&self, key: &[u8]) -> QuadraticProbing {
+        QuadraticProbing::compute(key, self.mask)
+    }
+
+    pub fn is_saturated(&self) -> bool {
+        self.table.len() < self.occupied.len() * 5
```
arthurprs, line 106: You are enforcing a max load factor of 20%. That's way too aggressive.

fulmicoton: You are right. I'll test different ratio.

fulmicoton:
> Is this strategy general? If you have high cardinality data it can potentially generate much more segments than expected.

You mean not resizing the hash table ? Yes it is general.

> If you have high cardinality data it can potentially generate much more segments than expected.

It is not as bad as you think it is.
So imagine we allocate half of the budget to hashmaps, and half to the heap.
By definition, by not knowing what should have been the perfect balance, we wasted **at most** half of the memory allocated.

Now what about the more sophisticated solution. Surely we do not want to rehash the table, so we'll start storing part of the hashes... It's a lot of complexity.

arthurprs:
I guess the key thing then is splitting the memory between the memory and index. If that's done in advance or tracked as the indexing progresses, etc..

> Now what about the more sophisticated solution. Surely we do not want to rehash the table, so we'll start storing the hashes... That suddenly makes every entry twice as big.

Well, it's not as clear cut as that. You are already storing somewhat equivalent metadata by keeping track of buckets with entries. I assume it's to make iteration faster due to the sparse table. There's tradeoffs here and I understand the one you choose.

src/indexer/merger.rs
```diff
@@ -201,6 +201,8 @@ impl IndexMerger {
             }
             merged_doc_id_map.push(segment_local_map);
         }
+
+        let mut field = Field(u32::max_value());
```
KodrAus: What do we use `u32::max_value` for?

src/indexer/segment_writer.rs
```diff
@@ -129,6 +94,16 @@ impl<'a> SegmentWriter<'a> {
 		self.heap.num_free_bytes() <= MARGIN_IN_BYTES
 	}
 	
+	
+	/// Return true if the term dictionary hashmap is reaching capacity.
+	/// It is one of the condition that triggers a `SegmentWriter` to
+	/// be finalized.
+	#[doc(hidden)]
+	pub fn is_termdic_saturated(&self,) -> bool {
```
KodrAus: Should this be public?

src/indexer/segment_writer.rs
```diff
@@ -214,14 +188,16 @@ impl<'a> SegmentWriter<'a> {
...
-			try!(per_field_postings_writer.serialize(serializer.get_postings_serializer(), heap));
-		}
+fn write<'a>(
+		multifield_postings: &MultiFieldPostingsWriter,
+		fast_field_writers: &FastFieldsWriter,
+		fieldnorms_writer: &FastFieldsWriter,
+	  	mut serializer: SegmentSerializer) -> Result<()> {
+		
+		try!(multifield_postings.serialize(serializer.get_postings_serializer()));
+		// for per_field_postings_writer in per_field_postings_writers {
```
KodrAus: Do you still need this commented out code?

src/schema/term.rs
```diff
@@ -15,6 +15,12 @@ const INT_TERM_LEN: usize = 4 + 8;
 #[derive(Clone, PartialEq, PartialOrd, Ord, Eq, Hash)]
 pub struct Term(Vec<u8>);
 
+/// Extract `field` from Term.
+#[doc(hidden)]
```
KodrAus: Do missing doc checks not understand `pub(restricted)`? If this is only public to the crate it shouldn't emit a missing doc warning.

KodrAus (commented):
I don't fully grok the change so will leave that discussion to the experts :) I'm happy when you are.

Just left a few nits.

src/schema/term.rs
fulmicoton: will clean up!

src/indexer/segment_writer.rs
fulmicoton: will clean up!

src/indexer/segment_writer.rs
fulmicoton: will clean up!

src/indexer/merger.rs
fulmicoton: I'll change it for an Option<>

Right now `u32::max_value()` is here to make sure the first field is added.

PR #159, opened 2017-05-17
Clippy fixes

lnicola:
These are just some one-time fixes. I'm not sure whether it's worth enabling it on Travis.

### DIFF
--- a/src/collector/top_collector.rs
+++ b/src/collector/top_collector.rs
@@ -27,7 +27,7 @@ impl Ord for GlobalScoredDoc {
         other
             .score
             .partial_cmp(&self.score)
-            .unwrap_or(other.doc_address.cmp(&self.doc_address))
+            .unwrap_or_else(|| other.doc_address.cmp(&self.doc_address))
     }
 }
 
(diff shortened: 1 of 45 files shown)


src/collector/top_collector.rs
```diff
@@ -27,7 +27,7 @@ impl Ord for GlobalScoredDoc {
         other
             .score
             .partial_cmp(&self.score)
-            .unwrap_or(other.doc_address.cmp(&self.doc_address))
+            .unwrap_or_else(|| other.doc_address.cmp(&self.doc_address))
```
fulmicoton, line 30: cool

src/schema/field_entry.rs
```diff
@@ -63,15 +63,15 @@ impl FieldEntry {
     pub fn is_indexed(&self) -> bool {
         match self.field_type {
             FieldType::Str(ref options) => options.get_indexing_options().is_indexed(),
-            FieldType::U64(ref options) => options.is_indexed(),
+            FieldType::U64(ref options) |
```
fulmicoton, line 66: Oh i wondered if it was even possible to do that.

src/collector/top_collector.rs
lnicola: I was a bit worried about that one, but it actually does what it should and I don't expect the `score`s to be `NaN`.

PR #160, opened 2017-05-17
Use VInt in IntFastFieldWriter

### LINKED ISSUE #131: Reduce the RAM usage of FastFieldWriter
Currently fast field writer are using u32 ints. This is often overkill, especially for the fieldnorms. Alternatively, we could use `VInt` for instance.

(This is rather low priority, but could be an interesting optimization for an analytics use case with a lot of fastfields)

lnicola:
Closes #131

### DIFF
--- a/src/fastfield/writer.rs
+++ b/src/fastfield/writer.rs
@@ -3,8 +3,10 @@ use fastfield::FastFieldSerializer;
 use std::io;
 use schema::Value;
 use DocId;
-use common;
 use schema::FieldType;
+use common;
+use common::VInt;
+use common::BinarySerializable;
 
 /// The fastfieldswriter regroup all of the fast field writers.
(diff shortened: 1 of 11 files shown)


src/fastfield/writer.rs
```diff
@@ -183,13 +203,19 @@ impl IntFastFieldWriter {
...
-        let max = *self.vals.iter().max().unwrap_or(&min);
+        let (min, max) = if self.val_min > self.val_max {
+            (0, 0)
+        } else {
+            (self.val_min, self.val_max)
+        };
+
         serializer.new_u64_fast_field(self.field, min, max)?;
-        for &val in &self.vals {
+        let mut cursor = Cursor::new(&self.vals);
```
fulmicoton: I think this could be &self.vals[..]

PR #169, opened 2017-05-21
issue/168 Make doc! macro usable outside tantivy

fulmicoton:


src/macros.rs
```diff
@@ -0,0 +1,49 @@
+
+macro_rules! get(
+    ($e:expr) => (match $e { Some(e) => e, None => return None })
+);
+
+
+/// `doc!` is a shortcut that helps building `Document`
+/// object, assuming you have the field object.
```
KodrAus: I think we should point out specifically that `$field` is an ident for a `Field` instance and `$value` is an expression that can be converted into a `Value` so it's totally clear to callers what they are.

KodrAus (approved):
Just a simple, optional nit for the documentation. Otherwise LGTM

PR #177, opened 2017-05-26
Use error-chain for root Error type

KodrAus:
Opening this up for visibility, but I've still got plenty of work to do on it before it's ready for review. The plan is:

- Use `error-chain` for the main `tantivy::Error` type
- Ensure all errors implement `std::Error`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ Cargo.lock
 benchmark
 .DS_Store
 cpp/simdcomp/bitpackingbenchmark
+*.bk
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ version = "2"
 crossbeam = "0.2"
 futures = "0.1.9"
 futures-cpupool = "0.1.2"
+error-chain = "0.8.1"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,9 +37,9 @@ uuid = { version = "0.5", features = ["v4", "serde"] }
 chan = "0.1"
 version = "2"
 crossbeam = "0.2"
-futures = "0.1.9"
-futures-cpupool = "0.1.2"
-error-chain = "0.8.1"
+futures = "0.1"
+futures-cpupool = "0.1"
+error-chain = "0.8"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.2"
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -29,8 +29,7 @@ const NUM_SEARCHERS: usize = 12;
 fn load_metas(directory: &Directory) -> Result<IndexMeta> {
     let meta_data = directory.atomic_read(&META_FILEPATH)?;
     let meta_string = String::from_utf8_lossy(&meta_data);
-    serde_json::from_str(&meta_string)
-        .chain_err(|| ErrorKind::CorruptedFile(META_FILEPATH.clone()))
+    serde_json::from_str(&meta_string).chain_err(|| ErrorKind::CorruptedFile(META_FILEPATH.clone()))
 }
 
 /// Tantivy's Search Index
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -1,5 +1,5 @@
 use Result;
-use error::*;
+use error::{ErrorKind, ResultExt};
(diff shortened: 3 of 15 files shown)


fulmicoton:
Thank you

.gitignore
```diff
@@ -5,4 +5,5 @@ target/release
 Cargo.lock
 benchmark
 .DS_Store
-cpp/simdcomp/bitpackingbenchmark
\ No newline at end of file
+cpp/simdcomp/bitpackingbenchmark
+*.bk
```
fulmicoton, line 9: hehe :)

Cargo.toml
```diff
@@ -39,6 +39,7 @@ version = "2"
 crossbeam = "0.2"
 futures = "0.1.9"
 futures-cpupool = "0.1.2"
+error-chain = "0.8.1"
```
fulmicoton: I'm afraid I'm guilty for that bad habit, but I think we should probably avoid putting minor in there. So 0.8 should be fine. I should have done it for futures, and futures-cpupool as well

src/core/index.rs
```diff
@@ -1,5 +1,5 @@
 use Result;
-use Error;
+use error::*;
```
fulmicoton: do we need the * there ? It is only for ErrorKind isn't it?

src/error.rs
```diff
@@ -3,105 +3,129 @@
...
-    ErrorInThread(String),
-    /// An Error appeared related to the lack of a field.
-    SchemaError(String),
-    /// Tried to access a fastfield reader for a field not configured accordingly.
-    FastFieldError(FastFieldNotAvailableError),
-}
+error_chain!(
+    errors {
+        PathDoesNotExist(buf: PathBuf) {
+            description("path does not exist")
```
fulmicoton: have you checked that the rustdoc looked ok? On the first PR, the doc was not generated for the error kind

src/indexer/index_writer.rs
```diff
@@ -9,7 +9,7 @@ use core::SegmentReader;
 use indexer::stamper::Stamper;
 use datastruct::stacker::Heap;
 use directory::FileProtection;
-use Error;
+use error::*;
```
fulmicoton: no star, if possible

src/directory/managed_directory.rs
```diff
@@ -12,8 +12,7 @@ use std::io::Write;
 use core::MANAGED_FILEPATH;
 use std::collections::HashMap;
 use std::fmt;
-use Result;
-use Error;
+use error::*;
```
fulmicoton: no star, if possible

src/indexer/index_writer.rs
```diff
@@ -603,7 +601,7 @@ mod tests {
     use schema::{self, Document};
     use Index;
     use Term;
-    use Error;
+    use error::*;
```
fulmicoton, line 604: no star, if possible

src/indexer/segment_updater.rs
```diff
@@ -7,7 +7,7 @@ use core::SegmentMeta;
 use core::SerializableSegment;
 use directory::Directory;
 use indexer::stamper::Stamper;
-use Error;
+use error::*;
```
fulmicoton: no star, if possible

fulmicoton:
LGTM, added a bunch of minor comments

src/core/index.rs
KodrAus: There's also a `ResultExt` trait that gives us the `chain_err` method. 

It's not uncommon to `use error::*` but if you'd rather avoid it I think that's reasonable too, I can just import the bits from `error` I actually need.

src/directory/error.rs
```diff
@@ -45,8 +175,46 @@ pub enum DeleteError {
...
+        match *self {
+            DeleteError::FileDoesNotExist(ref path) => {
+                write!(f, "the file '{:?}' does not exist", path)
+            }
+            DeleteError::FileProtected(ref path) => {
+                write!(f, "the file '{:?}' is protected and can't be deleted", path)
+            }
+            DeleteError::IOError(ref err) => {
+                write!(f,
+                       "an io error occurred while opening a file for reading: '{}'",
```
KodrAus: Wrong message

src/core/index.rs
fulmicoton: I'd rather avoid it.

PR #203, opened 2017-08-03
Remove submodule information in README as subtree is now used

king6cong:


fulmicoton:
Oh awesome catch, thank you @king6cong !

PR #209, opened 2017-08-07
WIP: S3

drusellers:
# At this point this PR is more about getting feedback then it is about being merged in.

My goal in opening this PR is to get feedback on my `Directory::read` implementation. I'm looking to learn as much as possible so please don't hold back from any review comments.


Cargo.toml
```diff
@@ -42,6 +42,8 @@ futures-cpupool = "0.1"
 error-chain = "0.8"
 owning_ref = "0.3"
 stable_deref_trait = "1.0.0"
+rusoto_core = "*"
+rusoto_s3 = "*"
```
drusellers: Once I get closer to done, I'll lock down the version numbers

src/directory/s3_directory.rs
```diff
@@ -0,0 +1,248 @@
...
+        InnerDirectory(Arc::new(RwLock::new(HashMap::new())))
+    }
+}
+
+/// Directory storing data in files, read via mmap.
+///
+/// The Mmap object are cached to limit the
+/// system calls.
+#[derive(Clone)]
+pub struct S3Directory {
```
drusellers, line 313: I think I want to cache the `S3` client here rather than just create it all the time.

src/directory/s3_directory.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    /// Joins a relative_path to the directory `root_path`
+    /// to create a proper complete `filepath`.
+    fn resolve_path(&self, relative_path: &Path) -> PathBuf {
+        self.root_path.join(relative_path)
+    }
+}
+
+impl Directory for S3Directory {
+    fn open_read(&self, path: &Path) -> result::Result<ReadOnlySource, OpenReadError> {
```
drusellers, line 373: This method has been implemented in a very brute force procedural way. My goals are to make sure I understand all of the primitives I'm going to be using. I'll then look at extracting methods to better leverage rust's option/result chaining methods.

If you can help invest in my understanding of rust here, I feel that it will carry out through the rest of the code.

Cargo.toml
```diff
@@ -64,6 +66,7 @@ debug-assertions = false
 default = ["simdcompression"]
 simdcompression = ["libc", "gcc"]
 streamdict = []
+s3 = []
```
fulmicoton: rusoto should be a s3 feature dependency.

drusellers: got it and made it optional

src/directory/s3_directory.rs
```diff
@@ -0,0 +1,285 @@
...
+            let io_err = make_io_err(msg);
+            OpenReadError::IOError(IOError::with_path(path.to_owned(), io_err))
+        })?;
+
+        if !cache.contains_key(path) {
+            self.load_missing_key(client, path)?;
+        }
+
+        //TODO: map_err
+        let data = cache.get(path).unwrap();
```
fulmicoton: Supposedly it is not possible, because if it was the case, load_missing_key would have returned an Error right? 
But this assumes load_missing_key has been implemented correctly of course... And having to deal with the None branch feels a bit clumsy...

What if `load_missing_key` was called `fetch` and was returning the data? You could then use the entry interface of the map in this method.

drusellers: That will by my next commit! Thank you.

src/directory/s3_directory.rs
```diff
@@ -0,0 +1,285 @@
...
+    let client = default_tls_client()?;
+    let provider = DefaultCredentialsProvider::new()?;
+
+    Ok(Box::new(S3Client::new(client, provider, region)))
+}
+
+#[derive(Clone)]
+struct InnerDirectory {
+    bucket: String,
+    cache: Arc<RwLock<HashMap<PathBuf, Arc<Vec<u8>>>>>,
```
fulmicoton: I think you can make the directory even nicer if instead of using this cache in memory, you wrapped another directory.

So to get the same functionality at this one, you would pass it the RAMDirectory, but if you want your cache to be persisting, you could use the MmapDirectory.

drusellers: that would also handle some repetitive code i have between the two. Nice.

src/directory/s3_directory.rs
```diff
@@ -0,0 +1,285 @@
...
+                "Failed to acquire write lock for the \
+                                            directory when trying to read {:?}",
+                path
+            );
+            let io_err = make_io_err(msg);
+            OpenReadError::IOError(IOError::with_path(path.to_owned(), io_err))
+        })?;
+
+        // TODO: this is comical and I'm more than likely over thinking it
+        let key = path.as_os_str().to_os_string().into_string().map_err(|_| {
```
fulmicoton: What is the reason of those?

src/directory/s3_directory.rs
```diff
@@ -0,0 +1,285 @@
...
+            let msg = format!("Could not build key path");
+            let io_err = make_io_err(msg);
+            OpenReadError::IOError(IOError::with_path(path.to_owned(), io_err))
+        })?;
+
+        let obj = client
+            .get_object(&GetObjectRequest {
+                bucket: self.bucket.clone(),
+                key,
+                ..Default::default()
```
fulmicoton, line 191: Oh I have no idea what this does, what is `..` doing here?

src/directory/s3_directory.rs
```diff
@@ -0,0 +1,285 @@
...
+        let cache_exist = self.cache
+            .read()
+            .expect("Failed to get read lock directory.")
+            .contains_key(path);
+
+        let mut exist = cache_exist;
+        if !cache_exist {
+            let r = self.load_missing_key(client, path);
+
+            exist = r.is_ok();
```
fulmicoton: the logic of the if block can be simplified if you remove the mutating bool.

src/directory/s3_directory.rs
drusellers: Trying to convert a `PathBuf` / `Path` from the `Directory` api into a string for usage with the `s3` api

src/directory/s3_directory.rs
drusellers: The `rusoto` api has a large number of property fields, and each `rusoto` request object implements the default trait. This way you can just specify the 1 or 2 fields that you need then apply the `..` operator + `Default::default` to set all the other fields to their default value. Often this is `None` in `rusoto`. Saved me a lot of boiler plate typing for sure.

src/directory/s3_directory.rs
drusellers: this changed quite a bit with the `Entry` api changes

src/directory/s3_directory.rs
drusellers: If I make that move, I lose access to the `Entry` api of `HashMap`

PR #211, opened 2017-08-19
fix segments sorting in log_merge_policy

rclaude:
bug: segments were sorted on their indices (first field in the tuples)
fix: sort on the segments size


fulmicoton:
Awesome thanks!

PR #217, opened 2017-10-06
Updates crate gcc to cc v1

j4in:


KodrAus (approved):
Thanks @j4in! This looks good to me. I don't think this is a breaking change since `cc` is just a build dependency.

PR #275, opened 2018-04-24
Add fast field for associating arbitrary bytes to a document

### LINKED ISSUE #270: Create `&[u8]` fast fields
They can use the double array strategy used for multivalued int.
This requires to add another type in the schema.

We can use base64 for its JSON encoding.

jason-wolfe:
Closes #270

### DIFF
--- a/src/core/segment_reader.rs
+++ b/src/core/segment_reader.rs
@@ -12,7 +12,6 @@ use fastfield::FastFieldReader;
 use fastfield::{self, FastFieldNotAvailableError};
 use fastfield::{BytesFastFieldReader, FastValue, MultiValueIntFastFieldReader};
 use fieldnorm::FieldNormReader;
-use owning_ref::OwningRef;
 use schema::Cardinality;
 use schema::Document;
 use schema::Field;
@@ -151,17 +150,19 @@ impl SegmentReader {
     }
 
     /// Accessor to the `BytesFastFieldReader` associated to a given `Field`.
-    /// May panic if the field is not a Bytes fast field.
     pub fn bytes_fast_field_reader(&self, field: Field) -> fastfield::Result<BytesFastFieldReader> {
         let field_entry = self.schema.get_field_entry(field);
+        match field_entry.field_type() {
+            &FieldType::Bytes => {},
+            _ => return Err(FastFieldNotAvailableError::new(field_entry)),
+        }
         let idx_reader = self.fast_fields_composite
             .open_read_with_idx(field, 0)
             .ok_or_else(|| FastFieldNotAvailableError::new(field_entry))
             .map(FastFieldReader::open)?;
         let values = self.fast_fields_composite
             .open_read_with_idx(field, 1)
-            .ok_or_else(|| FastFieldNotAvailableError::new(field_entry))
-            .map(|x| OwningRef::new(x).map(|source| &source[..]))?;
+            .ok_or_else(|| FastFieldNotAvailableError::new(field_entry))?;
         Ok(BytesFastFieldReader::open(idx_reader, values))
     }
 
--- a/src/core/segment_reader.rs
+++ b/src/core/segment_reader.rs
@@ -10,8 +10,9 @@ use fastfield::DeleteBitSet;
 use fastfield::FacetReader;
 use fastfield::FastFieldReader;
 use fastfield::{self, FastFieldNotAvailableError};
-use fastfield::{FastValue, MultiValueIntFastFieldReader};
+use fastfield::{BytesFastFieldReader, FastValue, MultiValueIntFastFieldReader};
 use fieldnorm::FieldNormReader;
+use owning_ref::OwningRef;
 use schema::Cardinality;
 use schema::Document;
 use schema::Field;
@@ -140,6 +141,21 @@ impl SegmentReader {
         } else {
             Err(FastFieldNotAvailableError::new(field_entry))
         }
+    }
+
+    /// Accessor to the `BytesFastFieldReader` associated to a given `Field`.
+    /// May panic if the field is not a Bytes fast field.
+    pub fn bytes_fast_field_reader(&self, field: Field) -> fastfield::Result<BytesFastFieldReader> {
+        let field_entry = self.schema.get_field_entry(field);
+        let idx_reader = self.fast_fields_composite
+            .open_read_with_idx(field, 0)
+            .ok_or_else(|| FastFieldNotAvailableError::new(field_entry))
+            .map(FastFieldReader::open)?;
+        let values = self.fast_fields_composite
+            .open_read_with_idx(field, 1)
+            .ok_or_else(|| FastFieldNotAvailableError::new(field_entry))
+            .map(|x| OwningRef::new(x).map(|source| &source[..]))?;
+        Ok(BytesFastFieldReader::open(idx_reader, values))
     }
 
     /// Accessor to the `FacetReader` associated to a given `Field`.
--- a/src/fastfield/bytes/reader.rs
+++ b/src/fastfield/bytes/reader.rs
@@ -1,0 +1,36 @@
+use owning_ref::OwningRef;
+
+use directory::ReadOnlySource;
+use fastfield::FastFieldReader;
+use DocId;
+
+/// Reader for byte array fast fields
+///
+/// The reader is implemented as a `u64` fast field and a separate collection of bytes.
+///
+/// The `vals_reader` will access the concatenated list of all values for all documents.
+///
+/// The `idx_reader` associates, for each document, the index of its first value.
+///
+/// Reading the value for a document is done by reading the start index for it,
+/// and the start index for the next document, and keeping the bytes in between.
+pub struct BytesFastFieldReader {
+    idx_reader: FastFieldReader<u64>,
+    values: OwningRef<ReadOnlySource, [u8]>,
+}
+
+impl BytesFastFieldReader {
+    pub(crate) fn open(
+        idx_reader: FastFieldReader<u64>,
+        values: OwningRef<ReadOnlySource, [u8]>,
+    ) -> BytesFastFieldReader {
+        BytesFastFieldReader { idx_reader, values }
+    }
+
+    /// Returns the bytes associated to the given `doc`
+    pub fn get_val(&self, doc: DocId) -> &[u8] {
+        let start = self.idx_reader.get(doc) as usize;
+        let stop = self.idx_reader.get(doc + 1) as usize;
+        &self.values[start..stop]
+    }
+}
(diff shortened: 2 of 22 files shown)


jason-wolfe:
WIP because I haven't actually made non-test changes that depend on this to test it for real.

jason-wolfe:
Also still missing merging support.

src/fastfield/bytes/reader.rs
```diff
@@ -0,0 +1,36 @@
...
+/// Reader for byte array fast fields
+///
+/// The reader is implemented as a `u64` fast field and a separate collection of bytes.
+///
+/// The `vals_reader` will access the concatenated list of all values for all documents.
+///
+/// The `idx_reader` associates, for each document, the index of its first value.
+///
+/// Reading the value for a document is done by reading the start index for it,
+/// and the start index for the next document, and keeping the bytes in between.
```
fulmicoton, line 16: Awesome doc.

src/fastfield/bytes/writer.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// Shift to the next document and add all of the
+    /// matching field values present in the document.
+    pub fn add_document(&mut self, doc: &Document) {
+        self.next_doc();
+        for field_value in doc.field_values() {
+            if field_value.field() == self.field {
+                if let &Value::Bytes(ref bytes) = field_value.value() {
+                    self.vals.extend_from_slice(bytes);
+                } else {
+                    panic!("MultiValued Bytes contained non-Bytes Value!: {:?}", field_value);
```
fulmicoton: maybe add the field id in the error msg.

src/indexer/segment_writer.rs
```diff
@@ -202,6 +202,9 @@ impl<'a> SegmentWriter<'a> {
                         }
                     }
                 }
+                FieldType::Bytes => {
+                    // Do nothing. Bytes only supports fast fields.
```
fulmicoton, line 206: cool! THanks for not using `_`

src/schema/field_type.rs
```diff
@@ -120,7 +139,7 @@ impl FieldType {
                         Err(ValueParsingError::OverflowError(msg))
                     }
                 }
-                FieldType::Str(_) | FieldType::HierarchicalFacet => {
+                FieldType::Str(_) | FieldType::HierarchicalFacet | FieldType::Bytes => {
```
fulmicoton: I think you want bytes as another case, and panic or return an error.

src/schema/field_type.rs
```diff
@@ -135,3 +154,17 @@ impl FieldType {
...
+
+#[cfg(test)]
+mod tests {
+    use super::FieldType;
+    use schema::Value;
+
+    #[test]
+    fn test_bytes_value_from_json() {
+        let result = FieldType::Bytes
+            .value_from_json(&json!("dGhpcyBpcyBhIHRlc3Q="))
```
fulmicoton, line 167: cool. maybe test the errors as well.

src/schema/field_entry.rs
```diff
@@ -168,8 +181,16 @@ impl<'de> Deserialize<'de> for FieldEntry {
                                 return Err(de::Error::duplicate_field("type"));
                             }
                             ty = Some(map.next_value()?);
-                            if ty == Some("hierarchical_facet") {
-                                field_type = Some(FieldType::HierarchicalFacet);
+                            match ty {
```
fulmicoton: something is weird in this code (not yours). We wrap it in `Some` and match it right away?

fulmicoton: Can you have a look at the bigger picture and see if there is something wrong here?

src/postings/postings_writer.rs
```diff
@@ -36,7 +36,7 @@ fn posting_from_field_entry<'a>(
                 }
             })
             .unwrap_or_else(|| SpecializedPostingsWriter::<NothingRecorder>::new_boxed(heap)),
-        FieldType::U64(_) | FieldType::I64(_) | FieldType::HierarchicalFacet => {
+        FieldType::U64(_) | FieldType::I64(_) | FieldType::HierarchicalFacet | FieldType::Bytes => {
```
fulmicoton: `FieldType::Bytes` cannot be indexed at all. So you probably want a specific case.
`NothingRecorder` is a bit misleading here. It deos not "not index" the field. It indexes it with no extra informaiton (like docfreq or position)

src/schema/field_entry.rs
jason-wolfe: Yeah, will do.

src/postings/postings_writer.rs
jason-wolfe: I'm not sure what the right solution is here.

One option is to return on `Option<Box<PostingsWriter + 'a>>` and never create a postings writer for the field. I don't know enough to know if this will cause downstream blowups.

Another option is to return a `ReallyDoesNothingRecorder` (obviously not the real name :smiley:).

src/schema/field_type.rs
```diff
@@ -135,3 +154,32 @@ impl FieldType {
...
+        match result {
+            Err(ValueParsingError::TypeError(_)) => {}
+            _ => panic!("Expected parse failure for wrong type")
+        }
+
+        let result = FieldType::Bytes
+            .value_from_json(&json!("-"));
+        match result {
+            Err(ValueParsingError::InvalidBase64(_)) => {}
+            _ => panic!("Expected parse failure for invalid base64")
```
fulmicoton, line 182: awesome.

src/postings/postings_writer.rs
fulmicoton: I think you are right. The whole thing will stay ugly as long as it does not under go a rather large refactoring.

src/schema/field_type.rs
fulmicoton: thx

src/postings/postings_writer.rs
fulmicoton: Can you just add the following comment :
// `FieldType::Bytes` cannot actually be indexed.
// TODO fix during the indexer refactoring described in #276

fulmicoton:
Please edit `CHANGELOG.md` for `0.6.0` and make sure to add your username on the feature you add. It makes the release process much smoother.

jason-wolfe:
Ready for merge from my side, pending CR.

src/postings/postings_writer.rs
jason-wolfe: Done

fulmicoton:
Sweet． Will have at that tomorrow

src/indexer/merger.rs
```diff
@@ -284,25 +287,27 @@ impl IndexMerger {
...
         // what should be the bit length use for bitpacking.
         for reader in &self.readers {
-            let multi_ff_reader = reader.multi_fast_field_reader::<u64>(field)?;
+            let idx_reader = reader.fast_field_reader_with_idx::<u64>(field, 0)?;
             if let Some(delete_bitset) = reader.delete_bitset() {
                 for doc in 0u32..reader.max_doc() {
                     if !delete_bitset.is_deleted(doc) {
-                        total_num_vals += multi_ff_reader.num_vals(doc) as u64;
+                        let start = idx_reader.get(doc);
+                        let end = idx_reader.get(doc + 1);
```
fulmicoton, line 305: why did you remove `.num_vals` ?

src/indexer/merger.rs
```diff
@@ -311,11 +316,13 @@ impl IndexMerger {
...
         let mut idx = 0;
         for reader in &self.readers {
-            let multi_ff_reader = reader.multi_fast_field_reader::<u64>(field)?;
+            let idx_reader = reader.fast_field_reader_with_idx::<u64>(field, 0)?;
             for doc in 0u32..reader.max_doc() {
                 if !reader.is_deleted(doc) {
                     serialize_idx.add_val(idx)?;
-                    idx += multi_ff_reader.num_vals(doc) as u64;
+                    let start = idx_reader.get(doc);
+                    let end = idx_reader.get(doc + 1);
```
fulmicoton, line 324: why did you remove .num_vals ?

src/core/segment_reader.rs
```diff
@@ -149,6 +150,21 @@ impl SegmentReader {
         }
     }
 
+    /// Accessor to the `BytesFastFieldReader` associated to a given `Field`.
+    /// May panic if the field is not a Bytes fast field.
```
fulmicoton: Does it panic though?

src/core/segment_reader.rs
```diff
@@ -149,6 +150,21 @@ impl SegmentReader {
...
+    pub fn bytes_fast_field_reader(&self, field: Field) -> fastfield::Result<BytesFastFieldReader> {
+        let field_entry = self.schema.get_field_entry(field);
```
fulmicoton: maybe test if the type is right now that you have the field entry, and return the error here.

The benefit is then that you `panic!` only if the index is corrupted, but you return a nice error if the field just does not have the right type.

src/indexer/merger.rs
jason-wolfe: Because it's no longer used. I could move it to FastFieldReader<u64> (which is all this method uses now), but it's no longer used by anything on the multi reader. It seems weird on FastFieldReader<u64>, however, because it's only meaningful for indexes.

src/core/segment_reader.rs
```diff
@@ -149,6 +150,21 @@ impl SegmentReader {
...
+        let idx_reader = self.fast_fields_composite
+            .open_read_with_idx(field, 0)
+            .ok_or_else(|| FastFieldNotAvailableError::new(field_entry))
+            .map(FastFieldReader::open)?;
+        let values = self.fast_fields_composite
+            .open_read_with_idx(field, 1)
+            .ok_or_else(|| FastFieldNotAvailableError::new(field_entry))
+            .map(|x| OwningRef::new(x).map(|source| &source[..]))?;
```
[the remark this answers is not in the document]
jason-wolfe: Sure.

src/core/segment_reader.rs
jason-wolfe: It doesn't. Will remove.

src/indexer/merger.rs
jason-wolfe: Answered in other comment.

src/indexer/merger.rs
fulmicoton: ok I see

src/core/segment_reader.rs
fulmicoton: Right now I think it does panic but in some specific corner case.. You can remove the panic if you address the other comment below.

src/core/segment_reader.rs
jason-wolfe: Should I turn FastFieldNotAvailableError into a FastFieldError enum to indicate the fast field is the wrong type? Just saying it's not available seems confusing.
EDIT: Actually I don't think there's a good way to do this cleanly. I'll stick with FastFieldNotAvailableError. But maybe should file a follow-up to do this more nicely.

fulmicoton: ok. I also think making the two unindistiguishable is not the best but we can keep it that way for the moment.

PR #278, opened 2018-04-29
Simple Implementation of NGram Tokenizer

drusellers:
It does not yet support edges
It could probably be better in many "rusty" ways
But the test is passing, so I'll call this a good stopping point for
the day.

- [x] support edgegrams
- [x] how should a user configure the min/max gram stuffs
- [x] should it be in the default token manager (I think no)
- [ ] is this following tantivy best practices
- [x] how should we support white space - Lucene doesn't address it. It uses other filters

### DIFF
--- a/src/tokenizer/mod.rs
+++ b/src/tokenizer/mod.rs
@@ -155,29 +155,34 @@ pub use self::tokenizer::BoxedTokenizer;
 pub use self::tokenizer::{Token, TokenFilter, TokenStream, Tokenizer};
 pub use self::tokenizer_manager::TokenizerManager;
 
+/// This is a function that can be used in tests and doc tests
+/// to assert a token's correctness.
+/// TODO: can this be wrapped in #[cfg(test)] so as not to be in the
+/// public api?
+pub fn assert_token(token: &Token, position: usize, text: &str, from: usize, to: usize) {
+    assert_eq!(
+        token.position, position,
+        "expected position {} but {:?}",
+        position, token
+    );
+    assert_eq!(token.text, text, "expected text {} but {:?}", text, token);
+    assert_eq!(
+        token.offset_from, from,
+        "expected offset_from {} but {:?}",
+        from, token
+    );
+    assert_eq!(
+        token.offset_to, to,
+        "expected offset_to {} but {:?}",
+        to, token
+    );
+}
+
 #[cfg(test)]
-mod test {
+pub mod test {
+    use super::assert_token;
     use super::Token;
     use super::TokenizerManager;
-
-    fn assert_token(token: &Token, position: usize, text: &str, from: usize, to: usize) {
-        assert_eq!(
-            token.position, position,
-            "expected position {} but {:?}",
-            position, token
-        );
-        assert_eq!(token.text, text, "expected text {} but {:?}", text, token);
-        assert_eq!(
-            token.offset_from, from,
-            "expected offset_from {} but {:?}",
-            from, token
-        );
-        assert_eq!(
-            token.offset_to, to,
-            "expected offset_to {} but {:?}",
-            to, token
-        );
-    }
 
     #[test]
     fn test_raw_tokenizer() {
@@ -242,6 +247,7 @@ mod test {
     #[test]
     fn test_ngram_tokenizer() {
         use super::{LowerCaser, NgramTokenizer};
+        use tokenizer::tokenizer::TokenStream;
         use tokenizer::tokenizer::Tokenizer;
 
         let tokenizer_manager = TokenizerManager::default();
@@ -255,7 +261,7 @@ mod test {
             NgramTokenizer::new(2, 5, true).filter(LowerCaser),
         );
 
-        let tokenizer = tokenizer_manager.get("ngram12").unwrap();
+        let tokenizer = NgramTokenizer::new(1, 2, false);
         let mut tokens: Vec<Token> = vec![];
         {
             let mut add_token = |token: &Token| {
--- a/src/tokenizer/mod.rs
+++ b/src/tokenizer/mod.rs
@@ -226,20 +226,15 @@ mod test {
         use tokenizer::tokenizer::Tokenizer;
 
         let tokenizer_manager = TokenizerManager::default();
+        tokenizer_manager.register("ngram12", NgramTokenizer::new(1, 2, false));
         tokenizer_manager.register(
-            "ngram12",
-            NgramTokenizer {
-                min_gram: 1,
-                max_gram: 2,
-                edges_only: false,
-            },
+            "ngram3",
+            NgramTokenizer::new(3, 3, false).filter(LowerCaser),
         );
-        let combined = NgramTokenizer {
-            min_gram: 3,
-            max_gram: 3,
-            edges_only: false,
-        };
-        tokenizer_manager.register("ngram3", combined.filter(LowerCaser));
+        tokenizer_manager.register(
+            "edgegram5",
+            NgramTokenizer::new(2, 5, true).filter(LowerCaser),
+        );
 
         let en_tokenizer = tokenizer_manager.get("ngram12").unwrap();
         let mut tokens: Vec<String> = vec![];
@@ -273,6 +268,23 @@ mod test {
         assert_eq!(&tokens[0], "hel");
         assert_eq!(&tokens[1], "ell");
         assert_eq!(&tokens[2], "llo");
+
+        let en_tokenizer = tokenizer_manager.get("edgegram5").unwrap();
+        let mut tokens: Vec<String> = vec![];
+        {
+            let mut add_token = |token: &Token| {
+                tokens.push(token.text.clone());
+            };
+            en_tokenizer
+                .token_stream("Frankenstein")
+                .process(&mut add_token);
+        }
+
+        assert_eq!(tokens.len(), 4);
+        assert_eq!(&tokens[0], "fr");
+        assert_eq!(&tokens[1], "fra");
+        assert_eq!(&tokens[2], "fran");
+        assert_eq!(&tokens[3], "frank");
     }
 
     #[test]
--- a/src/tokenizer/mod.rs
+++ b/src/tokenizer/mod.rs
@@ -160,21 +160,40 @@ mod test {
     use super::Token;
     use super::TokenizerManager;
 
+    fn assert_token(token: &Token, position: usize, text: &str, from: usize, to: usize) {
+        assert_eq!(
+            token.position, position,
+            "expected position {} but {:?}",
+            position, token
+        );
+        assert_eq!(token.text, text, "expected text {} but {:?}", text, token);
+        assert_eq!(
+            token.offset_from, from,
+            "expected offset_from {} but {:?}",
+            from, token
+        );
+        assert_eq!(
+            token.offset_to, to,
+            "expected offset_to {} but {:?}",
+            to, token
+        );
+    }
+
     #[test]
     fn test_raw_tokenizer() {
         let tokenizer_manager = TokenizerManager::default();
         let en_tokenizer = tokenizer_manager.get("raw").unwrap();
-        let mut tokens: Vec<String> = vec![];
+        let mut tokens: Vec<Token> = vec![];
         {
             let mut add_token = |token: &Token| {
-                tokens.push(token.text.clone());
+                tokens.push(token.clone());
             };
             en_tokenizer
                 .token_stream("Hello, happy tax payer!")
                 .process(&mut add_token);
         }
         assert_eq!(tokens.len(), 1);
-        assert_eq!(&tokens[0], "Hello, happy tax payer!");
+        assert_token(&tokens[0], 0, "Hello, happy tax payer!", 0, 23);
     }
 
     #[test]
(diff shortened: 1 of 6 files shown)


drusellers:
@fulmicoton i'm ready for any feedback you might have on this PR

fulmicoton:
@drusellers  Cool.  Will have a look at that this weekend then.

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -0,0 +1,121 @@
...
+  }
+}
+
+impl<'a> TokenStream for NgramTokenStream<'a> {
+  fn advance(&mut self) -> bool {
+    // clear out working token text
+    self.token.text.clear();
+
+    loop {
+      // can we proceed?
```
fulmicoton: this comment could be more informative.

e.g. Check if the token we are about to emit is not exceeding the text length.

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -0,0 +1,121 @@
...
+    NgramTokenizer {
+      min_gram,
+      max_gram,
+      edges_only,
+    }
+  }
+}
+pub struct NgramTokenStream<'a> {
+  text: &'a str,
+  location: usize,
```
fulmicoton: rename as position

fulmicoton:
I need more information here. Can you explain the use case? Especially do we want a `Tokenizer` or a `TokenFilter` ?

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -0,0 +1,121 @@
+use super::{Token, TokenStream, Tokenizer};
+
+///Tokenize the text by splitting words into ngrams of the given size
+#[derive(Clone)]
```
fulmicoton: We need more info about what it does.

Especially what are the resulting token's `position` and what `edges_only` is about.

Maybe even move the example as a doctest.

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -0,0 +1,121 @@
...
+  edges_only: bool,
+}
+
+impl<'a> Tokenizer<'a> for NgramTokenizer {
+  type TokenStreamImpl = NgramTokenStream<'a>;
+
+  fn token_stream(&self, text: &'a str) -> Self::TokenStreamImpl {
+    NgramTokenStream {
+      text,
+      location: 0,
```
fulmicoton, line 86: Rename `location` to `offset_from`, or `cur_offset_from`

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -0,0 +1,121 @@
...
+        // cycle through 1,2 (min...max)
+        match self.cycle() {
+          Some(size) => {
+            let offset_from = self.location;
+            let offset_to = offset_from + size;
+
+            self.token.offset_from = offset_from;
+            self.token.offset_to = offset_to;
+
+            self.token.text.push_str(&self.text[offset_from..offset_to]);
```
fulmicoton: Alternatively we could push one new char at each iteration, or send the ngrams in an inuitive order and remove one char at each time.
  
But no need to fix your code... This is probably not performance sensitive.

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+}
+
+impl<'a> TokenStream for NgramTokenStream<'a> {
+  fn advance(&mut self) -> bool {
+    // clear out working token text
+    self.token.text.clear();
+
+    loop {
```
fulmicoton: This does not need to be a loop.

src/tokenizer/mod.rs
```diff
@@ -218,6 +220,73 @@ mod test {
...
+
+        assert_eq!(tokens.len(), 8);
+        assert_eq!(&tokens[0], "h");
+        assert_eq!(&tokens[1], "he");
+        assert_eq!(&tokens[2], "e");
+        assert_eq!(&tokens[3], "el");
+        assert_eq!(&tokens[4], "l");
+        assert_eq!(&tokens[5], "ll");
+        assert_eq!(&tokens[6], "l");
+        assert_eq!(&tokens[7], "lo");
```
fulmicoton: `assert` on token positions

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -0,0 +1,121 @@
+use super::{Token, TokenStream, Tokenizer};
+
+///Tokenize the text by splitting words into ngrams of the given size
+#[derive(Clone)]
+pub struct NgramTokenizer {
+  min_gram: usize,
+  max_gram: usize,
+  edges_only: bool,
```
fulmicoton: Should `edges_only` be called `prefix_only`?
I don't think the code does suffix right now, does it?

drusellers:
Thank you! I'll square this away this weekend!

src/tokenizer/ngram_tokenizer.rs
drusellers: Will do, but one question. I chose location so it would match the `JapaneseTokenizer` and the `RawTokenizer`. Would you still like it to be changed?

src/tokenizer/ngram_tokenizer.rs
drusellers: that was a fun brain bender! Code is much better now. thank you.

src/tokenizer/mod.rs
drusellers: I'm going to derive `Clone` and `Debug` for `Token` to make this a bit easier.

drusellers: OMG, clutch. I added token tests to the other tokenizers. I get it now. :)

src/tokenizer/ngram_tokenizer.rs
drusellers: It does not.  I was following the naming of [Lucene](https://github.com/apache/lucene-solr/blob/master/lucene/analysis/common/src/java/org/apache/lucene/analysis/ngram/NGramTokenizer.java#L69) - I'm happy to name it what makes sense for tantivy though.

drusellers:
> I need more information here. Can you explain the use case? Especially do we want a Tokenizer or a TokenFilter ?

Looks like Lucene has both. I just started here, but can add the Filter too.

I originally desired this functionality so I could implement EdgeGramTokens so I could do type ahead search for my project. Also, it looked like a relatively easy Tokenizer to port from Lucene for Tantivy.

src/tokenizer/ngram_tokenizer.rs
fulmicoton: I don't see `location` in the `JapaneseTokenizer` nor the `RawTokenizer`.

src/tokenizer/mod.rs
```diff
@@ -160,64 +160,83 @@ mod test {
...
                 .token_stream("野菜食べないとやばい!")
                 .process(&mut add_token);
         }
         assert_eq!(tokens.len(), 5);
-        assert_eq!(&tokens[0], "野菜");
-        assert_eq!(&tokens[1], "食べ");
-        assert_eq!(&tokens[2], "ない");
-        assert_eq!(&tokens[3], "と");
-        assert_eq!(&tokens[4], "やばい");
+        assert_token(&tokens[0], 0, "野菜", 0, 6);
```
fulmicoton, line 240: thank you for all these!

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -74,41 +146,21 @@ impl<'a> TokenStream for NgramTokenStream<'a> {
...
-            continue;
-          }
-        }
-
-        return true;
-      } else {
-        break;
-      }
-    }
+    if let Some((location, size)) = self.chomp() {
```
fulmicoton: sweet. Thank you

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -74,41 +146,21 @@ impl<'a> TokenStream for NgramTokenStream<'a> {
...
+      // this token's position is our current location
```
fulmicoton: let's  not use the word `location`. `position` or `offset`

src/tokenizer/tokenizer.rs
```diff
@@ -260,3 +261,21 @@ pub trait TokenFilter<TailTokenStream: TokenStream>: Clone {
     /// Wraps a token stream and returns the modified one.
     fn transform(&self, token_stream: TailTokenStream) -> Self::ResultTokenStream;
 }
+
+#[cfg(test)]
+mod test {
+    use super::Token;
+
+    #[test]
+    fn clone() {
```
fulmicoton, line 270: Maybe test other attributes or remove the test entirely?

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -1,10 +1,67 @@
...
 
-///Tokenize the text by splitting words into ngrams of the given size
+/// Tokenize the text by splitting words into n-grams of the given size(s)
+///
+/// Example: `hello` would be tokenized as (min_gram: 2, max_gram: 3)
+///
+/// | Term     | he  | hel | el  | ell | ll  | llo | lo |
+/// |----------|-----|-----|-----|-----|-----|-----|----|
+/// | Position | 0   | 0   | 1   | 1   | 2   | 2   | 3  |
+/// | Offsets  | 0,2 | 0,3 | 1,3 | 1,4 | 2,4 | 2,5 | 3,5|
```
fulmicoton, line 13: The table is a great idea! Awesome!!!

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -0,0 +1,173 @@
+use super::{Token, TokenStream, Tokenizer};
+
+/// Tokenize the text by splitting words into n-grams of the given size(s)
+///
```
fulmicoton, line 4: Please rename `edges_only` to something that `prefix_only` and document it in the doc.

Position expresses the position of a token. The current spec is ok but requires a word or two.
Also, explicitely state that the position emitted will be the starting offset of the ngram as this is quite non-trivial.

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -0,0 +1,173 @@
...
+///       from, token
+///   );
+///   assert_eq!(
+///       token.offset_to, to,
+///       "expected offset_to {} but {:?}",
+///       to, token
+///   );
+/// }
+/// # fn main() {
+/// let tokenizer_manager = TokenizerManager::default();
```
fulmicoton: You can leave it as is, but this is a bit noisy for a doctest.

fulmicoton:
Good job. I can merge it as soon as you rename the `edges_only` and `location` stuff.

src/tokenizer/ngram_tokenizer.rs
drusellers: Ok, should it be something else? I can see now how the other's are working with their `position` fields, and this one is deviating. 🤔

src/tokenizer/tokenizer.rs
drusellers: Added. Thank you.

src/tokenizer/ngram_tokenizer.rs
drusellers: Fixed

src/tokenizer/mod.rs
drusellers: It was super helpful for me to see. I can now see why my `position` usage is divergent. I'll try and dig into this a bit more and see if I can't come up with something better.

src/tokenizer/ngram_tokenizer.rs
drusellers: Yeah, I hear that. :/ I'm not a huge fan, but not sure how to best expose the desired output. I pulled the `assert_token` out of the `#[cfg(test)]` so that I could reference it in the doctest. That helped a little bit. I'm thinking of trying to build some common collector as well, to reduce the noise and around that. Hmm.

src/tokenizer/ngram_tokenizer.rs
fulmicoton: I think that's ok for a tokenizer. That's also Lucene behavior. (not for the TokenFilter but for thisf Tokenizer).

FYI I just added #291 too. It's about adding a length to tokens.

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -2,72 +2,57 @@ use super::{Token, TokenStream, Tokenizer};
...
-/// assert_token(&tokens[1], 0, "hel", 0, 3);
-/// assert_token(&tokens[2], 1, "el", 1, 3);
-/// assert_token(&tokens[3], 1, "ell", 1, 4);
-/// assert_token(&tokens[4], 2, "ll", 2, 4);
-/// assert_token(&tokens[5], 2, "llo", 2, 5);
-/// assert_token(&tokens[6], 3, "lo", 3, 5);
+/// let tokenizer = NgramTokenizer::new(2, 3, false);
+/// let mut stream = tokenizer.token_stream("hello");
+///
+/// assert_token(stream.next().unwrap(), 0, "he", 0, 2);
```
fulmicoton: thanks!

fulmicoton:
I think we are good to go. Ideally location would have been renamed but this is private anway.

I added you as a member of `tantivy`. In the future you can push your work as feature branch in `tantivy-search/tantivy`.

drusellers:
Oh, I thought I did. :( re: `location`

I'll square that away.

And thank you. 🎉

PR #282, opened 2018-05-02
Compute space usage of a Searcher / SegmentReader / CompositeFile

### LINKED ISSUE #281: Add built in functionality for understanding per-field storage size
Add libraries to tantivy core for exposing per-thing (where "thing" could be postings, fast field, etc.), per-field information on disk usage.

This should be designed in a way to be able to be called and exposed to the user in tantivy-cli, which should also mean it's sufficient for other programmatic usages.

jason-wolfe:
Closes #281


jason-wolfe:
I opted to not call anything relating to this "weighing" or "weight", due to terminological overlap with Tantivy/Lucene `Weight` in terms of `Query` evaluation. `SpaceUsage` is clearer anyway.

One thing I'm not quite certain on is exposing `ByteCount`. It's helpful for type safety internally but other than that I'm not sure how I feel.

I also haven't performed any `tantivy-cli` integration so I'm not sure about whether the goals in #281 are entirely fulfilled by this API.

jason-wolfe:
I added serde support and tried it locally with a really simple program hacked together quickly.

```
extern crate env_logger;
extern crate serde_json;
extern crate tantivy;

use std::env;
use std::path::Path;
use tantivy::Index;
use tantivy::Result;


fn main() {
    env_logger::init();
    let args = env::args().collect::<Vec<_>>();
    let path = Path::new(&args[1]);
    main_inner(&path).unwrap();
}

fn main_inner(path: &Path) -> Result<()> {
    let index = Index::open(path)?;
    let searcher = index.searcher();
    let space_usage = searcher.space_usage();

    println!("{}", serde_json::to_string_pretty(&space_usage).expect("failed to serialize"));

    Ok(())
}
```

The output results are near unusable. Unreadable field numbers everywhere, not sorted in any way whatsoever, either by name, number, or size. Just generally not great.

Things that would be more helpful:
Grouping things by field, even though it doesn't represent how things are stored in the index.
Sorting fields by total space usage. Sorting within a field by usage.
Displaying real field names everywhere.

jason-wolfe:
The methods as is are sufficient for powering tantivy-viewer now.

I think there's no point in abstracting these at this time.

jason-wolfe:
@fulmicoton please review!

src/space_usage/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+        self.total
+    }
+}
+
+/// Represents combined space usage for all of the large components comprising a segment.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct SegmentSpaceUsage {
+    num_docs: u32,
+
+    termdict: PerFieldSpaceUsage,
```
fulmicoton, line 68: Can we use a HashMap<SegmentComponent, PerFieldSpaceUsage> instead?

src/space_usage/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+        }
+    }
+
+    /// Num docs in segment
+    pub fn num_docs(&self) -> u32 {
+        self.num_docs
+    }
+
+    /// Space usage for term dictionary
+    pub fn termdict(&self) -> &PerFieldSpaceUsage {
```
fulmicoton, line 140: Could we replace all of the accessors by a `.get(&self, component: SegmentComponent) ` ?

src/space_usage/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+use std::collections::HashMap;
+use std::ops::{Add, AddAssign};
+use serde::Serialize;
+use serde::Serializer;
+use serde::Deserialize;
+use serde::Deserializer;
+
+/// Indicates space usage in bytes
+#[derive(Clone, Copy, Debug)]
+pub struct ByteCount(pub usize);
```
fulmicoton: Do we want the inner value to be `pub` here?

fulmicoton: It might be worth investigating if there might be some usage of this struct elsewhere or if this is overkill. (That should not be part of this issue though.)

src/space_usage/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+}
+
+/// Represents space usage of a given field, breaking it down into the (field, index) pairs that
+/// comprise it.
+///
+/// See documentation for PerFieldSpaceUsage for slightly more information.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct FieldUsage {
+    field: Field,
+    weight: ByteCount,
```
fulmicoton: `weight` is not really explicit and overlaps with another concept in `tantivy` . Can we find a better name?

src/space_usage/mod.rs
jason-wolfe: This is possible, but gives up some static safety of having actually filled in all the values.

src/space_usage/mod.rs
jason-wolfe: Not exactly, since `SegmentComponent::DELETE` and `SegmentComponent::STORE` are not per field things. So there would have to be an enum result type.

src/space_usage/mod.rs
jason-wolfe: Yeah this should probably just be a `type ByteCount` instead of a `struct`.

src/space_usage/mod.rs
jason-wolfe: Sure. `bytes` or `byte_count` or something.

src/space_usage/mod.rs
fulmicoton: That's a matter of perspective. If I add a new component today, nothing enforces fixing the `SegmentSpaceUsage` structure.

You can get the best of two world (kind of) by populating it with a switch case.

```rust
fn compute_segment_space_usage(&self, component: SegmentComponent) {
    match component {
      // forces you to handle all components.
   }
}
```

Populating the HashMap can then be done by iterating over the SegmentComponent values.

```rust
for component in SegmentComponent::iterator() {
    let component_space_usage = self.compute_segment_space_usage(component);
    map.insert(component, component_space_usage )
}
```

src/space_usage/mod.rs
fulmicoton: Right.

jason-wolfe:
Yeah, I think you're right about using the component enum. I'll make a new enum for the result type and see how the ergonomics are.

src/space_usage/mod.rs
jason-wolfe: So I kind of went the opposite way. But it supports both use cases as well as makes updating almost certain to happen when a component is added.

Something about having no choice other than to consume the enum makes me feel dirty. I don't know why :\

Let me know if you feel strongly about deleting the other methods.

fulmicoton: ok...

fulmicoton:
Can you add unit tests?

fulmicoton:
@jason-wolfe  Is this dead?

jason-wolfe:
Just needs unit tests to merge.

I wouldn't say dead as it's important functionality and required for `tantivy-viewer`.

I can maybe find some time to wrap up in September :(

I also don't know what exactly to unit test.

jason-wolfe:
@fulmicoton I finally added some tests, and made it build again. Sorry for the ridiculous latency.

I did encounter a surprise when writing tests in which the store uses a non-zero number of bytes even when no fields are stored. Is this expected?

fulmicoton:
pinging @hntd187 as this may interest you.

Issue #277, opened 2018-04-25
Make indexing fieldnorms optional


fulmicoton:
Fieldnorms is only useful for users using the vanilla BM25 scoring.
For other users, this is a waste of resources.

Also, currently int fields can be indexed but do not have any fieldnorms. In some use case, one might want to use the regular scoring on that kind of field. Fieldnorm should be made optionl for int fields as well.

fulmicoton:
Update: In master (0.6.0-dev),  `fieldnorm` is computed for int fields.

fulmicoton:
Closed see https://github.com/quickwit-oss/tantivy/pull/1001

PR #284, opened 2018-05-02
Issue/274 merge facets

fulmicoton:


fulmicoton:
@jason-wolfe can you have a look at it please? I need to add unit test for the multivalued index field stuff.

This is forked from tag 0.5.1. Some of the code/bugfixes in here is irrelevant ot master.

src/query/all_query.rs
```diff
@@ -26,28 +27,52 @@ pub struct AllWeight;
...
+enum State {
+    NotStarted,
+    Started,
+    Finished
+}
+
 /// Scorer associated to the `AllQuery` query.
 pub struct AllScorer {
-    started: bool,
+    state: State,
```
fulmicoton, line 46: irrelevant for master but AllQuery was ignoring deletes.

src/indexer/merger.rs
```diff
@@ -17,6 +17,12 @@ use store::StoreWriter;
 use std::cmp::{max, min};
 use termdict::TermDictionary;
 use termdict::TermStreamer;
+use schema::FieldType;
+use termdict::TermOrdinal;
+use schema::Cardinality;
+use std::collections::HashMap;
+use fastfield::MultiValueIntFastFieldReader;
+use std::cmp;
```
jason-wolfe, line 24: nit: Already importing `cmp::{max, min}`. Side-note: `val.min(other_val)` also exists now.

Also imports in general need some re-ordering. Do we use `cargo fmt`?

src/indexer/merger.rs
```diff
@@ -103,204 +146,387 @@ impl IndexMerger {
...
-            .filter(|&(_, field_entry)| field_entry.is_indexed())
-            .map(|(field_id, _)| Field(field_id as u32))
-            .collect();
-        self.generic_write_fast_field(
-            fieldnorm_fastfields,
-            &extract_fieldnorm_reader,
-            fast_field_serializer,
-        )
+        for (field_id, field_entry) in self.schema.fields().iter().enumerate() {
+            if let FieldType::Str(ref text_options) = *field_entry.field_type() {
```
jason-wolfe: Are fieldnorms only supported for string fields?

src/indexer/merger.rs
```diff
@@ -103,204 +146,387 @@ impl IndexMerger {
...
-            .enumerate()
-            .filter(|&(_, field_entry)| field_entry.is_indexed())
-            .map(|(field_id, _)| Field(field_id as u32))
-            .collect();
-        self.generic_write_fast_field(
-            fieldnorm_fastfields,
-            &extract_fieldnorm_reader,
-            fast_field_serializer,
-        )
+        for (field_id, field_entry) in self.schema.fields().iter().enumerate() {
```
jason-wolfe, line 148: Is there any reason for this to contain field_ids instead of `Field`s?

src/indexer/merger.rs
```diff
@@ -103,204 +146,387 @@ impl IndexMerger {
...
+                         fast_field_serializer: &mut FastFieldSerializer,
+                         mut term_ord_mappings: HashMap<Field, TermOrdinalMapping>) -> Result<()> {
+        for (field_id, field_entry) in self.schema.fields().iter().enumerate() {
+            let field = Field(field_id as u32);
+            let field_type = field_entry.field_type();
+            match *field_type {
+                FieldType::HierarchicalFacet => {
+                    let term_ordinal_mapping = term_ord_mappings
+                        .remove(&field)
+                        .expect("Logic Error in Tantivy (Please report). HierarchicalFact field should have requires a\
```
jason-wolfe: nit: s/requires/required/

jason-wolfe (commented):
Can't finish review right now, left some comments.

src/indexer/merger.rs
fulmicoton: yes... Rethinking that... Maybe it makes sense to enable them for ints as well. What do you think?

src/indexer/merger.rs
fulmicoton: changed

src/indexer/merger.rs
```diff
@@ -103,204 +146,387 @@ impl IndexMerger {
...
+                }
+            }
+        }
+        Ok(())
     }
 
-    // used both to merge field norms and regular u64 fast fields.
-    fn generic_write_fast_field(
+
+    // used both to merge field norms, u64 fast fields, and i64 fast fields.
```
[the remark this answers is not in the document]
fulmicoton: changed. it is quite different in master.

src/indexer/merger.rs
```diff
@@ -103,204 +146,387 @@ impl IndexMerger {
...
-                        let error_msg =
-                            format!("Failed to find a u64_reader for field {:?}", field);
-                        error!("{}", error_msg);
-                        bail!(ErrorKind::SchemaError(error_msg));
+                }
+                None => {
+                    let error_msg =
+                        format!("Failed to find a u64_reader for field {:?}", field);
+                    error!("{}", error_msg);
+                    bail!(ErrorKind::SchemaError(error_msg));
```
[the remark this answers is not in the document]
fulmicoton: changed

src/indexer/merger.rs
fulmicoton: Right now you need to enumerate and create the Field object on your own. (That kind of sucks.)

That would be nice to clean that up in one way or another, but the refactoring is quite heavy and requires a lot of thinking ahead of time.
I'll move the creation of the `Field object` closer to the loop, if that helps a bit.

src/indexer/merger.rs
```diff
@@ -103,204 +146,387 @@ impl IndexMerger {
...
+                }
+            }
+        }
+
+        for reader in &self.readers {
```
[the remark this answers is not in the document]
fulmicoton: changed.

src/indexer/merger.rs
fulmicoton: You are right. There is no reason to not consider fieldnorm when scoring a query with `int fields`.

You actually have a small horse in this race, but its running the other way... You may end up having some index int fields, in which case field norms will just be a useless load for you (exactly one byte per doc in master). Note that this is also true for text fields so we probably want to make it configurable. Well the stuff we already discussed and spawned #277

jason-wolfe:
Sorry it took so long, finished reading everything.

src/indexer/merger.rs
fulmicoton: changed btw. I'll run cargo fmt once I integrate with master. Running it now might make the merge painful

src/indexer/merger.rs
```diff
@@ -103,204 +145,369 @@ impl IndexMerger {
...
-                        fast_single_field_serializer.add_val(val)?;
+        // First we compute the total number of vals.
+        //
+        // This is required by the bitpacker, as it needs to know
+        // what should be the bit length use for bitpacking.
+        for reader in &self.readers {
+            let multi_ff_reader = reader.multi_fast_field_reader::<u64>(field)?;
+            let delete_bitset = reader.delete_bitset();
+            if delete_bitset.has_deletes() {
+                for doc in 0u32..reader.max_doc() {
```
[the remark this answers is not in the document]
fulmicoton: Created https://github.com/tantivy-search/tantivy/issues/287

src/indexer/merger.rs
```diff
@@ -103,204 +145,369 @@ impl IndexMerger {
...
+                        max_value = cmp::max(val, max_value);
+                    }
+                }
+            }
+            // TODO optimize when no deletes
+        }
+
+        // then we serialize the values
+        {
+            let mut serialize_vals = fast_field_serializer.new_u64_fast_field_with_idx(field, min_value, max_value, 1)?;
```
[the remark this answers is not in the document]
fulmicoton: Excellent point!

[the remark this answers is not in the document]
fulmicoton: Fixed and added a unit test.

src/postings/postings_writer.rs
```diff
@@ -110,17 +111,24 @@ impl<'a> MultiFieldPostingsWriter<'a> {
...
-            {
-                mapping.insert(term_unord_id, term_ord);
+            let field_entry = self.schema.get_field_entry(field);
+
+            match field_entry.field_type() {
+                FieldType::Str(_) | FieldType::HierarchicalFacet => {
+                    // populating the unordered term ord -> ordered term ord mapping
+                    // for the field.
+                    let mut mapping = HashMap::new();
+                    for (term_ord, term_unord_id) in term_offsets[start..stop]
```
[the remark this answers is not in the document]
fulmicoton: I did not really do that, but broke the logic into two parts, and used the `TermOrdinal` type for term_ord.

PR #285, opened 2018-05-03
Add Index::open_directory function

### LINKED ISSUE #279: Create a `Index::open_directory(...)` function
Right now, one can open a MMapDirectory from a path, but we really should have a method that works for any directory.

drusellers:
Closes #279

- [x] How would you like this tested, and where?

Side note: happy to make a new issue, but it would be nice if there was a bit more consistency to the method names here. I'm going to attempt to propose a solution, but I'm not trying to cause more trouble than its worth either. :)

## Creation
```
Index::create_in_ram(schema) // uses ram dir -> Index::from_directory
Index::create(path, schema) // uses mmap dir -> Index::from_directory
Index::create_from_tempdir(schema) //uses mmap dir -> Index::from_directory
Index::from_directory(managed_directory, schema) // a little more universal - but creates a new setup
```

- Maybe? make from directory take a directory and wrap it in managed ourselves?
- Does managed need to be exposed outside of the crate?
- `Index::from_directory` -> `Index::create_directory` ?

## Opening
```
Index::open(path) // uses mmap
Index::open_directory(directory) // new one
```

Ok, random ideas aside. Hopefully the meat of the PR is what you were looking for.

### DIFF
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -103,6 +103,13 @@ impl Index {
         Ok(index)
     }
 
+    /// Open the index using the provided directory
+    pub fn open_directory<D: Directory>(directory: D) -> Result<Index> {
+        let directory = ManagedDirectory::new(directory)?;
+        let metas = load_metas(&directory)?;
+        Index::create_from_metas(directory, &metas)
+    }
+
     /// Create a new index from a directory.
     pub fn from_directory(mut directory: ManagedDirectory, schema: Schema) -> Result<Index> {
         save_new_metas(schema.clone(), 0, directory.borrow_mut())?;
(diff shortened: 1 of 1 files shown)


fulmicoton:
I agree with the concern. Also I wonder if we could entirely hide the concept of ManagedDirectory from all public API, remove the `impl Directory` and add the method to `ManagedDirectory` directly instead.

That would require a bit more thinking though.

fulmicoton:
Can you open a new ticket?

fulmicoton:
(merging this for the moment.)

drusellers:
done: https://github.com/tantivy-search/tantivy/issues/286

PR #292, opened 2018-05-06
Implement StopWords Filter

drusellers:
As I look through the codebase for low hanging fruit. :) I thought implementing a stop word filter would be a nice addition.

I'm currently using a `Vec<String>` which may or may not be ok, and then simply calling `Vec::contains` to check for the presence of a stop word, since the `token.text` is a String anyways.

The basic test is working, but there may be a better approach with regards to `str` and `String` usage etc.

- [x] What should the data structure of `words` be?
- [ ] provide a common list of stop words for various languages (does tantivy ship with these like Lucene?)
- [ ] do we want to provide a way to slurp in a new line delimited file?


src/tokenizer/stop_word_filter.rs
```diff
@@ -0,0 +1,86 @@
...
+    self.tail.token()
+  }
+
+  fn token_mut(&mut self) -> &mut Token {
+    self.tail.token_mut()
+  }
+
+  fn advance(&mut self) -> bool {
+    loop {
+      if self.tail.advance() {
```
fulmicoton: that's a `while self.tail.advance()` loop.

src/tokenizer/alphanum_only.rs
```diff
@@ -1,3 +1,28 @@
...
+//!
+//! let mut stream = tokenizer.token_stream("hello there");
+//! // is none because the raw filter emits one token that
+//! // contains a space
+//! assert!(stream.next().is_none());
+//!
+//! let tokenizer = SimpleTokenizer
+//!   .filter(AlphaNumOnlyFilter);
+//!
+//! let mut stream = tokenizer.token_stream("hello there 💣");
```
fulmicoton, line 19: Awesome thank you!

fulmicoton:
Sweet you are on a roll. I added a bunch of comments.

I don't do this everytime but, you can make release management nicer by adding a line for your feature in the `CHANGELOG.md` for version 0.6.0... And don't forget your username :)

src/tokenizer/stop_word_filter.rs
```diff
@@ -0,0 +1,86 @@
...
+//! assert_eq!(stream.next().unwrap().text, "crafty");
+//! assert!(stream.next().is_none());
+//! # }
+//! ```
+use super::{Token, TokenFilter, TokenStream};
+
+/// `TokenFilter` that removes stop words from a token stream
+#[derive(Clone)]
+pub struct StopWordFilter {
+  words: Vec<String>,
```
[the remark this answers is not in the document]
drusellers: YAS, much better!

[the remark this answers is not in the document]
fulmicoton: Make sure to not use the default Hash too... The FnvHasher should be ok. In the future we will probably want to have something a tad smarter to avoid hashing tokens twice (once for StopWords and once to append to the posting lists.)

src/tokenizer/stop_word_filter.rs
drusellers: ok, cool. fixed a few others as well.

src/tokenizer/stop_word_filter.rs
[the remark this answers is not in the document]
drusellers: Ok, added the `fnv` crate and now that is rocking.

src/tokenizer/alphanum_only.rs
drusellers: I'd like to do something like this for each one. Helps me to understand the code as well. :)

drusellers:
> Sweet you are on a roll.

Well, I tried to look at Synonym's but that proved to be a bit too much. So, just trying to go through what I know about search, and implement whatever I can. At some point I'll get to harder stuff, and have to slow down. ;p

fulmicoton:
@drusellers Let me know if you want help finding tickets. I think there are a bunch of interesting ones!

drusellers:
@fulmicoton feel free to assign a few to me. I'm about done today, but can pick back up later this week.

drusellers:
Ok, I think this one is ready. Holler if you want anything else done.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ Tantivy 0.6
 - Add &[u8] fastfield for associating arbitrary bytes to each document (@jason-wolfe) (#270)
     - Completely uncompressed
     - Internally: One u64 fast field for indexes, one fast field for the bytes themselves.
+- Add NGram token support (@drusellers)
```
fulmicoton, line 10: oh, I thought I added that one before. Thanks for spotting its absence.

PR #293, opened 2018-05-06
Alive docs iterator

### LINKED ISSUE #287: Add an iterator on `SegmentReader` to browse through non-deleted document ids.
As pointed out by @jason-wolfe in CR...

We often have the following pattern 

```rust 

for doc in 0..segment.max_doc() {
   if !segment.is_deleted(doc) {
       ...
   }
}
```

It would be nice to have an iterator to do this 

```rust
for doc in segment.doc_ids_alive() {}
```

drusellers:
This one has a bit of rustfmt in it. but it is isolated to a single commit

This PR closes #287 

If there are other areas that need clean up, just point them out and I'll scoop them up as well.

### DIFF
--- a/src/core/segment_reader.rs
+++ b/src/core/segment_reader.rs
@@ -396,7 +396,7 @@ impl<'a> Iterator for SegmentReaderAliveDocsIterator<'a> {
         while self.reader.is_deleted(self.current) {
             self.current += 1;
 
-            if self.current > self.max_doc {
+            if self.current >= self.max_doc {
                 return None;
             }
         }
@@ -413,26 +413,23 @@ impl<'a> Iterator for SegmentReaderAliveDocsIterator<'a> {
 
 #[cfg(test)]
 mod test {
-    use core::{Index, SegmentReader};
+    use core::Index;
     use schema::*;
 
     #[test]
     fn test_iterator() {
         let mut schema_builder = SchemaBuilder::new();
-        schema_builder.add_text_field("name", TEXT);
+        schema_builder.add_text_field("name", TEXT | STORED);
         let schema = schema_builder.build();
         let index = Index::create_in_ram(schema.clone());
         let name = schema.get_field("name").unwrap();
 
         {
-            let mut index_writer = index.writer(50_000_000).unwrap();
-            let mut doc1 = Document::default();
-            doc1.add_text(name, "tantivy");
-            index_writer.add_document(doc1);
-
-            let mut doc2 = Document::default();
-            doc2.add_text(name, "horse");
-            index_writer.add_document(doc2);
+            let mut index_writer = index.writer_with_num_threads(1, 3_000_000).unwrap();
+            index_writer.add_document(doc!(name => "tantivy"));
+            index_writer.add_document(doc!(name => "horse"));
+            index_writer.add_document(doc!(name => "jockey"));
+            index_writer.add_document(doc!(name => "cap"));
 
             // we should now have one segment with two docs
             index_writer.commit().unwrap();
@@ -441,26 +438,22 @@ mod test {
         {
             let mut index_writer2 = index.writer(50_000_000).unwrap();
             index_writer2.delete_term(Term::from_field_text(name, "horse"));
+            index_writer2.delete_term(Term::from_field_text(name, "cap"));
 
             // ok, now we should have a deleted doc
             index_writer2.commit().unwrap();
         }
 
-        let segments = index.searchable_segments().unwrap();
-        assert_eq!(segments.len(), 2);
+        index.load_searchers().unwrap();
+        let searcher = index.searcher();
 
-        // ok, we have two segments, lets grab the first segment and walk it
-        let sr = SegmentReader::open(&segments[0]).unwrap();
-        let mut it = sr.doc_ids_alive();
-        let f = it.next();
-        assert!(f.is_some());
-        assert_eq!(f.unwrap(), 0);
-        assert!(it.next().is_none());
+        let mut doc_count = 0;
+        for sr in searcher.segment_readers() {
+            for _alive in sr.doc_ids_alive() {
+                doc_count += 1;
+            }
+        }
 
-        // ok, the second segment should have no docs
-        let sr2 = SegmentReader::open(&segments[1]).unwrap();
-        let mut it2 = sr2.doc_ids_alive();
-        assert_eq!(it2.next().unwrap(), 1);
-        assert!(it2.next().is_none());
+        assert_eq!(doc_count, 2)
     }
 }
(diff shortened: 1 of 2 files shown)


fulmicoton:
I think you have now permissions to assign yourself a ticket. Make sure to do so, to make sure we don't have two people working on the same thing.

fulmicoton:
Ideally do not run `cargo fmt`. Instead ping to remind me to do so in master.

src/indexer/merger.rs
```diff
@@ -441,20 +426,17 @@ impl IndexMerger {
...
         }
 
-        let mut term_ord_mapping_opt =
-            if *field_type == FieldType::HierarchicalFacet {
-                Some(TermOrdinalMapping::new(max_term_ords))
-            } else {
-                None
-            };
-
+        let mut term_ord_mapping_opt = if *field_type == FieldType::HierarchicalFacet {
```
fulmicoton: I wonder if we can configure rustfmt to avoid this?

src/core/segment_reader.rs
```diff
@@ -335,10 +335,60 @@ impl SegmentReader {
...
 
 impl fmt::Debug for SegmentReader {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "SegmentReader({:?})", self.segment_id)
     }
 }
+
+/// Implements the iterator trait to allow easy iteration
+/// over non-deleted ("alive") DocIds in a SegmentReader
+pub struct SegmentReaderAliveDocsIterator<'a> {
```
fulmicoton, line 370: Please add a simple unit test or two for this iterator.

fulmicoton:
@drusellers I don't know if you saw that but I wrote a bunch of comment in the CR. Most of them are not really actionnable, but I need a `// TODO` line and a simple unit test before merging this.

drusellers:
Roger that! I should be able to square this one away tonight. :)

drusellers:
> Ideally do not run cargo fmt. Instead ping to remind me to do so in master.

My editor just applies it when I save. So, i'm not doing anything by hand. I can cherry pick my commits out from the rustfmt and give you a clean PR i'm pretty sure. I can verify all is well in another editor that I don't use to write rust.

src/indexer/merger.rs
drusellers: Test has been added but I'm not super happy with it. 

1. Should I do this much setup, or is there an easier :tm: way?
2. Since I don't understand some of the internals of the meta, etc. I may be missing an edge case.

drusellers:
Once this looks good, I'll rebase my commits and get rid of the rustfmt changes. :)

src/core/segment_reader.rs
```diff
@@ -393,3 +393,57 @@ impl<'a> Iterator for SegmentReaderAliveDocsIterator<'a> {
...
+    #[test]
+    fn test_iterator() {
+        let mut schema_builder = SchemaBuilder::new();
+        schema_builder.add_text_field("name", TEXT);
+        let schema = schema_builder.build();
+        let index = Index::create_in_ram(schema.clone());
+        let name = schema.get_field("name").unwrap();
+
+        {
+            let mut index_writer = index.writer(50_000_000).unwrap();
```
fulmicoton: `index.writer_with_num_threads(1, 3_000_000)`...

`.writer(...)` is sniffing your number of cores to choose a number of threads and splits your heap accordingly. If the resulting heap is too small it will fail, so your test might not pass on a crazy high-end CPU.

On one thread, you can get away with using only 3MB.

src/core/segment_reader.rs
```diff
@@ -393,3 +393,57 @@ impl<'a> Iterator for SegmentReaderAliveDocsIterator<'a> {
...
+            let mut doc1 = Document::default();
+            doc1.add_text(name, "tantivy");
```
fulmicoton: Alternatively you can use the following syntax
`index_writer.add_document(doc!(name=>"tantivy"));`

You can leave it that way it is fine as is.

src/core/segment_reader.rs
```diff
@@ -393,3 +393,57 @@ impl<'a> Iterator for SegmentReaderAliveDocsIterator<'a> {
...
+        {
+            let mut index_writer2 = index.writer(50_000_000).unwrap();
+            index_writer2.delete_term(Term::from_field_text(name, "horse"));
+
+            // ok, now we should have a deleted doc
+            index_writer2.commit().unwrap();
+        }
+
+        let segments = index.searchable_segments().unwrap();
+        assert_eq!(segments.len(), 2);
```
fulmicoton: It is actually possible to end up with 1 segments with this setting.

src/core/segment_reader.rs
drusellers: neat! Changed.

src/core/segment_reader.rs
drusellers: changed

src/core/segment_reader.rs
drusellers: ok. i'll write something more robust. :) that was from me sniffing around and not really applicable to the test. :)

src/core/segment_reader.rs
```diff
@@ -393,3 +393,57 @@ impl<'a> Iterator for SegmentReaderAliveDocsIterator<'a> {
...
+
+        // ok, we have two segments, lets grab the first segment and walk it
+        let sr = SegmentReader::open(&segments[0]).unwrap();
```
fulmicoton: I'd prefer you avoid calling `SegmentReader::open` and use the user path instead.

Call `.load_searcher()`, then acquire a searcher using
`index.searcher()`.

The readers are then accessible via  `searcher.segment_reader(0)` and `searcher.segment_reader(1)`.

src/core/segment_reader.rs
fulmicoton: It should be quite deterministic with the 1 thread setting.

fulmicoton: Ok so to sum it up, as it might help you understanding the behavior.

When you use several threads, each thread is building its own little segment by picking off docs from the same queue. When you call commit all of the workers who have processed at least one document flush their work to disk and create one segment.

Also if you add a lot of document without committing, the worker have a RAM budget and will start flushing to disk once their RAM budget is exhausted.

fulmicoton:
LGTM. Can you extrude remove the rustfmt stuff?

drusellers:
Sure can. I’ll do that in the morning!

drusellers:
Ok, rebased on master, squared away all merge conflicts, etc.

fulmicoton:
Thanks!

fulmicoton:
I modified the unit test a tad to actually test for the doc ids.
https://github.com/tantivy-search/tantivy/commit/96b2c2971ea8538644844fb103ba100a88a7af62

Once you switch to single thread, the doc id / segment allocation is pretty much deterministic. They get their doc ids in order.

drusellers:
even better!

PR #295, opened 2018-05-08
#294: Make fieldnorm module public, add documentation

### LINKED ISSUE #294: Make fieldnorm module public
Necessary for the term-wise index rewriting that I'm currently doing.

Will add documentation to not have warnings from it.

jason-wolfe:
Closes #294

### DIFF
--- a/src/fieldnorm/writer.rs
+++ b/src/fieldnorm/writer.rs
@@ -6,13 +6,20 @@ use schema::Field;
 use schema::Schema;
 use std::io;
 
+/// The `FieldNormsWriter` is in charge of tracking the fieldnorm byte
+/// of each document for each field with field norms.
+///
+/// `FieldNormsWriter` stores a Vec<u8> for each tracked field, using a
+/// byte per document per field.
 pub struct FieldNormsWriter {
     fields: Vec<Field>,
     fieldnorms_buffer: Vec<Vec<u8>>,
 }
 
 impl FieldNormsWriter {
-    pub fn fields_with_fieldnorm(schema: &Schema) -> Vec<Field> {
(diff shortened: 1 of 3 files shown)


src/fieldnorm/writer.rs
```diff
@@ -6,13 +6,20 @@ use schema::Field;
...
 pub struct FieldNormsWriter {
     fields: Vec<Field>,
     fieldnorms_buffer: Vec<Vec<u8>>,
 }
 
 impl FieldNormsWriter {
-    pub fn fields_with_fieldnorm(schema: &Schema) -> Vec<Field> {
+    /// Returns the fields that should have field norms computed
+    /// according to the given schema.
+    pub(crate) fn fields_with_fieldnorm(schema: &Schema) -> Vec<Field> {
```
jason-wolfe, line 22: Just want to draw attention to the `pub(crate)` bit here @fulmicoton

src/fieldnorm/writer.rs
```diff
@@ -6,13 +6,20 @@ use schema::Field;
 use schema::Schema;
 use std::io;
 
+/// The `FieldNormsWriter` is in charge of tracking the fieldnorm byte
+/// of each document for each field with field norms.
+///
+/// `FieldNormsWriter` stores a Vec<u8> for each tracked field, using a
```
fulmicoton: I totally missed that. I'm not fond at all of the `Vec<Vec<u8>>`. We should probably open another issue.

src/fieldnorm/writer.rs
fulmicoton: good idea.

PR #297, opened 2018-05-09
Make it possible to stream the terms matching an Automaton

### LINKED ISSUE #273: Add a `.search(automaton: fst::Automaton)` to the `TermDictionary`
This is required for `#219` (sublime text like search) and `#272` (FuzzyTermQuery).

Note that while similar, they are quite different in that 
- in `#219`, the user probably want the list of terms that matched, possibly sorted by a function of the doc frequencies and some function of the query and the term`. (See the ticket comments for some links)
- in `#272`, this is effectively a query. The user wants the list of ranked documents.

The method already exists is already implemented in the `fst` crate.
Unfortunately it does not give access to the automaton state in the resulting `Streamer`.

I [opened a ticket](https://github.com/BurntSushi/fst/issues/60) in the `fst` crate to discuss whether this feature should be added upstream.
Otherwise, the `raw` module of the `fst` crate exposes enough public API to reimplement the intersection in tantivy if required.

drusellers:
Closes #273

Ok, so I've started down the path, but I'm for sure lost.

1. Not sure how I would test with Automaton - still reading up on the fst code there
2. Not sure how to use the fst StreamBuilder, but it feels close on the code implementation, but I have compiler issues, which could just be about spreading the generic type love around more.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,3 +72,6 @@ required-features = ["mmap"]
 
 [[example]]
 name = "custom_tokenizer"
+
+[[example]]
+name = "fuzzy_search"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,6 +72,3 @@ required-features = ["mmap"]
 
 [[example]]
 name = "custom_tokenizer"
-
-[[example]]
-name = "fuzzy_search"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,6 @@ lazy_static = "0.2.1"
 tinysegmenter = "0.1.0"
 regex = "0.2"
 fst = {version="0.3", default-features=false}
-fst-levenshtein = "0.2"
 atomicwrites = {version="0.1", optional=true}
 tempfile = "2.1"
 log = "0.3.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,8 @@ byteorder = "1.0"
 lazy_static = "0.2.1"
 tinysegmenter = "0.1.0"
 regex = "0.2"
-fst = {version="0.2", default-features=false}
+fst = {version="0.3", default-features=false}
+fst-levenshtein = "0.2"
 atomicwrites = {version="0.1", optional=true}
 tempfile = "2.1"
 log = "0.3.6"
@@ -28,6 +29,7 @@ serde_derive = "1.0"
 serde_json = "1.0"
 num_cpus = "1.2"
 itertools = "0.5.9"
+levenshtein_automata = {version="0.1", features=["fst_automaton"]}
 lz4 = "1.20"
 bit-set = "0.4.0"
 uuid = { version = "0.6", features = ["v4", "serde"] }
--- a/src/directory/mmap_directory.rs
+++ b/src/directory/mmap_directory.rs
@@ -40,9 +40,11 @@ fn open_mmap(full_path: &Path) -> result::Result<Option<MmapReadOnly>, OpenReadE
         // instead.
         return Ok(None);
     }
-    MmapReadOnly::open(&file)
-        .map(Some)
-        .map_err(|e| From::from(IOError::with_path(full_path.to_owned(), e)))
+    unsafe {
+        MmapReadOnly::open(&file)
+            .map(Some)
+            .map_err(|e| From::from(IOError::with_path(full_path.to_owned(), e)))
+    }
 }
 
 #[derive(Default, Clone, Debug, Serialize, Deserialize)]
--- a/src/directory/read_only_source.rs
+++ b/src/directory/read_only_source.rs
@@ -42,7 +42,7 @@ impl ReadOnlySource {
     pub fn as_slice(&self) -> &[u8] {
         match *self {
             #[cfg(feature = "mmap")]
-            ReadOnlySource::Mmap(ref mmap_read_only) => unsafe { mmap_read_only.as_slice() },
+            ReadOnlySource::Mmap(ref mmap_read_only) => mmap_read_only.as_slice(),
             ReadOnlySource::Anonymous(ref shared_vec) => shared_vec.as_slice(),
         }
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,7 +136,6 @@ extern crate combine;
 extern crate crossbeam;
 extern crate fnv;
 extern crate fst;
-extern crate fst_levenshtein;
 extern crate futures;
 extern crate futures_cpupool;
 extern crate itertools;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -140,6 +140,7 @@ extern crate fst_levenshtein;
 extern crate futures;
 extern crate futures_cpupool;
 extern crate itertools;
+extern crate levenshtein_automata;
 extern crate lz4;
 extern crate num_cpus;
 extern crate owning_ref;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,6 +136,7 @@ extern crate combine;
 extern crate crossbeam;
 extern crate fnv;
 extern crate fst;
+extern crate fst_levenshtein;
 extern crate futures;
 extern crate futures_cpupool;
 extern crate itertools;
--- a/src/query/range_query.rs
+++ b/src/query/range_query.rs
@@ -2,6 +2,7 @@ use common::BitSet;
 use core::Searcher;
 use core::SegmentReader;
 use error::ErrorKind;
+use fst::automaton::AlwaysMatch;
 use query::BitSetDocSet;
 use query::ConstScorer;
 use query::{Query, Scorer, Weight};
@@ -215,7 +216,7 @@ pub struct RangeWeight {
 }
 
 impl RangeWeight {
-    fn term_range<'a>(&self, term_dict: &'a TermDictionary) -> TermStreamer<'a> {
+    fn term_range<'a>(&self, term_dict: &'a TermDictionary) -> TermStreamer<'a, AlwaysMatch> {
         use std::collections::Bound::*;
         let mut term_stream_builder = term_dict.range();
         term_stream_builder = match self.left_bound {
--- a/src/query/range_query.rs
+++ b/src/query/range_query.rs
@@ -2,7 +2,6 @@ use common::BitSet;
 use core::Searcher;
 use core::SegmentReader;
 use error::ErrorKind;
-use fst::automaton::AlwaysMatch;
 use query::BitSetDocSet;
 use query::ConstScorer;
 use query::{Query, Scorer, Weight};
@@ -216,7 +215,7 @@ pub struct RangeWeight {
 }
 
 impl RangeWeight {
-    fn term_range<'a>(&self, term_dict: &'a TermDictionary) -> TermStreamer<'a, AlwaysMatch> {
+    fn term_range<'a>(&self, term_dict: &'a TermDictionary) -> TermStreamer<'a> {
         use std::collections::Bound::*;
         let mut term_stream_builder = term_dict.range();
         term_stream_builder = match self.left_bound {
--- a/src/termdict/merger.rs
+++ b/src/termdict/merger.rs
@@ -1,4 +1,3 @@
-use fst::automaton::AlwaysMatch;
 use schema::Term;
 use std::cmp::Ordering;
 use std::collections::BinaryHeap;
@@ -6,7 +5,7 @@ use termdict::TermOrdinal;
 use termdict::TermStreamer;
 
 pub struct HeapItem<'a> {
-    pub streamer: TermStreamer<'a, AlwaysMatch>,
+    pub streamer: TermStreamer<'a>,
     pub segment_ord: usize,
 }
 
@@ -45,7 +44,7 @@ pub struct TermMerger<'a> {
 impl<'a> TermMerger<'a> {
     /// Stream of merged term dictionary
     ///
-    pub fn new(streams: Vec<TermStreamer<'a, AlwaysMatch>>) -> TermMerger<'a> {
+    pub fn new(streams: Vec<TermStreamer<'a>>) -> TermMerger<'a> {
         TermMerger {
             heap: BinaryHeap::new(),
             current_streamers: streams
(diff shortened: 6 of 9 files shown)


src/termdict/termdict.rs
```diff
@@ -101,7 +102,7 @@ fn open_fst_index(source: ReadOnlySource) -> fst::Map {
 /// The term dictionary contains all of the terms in
 /// `tantivy index` in a sorted manner.
 ///
-/// The `Fst` crate is used to assoicated terms to their
+/// The `Fst` crate is used to assoicate terms to their
```
fulmicoton: hehe there is still the typo :)

src/termdict/termdict.rs
```diff
@@ -199,4 +200,13 @@ impl TermDictionary {
     pub fn stream<'a>(&'a self) -> TermStreamer<'a> {
         self.range().into_stream()
     }
+
+    pub fn search<'a, A: Automaton>(&'a self, automaton: A) -> TermStreamerBuilder<'a, A> {
+        use fst::map;
+
+        // given an Automaton and a fst::Map (fst_index)
+        // how can I generate a streambuilder?
```
fulmicoton: I don't understand the question here???

fulmicoton:
Can you push those to the tantivy repo as  feature branches in the future? This will make it possible for me to commit stuff.

src/termdict/termdict.rs
fulmicoton: oh gotcha:
```rust
    pub fn search<'a, A: Automaton>(&'a self, automaton: A) -> TermStreamerBuilder<'a, A> {
        let sb = self.fst_index.search(automaton);
        TermStreamerBuilder::<A>::new(self, sb)
    }
```

src/termdict/termdict.rs
drusellers: ha! that's what I get. ;p

src/termdict/termdict.rs
drusellers: Thank you.

drusellers:
> Can you push those to the tantivy repo as feature branches in the future? This will make it possible for me to commit stuff.

Absolutely. :) Do you have any preferences for branch names in the tantivy repo?

Also, for the life of me I can't get the test to compile. :/

```
error[E0277]: the trait bound `fst_levenshtein::Levenshtein: fst::Automaton` is not satisfied
   --> src/termdict/mod.rs:459:31
    |
459 |         let range = term_dict.search(automaton).into_stream();
    |                               ^^^^^^ the trait `fst::Automaton` is not implemented for `fst_levenshtein::Levenshtein`
```

But it appears to be implemented right here.

https://github.com/BurntSushi/fst/blob/master/fst-levenshtein/src/lib.rs#L142

drusellers:
Should I be using `fst_levenshtein::Levenshtein` to test this code or something else?

fulmicoton:
Alternatively you can use `https://github.com/tantivy-search/levenshtein-automata` or write your own. You still have a compilation error?

drusellers:
Yeah, same error. Then I saw your impl of Levenshtein and wondered if that is a better "Test"

drusellers:
```
error[E0277]: the trait bound `levenshtein_automata::DFA: fst::Automaton` is not satisfied
   --> src/termdict/mod.rs:464:31
    |
464 |         let range = term_dict.search(dfa).into_stream();
    |                               ^^^^^^ the trait `fst::Automaton` is not implemented for `levenshtein_automata::DFA`
```

:(

fulmicoton:
Hmmm.... The only think I can think of, is version of fst conflicting for some reason.
Let me try on my computer.

fulmicoton:
Yeah it compiles on my computer. Maybe try `rm -f Cargo.lock`?

drusellers:
Ok, did a cargo clean, cargo update - no love. I could try updating `fst` to `0.3.0`

drusellers:
@fulmicoton sweet. one step in the right direction. :)

drusellers:
`cargo 1.26.0 (0e7c5a931 2018-04-06)` FWIW

drusellers:
`cargo build` is 👍 
`cargo test` is 👎 

just to be clear

fulmicoton:
Ok `cargo test` fails on my computer as well. Let me check if I can solve that riddle.

fulmicoton:
It works if you change tantivy's `fst` version to `0.3`.

If you want to use `tantivy`'s `levenshtein_automata` crate, be careful. 
The implementation of `fst::Automaton` is guarded by a feature called `fst_automaton`.

So 

`levenshtein_automata = {version="0.1", features=["fst_automaton"]}`

is the right way to add the dependency in tantivy's `Cargo.toml`

Cargo.toml
```diff
@@ -28,6 +29,7 @@ serde_derive = "1.0"
 serde_json = "1.0"
 num_cpus = "1.2"
 itertools = "0.5.9"
+levenshtein_automata = {version="0.1", features=["fst_automaton"]}
```
drusellers, line 31: I should probably make this optional?

Cargo.toml
```diff
@@ -17,7 +17,8 @@ byteorder = "1.0"
 lazy_static = "0.2.1"
 tinysegmenter = "0.1.0"
 regex = "0.2"
-fst = {version="0.2", default-features=false}
+fst = {version="0.3", default-features=false}
+fst-levenshtein = "0.2"
```
drusellers: Should this be optional - maybe behind a `fuzzy` feature gate until the whole feature is done?

src/directory/mmap_directory.rs
```diff
@@ -40,9 +40,11 @@ fn open_mmap(full_path: &Path) -> result::Result<Option<MmapReadOnly>, OpenReadE
...
         return Ok(None);
     }
-    MmapReadOnly::open(&file)
-        .map(Some)
-        .map_err(|e| From::from(IOError::with_path(full_path.to_owned(), e)))
+    unsafe {
+        MmapReadOnly::open(&file)
+            .map(Some)
+            .map_err(|e| From::from(IOError::with_path(full_path.to_owned(), e)))
+    }
```
drusellers, line 47: I don't understand why this change was required when I updated fst -> 0.3 (could be new rust stable too) - but would want someone else's eyes on it.

src/directory/read_only_source.rs
```diff
@@ -42,7 +42,7 @@ impl ReadOnlySource {
     pub fn as_slice(&self) -> &[u8] {
         match *self {
             #[cfg(feature = "mmap")]
-            ReadOnlySource::Mmap(ref mmap_read_only) => unsafe { mmap_read_only.as_slice() },
+            ReadOnlySource::Mmap(ref mmap_read_only) => mmap_read_only.as_slice(),
```
drusellers, line 45: I don't understand why this change was required when I updated fst -> 0.3 (could be new rust stable too) - but would want someone else's eyes on it.

src/directory/mmap_directory.rs
fulmicoton: You mean the `unsafe`  ? There is a comment about it in the fst crate.
It was modified in a non-backward compatible way in `0.3`. As a general rule any change from `0.x` to `0.y` may change backward compatibility.
For version > 1.x, backward compatilibility is respected as long as the major version does not change.

src/termdict/streamer.rs
```diff
@@ -1,18 +1,25 @@
...
+use fst::Automaton;
 use fst::{IntoStreamer, Streamer};
 use postings::TermInfo;
 use termdict::TermOrdinal;
 
-/// `TermStreamerBuilder` is an helper object used to define
+/// `TermStreamerBuilder` is a helper object used to define
 /// a range of terms that should be streamed.
-pub struct TermStreamerBuilder<'a> {
+pub struct TermStreamerBuilder<'a, A>
```
fulmicoton: make `A=AlwaysMatch` a default

src/termdict/streamer.rs
```diff
@@ -58,15 +65,21 @@ impl<'a> TermStreamerBuilder<'a> {
 
 /// `TermStreamer` acts as a cursor over a range of terms of a segment.
 /// Terms are guaranteed to be sorted.
-pub struct TermStreamer<'a> {
+pub struct TermStreamer<'a, A>
```
fulmicoton: make `A=AlwaysMatch` a default

src/termdict/merger.rs
```diff
@@ -1,11 +1,12 @@
+use fst::automaton::AlwaysMatch;
 use schema::Term;
 use std::cmp::Ordering;
 use std::collections::BinaryHeap;
 use termdict::TermOrdinal;
 use termdict::TermStreamer;
 
 pub struct HeapItem<'a> {
-    pub streamer: TermStreamer<'a>,
+    pub streamer: TermStreamer<'a, AlwaysMatch>,
```
fulmicoton: useless once you set the default `make `A=AlwaysMatch`

src/termdict/merger.rs
```diff
@@ -44,7 +45,7 @@ pub struct TermMerger<'a> {
 impl<'a> TermMerger<'a> {
     /// Stream of merged term dictionary
     ///
-    pub fn new(streams: Vec<TermStreamer<'a>>) -> TermMerger<'a> {
+    pub fn new(streams: Vec<TermStreamer<'a, AlwaysMatch>>) -> TermMerger<'a> {
```
fulmicoton: useless once you set the default `make `A=AlwaysMatch`

src/termdict/mod.rs
```diff
@@ -381,7 +383,7 @@ mod tests {
         let source = ReadOnlySource::from(buffer);
         let term_dictionary: TermDictionary = TermDictionary::from_source(source);
 
-        let value_list = |mut streamer: TermStreamer| {
+        let value_list = |mut streamer: TermStreamer<AlwaysMatch>| {
```
fulmicoton: useless once you set the default `make `A=AlwaysMatch`

src/termdict/mod.rs
```diff
@@ -366,6 +366,8 @@ mod tests {
 
     #[test]
     fn test_stream_range_boundaries() {
+        use fst::automaton::AlwaysMatch;
```
fulmicoton: useless once you set the default `make `A=AlwaysMatch`

src/termdict/termdict.rs
```diff
@@ -191,12 +193,19 @@ impl TermDictionary {
 
     /// Returns a range builder, to stream all of the terms
     /// within an interval.
-    pub fn range<'a>(&'a self) -> TermStreamerBuilder<'a> {
+    pub fn range<'a>(&'a self) -> TermStreamerBuilder<'a, AlwaysMatch> {
```
fulmicoton: useless once you set the default `make `A=AlwaysMatch`

src/termdict/termdict.rs
```diff
@@ -191,12 +193,19 @@ impl TermDictionary {
...
         TermStreamerBuilder::new(self, self.fst_index.range())
     }
 
     /// A stream of all the sorted terms. [See also `.stream_field()`](#method.stream_field)
-    pub fn stream<'a>(&'a self) -> TermStreamer<'a> {
+    pub fn stream<'a>(&'a self) -> TermStreamer<'a, AlwaysMatch> {
```
fulmicoton: useless once you set the default `make `A=AlwaysMatch`

src/query/range_query.rs
```diff
@@ -215,7 +216,7 @@ pub struct RangeWeight {
 }
 
 impl RangeWeight {
-    fn term_range<'a>(&self, term_dict: &'a TermDictionary) -> TermStreamer<'a> {
+    fn term_range<'a>(&self, term_dict: &'a TermDictionary) -> TermStreamer<'a, AlwaysMatch> {
```
fulmicoton: useless once you set the default `make `A=AlwaysMatch`

src/directory/mmap_directory.rs
drusellers: Cool, since it was an unsafe change I just wanted someone else to look at it. :)

fulmicoton: (To be really "philosophically" correct here, we should probably make the caller `unsafe`... I think this is borderline enough to not care.)

Cargo.toml
fulmicoton: No that's ok.

src/termdict/termdict.rs
fulmicoton: can you fix the typo:

`assoicate` => `associate`

drusellers:
@fulmicoton thank you for helping me with this. It was cool to see this work. I'm now excited to try and close out some more of the fuzzy related issues. 👍

src/termdict/termdict.rs
```diff
@@ -199,4 +200,11 @@ impl TermDictionary {
     pub fn stream<'a>(&'a self) -> TermStreamer<'a> {
         self.range().into_stream()
     }
+
+    /// Returns a search builder, to stream all of the terms
+    /// within the Automaton
+    pub fn search<'a, A: Automaton>(&'a self, automaton: A) -> TermStreamerBuilder<'a, A> {
+        let sb = self.fst_index.search(automaton);
```
fulmicoton: Let's avoid riddly short names like `sb`. especially where the type is non-trviail to the reader. 
`sb` => `stream_builder`.

fulmicoton:
kudos! We're almost there.

fulmicoton:
@drusellers  That was the hardest part, right there! I'll fill another ticket to get the next step toward a fuzzy search.

src/lib.rs
```diff
@@ -136,9 +136,11 @@ extern crate combine;
 extern crate crossbeam;
 extern crate fnv;
 extern crate fst;
+extern crate fst_levenshtein;
```
fulmicoton: we can remove `fst_levenshtein` now?

Cargo.toml
fulmicoton: we can remove `fst-levenshtein` and keep only `levenshtein-automata`. The latter does the same thing, but faster, better and stronger.

fulmicoton:
We're almost good to go. Can you just remove the deps to `fst-levenshtein`?

Issue #272, opened 2018-04-21
Add a `FuzzyTermQuery`


fulmicoton:
This won't properly address the need for a simple typo-proof search story but that's a beginning.

PR #300, opened 2018-05-11
Add AutomatonWeight to a fuzzy_search module and FuzzyQuery

### LINKED ISSUE #299: Implement the `AutomatonWeight`
In #297, we made it possible to stream through the terms matching an automaton.

We want an `AutomatonWeight` that will take an automaton, browse through these terms, opens the associated inverted list, and appends the document in a `BitSet` and producsest a `BitSetDocSet` as scorer.

A very similar logic is already implemented in the `RangeWeight`.

The `AutomatonWeight` can then serve as the base for a `FuzzyTermQuery` or a `RegexTermQuery` or a #219

drusellers:
Closes #299

So, I was able to review the range code - I "think" I have the way to go about this, where the `AutomatonWeight` has a field of type `Automaton` - but then promptly ran into ownership issues. The other option would be some kind of `Fn<Automaton>` that would build it up? Def looking for some rust guidance here. :)

### DIFF
--- a/src/query/fuzzy_query.rs
+++ b/src/query/fuzzy_query.rs
@@ -1,57 +1,140 @@
 use common::BitSet;
 use core::SegmentReader;
+use fst::automaton::AlwaysMatch;
 use fst::Automaton;
+use levenshtein_automata::{LevenshteinAutomatonBuilder, DFA};
 use query::BitSetDocSet;
 use query::ConstScorer;
 use query::{Query, Scorer, Weight};
 use schema::{Field, IndexRecordOption, Term};
 use termdict::{TermDictionary, TermStreamer};
 use Result;
+use Searcher;
+
+/// A Fuzzy Query matches all of the documents
+/// containing a specific term that is with in
+/// Levenshtein distance
+#[derive(Debug)]
+pub struct FuzzyQuery {
+    term: Term,
+    distance: u8,
+}
+
+impl FuzzyQuery {
+    /// Creates a new Fuzzy Query
+    pub fn new(term: Term, distance: u8) -> FuzzyQuery {
+        FuzzyQuery { term, distance }
+    }
+
+    pub fn specialized_weight(&self) -> AutomatonWeight<DFA> {
+        AutomatonWeight {
+            term: self.term.clone(),
+            field: self.term.field(),
+            builder: self,
+        }
+    }
+}
+
+impl Query for FuzzyQuery {
+    fn weight(&self, _searcher: &Searcher, _scoring_enabled: bool) -> Result<Box<Weight>> {
+        Ok(Box::new(self.specialized_weight()))
+    }
+}
+
+impl AutomatonBuilder<DFA> for FuzzyQuery {
+    fn build_automaton(&self) -> Box<DFA> {
+        let lev_automaton_builder = LevenshteinAutomatonBuilder::new(self.distance, true);
+        let automaton = lev_automaton_builder.build_dfa(self.term.text());
+        Box::new(automaton)
+    }
+}
+
+trait AutomatonBuilder<A>
+where
+    A: Automaton,
+{
+    fn build_automaton(&self) -> Box<A>;
+}
 
 pub struct AutomatonWeight<A>
 where
-  A: Automaton,
+    A: Automaton,
 {
-  field: Field,
-  automaton: A,
+    term: Term,
+    field: Field,
+    builder: AutomatonBuilder<A>,
 }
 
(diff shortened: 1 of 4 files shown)


drusellers:
@fulmicoton if you have some time I would appreciate any ideas for the ownership issue I've walked smack into around the automation. :) Thank you.

fulmicoton:
Thanks for pinging me. I rapidly hacked a solution and pushed it to the branch called `dss/automaton-weight-suggestion`.
This is not "ready as is", but the solution is there.

So... I think it is ok to build the automaton when creating the Weight. `Query` object are rarely called more than once anyway. We don't want to build the automaton for every single segment though.
I rely on the fact that a reference to an `Automaton` is also an `Automaton` thanks to https://docs.rs/fst/0.3.0/fst/automaton/trait.Automaton.html#impl-Automaton.

fulmicoton:
@drusellers ping me if you are still stuck.

drusellers:
Ok, the code is now compiling. But it seems I have something wrong, as I have to set my distance to `2` to match `Japan` to `Japon` when I would expect `1` to be sufficient. Am I understanding the Levenshtein correctly?

fulmicoton:
t found your bug... In your test case, your index uses the default tokenizer to process your text. This includes a lowercaser, so the term in your term dictionary is in fact "japan" as opposed to "Japan".

In practise, users of your automaton need to make sure to use the same tokenizer to process queries. 

You can fix your text by lowercasing "Japon" to "japon"

drusellers:
HAHAHA, yes. Facepalm. I need to start seeing `J` as being different from `j` - good stuff, thank you.

drusellers:
Ok, this one is green. Assuming you are happy with the code. I'm ready for the next task. :)

drusellers:
Closes #272

src/query/fuzzy_query.rs
```diff
@@ -0,0 +1,144 @@
...
+use query::BitSetDocSet;
+use query::ConstScorer;
+use query::{Query, Scorer, Weight};
+use schema::{Field, IndexRecordOption, Term};
+use termdict::{TermDictionary, TermStreamer};
+use Result;
+use Searcher;
+
+/// A Fuzzy Query matches all of the documents
+/// containing a specific term that is with in
```
fulmicoton: `within` in one word

src/query/fuzzy_query.rs
```diff
@@ -0,0 +1,144 @@
...
+/// Levenshtein distance
+#[derive(Debug, Clone)]
+pub struct FuzzyQuery {
```
fulmicoton: Please renamed it to `FuzzyTermQuery` to make it clear that it works over a single term. This is to avoid confusion with how we will try and handle queries with multiple terms.

src/query/fuzzy_query.rs
```diff
@@ -0,0 +1,144 @@
...
+    term: Term,
+    distance: u8,
+}
+
+impl FuzzyQuery {
+    /// Creates a new Fuzzy Query
+    pub fn new(term: Term, distance: u8) -> FuzzyQuery {
```
fulmicoton: add options to handle 
- `Damerau` (that is consider transposition as having a cost of 1)
- `prefix queries`

Both are options available in the LevenshteinBuilder.

src/query/fuzzy_query.rs
```diff
@@ -0,0 +1,144 @@
...
+
+impl Query for FuzzyQuery {
+    fn weight(&self, _searcher: &Searcher, _scoring_enabled: bool) -> Result<Box<Weight>> {
+        Ok(Box::new(self.specialized_weight()))
+    }
+}
+
+impl AutomatonBuilder<DFA> for FuzzyQuery {
+    fn build_automaton(&self) -> Box<DFA> {
+        let lev_automaton_builder = LevenshteinAutomatonBuilder::new(self.distance, true);
```
fulmicoton: The `LevenshteinAutomatonBuilder` can be cached with a `lazy_static!` HashMap

src/query/fuzzy_query.rs
```diff
@@ -0,0 +1,144 @@
...
+        let automaton = lev_automaton_builder.build_dfa(self.term.text());
+        Box::new(automaton)
+    }
+}
+
+trait AutomatonBuilder<A>
```
fulmicoton: Can we put this guy in its own file?

drusellers:
Ok, all but the lazy static changes have been made. :)

drusellers:
Ok, ready for the next review! :)

Issue #302, opened 2018-05-17
Improve ergonomics of Query


jason-wolfe:
There's a few things that could make Query more usable for programmatic manipulations.

My particular desired use-cases are:

1. Be able to nicely display a query and its terms. Having `downcast::Any` implemented on `Query` would help with this. `fmt::Debug` is almost enough, except that `Term`s display raw byte slices and `Field` numbers, neither of which is particularly readable.
2. Be able to debug why queries match / do not match a given document. So for a complicated query I want to see which of its clauses matched a given document, and evaluate this recursively.

The changes to `Query` necessary to do this are:

1. Add `downcast::Any` for checking if we have a particular type. Necessary for recursive pattern matching without turning everything into enums.
2. Add a `fn box_clone(&self) -> Box<Query>` for turning a `&Query` into a `Box<Query>`, in order to deconstruct queries and put them back together into other queries.
3. Edit the common default built-in `Query` impls to support introspection. This means being able to see the clauses of a `BooleanQuery`, the `Term` of a `TermQuery`, and the `Field` + `Term`s of a `PhraseQuery`.

Will first add the capabilities to `Query` necessary to make this happen, and then after API discussion add the other two things. I imagine `println!` is not the right target :smile:.

fulmicoton:
This is closed isn't it? @jason-wolfe

PR #303, opened 2018-05-17
Add box_clone() and downcast::Any to Query

jason-wolfe:
This partially completes the work of #302 .


jason-wolfe:
Oops this needs docs.

src/query/boolean_query/boolean_query.rs
```diff
@@ -23,6 +23,16 @@ pub struct BooleanQuery {
     subqueries: Vec<(Occur, Box<Query>)>,
 }
 
+impl Clone for BooleanQuery {
+    fn clone(&self) -> Self {
```
fulmicoton, line 27: Alternatively we could have just made it `Arc<_>`.

PR #304, opened 2018-05-17
Optimize segment readers alive doc iterator

### LINKED ISSUE #296: Optimize Segment Readers Alive Doc Iterator
Noted in PR #293

Use the same trick as what is used in the [BitSetDocSet]
(https://github.com/tantivy-search/tantivy/blob/78673172d001a0c4c7c73cdd3d9923fc43fc0312/src/query/bitset/mod.rs)
which relies on TinySet

In a nutshell : instead of looping on the bitset to find the next thing alive, it uses a special cpu instruction to get
the next bit set to 1 in a u64. The instruction is not free, but except if your bitset is EXTREMELY saturated it is generally faster.

https://github.com/tantivy-search/tantivy/pull/293#discussion-diff-186488306R373

kphelps:
I took a stab at tackling https://github.com/tantivy-search/tantivy/issues/296. The `DeleteBitSet` now builds a slice of TinySets, then the `SegmentReaderAliveDocsIterator` leverages the tinysets to iterate through the alive documents more efficiently. As I'm not very familiar with the codebase, there may have been a cleaner way to do some parts of this - feedback is appreciated!

Closes #296

### DIFF
--- a/src/core/segment_reader.rs
+++ b/src/core/segment_reader.rs
@@ -1,5 +1,6 @@
 use common::CompositeFile;
 use common::HasLen;
+use common::TinySet;
 use core::InvertedIndexReader;
 use core::Segment;
 use core::SegmentComponent;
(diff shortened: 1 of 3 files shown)


src/fastfield/delete.rs
```diff
@@ -62,6 +79,26 @@ impl DeleteBitSet {
...
+            .cloned()
+            .position(|tinyset| !tinyset.is_empty())
+            .map(|delta_bucket| bucket + delta_bucket as u32)
+    }
+
+    /// Returns the tiny bitset representing the
+    /// the set restricted to the number range from
+    /// `bucket * 64` to `(bucket + 1) * 64`.
+    pub(crate) fn tinyset(&self, bucket: u32) -> TinySet {
+        self.tinysets[bucket as usize]
```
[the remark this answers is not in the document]
fulmicoton: (You can double check that rapidly on rust playground. It makes it possible to import the byteorder crate.)

src/fastfield/delete.rs
```diff
@@ -45,8 +47,23 @@ impl DeleteBitSet {
             .iter()
             .map(|b| b.count_ones() as usize)
             .sum();
+        let mut tinysets = Vec::new();
+        let mut value_accumulator = 0u64;
+        let byte_count = data.as_slice().len();
+        for index in 0..byte_count {
+            value_accumulator <<= 1;
+            value_accumulator += data.as_slice()[byte_count - index - 1] as u64;
```
fulmicoton, line 55: Using the ByteOrder crate would probably be more readable and more efficient as described below....

src/core/segment_reader.rs
```diff
@@ -369,45 +370,64 @@ impl fmt::Debug for SegmentReader {
...
-
-        // move down the chain
-        self.current += 1;
+impl<'a> Iterator for SegmentReaderAliveDocsIterator<'a> {
+    type Item = DocId;
 
-        result
+    fn next(&mut self) -> Option<Self::Item> {
+        self.next_possible_doc().and_then(|doc_id| {
+            if doc_id >= self.max_doc {
```
fulmicoton, line 425: why / when does that happen?

src/fastfield/delete.rs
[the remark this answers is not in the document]
fulmicoton: Padding the delete bit set to make sure we have a multiple of 8 bytes is ok..

src/core/segment_reader.rs
fulmicoton: This happens because of the `complement` I assume. We probably want to fix this corner case when we create the tinyset or even more upstream, in the delete bitset representation. (by padding with 1 or switching the meaning of alive `=1`)

src/fastfield/delete.rs
```diff
@@ -62,6 +79,26 @@ impl DeleteBitSet {
...
+            .map(|delta_bucket| bucket + delta_bucket as u32)
+    }
+
+    /// Returns the tiny bitset representing the
+    /// the set restricted to the number range from
+    /// `bucket * 64` to `(bucket + 1) * 64`.
+    pub(crate) fn tinyset(&self, bucket: u32) -> TinySet {
+        self.tinysets[bucket as usize]
+    }
 }
```
fulmicoton, line 103: This needs a benchmark with different density to double check that this is effectively faster.

src/core/segment_reader.rs
kphelps: This if statement is also needed in the case where we are iterating through a segment that has 0 deleted documents. I looked into padding the delete bitset or toggling the semantics of the bitset, but I think that makes it impossible to determine how many deleted documents there are from the bitset alone.

fulmicoton: > This if statement is also needed in the case where we are iterating through a segment that has 0 deleted documents

How come? That probably means that some other object in here is not what it pretends to be.

fulmicoton:
@kphelps any update on this?

kphelps:
Sorry, I haven't had time to get back to this. I'll re-open it when I have some free time to tackle your feedback

fulmicoton:
Thanks for the update!

PR #309, opened 2018-05-18
Expose parameters of RangeQuery for external usage

jason-wolfe:
More things related to #302 

Sorry this functionality has been so spread out. I kept missing stuff until they came up in my personal use-case :blush:


jason-wolfe:
@fulmicoton: Note that this changes a bunch more than the title implies.

PR #310, opened 2018-05-20
Issue/257

fulmicoton:


src/store/compression_lz4.rs
```diff
@@ -0,0 +1,19 @@
+extern crate lz4;
```
drusellers, line 1: The goal (of putting the extern crate call here rather than lib.rs) being that we only "include the crate" in the library if that feature was enabled. Correct?

src/store/mod.rs
```diff
@@ -38,6 +38,22 @@ mod writer;
 pub use self::reader::StoreReader;
 pub use self::writer::StoreWriter;
 
+trait Compression {
+    fn compress(uncompressed: &[u8], compress: &mut Vec<u8>);
+    fn decompress(compressed: &[u8], decompressed: &mut Vec<u8>);
+}
```
drusellers: Is this to allow an eventual "plug and play"? I didn't see it being used.

src/store/mod.rs
```diff
@@ -38,6 +38,22 @@ mod writer;
...
+
+#[cfg(feature="lz4")]
+mod compression_lz4;
+#[cfg(feature="lz4")]
+use self::compression_lz4::*;
+
+#[cfg(not(feature="lz4"))]
+mod compression_snap;
+#[cfg(not(feature="lz4"))]
+use self::compression_snap::*;
```
drusellers, line 50: Cool, so by default we get [snappy](http://google.github.io/snappy/), but someone can turn on lz4 with the feature flag. Our snappy lib is a pure rust implementation, but `lz4` requires a C library. Sweet.

drusellers (approved):

src/store/compression_lz4.rs
fulmicoton: yes... Alternative I could have added a `#[cfg()]` thing before hte statement in lib.rs.

src/store/mod.rs
fulmicoton: Removing this.

src/store/mod.rs
fulmicoton: yes... LZ4 is probably a nicer compresson/performance trade-off but not by a large margin.
This makes it easier to compile tantivy to web assembly.

Issue #286, opened 2018-05-03
Index API Consistency


drusellers:
While working on #279 I noticed a wee bit of inconsistency in the API that I thought would be good to clean up.

## Creation
```
Index::create_in_ram(schema) // uses ram dir -> Index::from_directory
Index::create(path, schema) // uses mmap dir -> Index::from_directory
Index::create_from_tempdir(schema) //uses mmap dir -> Index::from_directory
Index::from_directory(managed_directory, schema) // a little more universal - but creates a new setup
```

- Maybe? make from directory take a directory and wrap it in managed ourselves?
- Does managed need to be exposed outside of the crate?
- `Index::from_directory` -> `Index::create_directory` ?

## Opening
```
Index::open(path) // uses mmap
Index::open_directory(directory) // new one
```

It was additionally noted:

> I agree with the concern. Also I wonder if we could entirely hide the concept of ManagedDirectory from all public API, remove the impl Directory and add the method to ManagedDirectory directly instead.
>
> That would require a bit more thinking though.

fulmicoton:
@drusellers  Can I close this?

drusellers:
Let me take a pass, i think its a quick PR if its not done already.

fulmicoton:
Apparently at least one branch was merged here.

fulmicoton:
Thanks for this!

drusellers:
:+1: now on to the harder regex stuff. ;p

PR #317, opened 2018-06-13
Standardizes the Index::create_* APIs

drusellers:
This pull request addresses the create half of #286

The goal here is to create a consistent api that will flow well in IDE's that support code completion since there are grouped together.

The docs now have them next to each other as well

image

Ok, this is an "important" change since its the API for creating indices. Let me have it with the questions and comments. :)

After this I'll do a PR for the `open_*` API's with what was learned from here.

### DIFF
--- a/examples/custom_tokenizer.rs
+++ b/examples/custom_tokenizer.rs
@@ -61,7 +61,7 @@ fn run_example(index_path: &Path) -> tantivy::Result<()> {
     //
     // This will actually just save a meta.json
     // with our schema in the directory.
-    let index = Index::create(index_path, schema.clone())?;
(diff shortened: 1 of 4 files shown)


fulmicoton:
I was ill at ease with the spec of create (wiping off the previous data) but it has the merit of being in line with the `File::create` API.

In this case, I think the previous files should be garbage collected correctly but we probably want to double check that. Also, it might be a good idea to trigger GC right after the `meta.json` have been overwritten.

fulmicoton (commented):
LGTM

fulmicoton (approved):
LGTM

fulmicoton:
Let's allow ourselves to break public API until 1.0.

I'll merge this right now and look forward your PR for the open API.

PR #318, opened 2018-06-15
Standardizes the Index::open_* APIs

### LINKED ISSUE #286: Index API Consistency

drusellers:
Follow on to #317 
Closes #286 

One positive change of this that I like, is that post merge, you will always have `create` and `open` no matter what features you turn on. And for each feature you turn on, you get a `create_in_*` and a `open_in_*` type methods.

# Open Methods

```
Index::open(path) -> Index::open_in_dir
Index::open(directory) // universal
```

There were fewer of these, so it wasn't as "hard".

# Create Methods

```
Index::create_in_ram(schema)  -> Index::create_in_ram
Index::create(path, schema)  -> Index::create_in_dir
Index::create_from_tempdir(schema) -> Index::create_from_tempdir * I missed this one
Index::from_directory -> hidden as it uses ManagedDirectory
Index::create(D: Directory, schema) // added universal method
```

Question: Should I take this moment to rename `create_from_tempdir` to `create_in_tempdir` to better align?

Ok, this is an "important" change since its the API for creating indices. Let me have it with the questions and comments. :)

### DIFF
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -111,13 +111,13 @@ impl Index {
 
     /// Opens a new directory from an index path.
     #[cfg(feature = "mmap")]
-    pub fn open<P: AsRef<Path>>(directory_path: P) -> Result<Index> {
+    pub fn open_in_dir<P: AsRef<Path>>(directory_path: P) -> Result<Index> {
         let mmap_directory = MmapDirectory::open(directory_path)?;
-        Index::open_directory(mmap_directory)
+        Index::open(mmap_directory)
     }
 
     /// Open the index using the provided directory
-    pub fn open_directory<D: Directory>(directory: D) -> Result<Index> {
+    pub fn open<D: Directory>(directory: D) -> Result<Index> {
         let directory = ManagedDirectory::new(directory)?;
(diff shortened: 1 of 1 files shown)


fulmicoton (approved):
LGTM

PR #319, opened 2018-06-15
Add in simple regex query support

### LINKED ISSUE #84: Handle regular expressions
This should be fairly easy to get something working thanks to the `fst` crate.

It could be exposed in the query parse (not by default) with some nice syntax.
`field:r"http://[\s]+"` for instance ?

drusellers:
closes #84

There may be more efficient ways to do this. But this was my naive approach based on the fuzzy term query. 

Please let me know what you would like changed. I'll be adding the document strings after work, but wanted to get this in your hands.

I did not handle the string parsing aspect mentioned in the issue. If this PR looks good I'd like to take a pass through a more comprehensive "query" parsing from a string so we can do regex, distance, phrase, etc from a single string. If that makes sense to do so. :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ lazy_static = "0.2.1"
 tinysegmenter = "0.1.0"
 regex = "0.2"
 fst = {version="0.3", default-features=false}
+fst-regex = { version="0.2" }
 lz4 = {version="1.20", optional=true}
 snap = {version="0.2"}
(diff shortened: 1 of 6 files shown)


fulmicoton:
I think this looks good. 
Let me have a closer look tomorrow, I need to sleep :)

The story about string parsing was about a different ticket. Sorry for the confusion.

fulmicoton:
LGTM merging

PR #320, opened 2018-06-16
Add some simple doc tests

### LINKED ISSUE #166: Have proper examples / test in the rustdoc
Especially for collectors and queries

drusellers:
Closes #166

The FuzzyTermQuery doc test is failing. I'm not 100% sure why, it seems like it should pass.

FuzzyTermQuery using the standard analyzers I would expect `Diary` with Distance 1 to match 3 of the 4 documents.

### DIFF
--- a/src/collector/top_collector.rs
+++ b/src/collector/top_collector.rs
@@ -44,6 +44,59 @@ impl Eq for GlobalScoredDoc {}
 ///
 /// The implementation is based on a `BinaryHeap`.
 /// The theorical complexity is `O(n log K)`.
+///
+/// ```rust
+/// #[macro_use]
+/// extern crate tantivy;
+/// use tantivy::schema::{SchemaBuilder, TEXT};
+/// use tantivy::{Index, Result, DocId, Score};
+/// use tantivy::collector::TopCollector;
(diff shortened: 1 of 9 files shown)


fulmicoton:
Test are still not passing. The doc id is set to `0` in the test apparently.

drusellers:
@fulmicoton yeah, it used to be two failures, but I found my mistake in the first one. Now I'm down to one failure that I don't fully understand. I need to do some playing around so I can get what I want to have happen. Which is a FuzzyTermSearch on Diary that gets three hits on Diary, Diary, and Dairy.

src/collector/top_collector.rs
```diff
@@ -61,7 +61,7 @@ impl Eq for GlobalScoredDoc {}
 ///     let schema = schema_builder.build();
 ///     let index = Index::create_in_ram(schema);
 ///     {
-///         let mut index_writer = index.writer(3_000_000)?;
+///         let mut index_writer = index.writer_with_num_threads(1, 3_000_000)?;
```
drusellers: DOH! I will remember to do that in tests. Thank you.

PR #323, opened 2018-06-20
Support parsing RangeQuery and AllQuery in Queryparser

### LINKED ISSUE #321: QueryParser support for RangeQuery and AllQuery
Currently there is no way to generate either of these from a `QueryParser`. It would be helpful to have a way to write these down as a human user.

Implemented in (#323):

- Range query as `[low to high]`, `{lowExclusive to highExclusive}` (can also mix and match `[}`, `{]`).
- AllQuery as `*`

Possible future work for RangeQuery for a separate issue: Support for unbounded lower or upper bounds. Maybe via `[X TO]` and `[TO X]` or something (while being sure to still parse `[TO TO TO}`...).

jason-wolfe:
Closes #321

### DIFF
--- a/src/query/query_parser/query_grammar.rs
+++ b/src/query/query_parser/query_grammar.rs
@@ -66,6 +66,7 @@ where
         .map(|(_, expr)| UserInputAST::Not(Box::new(expr)))
         .or((char('+'), parser(leaf)).map(|(_, expr)| UserInputAST::Must(Box::new(expr))))
         .or((char('('), parser(parse_to_ast), char(')')).map(|(_, expr, _)| expr))
+        .or(char('*').map(|_| UserInputAST::All))
         .or(try(parser(range)))
         .or(parser(literal))
         .parse_stream(input)
--- a/src/query/query_parser/query_grammar.rs
+++ b/src/query/query_parser/query_grammar.rs
@@ -1,29 +1,33 @@
 use super::user_input_ast::*;
 use combine::char::*;
 use combine::*;
+use query::query_parser::user_input_ast::UserInputBound;
+
+fn field<I: Stream<Item = char>>() -> impl Parser<Input = I, Output = String> {
+    (letter(), many(satisfy(|c: char| c.is_alphanumeric() || c == '_')))
+        .map(|(s1, s2): (char, String)| format!("{}{}", s1, s2))
+}
+
+fn word<I: Stream<Item = char>>() -> impl Parser<Input = I, Output = String> {
+    many1(satisfy(|c: char| c.is_alphanumeric()))
+}
 
 fn literal<I>(input: I) -> ParseResult<UserInputAST, I>
 where
     I: Stream<Item = char>,
 {
     let term_val = || {
-        let word = many1(satisfy(|c: char| c.is_alphanumeric()));
         let phrase = (char('"'), many1(satisfy(|c| c != '"')), char('"')).map(|(_, s, _)| s);
-        phrase.or(word)
+        phrase.or(word())
     };
 
     let negative_numbers = (char('-'), many1(satisfy(|c: char| c.is_numeric())))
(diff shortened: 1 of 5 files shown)


drusellers:
Unless I missed it, the only thing I spotted was a lack of update to the QueryParser doc strings.

jason-wolfe:
Good catch, thanks for the CR @drusellers !

jason-wolfe:
Writing the docs made me think about default fields. The range query parsing currently does not support them. I guess this should be fixed, but using that seems somehow weird to me.

I'll come back to that tomorrow probably.

jason-wolfe:
Added default field support to `RangeQuery`.  Would appreciate another pass of checks @drusellers, if you don't mind! 😃

src/query/query_parser/query_grammar.rs
```diff
@@ -115,6 +121,7 @@ mod test {
         test_parse_query_to_ast_helper("abc:a b", "(abc:\"a\" \"b\")");
         test_parse_query_to_ast_helper("abc:\"a b\"", "abc:\"a b\"");
         test_parse_query_to_ast_helper("foo:[1 TO 5]", "foo:[\"1\" TO \"5\"]");
+        test_parse_query_to_ast_helper("[1 TO 5]", "[\"1\" TO \"5\"]");
         test_parse_query_to_ast_helper("foo:{a TO z}", "foo:{\"a\" TO \"z\"}");
```
fulmicoton: Wow... why the hell did Lucene decide to use curly brackets for that O_o

jason-wolfe: Parens are already taken. Backwards braces probably cause parsing ambiguities. Better options?

src/query/query_parser/query_parser.rs
```diff
@@ -66,15 +73,26 @@ impl From<ParseIntError> for QueryParserError {
...
+/// * phrase terms: Quoted terms become phrase searches on fields that have positions indexed.
+///   e.g., `title:"Barack Obama"` will only find documents that have "barack" immediately followed
+///   by "obama".
+///
+/// * range terms: Range searches can be done by specifying the start and end bound. These can be
+///   inclusive or exclusive. e.g., `title:[a TO c}` will find all documents whose title contains
+///   a word lexicographically between `a` and `c` (inclusive lower bound, exclusive upper bound).
+///   Inclusive bounds are `[]`, exclusive are `{}`.
+///
+/// *  all docs query: A plain `*` will match all documents in the index.
```
drusellers, line 94: Just a question. Do we support `*:*`

src/query/query_parser/query_parser.rs
```diff
@@ -511,6 +591,31 @@ mod test {
              Term([0, 0, 0, 0, 98])]\"",
             false,
         );
+        test_parse_query_to_logical_ast_helper(
+            "title:[a TO b]",
+            "(Included(Term([0, 0, 0, 0, 97])) TO \
+             Included(Term([0, 0, 0, 0, 98])))",
```
drusellers, line 597: Possible Silly Q: Is it not possible to test the input text against the actual struct's? Rather than a string interpretation? I ask, because refactoring breaks would be caught by the compiler rather than at test run. That is all.

drusellers (commented):
Looks nice and I learned a little bit about the parser stuff! Thank you. I'll leave the full approval to @fulmicoton but otherwise :+1:

src/query/query_parser/query_parser.rs
jason-wolfe: Agreed that this would be great.

src/query/query_parser/query_parser.rs
jason-wolfe: We do not, in the current implementation.

jason-wolfe:
Thanks, @drusellers !

Issue #328, opened 2018-06-28
QueryParser support for unbounded RangeQuery


jason-wolfe:
I like this option from [ElasticSearch syntax](https://www.elastic.co/guide/en/elasticsearch/reference/6.3/query-dsl-query-string-query.html#_ranges):

`field:>=5`
`field:>2`
`field:<3`
`field:<=12`

The other supported syntax is `foo:[* TO whatever]` or `foo:[whatever TO *]`

I guess we can support both. Though I'm also happy to just do the more consistent (second) one.

fulmicoton:
That's cool

jason-wolfe:
Any thoughts about the `:>=`... syntax? I have not currently implemented it.

fulmicoton:
That would be sweet to add it I think. The lucene syntax is a bit ugly.

fulmicoton:
I'd like to close this. The work seems almost entirely done.  
@jason-wolfe @drusellers Can one of you check what might be missing and open a PR?

jason-wolfe:
Looks like #329 got merged long ago so the only thing missing is ElasticSearch syntax.
I consider that optional for completing this work, so this can be closed.

fulmicoton:
closing an opening another ticket to avoid confusion

PR #329, opened 2018-06-28
#328: Support parsing unbounded range queries

jason-wolfe:


CHANGELOG.md
```diff
@@ -1,6 +1,16 @@
-Tantivy 0.6
+Tantivy 0.???
 ==========================
 
+- Added support for parsing AllQuery and RangeQuery via QueryParser
+    - AllQuery: `*`
+    - RangeQuery:
+        - Inclusive `field:[startIncl to endIncl]`
```
fulmicoton, line 7: I think most of this has been released in tantivy0.6 actually but well...

jason-wolfe: Is this true? It didn't look like it when I checked.

fulmicoton: I think you are right. Perfect.

Issue #291, opened 2018-05-06
Add a position length to `Token`


fulmicoton:
It will be useful if we end up merging tokens in a `SynonymTokenFilter` for instance.
The synonym should span over several tokens so that it can still matches a `PhraseQuery`

This functionality will be used in #314 

- [x] Add the property to `Token` 
- [x] Add a new constructor in `PhraseQuery` to accept offset. (this will require updating `PhraseWeight` as well)
- [ ] Fix the QueryParser to pass the `position_length` information and use the above constructor when building a PhraseQuery.

fulmicoton:
Sorry my guidance was sloppy. Your PR looks good. Contrary to what I wrote in the ticket description this does not impact the existing `NgramTokenizer`. (I edited the ticket description accordingly)

So the main client right now is #314. Before merging your change, could you fix PhraseQuery so that it handles `position_length` correctly?

Currently all terms are assumed to have a length of 1, so the constructor just takes a [list of terms](https://github.com/tantivy-search/tantivy/blob/master/src/query/phrase_query/phrase_query.rs#L35).

We need to introduce another constructor `new_with_offsets` which takes a list of terms and their respective position offset: `&[(usize, Term)]` They should be sorted by offset.

In most case, each token should take a length of 1 and the argument query should look straightforward.
`&[(0, term_the), (1, term_three), (2, term_little), (3, term_pigs))]`. 

This will be handful when we face tokens with a token_length larger than 1. For instance, afer introducing `#314`, we can get the exact same result set by running the query as
`&[(0, term_the_three), (2, term_little), (3, term_pigs))]`.


Finally the `QueryParser` need to be fixed in order to use this new constructor. For instance, we will need to read the [token length information from the analyzer]([https://github.com/tantivy-search/tantivy/blob/master/src/query/query_parser/query_parser.rs#L256]) and eventually use the new `new_with_offsets` construction

vigneshsarma:
I have been looking at the code and am a little confused as to how the changes you suggested are supposed to work. To start with, `PhraseQuery` currently has a `Vec<Term>`. Wouldnt this have to be changed to `Vec<(usize, Term)>`, in which case the existing constructor will have to be changed right? Another alternative is to store the offset in `Term`.  Would that work?

I am also unsure of how the two new tasks are related to the original issue. The new tasks seems to be related to querying while the original one seems to be related to Indexing/Indexed data. Am I correct?

fulmicoton:
[...]

PR #337, opened 2018-07-19
add position_length to Token

vigneshsarma:
refer #291 

I am not completely sure this what was required. @fulmicoton what do you think?

### DIFF
--- a/src/query/phrase_query/phrase_query.rs
+++ b/src/query/phrase_query/phrase_query.rs
@@ -24,7 +24,7 @@ use Result;
 #[derive(Clone, Debug)]
 pub struct PhraseQuery {
     field: Field,
-    phrase_terms: Vec<Term>,
+    phrase_terms: Vec<(usize, Term)>,
 }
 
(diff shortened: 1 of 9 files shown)


src/query/phrase_query/phrase_query.rs
```diff
@@ -53,10 +63,11 @@ impl PhraseQuery {
         self.field
     }
 
-    /// The `Term`s in the phrase making up this `PhraseQuery`.
-    pub fn phrase_terms(&self) -> &[Term] {
-        &self.phrase_terms[..]
-    }
+    /// `Term`s in the phrase without the associated offsets.
+    pub fn phrase_terms_without_offsets(&self) -> Vec<Term> {
```
fulmicoton: You can call keep calling it `phrase_terms`

fulmicoton:
We will also need unit test for this. The new phrase query constructor is already pretty interesting as someone can now perform query such as `the <any word> man` .

PR #361, opened 2018-08-01
replace chan with crossbeam-channel

### LINKED ISSUE #360: Deprecate the use of chan
See the [following post from Burntsushi](https://users.rust-lang.org/t/ann-chan-is-deprecated-use-crossbeam-channel-instead/19251)

kardeiz:
Closes #360


kardeiz:
Do you mean the `crossbeam-channel` version? `0.2.3` is the latest version.

fulmicoton:
Can you set it to `0.2` ? Clients who do not include this dependency will download the freshest compatible version automatically (0.2.3 today). Clients who already include this dependency will be less likely to download two different versions.

fulmicoton:
Thank you!

PR #369, opened 2018-08-11
Moved NUM_SEARCHERS into a local variable

### LINKED ISSUE #262: NUM_SEARCHERS should not be hardcoded.
Right now it is hardcoded to 12.
Some CPU have more threads than this.

petr-tik:
fixes #262 

dynamically determined as the number of available cpus.

var name in lowercase (not a constant anymore).

updated it in docstring

### DIFF
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -102,9 +102,7 @@ impl Index {
             directory,
             schema,
             num_searchers: match logical_cores {
-                None => {
-                    num_cpus::get()
-                }
+                None => num_cpus::get(),
                 Some(num) => num,
             },
             searcher_pool: Arc::new(Pool::new()),
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -4,6 +4,7 @@ use schema::Schema;
 use serde_json;
 use std::borrow::BorrowMut;
 use std::fmt;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 use Result;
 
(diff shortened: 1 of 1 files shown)


petr-tik:
@fulmicoton 

If I understand the issue correctly, you mean number of logical cores, not physical ones? 

I grepped the codebase and couldn't find any cases of NUM_SEARCHERS used elsewhere. This made me confident that I could move it inside the function call and make it not static/const. As it stands, Rust doesn't have CTFE anyway, so it will have to be dynamic. Other option includes running a shell script and passing the result as feature/config var to cargo build

If it's the latter, I will change the call to `num_cpus::get_physical()`

What do you think?

fulmicoton:
logical cores is ok yes. This is a good default, but we probably want to let people change this value.
Can you make it a property of `Index`, that defaults to `num_cpus::get()`, and can modified using a setter?

fulmicoton:
no need to modify `create_from_metas()`.

I was just thinking of adding a setter method. 
``` 
fn set_num_searcher(&self, num_searchers: usize ) {
    // ...
}
```

Maybe store it in a `Arc<AtomicUsize>` to make sure it affects all clones of the `Index`.

Comment that it will only be effective after the next call to `.load_searcher()`.

petr-tik:
added a setter method, keen to hear your thoughts on when to best call the setter. Currently, create_from_metas calls `num_cpus::get()` and sets the value. 

Reading from ARC using Relaxed ordering, I couldn't find documentation on it a specific ordering type

Issue #375, opened 2018-08-15
Questions regarding custom weights and stopwords


kornelski:
I’m planning to use tantivy for search of crates on crates.rs. I have more data to include in the search. I’m not sure if tantivy handles it, or do I have to use pre/post-processing to include it.

* overall ranking of a crate. When multiple crates match a query, I'd like my rank/popularity score included in the ranking. Can I index documents as more or less important?

* I have categories. One crate can be in multiple categories, but has a different ranking/relevance in each. Can I tune the ranking (as above) per category (to be used when searching with a category)?

* Example for faceted search would be useful.

* crates have keywords, which should get much more weight/relevance than words from their description or readme. Can there be a separate weighting for the keywords field only? Can they have their own TF-IDF? (e.g. almost every crate mentions "cargo" in its description, so its a noisy word for the readme, but crates tagged "cargo" really mean it).

* Can I set custom stopwords? For crate search "rust" and "rs" are noise. Ideally I'd like stopwords per category (when search is narrowed to a category), since e.g. "ffi" keyword is meaningless in FFI bindings category. I guess I could have a wholly separate index per category.

fulmicoton:
- You can index some kind of importance score in a so called fast field. Right now floating point is not supported. Ideally you'd have to make it a `u64`... Then you'd need a custom scorer. 
It's not difficult and we can help with the implementation. In fact, this should probably be an example in tantivy crate.

- Yes. It requires a bit of coding (outside of tantivy) but anything can be done.

- Roger that. I'll try to add examples.

- Tantivy uses a popular flavor of Tf-Idf called BM25. The presence of a keyword increases the score, but this increase varies depending on 
   - is this considered keyword rare? 
   - is this keyword in a field that is short, (so that you get naturally tags  > title > text) ?
It will do the right thing for `cargo`. 

- You can set custom stop words. That would deserve an example too.

fulmicoton:
Closing

Issue #372, opened 2018-08-14
Add examples


fulmicoton:
Right now we have one large example that does not cover all use case. It would be great to instead, have many small examples covering one single use case.

- [x] deleting documents
- [ ] index payload and resuming indexing a log from a given checkpoint
- [x] integer ranges
- [x] define a custom collector
- [x] facetting
- [x] how to define a custom analyzer (what use case?)
- [x] iterating through documents/position directly
- [x] working with json

petr-tik:
Looks like we are only missing one example showing how to resume indexing from a checkpoint. 

Does tantivy support logs and checkpoints at the moment? If not, are they going to be added?

PR #377, opened 2018-08-17
Add an example of using the stopwords filter

drusellers:
As I was writing this example, I kept asking myself, "How much of the `simple_search` comments do I keep? So, I consider this an RFC as well as a working example of the `StopWordsFilter`. I'm open to all kinds of feedback on we can best demo items in Tantivy examples.

Related to #375

### DIFF
--- a/examples/stop_words.rs
+++ b/examples/stop_words.rs
@@ -1,0 +1,129 @@
+// # Stop Words Example
+//
+// This example covers the basic usage of stop words
+// with tantivy
+//
+// We will :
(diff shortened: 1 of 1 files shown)


drusellers:
related to #372 as well

fulmicoton:
I'm removing a lot of the comments off the other examples, and just referring to the freshly renamed basic search example

fulmicoton:
Merging. I might trim it later! Thanks for that!

PR #381, opened 2018-08-18
Add default EN stopwords to the default analyzer

drusellers:
I'd like to suggest adding a default set of EN stopwords to the standard tokenizers for Tantivy.

1. I'm not sure that `Stopwords::default` is the best method name and very open to renaming.
2. I'm pretty sure my implementation could be quite a bit better memory wise.
3. Is this a sensible add to the default tokenizers?

### DIFF
--- a/src/tokenizer/stop_word_filter.rs
+++ b/src/tokenizer/stop_word_filter.rs
@@ -98,3 +98,15 @@ where
         false
     }
 }
+
+impl Default for StopWordFilter {
+    fn default() -> StopWordFilter {
+        let words = vec![
+            "a", "an", "and", "are", "as", "at", "be", "but", "by", "for", "if", "in", "into",
+            "is", "it", "no", "not", "of", "on", "or", "such", "that", "the", "their", "then",
+            "there", "these", "they", "this", "to", "was", "will", "with",
+        ];
+
+        StopWordFilter::remove(words.iter().map(|s| s.to_string()).collect())
+    }
+}
--- a/src/tokenizer/stop_word_filter.rs
+++ b/src/tokenizer/stop_word_filter.rs
@@ -39,6 +39,16 @@ impl StopWordFilter {
 
         StopWordFilter { words: set }
     }
+
+    fn english() -> StopWordFilter {
+        let words: [&'static str; 33] = [
+            "a", "an", "and", "are", "as", "at", "be", "but", "by", "for", "if", "in", "into",
+            "is", "it", "no", "not", "of", "on", "or", "such", "that", "the", "their", "then",
+            "there", "these", "they", "this", "to", "was", "will", "with",
+        ];
+
+        StopWordFilter::remove(words.iter().map(|s| s.to_string()).collect())
+    }
 }
 
 pub struct StopWordFilterStream<TailTokenStream>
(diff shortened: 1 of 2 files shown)


src/tokenizer/tokenizer_manager.rs
```diff
@@ -64,12 +65,14 @@ impl Default for TokenizerManager {
...
             "default",
             SimpleTokenizer
+                .filter(StopWordFilter::default())
                 .filter(RemoveLongFilter::limit(40))
                 .filter(LowerCaser),
         );
         manager.register(
             "en_stem",
             SimpleTokenizer
+                .filter(StopWordFilter::default())
```
fulmicoton: I don't think applying stop words by default is a good idea.

src/tokenizer/tokenizer_manager.rs
```diff
@@ -64,12 +65,14 @@ impl Default for TokenizerManager {
         manager.register(
             "default",
             SimpleTokenizer
+                .filter(StopWordFilter::default())
```
fulmicoton: same as below

src/tokenizer/stop_word_filter.rs
```diff
@@ -98,3 +98,15 @@ where
         false
     }
 }
+
+impl Default for StopWordFilter {
+    fn default() -> StopWordFilter {
```
fulmicoton, line 113: Maybe have a specific function called `.english()` and have `.default()` call that one.

src/tokenizer/stop_word_filter.rs
```diff
@@ -98,3 +98,15 @@ where
...
+        let words = vec![
+            "a", "an", "and", "are", "as", "at", "be", "but", "by", "for", "if", "in", "into",
```
fulmicoton: this could be a `[&'static str; N]`.

src/tokenizer/tokenizer_manager.rs
drusellers: Coming from Lucene I would expect to see a default list of stop words removed when indexing content. The words in this list are pretty low value. Do you have a use case in mind that I'm not considering? Either way, I can remove it - it would be pretty quick.

fulmicoton: Then that's an error in lucene... How do you search for the band "the who" in such an index?

fulmicoton: You can add another analyzer called en_stem_nostopwords or something like that maybe?

src/tokenizer/stop_word_filter.rs
drusellers: Oh nice. that should be "leaner" rather than building out a vec.

src/tokenizer/stop_word_filter.rs
drusellers: got it

src/tokenizer/stop_word_filter.rs
```diff
@@ -98,3 +98,15 @@ where
...
+            "is", "it", "no", "not", "of", "on", "or", "such", "that", "the", "their", "then",
+            "there", "these", "they", "this", "to", "was", "will", "with",
+        ];
+
+        StopWordFilter::remove(words.iter().map(|s| s.to_string()).collect())
```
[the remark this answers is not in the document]
fulmicoton: Depending on the use case there are slightly better ways. Here I think this is perfectly fine and easy to read.

src/tokenizer/tokenizer_manager.rs
drusellers: I think we can totally wait for our user base to let us know. ;p But yes, those types of phrases are tricky when using stopwords. :)

PR #382, opened 2018-08-18
RFC: Add lenient query mode to Tantivy

### LINKED ISSUE #5: Add a lenient mode to QueryParser
The role of the `QueryParser` is to give an off-the-shelf solution to people wanting to deliver a search box to end user.
Currently `QueryParser::parse` can return an error if the query does not match our query grammar.

e.g.: `+(happy`  : the parenthesis is not closed.

This is great, but how should somebody writing a site handle this error?
Maybe for a log analysis search engine, displaying an explicit _syntax error_ is a good idea, but for most use case, (ecommerce, document search, etc.) we want to never return an error and just do our best to handle the user query.

For this reason, we want to add a new `parse_lenient` that would never fail and always return a result.

```rust
pub fn parse_query_lenient(&self, query: &str) -> Box<Query> {
    // ...
}
``` 

The behavior of this lenient mode does NOT have to be very smart...
For instance, it could be a second very naive parser taking over when the initial parser failed.

For instance, this naive parser could remove all special characters and run the initial parser. 

Smarter attempts to interpret the faulty user query are also welcome, as long as we return `Box<Query>` and we are confident the code will never panic.

drusellers:
Closes #5 

Inspired by https://gist.github.com/bcoe/6505434

The goal is not to remove possibly poorly formatted queries but to try and take it as if the user was actually typing what they wanted, but didn't know that tantivy interprets some characters as special.

The goal is to escape these special characters and then pass them into the query parser. If that doesn't work - I'm thinking we should either return some kind of Empty query or the `AllQuery`. 

- [ ] does Tantivy have an escape character for special symbols?
- [ ] my current filter method is naive and could probably be written better with regards to memory and speed and all the things.


src/query/query_parser/query_parser.rs
```diff
@@ -144,18 +143,76 @@ impl QueryParser {
...
+    fn parse_lenient_query_to_logical_ast(
+        &self,
+        query: &str,
+    ) -> Result<LogicalAST, QueryParserError> {
+        // if we are here, we know we have a poorly formed
+        // query input
+
+        //  # Escape special characters: \\+-&|!(){}[]^~*?:\/
+        let special_chars = vec![
+            "\\", "+", "-", "&", "|", "!", "(", ")", "{", "}", "[", "]", "^", "~", "*", "?", ":",
```
fulmicoton: This could be a static array.
It could also be chars instead of &str, but interestingly, I actually wonder if it wouldn't end up being slower :).

src/query/query_parser/query_parser.rs
```diff
@@ -144,18 +143,76 @@ impl QueryParser {
...
+            "/",
+        ];
+
+        let mut scrubbed_query = query.to_string();
+        for c in special_chars.iter() {
+            scrubbed_query = scrubbed_query.replace(c, &format!("{}", c));
```
fulmicoton: I'm a bit lost. What does that do?

fulmicoton:
Just synced with master, and fixed a unit test.

src/query/query_parser/query_parser.rs
```diff
@@ -193,6 +190,61 @@ impl QueryParser {
         self.compute_logical_ast(user_input_ast)
     }
 
+    /// Parse a query
+    ///
+    /// Note that `parse_query_lenient` will NOT return an error
+    /// if the input is not a valid query.
+    ///
+    /// It will instead escape all special characters in the query body
+    /// retry to process the query, if it still fails will return the AllQuery
```
fulmicoton, line 199: `EmptyQuery` actually

src/query/query_parser/query_parser.rs
```diff
@@ -544,6 +596,26 @@ mod test {
...
+    }
+
+    #[test]
+    pub fn test_parse_query_lenient_escapes_bad_queries() {
+        let query_parser = make_query_parser();
+
+        let query = query_parser
+            .parse_lenient_query_to_logical_ast("+(happy")
+            .unwrap();
+        let query_str = format!("{:?}", query);
```
mauri870, line 615: What about matching this enums and assert_eq! `term.text()`?

PR #383, opened 2018-08-18
Add a basic faceted search example

drusellers:
This gives a basic example of faceted search

References #375 and #372


fulmicoton:
LGTM ! Cool!

PR #376, opened 2018-08-17
Switch from error-chain to failure crate

drusellers:
Closes #373 

[Failure](https://github.com/rust-lang-nursery/failure)

I implemented the custom fail type: https://boats.gitlab.io/failure/custom-fail.html

This was pretty easy and felt like it removed more code than it added.

PR #384, opened 2018-08-18
closes #235

### LINKED ISSUE #235: Display a nicer error when we acquire the IndexWriter lock
Currently the error message is

 thread 'main' panicked at 'Indexing failed: Error(FileAlreadyExists(".tantivy-indexer.lock"), State { next_error: None, backtrace: None })', /checkout/src/libcore/result.rs:916:4

petr-tik:
Closes #235

Added new error type when a lockfile already exists. Added possible
causes as seen in the comments to index_writer
It's is derived from OpenWriteError and specified if the filepath is
the LOCKFILE_FILEPATH

added a test that checks error message contains lockfile, already
exists and possible causes

Based on drusellers #376, happy to rebase from ffec038f819f8b3fda31f63f7ed6e12cc68a76af to master, if/when other PRs are merged


src/error.rs
```diff
@@ -9,130 +9,124 @@ use schema;
...
+pub enum TantivyError {
+    /// Path does not exist.
+    #[fail(display = "path does not exist: '{:?}'", _0)]
+    PathDoesNotExist(PathBuf),
+    /// File already exists, this is a problem when we try to write into a new file.
+    #[fail(display = "file already exists: '{:?}'", _0)]
+    FileAlreadyExists(PathBuf),
+    /// Lockfile already exists
+    #[fail(display = "Lockfile '{:?}' already exists. Possible causes: another IndexWriter instance or panic during previous lock drop.", _0)]
+    LockFileAlreadyExists(PathBuf),
```
fulmicoton, line 25: Maybe information about what to would be useful. Maybe we should link to a FAQ or something? What do you think?

petr-tik: Do we have a FAQ? I think an error message needs to balance between being descriptive and concise. If we have a FAQ, we can add a detailed response and troubleshooting steps and make it easily google-able by the error message

I thought creating another Error enum value is detailed enough and should give people information to debug.

fulmicoton: No FAQ yet. I'm trying to find the right solution for that.

fulmicoton:
I'm ok to merge once you fix the conflicts and fix the comment about the unit test.

There is a lot of spurious *plastic* diff in this CR. Next time, please limit yourself to the change you are doing.
If you want to `cargo format` or something like that, you can send a separate PR without any issue.

src/indexer/index_writer.rs
```diff
@@ -654,11 +657,27 @@ mod tests {
...
+            Err(TantivyError::LockFileAlreadyExists(_)) => {}
             _ => panic!("Expected FileAlreadyExists error"),
         }
     }
 
+    #[test]
+    fn test_lockfile_already_exists_error_msg() {
+        let schema_builder = schema::SchemaBuilder::default();
+        let index = Index::create_in_ram(schema_builder.build());
+        let _index_writer = index.writer(40_000_000).unwrap();
```
fulmicoton, line 669: use one thread and a `3MB` buffer

fulmicoton: @petr-tik up

petr-tik: Paul, thanks for following up. I have been travelling and jet lagged, should get some time tonight

fulmicoton:
thanks @petr-tik

Issue #368, opened 2018-08-10
Add a simple SnippetGenerator 


fulmicoton:
(This is also called highlighter in Lucene.)

The idea is, for a given snippet and a given a number of characters limit, to return a text snippet of the search result.
If possible, the keywords should be present in the snippet and highlighted using the `<em>` tag.

The MVP could look like this :

Write a function that takes 
- a tokenstream
- a maximum length budget
- a set of keywords to highlight, with a given importance score (HashMap<String, f32>)

Go through the resulting `TokenStream`, and for each token, check if they are a token, and if so,
keep track of starting and stopping offsets... 

Identify the best fragment of text with a length shorter or equal to the budget, and return it with its score.

The score is defined as the `sum of the score of the keywords present in the document`.
A keyword appearing more than once does NOT get extra points.

A fragment must not cut any word.

(We will work in adding the plumbing to make it user-friendly in another ticket.)

vigneshsarma:
Thanks a lot for that. I was figuring out these interfaces my self, what I would have come up with would have been completely different.

fulmicoton:
I don't have a super strong opinion. Feel free to change this interface.

fulmicoton:
@vigneshsarma  are you doing ok there? do you need help?

vigneshsarma:
I am having a difficult time understanding the Lucene Java code. They are passing around a lot of mutable references and using that mutation to calculate various params. I am not sure what could help me. I think I just have to keep at it. For now, I am trying to run that Java code and run the tests as well.

That might help me understand what is going on.

fulmicoton:
No need to do something as complicated as lucene. Sometying working would
actually be good enough.

vigneshsarma:
Ok in that case, 

### a simpler method would be, 

* Take fragment string as the first stretch of text with max snippet length 
* if it has one of the expected keywords keep it and find the next string
* else find the next string
* at the end of it, the one with the highest score will be returned. 

### Things *not* to be done.
- Overlapping won't be tested. We can do that later.
- We won't be merging fragments for a higher score. 

What do you think?

fulmicoton:
Sure... Let s start by making something that works. We can make improve it
later.

fulmicoton:
@vigneshsarma Merged,l thanks. I added a `SnippetGenerator` to simplify the usage, an example, and some documentation.

PR #395, opened 2018-08-27
WIP: Issue/368

vigneshsarma:
Refer: #368

### DIFF
--- a/src/snippet/mod.rs
+++ b/src/snippet/mod.rs
@@ -55,10 +55,8 @@ impl FragmentCandidate {
 
         if let Some(score) = terms.get(&token.text.to_lowercase()) {
             self.score += score;
-            self.highlighted.push(HighlightSection {
-                start: token.offset_from,
-                stop: token.offset_to,
-            });
+            self.highlighted
+                .push(HighlightSection::new(token.offset_from, token.offset_to));
         }
     }
 }
@@ -163,9 +161,11 @@ fn select_best_fragment_combination<'a>(
         let highlighted = fragment
             .highlighted
             .iter()
-            .map(|item| HighlightSection {
-                start: item.start - fragment.start_offset,
-                stop: item.stop - fragment.start_offset,
+            .map(|item| {
+                HighlightSection::new(
+                    item.start - fragment.start_offset,
+                    item.stop - fragment.start_offset,
+                )
             }).collect();
         Snippet {
             fragments: fragment_text.to_owned(),
--- a/src/snippet/mod.rs
+++ b/src/snippet/mod.rs
@@ -275,6 +275,31 @@ Rust won first place for \"most loved programming language\" in the Stack Overfl
     }
 
     #[test]
+    fn test_snippet_with_second_fragment_has_the_highest_score() {
+        let boxed_tokenizer = box_tokenizer(TOKENIZER);
+
+        let text = "a b c d e f g";
+
+        let mut terms = BTreeMap::new();
+        terms.insert(String::from("f"), 1.0);
+        terms.insert(String::from("a"), 0.9);
+
+        let fragments = search_fragments(boxed_tokenizer, &text, terms, 7);
+
+        assert_eq!(fragments.len(), 2);
+        {
+            let first = fragments.iter().nth(0).unwrap();
+            assert_eq!(first.score, 0.9);
+            assert_eq!(first.stop_offset, 7);
+            assert_eq!(first.start_offset, 0);
+        }
+
+        let snippet = select_best_fragment_combination(fragments, &text);
+        assert_eq!(snippet.fragments, "e f g");
+        assert_eq!(snippet.to_html(), "e <b>f</b> g");
+    }
+
+    #[test]
     fn test_snippet_with_term_not_in_text() {
         let boxed_tokenizer = box_tokenizer(TOKENIZER);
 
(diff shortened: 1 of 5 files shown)


src/snippet/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+    use tokenizer::{SimpleTokenizer, box_tokenizer};
+    use std::iter::Iterator;
+    use std::collections::BTreeMap;
+    use super::{search_fragments, select_best_fragment_combination};
+
+    #[test]
+    fn test_snippet() {
+        let tokenizer = SimpleTokenizer;
+
+        let t = box_tokenizer(tokenizer);
```
fulmicoton: let's avoid one letter variables. 
s/t/tokenizer/g

src/snippet/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+
+impl Snippet {
+
+    /// Returns a hignlightned html from the `Snippet`.
+    pub fn to_html(&self) -> String {
+        let mut html = String::new();
+        let mut start_from: usize = 0;
+
+        for item in self.highlighted.iter() {
+            html.push_str(&self.fragments[start_from..item.start]);
```
fulmicoton: Please encode the html entities in the original text. You can add [htmlescape](https://crates.io/crates/htmlescape) as a dependency.

src/snippet/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+                          start_offset: start_offset,
+                          stop_offset: end_offset,
+                          num_chars: 0,
+                          highlighted: vec![]}
+    }
+
+    /// Updates `score` and `highlighted` fields of the objects.
+    ///
+    ///
+    fn calculate_score(&mut self, token: &Token, terms: &BTreeMap<String, f32>) {
```
fulmicoton: this should be called add_token.

fulmicoton: This should be called try_add_token`.

The comment above should explain that if the token is one of the keyword, then the fragment length will be extended.

src/snippet/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+    let mut fragments:Vec<FragmentCandidate> = vec![];
+
+    loop {
+        if let Some(next) = token_stream.next() {
+            if (next.offset_to - fragment.start_offset) > max_num_chars {
+                let txt = &text[fragment.start_offset..fragment.stop_offset];
+                if fragment.score > 0.0 {
+                    fragments.push(fragment)
+                };
+                fragment = FragmentCandidate::new(next.offset_from, next.offset_to);
```
fulmicoton: I think there is a bug if `next` is a keyword. Then it never contributes to the score does it.

src/tokenizer/tokenizer.rs
```diff
@@ -130,7 +130,7 @@ where
     }
 }
 
-pub(crate) fn box_tokenizer<A>(a: A) -> Box<BoxedTokenizer>
+pub fn box_tokenizer<A>(a: A) -> Box<BoxedTokenizer>
```
fulmicoton, line 133: why remove `(crate)`?

src/snippet/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+    use std::collections::BTreeMap;
+    use super::{search_fragments, select_best_fragment_combination};
+
+    #[test]
+    fn test_snippet() {
+        let tokenizer = SimpleTokenizer;
+
+        let t = box_tokenizer(tokenizer);
+
+        let text = "Rust is a systems programming language sponsored by Mozilla which describes it as a \"safe, concurrent, practical language\", supporting functional and imperative-procedural paradigms. Rust is syntactically similar to C++[according to whom?], but its designers intend it to provide better memory safety while still maintaining performance.
```
fulmicoton: That's a good unit test. Keep it. Can you add more simpler unit test (the words can be one letter words) that explore the specification.

For instance :
- What happens if the first token is a keyword?
- if the text is empty?
- if the text contains no keyword?

(It's ok if the current specification is not great. We'll fix the unit tests as we improve the code.)

fulmicoton:
Thanks. I added some commetns to the CR. I understand it might feel a bit confusing without being able to run it with a query. I'll try to fill the gap later this week so we can see the result of your work on a actual queries.

src/tokenizer/tokenizer.rs
vigneshsarma: With this, I get the following error when I compile, 

```rust
error[E0364]: `box_tokenizer` is private, and cannot be re-exported
   --> src/tokenizer/mod.rs:156:43
    |
156 | pub use self::tokenizer::{BoxedTokenizer, box_tokenizer};
```

vigneshsarma:
@fulmicoton What do you think of the current PR?
To fix 
> if the term is repeated twice, it will add the score twice.

I think we will have to introduce a separate struct, which maintains the tokens that have been counted already. 
Also shouldn't we use a better comparison than a simple `terms.get()`?

> not calculating num_chars.

I am not completely sure how to fix this.

fulmicoton:
@vigneshsarma I need a little bit more time to review your code, and finishing the "connecting-the-dot" code to make your highlighter usable by users.

Please give me some more time... Overall we should be able to merge somewhere next week.

Issue #401, opened 2018-08-30
Custom weight/score calculation


tailhook:
Hi!

For my use case, I need to have custom BM25-like formula. In particular one that makes term frequency non-significant (but still applying idf). 

As far as I understand, with current codebase I need to implement my own TermQuery, TermWeight and QueryParser, which is too much work for the task. I'd like to substitute only B25Weight to another similarity weight without changing everything else.

I can make a PR, if you provide me some guidance. At a glance we need, `QueryParser` and `TermQuery` generic over similarityy weight type. Or maybe QueryParser generic over TermQuery (+PhraseQuery, RangeQuery)? Or maybe refactor query parser to a tool that makes it easier to make  your own parser?

fulmicoton:
Can you tell more about your use case? Why doesn't BM25 apply in your case?
If you don't have phrase queries, not encoding term frequency will apply BM25 with a term freq of 1.

Depending on your use case, we could introduce a `Similarity` enum and pass it to the term query.
I doubt the performance impact would be big.

tailhook:
> Can you tell more about your use case? Why doesn't BM25 apply in your case?

We are searching for goods on the marketplace. Things to note: (1) if someone searches "refrigerator" or "refrigerator samsung" we want users to see things like "refrigerator samsung 2345S". And (2) we don't control titles for products. So if we were using BM25, if some company publishing products wants their product to be in top results they can set title "refrigerator". Which is good for them for getting lots of impressions on generic search but isn't good for users and the marketplace as a whole.

*(to be clear we're just experimenting with tantivy, not using it for real marketplace yet, but we have similar tweaks in elasticsearch's ranking)*

> If you don't have phrase queries, not encoding term frequency will apply BM25 with a term freq of 1.

Not sure I understand, what phrase search means. But we have multi-term searches. And want `"quoted search"` to work too.

Also we're using `IndexRecordOption::Basic` for indexing that column.

> Depending on your use case, we could introduce a Similarity enum and pass it to the term query.
I doubt the performance impact would be big.

I'm quite new with full text search, so maybe I misunderstand something. I used to think that there might be different algorithms to score search results, that's why I've proposed to have generics.

fulmicoton:
See CR.

josemerca:
## A class of search workloads where the default BM25 length-norm breaks structurally: short-text catalog search

[...]

PR #411, opened 2018-09-10
Add constructors for setting bm25 coefficients

tailhook:
This is an attempt to solve #401 

I'm not sure about:
1. Maybe we need a stucture instead of passing two vars around
2. `_with_settings` suffix isn't particularly good. Alternatives?
3. Maybe we already need builder/fluent interface instead of many constructors for `QueryParser`, `TermQuery` and `PhraseQuery`?

### DIFF
--- a/examples/custom_bm25.rs
+++ b/examples/custom_bm25.rs
@@ -1,0 +1,195 @@
+// # Basic Example
+//
+// This example covers the basic functionalities of
+// tantivy.
+//
+// We will :
+// - define our schema
+// = create an index in a directory
+// - index few documents in our index
+// - search for the best document matchings "sea whale"
+// - retrieve the best document original content.
+
+
+extern crate tempdir;
+
+// ---
+// Importing tantivy...
+#[macro_use]
+extern crate tantivy;
+use tantivy::collector::TopCollector;
+use tantivy::query::QueryParser;
+use tantivy::schema::*;
+use tantivy::Index;
+
+fn main() -> tantivy::Result<()> {
+    // Let's create a temporary directory for the
+    // sake of this example
+    let index_path = TempDir::new("tantivy_example_dir")?;
+
+    // # Defining the schema
+    //
+    // The Tantivy index requires a very strict schema.
+    // The schema declares which fields are in the index,
+    // and for each field, its type and "the way it should
+    // be indexed".
+
+    // first we need to define a schema ...
+    let mut schema_builder = SchemaBuilder::default();
+
+    // Our first field is title.
+    // We want full-text search for it, and we also want
+    // to be able to retrieve the document after the search.
+    //
+    // `TEXT | STORED` is some syntactic sugar to describe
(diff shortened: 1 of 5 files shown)


examples/custom_bm25.rs
```diff
@@ -0,0 +1,195 @@
...
+use tantivy::query::QueryParser;
+use tantivy::schema::*;
+use tantivy::Index;
+
+fn main() -> tantivy::Result<()> {
+    // Let's create a temporary directory for the
+    // sake of this example
+    let index_path = TempDir::new("tantivy_example_dir")?;
+
+    // # Defining the schema
```
fulmicoton, line 30: You can remove the comments that are not really on point with the feature you are trying to highlight.
Also I don't think this feature requires an individual example. A rust doc example would be awesome though.

src/query/term_query/term_query.rs
```diff
@@ -96,7 +115,8 @@ impl TermQuery {
     /// This is useful for optimization purpose.
     pub fn specialized_weight(&self, searcher: &Searcher, scoring_enabled: bool) -> TermWeight {
```
fulmicoton, line 116: What about we replace `scoring_enabled` by an `Option<BM25Similiary>` ?

src/query/term_query/term_query.rs
```diff
@@ -81,6 +84,22 @@ impl TermQuery {
         TermQuery {
             term,
             index_record_option: segment_postings_options,
+            bm25_k1: bm25::DEFAULT_K1,
```
fulmicoton, line 87: I don't think this should be part of the `Query`

fulmicoton:
Cool, so changing the BM25 coefficients are solving your problem?

I am not fond of embedding the similarity object in `TermQuery` and `PhraseQuery`,
What about we have a default `BM25` in the index, in the query parser, and in the `Searcher`
and pass it 
as an `Option` in place of the `scoring_enabled  argument in the `Query::weight(...)` call.

fulmicoton (changes requested):
@tailhook 

Cool, so changing the BM25 coefficients are solving your problem?
I am not fond of embedding the similarity object in `TermQuery` and `PhraseQuery`,

What about the following solution... 

Have a default `BM25` object in the index that only contains the 2 parameters.

Add a method  `searcher::search_with_similarity(&self, query: &Query, collector: &mut C, similiary: BM25)`.  Make `BM25` `Copy`.

`Query::search` then could get an extra `Option<BM25>`. 
`Query::weight` then would replace the `scoring_enabled` by an `Option<BM25>` argument.

This is close to what Lucene does. 
I don't want to embed the `Similarity` in the `Searcher` because we recycle those and I am scared of mutating them.
With the following setting, we leave the door open in the future to having different `Similarity` implementation. If the dynamic dispatch hurts performance (which I doubt), we can consider batching and remove the dynamic dispatch on `Collector::collect` too.

PR #420, opened 2018-09-14
Cargo Format

fulmicoton:


petr-tik:
Hey, I feat this is going to make `git blame` less informative and maybe we should incorporate cargo fmt and only run it on the files in every commit from now on?

fulmicoton:
It used to be part of tantivy's CI, but the output of `rustfmt` was not really stable. it was annoying so I removed it. Is it stable nowadays?

petr-tik:
If you don't like the inconsistencies (I agree with that), why format the whole codebase now?

fulmicoton:
`cargo fmt` was giving different result for each version.

CI was checking that the code was committed formatted using the last version of `rustfmt`.
So you'd end up with CI broken just because your version was not the most up to date.
Also the size of the diff would be gigantic everyone it would happen.

Doing that once every 6 months does not have this problem.

(Maybe rustfmt is more stable now and this is not a problem anymore.)

PR #428, opened 2018-09-22
Adds open_or_create method

### LINKED ISSUE #416: Change the semantics associated to opening/creating indexes.

See #415 

- [ ] Change the semantic of `Index::create_in_dir`. It should return an error if the directory already contains an `Index`. 
- [ ] Add a `Index::open_or_create(dir, schema)` method.

drusellers:
closes #416 

- adds open_or_create
- makes create throw a new error if an index already exists

### DIFF
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -107,6 +107,8 @@ impl Index {
     }
 
     /// Create a new index from a directory.
+    ///
+    /// This will overwrite existing meta.json
     fn from_directory(mut directory: ManagedDirectory, schema: Schema) -> Result<Index> {
         save_new_metas(schema.clone(), 0, directory.borrow_mut())?;
         let metas = IndexMeta::with_schema(schema);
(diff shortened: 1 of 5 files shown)


src/core/index.rs
```diff
@@ -65,9 +70,22 @@ impl Index {
...
+            return Err(TantivyError::IndexAlreadyExists);
+        }
+
         Index::create(mmap_directory, schema)
     }
 
+    /// Opens or creates a new index in the provided directory
+    #[cfg(feature = "mmap")]
+    pub fn open_or_create<Dir: Directory>(dir: Dir, schema: Schema) -> Result<Index> {
+        if Index::exists(&dir) {
```
fulmicoton, line 83: check that the schema is not different

fulmicoton:
Sorry for the late answer.

I added a commit to check that Schema matches, and used RAMDirectory in the unit tests. Does that work with you @drusellers ?

src/core/index.rs
fulmicoton: added that

fulmicoton:
Merging as it looks inoccuous

PR #430, opened 2018-09-29
Fix NGram processing of non-ascii characters

### LINKED ISSUE #429: NgramTokenizer panics on non-ASCII inputs
NgramTokenizer panics when used with non-ASCII inputs. To reproduce, simply change `"hello"` to `"hεllo"` in `tokenizer/mod.rs:250`:

`thread 'tokenizer::test::test_ngram_tokenizer' panicked at 'byte index 2 is not a char boundary; it is inside 'ε' (bytes 1..3) of `hεllo`', libcore/str/mod.rs:2111:5`

Currently, the computation of n-grams is based on incrementing the current position by one, and adding the desired length. Both would need to account for character widths to respect character boundariers.

I can work on a pull request if desired.

Tested on master (e0cdd3114d05c927a349eb07a5daaf17360661ae)

drusellers:
closes #429 

- [ ] does this design pass "performance" demands

i'm calling clone twice to iterate the chars and get the physical offsets

- [ ] did i correctly map the concept of position and other token properties

i need to reread the docs to make sure i wrote the test code correctly when asserting non-ascii tokens

### DIFF
--- a/src/tokenizer/mod.rs
+++ b/src/tokenizer/mod.rs
@@ -291,6 +291,46 @@ pub mod test {
     }
 
     #[test]
+    fn test_ngram_non_ascii_tokenizer() {
+        use super::NgramTokenizer;
+        use tokenizer::tokenizer::TokenStream;
+        use tokenizer::tokenizer::Tokenizer;
+
+        let tokenizer = NgramTokenizer::new(1, 2, false);
+        let mut tokens: Vec<Token> = vec![];
+        {
+            let mut add_token = |token: &Token| {
+                tokens.push(token.clone());
+            };
+            tokenizer.token_stream("hεllo").process(&mut add_token);
+        }
+        assert_eq!(tokens.len(), 9);
+        assert_token(&tokens[0], 0, "h", 0, 1);
+        assert_token(&tokens[1], 0, "hε", 0, 3);
+        assert_token(&tokens[2], 1, "ε", 1, 3);
+        assert_token(&tokens[3], 1, "εl", 1, 4);
+        assert_token(&tokens[4], 2, "l", 3, 4);
+        assert_token(&tokens[5], 2, "ll", 3, 5);
+        assert_token(&tokens[6], 3, "l", 4, 5);
+        assert_token(&tokens[7], 3, "lo", 4, 6);
+        assert_token(&tokens[8], 4, "o", 5, 6);
+
+        let tokenizer = NgramTokenizer::new(2, 5, true);
+        let mut tokens: Vec<Token> = vec![];
+        {
+            let mut add_token = |token: &Token| {
+                tokens.push(token.clone());
+            };
+            tokenizer.token_stream("hεllo").process(&mut add_token);
+        }
+        assert_eq!(tokens.len(), 4);
+        assert_token(&tokens[0], 0, "hε", 0, 3);
+        assert_token(&tokens[1], 0, "hεl", 0, 4);
+        assert_token(&tokens[2], 0, "hεll", 0, 5);
+        assert_token(&tokens[3], 0, "hεllo", 0, 6);
+    }
+
+    #[test]
     fn test_tokenizer_empty() {
         let tokenizer_manager = TokenizerManager::default();
         let en_tokenizer = tokenizer_manager.get("en_stem").unwrap();
--- a/src/tokenizer/mod.rs
+++ b/src/tokenizer/mod.rs
@@ -309,11 +309,11 @@ pub mod test {
         assert_token(&tokens[1], 0, "hε", 0, 3);
         assert_token(&tokens[2], 1, "ε", 1, 3);
         assert_token(&tokens[3], 1, "εl", 1, 4);
-        assert_token(&tokens[4], 2, "l", 3, 4);
-        assert_token(&tokens[5], 2, "ll", 3, 5);
-        assert_token(&tokens[6], 3, "l", 4, 5);
-        assert_token(&tokens[7], 3, "lo", 4, 6);
-        assert_token(&tokens[8], 4, "o", 5, 6);
+        assert_token(&tokens[4], 3, "l", 3, 4);
+        assert_token(&tokens[5], 3, "ll", 3, 5);
+        assert_token(&tokens[6], 4, "l", 4, 5);
+        assert_token(&tokens[7], 4, "lo", 4, 6);
+        assert_token(&tokens[8], 5, "o", 5, 6);
 
         let tokenizer = NgramTokenizer::new(2, 5, true);
         let mut tokens: Vec<Token> = vec![];
--- a/src/tokenizer/mod.rs
+++ b/src/tokenizer/mod.rs
@@ -157,34 +157,33 @@ pub use self::tokenizer::BoxedTokenizer;
 pub use self::tokenizer::{Token, TokenFilter, TokenStream, Tokenizer};
 pub use self::tokenizer_manager::TokenizerManager;
 
-/// This is a function that can be used in tests and doc tests
-/// to assert a token's correctness.
-/// TODO: can this be wrapped in #[cfg(test)] so as not to be in the
-/// public api?
-pub fn assert_token(token: &Token, position: usize, text: &str, from: usize, to: usize) {
-    assert_eq!(
-        token.position, position,
-        "expected position {} but {:?}",
-        position, token
-    );
-    assert_eq!(token.text, text, "expected text {} but {:?}", text, token);
-    assert_eq!(
-        token.offset_from, from,
-        "expected offset_from {} but {:?}",
-        from, token
-    );
-    assert_eq!(
-        token.offset_to, to,
-        "expected offset_to {} but {:?}",
-        to, token
-    );
-}
 
 #[cfg(test)]
-pub mod test {
-    use super::assert_token;
+pub mod tests {
     use super::Token;
     use super::TokenizerManager;
+
+
+    /// This is a function that can be used in tests and doc tests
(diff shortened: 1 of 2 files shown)


src/tokenizer/mod.rs
```diff
@@ -290,6 +290,46 @@ pub mod test {
         assert_token(&tokens[3], 0, "frank", 0, 5);
     }
 
+    #[test]
+    fn test_ngram_non_ascii_tokenizer() {
```
fulmicoton: please add a test where codepoint != grapheme.

You can use this string for instance.
https://play.rust-lang.org/?gist=4ca51df298f3e2d1008872bb0b2b503d&version=stable&mode=debug&edition=2015

fulmicoton: I think it is ok to not handle grapheme for this functionality as long as we document it and we have a unit test for it.

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -113,8 +126,21 @@ impl<'a> NgramTokenStream<'a> {
...
+            let raw_position = chars.take(self.position).map(|c| c.len_utf8()).sum();
+
+            let chars = self.chars.clone();
+            let raw_size = chars
+                .skip(self.position)
+                .take(self.gram_size)
+                .map(|c| c.len_utf8())
+                .sum();
+
+            println!("logical: {},{}", self.position, self.gram_size);
```
fulmicoton: remaining println

fulmicoton:
@drusellers I'm ok to merge that without any work on perf. Remove the `println` and let me know whne your code is ready to ship.

drusellers:
whoot! I think I got it.

drusellers:
This was a super fun "challenge" - please give me any and all feedback. all the learning.

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -83,8 +91,9 @@ impl<'a> Tokenizer<'a> for NgramTokenizer {
     fn token_stream(&self, text: &'a str) -> Self::TokenStreamImpl {
         NgramTokenStream {
             text,
+            bytes: text.as_bytes(),
+            bytes_length: text.as_bytes().len(),
```
fulmicoton: this is the same as str.len

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -83,8 +91,9 @@ impl<'a> Tokenizer<'a> for NgramTokenizer {
     fn token_stream(&self, text: &'a str) -> Self::TokenStreamImpl {
         NgramTokenStream {
             text,
+            bytes: text.as_bytes(),
```
fulmicoton: not needed. This accessor is zero-cost.

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -132,14 +160,15 @@ impl<'a> TokenStream for NgramTokenStream<'a> {
         self.token.text.clear();
 
         if let Some((position, size)) = self.chomp() {
```
fulmicoton: Naming this position is very confusing

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -100,6 +109,7 @@ impl<'a> NgramTokenStream<'a> {
     /// returning None if processing should stop
     fn chomp(&mut self) -> Option<(usize, usize)> {
```
fulmicoton: The `start`, `len` is ok semantics, but it does interface well with rust slicing. 
 `start_offset`, `stop_offset` could have been better.

It is important to document the meaning of the returned value in that kind of function, especially when it is ambiguous like this. Even if it is private.

fulmicoton:
@drusellers I will fix a bunch of things in this struct directly on the branch.

I added the most important points as comments in the review, just for infomration.
Don't address them, I will.

drusellers:
Roger that. Thank you. Your feedback is much appreciated.

jvcop:
@fulmicoton Interesting work!

I put together a small benchmark based on a German news article (~7000 characters). I compared four candidates:

- Lookup table (`lut_256`)
- Condensed lookup table (`lut_16`) 
- Magic table (`magic_table`)
- Leading Zeros (`leading_zeros`)

On my machine, I get the following results:

```
leading_zeros ... bench:       8,891 ns/iter (+/- 441) = 775 MB/s
lut_16        ... bench:       3,334 ns/iter (+/- 3,621) = 2067 MB/s
lut_256       ... bench:       2,816 ns/iter (+/- 729) = 2448 MB/s
magic_table   ... bench:      13,192 ns/iter (+/- 1,843) = 522 MB/s
```

Rerunning it on a smaller portion of the article (~300 characters) gives:

```
leading_zeros ... bench:         368 ns/iter (+/- 73) = 771 MB/s
lut_16        ... bench:         137 ns/iter (+/- 34) = 2072 MB/s
lut_256       ... bench:         121 ns/iter (+/- 30) = 2347 MB/s
magic_table   ... bench:         552 ns/iter (+/- 26) = 514 MB/
```

It seems the versions based on lookup tables perform better. Maybe my benchmark is flawed/not representative enough, though.

fulmicoton:
Cool let's use the table then.

fulmicoton:
Cool let's use the table then.

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -181,3 +184,40 @@ impl<'a> TokenStream for NgramTokenStream<'a> {
...
+    0, // 0xBF
+    0, 0, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2,
+    2, // 0xDF
+    3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, // 0xEF
+    4, 4, 4, 4, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, // 0xFF
+];
+
+#[inline]
+pub fn utf8_char_width(b: u8) -> usize {
+    unsafe { *UTF8_CHAR_WIDTH.get_unchecked(b as usize) as usize }
```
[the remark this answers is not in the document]
drusellers: I totally copy/pasta'd that from the links pointing to the code. But good to know.

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -19,24 +20,63 @@ use super::{Token, TokenStream, Tokenizer};
...
 /// let tokenizer = NgramTokenizer::new(2, 3, false);
 /// let mut stream = tokenizer.token_stream("hello");
-///
-/// assert_token(stream.next().unwrap(), 0, "he", 0, 2);
-/// assert_token(stream.next().unwrap(), 0, "hel", 0, 3);
-/// assert_token(stream.next().unwrap(), 1, "el", 1, 3);
-/// assert_token(stream.next().unwrap(), 1, "ell", 1, 4);
-/// assert_token(stream.next().unwrap(), 2, "ll", 2, 4);
-/// assert_token(stream.next().unwrap(), 2, "llo", 2, 5);
-/// assert_token(stream.next().unwrap(), 3, "lo", 3, 5);
```
drusellers, line 39: Should we just remove `assert_token` ?

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -58,100 +98,376 @@ impl NgramTokenizer {
...
+    underlying: T,
+    min_gram: usize,
+    max_gram: usize,
+
+    memory: Vec<usize>,
+    cursor: usize,
+    gram_len: usize
+}
+
+impl<T> StutteringIterator<T>
```
drusellers, line 196: I <3 this name :joy:

src/tokenizer/ngram_tokenizer.rs
```diff
@@ -58,100 +98,376 @@ impl NgramTokenizer {
...
+        assert_token(&tokens[1], 0, "fra", 0, 3);
+        assert_token(&tokens[2], 0, "fran", 0, 4);
+        assert_token(&tokens[3], 0, "frank", 0, 5);
+    }
+
+    #[test]
+    fn test_ngram_non_ascii_1_2() {
+        let tokens = test_helper(NgramTokenizer::all_ngrams(1, 2).token_stream("hεllo"));
+        assert_eq!(tokens.len(), 9);
+        assert_token(&tokens[0], 0, "h", 0, 1);
```
drusellers, line 399: Ok, awesome. these are all from the first "token" so the position is 0. Thank you, that was a question I forgot to ask.

drusellers:
I can't approve in GitHub as the PR author, but I approve.

src/tokenizer/ngram_tokenizer.rs
fulmicoton: I think we discussed it a bit but my reply was stupid. Both of them kind of make no sense, but maybe this definition is slightly better.

src/tokenizer/ngram_tokenizer.rs
fulmicoton: I just prefer not to use it in rustdoc, to avoid making it public and to avoid forcing people to look for its documentation.

fulmicoton: `assert_token` is used in regular unit tests. This is a minor detail.

drusellers: Oh! Perfect. Totally tracking now.

fulmicoton:
@drusellers  Thanks!

Issue #386, opened 2018-08-20
Make the behavior of an empty query configurable in `QueryParser`


fulmicoton:
In some use case, most of the time, people want an empty query to match nothing, but for some analytics use case, it might make more sense for empty query to match all documents.

Add a setter ".set_empty_query_match_all(should_match_all: bool)" to the `QueryParser` to enforce the non default behavior.
(Also add a getter `.get_empty_query_match_all()` to read this property).

druzn3k:
Ok @fulmicoton , I wrote the code for this issue, but I need a little help on writing the tests. Could you help me?

fulmicoton:
For the unit tests, I think you can work from the logical ast.

Make tests on what you get when you call 
```rust
query_parser.parse_query_to_logical_ast("")
query_parser.parse_query_to_logical_ast("   ")
```

depending on whether query_parser.set_empty_query_match_all() was called or not.

druzn3k:
Thanks :+1:

druzn3k:
Ok, I wrote the PR. Is *far* from perfect (IMHO), but it can get an eventual discussion started. I can for sure expand (if needed) the test suite. Let me know, I'm really interested in knowing if there is any issue.

Thanks! :100:

PR #431, opened 2018-10-07
Add `set_empty_query_match_all` method

druzn3k:
This commit adds the ability to change the default behavior of an empty
search query, from returning `QueryEmpty` to returning `AllQuery`,
matching every value in the storage.

Closes issue #386

### DIFF
--- a/src/query/query_parser/query_parser.rs
+++ b/src/query/query_parser/query_parser.rs
@@ -182,15 +182,6 @@ impl QueryParser {
         self.empty_query_match_all = should_match_all;
     }
 
-    /// Get the current behavior of matching on an empty query
-    ///
-    /// This method get the value of `empty_query_match_all`, that regulate the
-    /// behavior of the matching on an empty query. When `true`, an empty query returns
-    /// all the value, opposite of the standard behavior `false` that return nothing.
-    pub fn get_empty_query_match_all(&self) -> bool {
-        self.empty_query_match_all
-    }
-
     /// Parse a query
     ///
     /// Note that `parse_query` returns an error if the input
@@ -570,6 +561,19 @@ mod test {
     pub fn test_parse_query_simple() {
         let query_parser = make_query_parser();
         assert!(query_parser.parse_query("toto").is_ok());
+    }
+
+    #[test]
+    pub fn test_parse_query_empty_match_all_field_default() {
+        let query_parser = make_query_parser();
+        assert_eq!(query_parser.empty_query_match_all, false);
+    }
+
+    #[test]
+    pub fn test_parse_query_empty_match_all_field_setter() {
+        let mut query_parser = make_query_parser();
+        query_parser.set_empty_query_match_all(true);
+        assert_eq!(query_parser.empty_query_match_all, true);
     }
 
     #[test]
--- a/src/query/query_parser/query_parser.rs
+++ b/src/query/query_parser/query_parser.rs
@@ -133,6 +133,7 @@ pub struct QueryParser {
     default_fields: Vec<Field>,
     conjunction_by_default: bool,
     tokenizer_manager: TokenizerManager,
+    empty_query_match_all: bool,
 }
 
 impl QueryParser {
@@ -150,6 +151,7 @@ impl QueryParser {
             default_fields,
             tokenizer_manager,
             conjunction_by_default: false,
+            empty_query_match_all: false,
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/query/query_parser/query_parser.rs
```diff
@@ -181,6 +202,10 @@ impl QueryParser {
     /// Implementing a lenient mode for this query parser is tracked
     /// in [Issue 5](https://github.com/fulmicoton/tantivy/issues/5)
     pub fn parse_query(&self, query: &str) -> Result<Box<Query>, QueryParserError> {
+        if query.trim().is_empty() && self.empty_query_match_all {
```
fulmicoton: Please work of the userinputast -> logicalinputast conversion. Not on the string itself.

The analyzer might transform a query string into an empty query even when `query.trim().is_empty()`.
For instance, if the query only contains punctuation or stop words.

src/query/query_parser/query_parser.rs
```diff
@@ -591,6 +616,19 @@ mod test {
         assert_eq!(format!("{:?}", query), "EmptyQuery");
     }
 
+    #[test]
+    pub fn test_parse_query_empty_match_all() {
+        test_parse_query_to_logical_ast_helper("", "<emptyclause>", false);
+        test_parse_query_to_logical_ast_helper(" ", "<emptyclause>", false);
+        let mut query_parser = make_query_parser();
```
fulmicoton, line 630: Why wasn't it possible to work off the logical ast to test the match all case too?

src/query/query_parser/query_parser.rs
druzn3k: Ok, no problem. I tried first with the naive solution to see if I could do it :)

src/query/query_parser/query_parser.rs
druzn3k: Lack of confidence with the language mostly. I tried first with a "good enough" solution to see if i could do it. Now I'll try to improve on it. Thanks for the patience

druzn3k:
I updated the code, and I want to punch myself in the face for not having seen the solution before. Oh well, the struggle is half the fun. Let me know if this is good or needs another iteration. Thanks! :+1:

src/query/query_parser/query_parser.rs
```diff
@@ -456,12 +468,12 @@ fn convert_literal_to_query(logical_literal: LogicalLiteral) -> Box<Query> {
     }
 }
 
-fn convert_to_query(logical_ast: LogicalAST) -> Box<Query> {
+fn convert_to_query(logical_ast: LogicalAST, should_match_all: bool) -> Box<Query> {
```
fulmicoton, line 471: `should_match_all` should be renamed `should_match_all_if_empty`.

src/query/query_parser/query_parser.rs
```diff
@@ -456,12 +468,12 @@ fn convert_literal_to_query(logical_literal: LogicalLiteral) -> Box<Query> {
...
     match trim_ast(logical_ast) {
         Some(LogicalAST::Clause(trimmed_clause)) => {
             let occur_subqueries = trimmed_clause
                 .into_iter()
-                .map(|(occur, subquery)| (occur, convert_to_query(subquery)))
+                .map(|(occur, subquery)| (occur, convert_to_query(subquery, should_match_all)))
```
fulmicoton, line 476: I don't think this is correct to recursively apply this value.

What happens if we have queries like :
`A AND NOT <some_stopword>`

This "what happens if empty logic should on" if the entire query is empty

src/query/query_parser/query_parser.rs
fulmicoton: This has not been fixed.

druzn3k: sorry I forgot! :(

src/query/query_parser/query_parser.rs
druzn3k: No problem

src/query/query_parser/query_parser.rs
druzn3k: Ok, I'll try to find a more correct way.

fulmicoton: To be honest I am not entirely sure of what the best definition should be. You just write a bunch of unit test beforehands, and we can discuss the best outcome when `should_match_all` is `true` or `false`.

- `something -<a stop word>`
- `something something_else`
- `something AND something_else`
- `something AND NOT something_else`
- `something OR something_else`

druzn3k: So, I wrote this at the moment:

```rust
pub fn test_logic_behind_should_match_all() {
    let query_parser = make_query_parser();

    // something -<a stop word>
    assert!(query_parser.parse_query("toto -such").is_ok());

    // something something_else
    assert!(query_parser.parse_query("toto teta").is_ok());

    // something AND something_else
    assert!(query_parser.parse_query("toto AND foo").is_ok());

    // something AND NOT something_else
    assert!(query_parser.parse_query("toto AND NOT bar").is_ok());

    // something OR something_else
    assert!(query_parser.parse_query("toto OR foobar").is_ok());
}
```

I plan to make the `query_parser` variable `mut` and toggle the `should_match_all` on/off on need. It's ok, or should I make a new `query_parser` from scratch for every test case? The queries are correct or did you expect something different? Sorry for not reaching out before, I got kind of busy :)

fulmicoton: @druzn3k 

Sorry for the 3month iatus. Of course all of these queries should pass.
My concern is more about what they should return...

For instance, if empty match all,  `term_that_exists -term_that_does_not_exist`  will return no document, which is rather unexpected.

I think the match_all_if_query should not apply if this is part of another query.

PR #437, opened 2018-11-09
Collector refactoring and multithreaded search

fulmicoton:
This is a rather large refactoring of the collector API.
This change include no change to the index format, but breaks user API.

The big picture is, implementing a collector now requires to implement 2 Traits.
They both share a `Fruit` associated types that represents the result type of the collector. (what you are trying to compute)
The `Collector` and the `SegmentCollector`.

The `SegmentCollector` does the actual computation over segments, and `Collector` acts as a factory of `SegmentCollector` and implements the logic to merge fruits.

- `collector::chain` is dropped and is replaced by just passing tuples of collectors.
- `search(...)` returns the fruit directly. There is no need to get the result from the collectors.

This unlocks multithreaded search, simplify the user API, and opens the door to distributed search (provided the fruit you are working with a serializable.)


src/collector/mod.rs
```diff
@@ -1,253 +1,314 @@
...
+Here the type of fruit is actually determined as an associated type of the collector (`Collector::Fruit`).
+
+
+# Combining several collectors
+
+
+# Implementing your own collectors.
+
+
+
```
drusellers, line 85: :clap:

src/collector/mod.rs
```diff
@@ -1,253 +1,314 @@
...
+    /// You should not use this method.
+    ///
+    /// Instead, please use [the `Searcher`'s search(...) method](../struct.Searcher.html#method.search).
+    #[doc(hidden)]
+    fn search_multithreads(&self, searcher: &Searcher, query: &Query, num_threads: usize) -> Result<Self::Fruit> {
+        let segment_readers = searcher.segment_readers();
+        let actual_num_threads = (segment_readers.len() + 1).max(num_threads);
+        let scoring_enabled = self.requires_scoring();
+        let weight = query.weight(searcher, scoring_enabled)?;
+        let thread_pool = rayon::ThreadPoolBuilder::new()
```
drusellers: So this builds a new threadpool with each search? Is this a perf concern?

src/collector/mod.rs
```diff
@@ -1,253 +1,314 @@
...
+            .build()
+            .map_err(|_| TantivyError::SystemError("Failed to spawn a search thread".to_string()))?;
+        let segment_fruits: Vec<Self::Fruit> = thread_pool.install(|| {
+            (0..segment_readers.len())
+                .into_par_iter()
+                .map(|segment_ord| {
+                    let segment_reader = &segment_readers[segment_ord];
+                    let mut scorer = weight.scorer(segment_reader).unwrap();
+                    let mut child = self
+                        .for_segment(segment_ord as u32, segment_reader)?;
```
drusellers: Is it worth refactoring this section here and the code in `search` method above into the same method?

src/collector/tests.rs
```diff
@@ -0,0 +1,202 @@
...
+use DocId;
+use Score;
+use SegmentLocalId;
+use DocAddress;
+
+/// Stores all of the doc ids.
+/// This collector is only used for tests.
+/// It is unusable in pr
+///
+/// actise, as it does not store
```
drusellers, line 15: typo

src/collector/tests.rs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+#[derive(Default)]
+pub struct TestFruit {
+    docs: Vec<DocAddress>,
+    scores: Vec<Score>
+}
+
+impl TestFruit {
+    /// Return the exhalist of documents.
```
drusellers: typo?

src/collector/top_collector.rs
```diff
@@ -3,54 +3,54 @@ use std::collections::BinaryHeap;
...
+
 
 /// Contains a feature (field, score, etc.) of a document along with the document address.
 ///
 /// It has a custom implementation of `PartialOrd` that reverses the order. This is because the
 /// default Rust heap is a max heap, whereas a min heap is needed.
-#[derive(Clone, Copy)]
-pub struct ComparableDoc<T> {
+///
+/// WARNING: equality is not what you would expect here
```
drusellers: what is it?

src/core/searcher.rs
```diff
@@ -86,9 +91,34 @@ impl Searcher {
...
+    }
+
+    /// Same as [`search(...)`](#method.search) but multithreaded.
+    ///
+    /// The current implementation is rather naive :
+    /// multithreading is by splitting search into as many task
+    /// as there are segments.
+    ///
+    /// It is powerless at making search faster if your index consists in
+    /// one large segment.
```
drusellers, line 142: great note

drusellers (approved):
I have one Q about performance around the thread pool which is more about my naïveté, than anything else. Over all this looks like both a solid new feature and some wonderful code clean up.

fulmicoton:
@hntd187 Added you because this refactoring is quite important for Toshi. It creates a simple solution to implementing sharded search.

src/collector/top_collector.rs
fulmicoton: adding a comment.

src/collector/mod.rs
fulmicoton: Did that and removed the confusing implementation in `Collector`

src/collector/tests.rs
fulmicoton: fixing

src/collector/mod.rs
fulmicoton: Excellent point.

fulmicoton:
Call for advise @jason-wolfe , @drusellers : as pointed out by @drusellers, it might not be a good idea to create the thread pool at every single query.

The trouble here is: we could let the user pass his own thread pool, but if so, we should have an object wrapping rayon's thread pool to avoid exposing rayon's object and create a dependency hell.

Alternatively, we could also have a thread pool at the scale of the Index, and use it by default.
The trouble then is that we could imagine a world where people want to run different queries on different thread pool and possibly single threaded too.

What's your take on this?

src/collector/mod.rs
```diff
@@ -1,253 +1,358 @@
...
-/// - `.collect(last_doc_of_last_segment)`
+/// The collection logic itself is in the `SegmentCollector`.
 ///
 /// Segments are not guaranteed to be visited in any specific order.
-pub trait Collector {
+pub trait Collector: Sync {
+
+    /// `Fruit` is the type for the result of our collection.
+    /// e.g. `usize` for the `Count` collector.
+    type Fruit: Fruit;
```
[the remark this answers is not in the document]
fulmicoton: Fruit as in *the fruit of your work*. 
I'm voluntarily choosing a word that is unambiguous with other words (as opposed to `Result` for instance).

src/collector/mod.rs
```diff
@@ -1,253 +1,358 @@
...
-    ) -> Result<()> {
-        (*self).set_segment(segment_local_id, segment)
-    }
+
+/// The `SegmentCollector` is the trait in charge of defining the
+/// collect operation at the scale of the segment.
+///
+/// `.collect(doc, score)` will be called for every documents
+/// matching the query.
+pub trait SegmentCollector: 'static {
```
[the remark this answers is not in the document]
fulmicoton: No. Can you have a look at oneof the collector implementation? It's not as if you have one struct implementing both. You have two structs respectively implementing one and the other.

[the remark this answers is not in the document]
hntd187: Oh okay, I missed that reading through it, so that makes a lot more sense then if it's 2 structs implementing different aspects of that. Disregard what I said before then.

src/collector/mod.rs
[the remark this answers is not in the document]
hntd187: Sounds good, I was merely curious what the thought was behind it.

fulmicoton:
@drusellers added an executor.

fulmicoton:
@hntd187 Nice catch, thanks.

petr-tik:
Question/suggestion re: Schema API since you are changing from `SchemaBuilder::default` to `Schema::Builder`

The last call to `schema_builder` is `schema_builder.build()`, which is a bit too verbose IMHO. 

This PR will break the API. Additionally, the examples folder includes user-facing code, which will form users' impression of the project/codebase, it would be great to discuss a change like to the `schema::build()` calls.  

We can change the `schema::build` to something like finish, complete or finalise, eg.
`schema_builder::finish()`

Alternatively, we can s/schema_builder/schema in all examples, which will make the last call site of `schema::build()` concise and to the point. 

Feel free to disregard, if you think it's a minor nit-pick. The rest of the PR looks great - thanks for the continuous work

fulmicoton:
@petr-tik I readded the Default implementation to make sure that we stay backward compatible.

I have no idea what you mean about the verbosity of `schema_builder.build()`.
I don't care about `build` vs `finish`.... Either way I don't think this is worth the change.

petr-tik:
As I said above, feel free to disregard - I felt it was minor

PR #442, opened 2018-11-15
Derive Clone for QueryParser

### LINKED ISSUE #441: QueryParser should have Clone
So I'm use Tantivy in a Lua binding (via rlua), and it required me that QueryParse implements Clone. Looking at the fields of QueryParser, I think it's ok to have Clone. What do you think?

juchiast:
fixes #441


fulmicoton:
Thanks!

Issue #439, opened 2018-11-10
Make sure the snippet result object is public.


fulmicoton:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**[Optional] describe alternatives you've considered**

trinity-1686a:
I think this issue is closed by #448 too

PR #448, opened 2018-12-01
Add accessors for Snippet and HighlightSection

### LINKED ISSUE #434: Getting subparts of Snippet
I just discovered the existence of Snippet, and I think I could use those in a project, however only be able to get results as html with `<b>` seems like a limitation. Depending on the context, using another html tag, or something totally different, like an ANSI escape code to color highlighted parts in a cli, would be more appropriate.

I'd like to be able to get the raw text stored in a Snippet, and highlighted sections, trough a getter or by marking them `pub`.

I could search for `<b>`s in the result of `to_html()`, and reconstruct the raw text and highlighted sections from that.

trinity-1686a:
Add accessors for Snippet and HighlightSection, and add example of a custom highlighter.

fix #434 and #439 I think

### DIFF
--- a/examples/snippet.rs
+++ b/examples/snippet.rs
@@ -14,7 +14,7 @@ use tantivy::collector::TopDocs;
 use tantivy::query::QueryParser;
 use tantivy::schema::*;
 use tantivy::Index;
-use tantivy::SnippetGenerator;
+use tantivy::{Snippet, SnippetGenerator};
 use tempdir::TempDir;
 
 fn main() -> tantivy::Result<()> {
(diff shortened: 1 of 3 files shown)


src/snippet/mod.rs
```diff
@@ -99,6 +108,18 @@ impl Snippet {
         ));
         html
     }
+
+    /// Returns a fragment from the `Snippet`.
+    #[inline(always)]
```
fulmicoton: no need for inline always.

src/snippet/mod.rs
```diff
@@ -99,6 +108,18 @@ impl Snippet {
...
+    pub fn fragments(&self) -> &str {
+        &self.fragments
+    }
+
+    /// Returns a list of higlighted positions from the `Snippet`.
+    #[inline(always)]
+    pub fn highlighted(&self) -> &[HighlightSection] {
```
fulmicoton, line 117: no need for inline always.

src/snippet/mod.rs
```diff
@@ -22,6 +22,12 @@ impl HighlightSection {
     fn new(start: usize, stop: usize) -> HighlightSection {
         HighlightSection { start, stop }
     }
+
+    /// Returns the bounds of the `HighlightSection`.
+    #[inline(always)]
```
fulmicoton: no need for inline always.

examples/snippet.rs
```diff
@@ -64,7 +64,27 @@ fn main() -> tantivy::Result<()> {
...
+
+fn highlight(snippet: Snippet) -> String {
+    let mut result = String::new();
+    let mut start_from = 0;
+
+    for (start, end) in snippet.highlighted().iter().map(|h| h.bounds()) {
+        result.push_str(&snippet.fragments()[start_from..start]);
+
+        result.push_str(" --> ");
+
```
fulmicoton: can you remove the empty lines?

examples/snippet.rs
```diff
@@ -64,7 +64,27 @@ fn main() -> tantivy::Result<()> {
         println!("Document score {}:", score);
         println!("title: {}", doc.get_first(title).unwrap().text().unwrap());
         println!("snippet: {}", snippet.to_html());
+        println!("custom highlighting: {}", highlight(snippet));
```
fulmicoton, line 67: Cool thanks for that.

fulmicoton:
LGTM. Added minor comments. Special thanks for completing the example.

PR #467, opened 2019-01-16
Closes 466. Removing mentions of the chain collector.

fulmicoton:


src/collector/multi_collector.rs
```diff
@@ -96,7 +96,10 @@ impl<TFruit: Fruit> FruitHandle<TFruit> {
 /// Multicollector makes it possible to collect on more than one collector.
 /// It should only be used for use cases where the Collector types is unknown
 /// at compile time.
-/// If the type of the collectors is known, you should prefer to use `ChainedCollector`.
+///
+/// If the type of the collectors is known, you can just group yours collectors
```
trinity-1686a, line 100: I think there is a typo, it should be `your` instead of `yours`

PR #473, opened 2019-01-21
Allow stemmers in languages other than English

### LINKED ISSUE #334: Stemming another languages
Hi! I found that there is already out of box fuzzy search, now there is the only need in morphological search to allow tantivy usage as a regular search engine for general purposes.
I found that you are using rust-stemmers library which already supports a lot of languages, but the only one allowed and used is English. What's wrong with another languages and is it possible to support another now?

pentlander:
Allow users to create stemmers for languages other than English. Add a
default stemmer for English. Closes #334.

### DIFF
--- a/src/tokenizer/mod.rs
+++ b/src/tokenizer/mod.rs
@@ -73,7 +73,7 @@
 //! let en_stem = SimpleTokenizer
 //!     .filter(RemoveLongFilter::limit(40))
 //!     .filter(LowerCaser)
-//!     .filter(Stemmer::new());
+//!     .filter(Stemmer::new(Language::English));
 //! # }
 //! ```
 //!
@@ -148,7 +148,7 @@ pub use self::ngram_tokenizer::NgramTokenizer;
 pub use self::raw_tokenizer::RawTokenizer;
 pub use self::remove_long::RemoveLongFilter;
 pub use self::simple_tokenizer::SimpleTokenizer;
-pub use self::stemmer::Stemmer;
+pub use self::stemmer::{Stemmer, Language};
 pub use self::stop_word_filter::StopWordFilter;
 pub(crate) use self::token_stream_chain::TokenStreamChain;
 pub(crate) use self::tokenizer::box_tokenizer;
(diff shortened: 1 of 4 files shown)


src/tokenizer/mod.rs
```diff
@@ -214,13 +222,41 @@ pub mod tests {
...
+                .filter(Stemmer::new(StemmerAlgorithm::Spanish)),
+        );
+        let en_tokenizer = tokenizer_manager.get("es_stem").unwrap();
+        let mut tokens: Vec<Token> = vec![];
+        {
+            let mut add_token = |token: &Token| {
+                tokens.push(token.clone());
+            };
+            en_tokenizer
+                .token_stream("Hola, feliz contribuyente!")
```
fulmicoton, line 250: Ha ha ha :)

src/tokenizer/stemmer.rs
```diff
@@ -1,9 +1,56 @@
...
+#![cfg_attr(feature = "cargo-clippy", allow(clippy::new_without_default, doc))]
 
 use super::{Token, TokenFilter, TokenStream};
 use rust_stemmers::{self, Algorithm};
 use std::sync::Arc;
 
+/// Available stemmer algorithms for different languanges.
+#[derive(Debug, Serialize, Deserialize, Eq, PartialEq, Copy, Clone)]
+#[allow(missing_docs)]
+pub enum StemmerAlgorithm {
```
fulmicoton: Maybe call that language?

pentlander: Will do

pentlander:
Should I do anything about the drop in code coverage? I'm not sure that adding a test for every language really adds value.

src/tokenizer/stemmer.rs
```diff
@@ -1,9 +1,56 @@
-#![cfg_attr(feature = "cargo-clippy", allow(clippy::new_without_default))]
+#![cfg_attr(feature = "cargo-clippy", allow(clippy::new_without_default, doc))]
```
fulmicoton: Do we need that?

fulmicoton (commented):
@pentlander Don't worry about the code coverage.

src/tokenizer/stemmer.rs
pentlander: Whoops, I didn't mean to leave that in, I'll remove it.

Issue #42, opened 2016-10-17
Implement date field type.


fulmicoton:

manuel-woelker:
What would implementation options be?

* u64: microsends since 1970
* String: ISO8601 i.e. "20170211T203212.323Z"  (160bits)

This type probably needs to support range queries, e.g. "only return events since midnight"

Possibly Relevant link:
https://www.elastic.co/blog/apache-lucene-numeric-filters

It seems Lucene uses block KD trees, which might also enable LatLon types.

fulmicoton:
@manuel-woelker sorry for the late answer.

We definitely want to use a 64 bit representation for that. Maybe i64 rather than u64 however?
We probably want to implement i64 or u64 and i64 fields and just wrap dates over such a field.
Then we should probably ditch the u32. Having a specific u32 field should not save any memory as tantivy use bitpacking in fastfields. (Implementing the reader in an efficient way might be a bit more challenging though)

Then comes another question which is how to make ranges searchable. 
I think this should a different ticket. 

I'll create another one.


Tantivy needs to support range for other numerical types as well, so the capacity to index a development over the integer.
I am unsure what we want here.

fulmicoton:
@manuel-woelker let me know if you want to work on these ticket.

PR #487, opened 2019-02-02
WIP: date field

barrotsteindev:
Initial version, still a work in progress.
An attempt to address #42

### DIFF
--- a/src/fastfield/multivalued/mod.rs
+++ b/src/fastfield/multivalued/mod.rs
@@ -7,7 +7,8 @@ pub use self::writer::MultiValueIntFastFieldWriter;
 #[cfg(test)]
 mod tests {
 
-    use std::time::{SystemTime, UNIX_EPOCH};
+    extern crate time;
+
     use query::QueryParser;
     use collector::TopDocs;
     use schema::Cardinality;
@@ -15,6 +16,7 @@ mod tests {
     use schema::IntOptions;
     use schema::Schema;
     use Index;
+    use self::time::Duration;
 
     #[test]
     fn test_multivalued_u64() {
@@ -51,13 +53,6 @@ mod tests {
         }
     }
 
-    fn get_current_time_stamp() -> i64 {
-        SystemTime::now()
-                               .duration_since(UNIX_EPOCH)
-                               .expect("could not get time stamp")
-                               .as_secs() as i64
-    }
-
     #[test]
     fn test_multivalued_date() {
         let mut schema_builder = Schema::builder();
@@ -76,13 +71,14 @@ mod tests {
         let schema = schema_builder.build();
         let index = Index::create_in_ram(schema);
         let mut index_writer = index.writer_with_num_threads(1, 3_000_000).unwrap();
-        let first_time_stamp = get_current_time_stamp();
-        index_writer.add_document(doc!(date_field=>get_current_time_stamp(), date_field=>get_current_time_stamp(), time_i=>1i64));
+        let first_time_stamp = chrono::Utc::now();
+        index_writer.add_document(doc!(date_field=>first_time_stamp, date_field=>first_time_stamp, time_i=>1i64));
         index_writer.add_document(doc!(time_i=>0i64));
         // add one second
-        index_writer.add_document(doc!(date_field=>first_time_stamp+1, time_i=>2i64));
+        index_writer.add_document(doc!(date_field=>first_time_stamp + Duration::seconds(1), time_i=>2i64));
         // add another second
-        index_writer.add_document(doc!(date_field=>first_time_stamp+2, date_field=>first_time_stamp+2,date_field=>first_time_stamp+2, time_i=>3i64));
+        let two_secs_ahead = first_time_stamp + Duration::seconds(2);
+        index_writer.add_document(doc!(date_field=>two_secs_ahead, date_field=>two_secs_ahead,date_field=>two_secs_ahead, time_i=>3i64));
         assert!(index_writer.commit().is_ok());
 
         index.load_searchers().unwrap();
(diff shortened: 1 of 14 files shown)


barrotsteindev:
This is far from done, or perfect.
Any help suggestion or building criticism is welcome.
I would like to get this up to par and tackle this issue.

fulmicoton:
@barrotsteindev if this is not already done, you might want to join the gitter chat. https://gitter.im/tantivy-search/tantivy ....

src/schema/document.rs
```diff
@@ -82,11 +82,16 @@ impl Document {
...
     }
 
-    /// Add a u64 field
+    /// Add a i64 field
     pub fn add_i64(&mut self, field: Field, value: i64) {
         self.add(FieldValue::new(field, Value::I64(value)));
     }
 
+    /// Add a date field
+    pub fn add_date(&mut self, field: Field, value: i64) {
```
fulmicoton: Internally we want to handle it as a `i64`, but the API could expose something nicer to the user. 

Ideally that would have been a `DateTime` , but that's in the `chrono` trait, and it is generally not a good idea to use object in libraries within our library (it can lead to dependency hell). Let's ask for advise on the rust user forum what is the best practise for that.

fulmicoton:
@barrotsteindev Repeating myself. As is, the `DateField` brings little benefit to the user as it is just a `i64` field with a differet name. Internally, I think this is what we want, but in the API the user should be able to give a date object that we convert right away.

Another nice refinement would be to handle automatically facetting for dates (`year/ month/day/hour/mm`), but that does not need to be part of the scope of this ticket.

The format the user should use to feed dates into tantivy is tricky.
`DateTime` seems natural but it is part of the `chrono` trait and it might not be a good idea to expose another crate in our API. Let's ask for the best practise in the user's community forum.

We also want to parse dates in the query parser. This might be tricky too, considering there are many possible date format, and we probably want to let people truncate them. For the scope of this ticket, we can probably stick to `ISO 8601`. I wouldn't be surprised if someone already made a crate to parse dates automatically. (Well some format are ambiguous, so maybe not)

fulmicoton:
https://users.rust-lang.org/t/advise-for-an-api-involving-dates/24949

barrotsteindev:
After reading the answer on rust-lang, I guess the game plan would be:

1. Adding chrono as a dependeny
2. Exposing it as tantivy::chrono
3. Changing Date Field's value type to tantivy::chorno and parsing it as an i64 at index time.

Does this sound alright @fulmicoton ?

barrotsteindev:
In which module would Chrono be exported?
I am kind of lost, as I am fairly new to rust.

barrotsteindev:
Rebased on latest master after adding chrono::DateTime as input for date fields

barrotsteindev:
@fulmicoton , would you be able to take another look at this?

src/lib.rs
```diff
@@ -187,10 +187,14 @@ pub use error::TantivyError as Error;
 
 extern crate census;
 extern crate owned_read;
+extern crate chrono;
 
 /// Tantivy result.
 pub type Result<T> = std::result::Result<T, error::TantivyError>;
 
+/// Tantivy DateTime
+pub type DateTime<Tz> = chrono::DateTime<Tz>;
```
fulmicoton: I think we should force UTC?

src/lib.rs
```diff
@@ -187,10 +187,14 @@ pub use error::TantivyError as Error;
 
 extern crate census;
 extern crate owned_read;
+extern crate chrono;
```
fulmicoton: I see you reexport the type. I think it might be useful to also reexport the entire crate as `tantivy::chrono`.

src/query/query_parser/query_parser.rs
```diff
@@ -229,6 +237,12 @@ impl QueryParser {
                 let term = Term::from_field_i64(field, val);
                 Ok(vec![(0, term)])
             }
+            FieldType::Date(_) => {
+                match chrono::DateTime::parse_from_rfc3339(phrase) {
```
fulmicoton, line 243: Can we document that rfc3339 is used in the `QueryParser` ?

fulmicoton (commented):
Please add some doc in `schema.rs`. You can mention implementation details such as "Internally, tantivy simply stores dates as i64 UTC timestamp"

fulmicoton:
@barrotsteindev Thanks for pinging me. I had totally forgot. I added some minor comments.

Can you also :
- add your feature in the CHANGELOG.md, ideally as part of this PR. It does not need to be long. Something like "- Added DateTime field (@barrotsteindev)" . I think we will ship that in 0.9.
- create an issue to implement range query for dates

src/lib.rs
barrotsteindev: Makes sense,
I'll add that.

barrotsteindev:
rebased on latest master, with date field now supporting INDEXED and STORED instead of INT_ prefixed variants.

src/fastfield/multivalued/mod.rs
```diff
@@ -48,6 +53,104 @@ mod tests {
...
+
+            for (_score, doc_address) in results {
+                let retrieved_doc = searcher.doc(doc_address).expect("cannot fetch doc");
+                assert_eq!(retrieved_doc.get_first(date_field).expect("cannot find value").date_value().timestamp(), two_secs_ahead.timestamp());
+                assert_eq!(retrieved_doc.get_first(time_i).expect("cannot find value").i64_value(), 3i64);
+            }
+        }
+
+
+        // TODO: support Date range queries
```
barrotsteindev, line 121: This test should pass once range queries are implemented for date field #516

PR #490, opened 2019-02-05
Add integer range search example

petr-tik:
Copied and simplified the example from the range_query mod

Related to #372


fulmicoton:
thx

PR #491, opened 2019-02-05
Replace MsQueue with SegQueue

petr-tik:
MsQueue was removed from the newest version of crossbeam - SegQueue is another
MPMC option that is supported.

TBH, I am confused as to you self.queue.pop().unwrap() works, but all tests
passed locally for me and I wanted to run a bigger travisCI job to see if I am
missing anything

add assert_matches to dev dependecies to make sure it builds locally.

Based on https://github.com/tantivy-search/tantivy/pull/485


src/core/pool.rs
```diff
@@ -64,7 +64,7 @@ impl<T> Pool<T> {
     pub fn acquire(&self) -> LeasedItem<T> {
         let generation = self.generation();
         loop {
-            let gen_item = self.queue.pop();
+            let gen_item = self.queue.pop().unwrap();
```
fulmicoton, line 110: Reading the documentation, the semantics are very different.

When the queue is empty, MsQueue was blocking on `.pop()`. `SegQueue`, on the other hand returns an error when it is empty.

We really want the earlier behavior here. The fact that it was not caught by unit test means... that we lack unit test :).
Possible way to test that. Create a pool with 2 elements. Have three threads pick an element and sleep for `1ms`. This should not panic.

fulmicoton:
I asked the following question in crossbeam.

fulmicoton:
`crossbeam-channel` has the right semantics.

Cargo.toml
```diff
@@ -58,6 +58,7 @@ winapi = "0.2"
 [dev-dependencies]
 rand = "0.6"
 maplit = "1"
+assert_matches = "1.3"
```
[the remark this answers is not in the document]
petr-tik: My build was failing without assert_matches with 11 errors like below

```md
error[E0433]: failed to resolve: could not find `panic` in `$crate`
--> src/schema/schema.rs:538:13
    |
538 |             assert_matches!(json_err, Err(NotJSON(_)));
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ could not find `panic` in `$crate`

```

Grepping showed that assert_matches is used in tests only, hence dev-dependency

rustc --version
rustc 1.32.0 (9fda7c223 2019-01-16)
cargo --version
cargo 1.32.0 (8610973aa 2019-01-02)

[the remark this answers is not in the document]
fulmicoton: I was importing it from the `matches` crate. Maybe they added a bug in their minor version? Can you investigate?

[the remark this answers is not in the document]
petr-tik: Ha! I was grepping for assert, not matches. Thanks for pointing out. 

Looks like there was an issue. 
https://github.com/SimonSapin/rust-std-candidates/issues/15

Closed in matches 0.1.8 - will update and see if it compiles

Cargo.toml
```diff
@@ -41,7 +41,7 @@ owning_ref = "0.4"
 stable_deref_trait = "1.0.0"
 rust-stemmers = "1"
 downcast-rs = { version="1.0" }
-matches = "0.1"
+matches = "0.1.8"
```
fulmicoton: Was the fix in `matches` or in `rustc` ? If `rustc` let's not fix the minor.

src/core/executor.rs
```diff
@@ -123,15 +123,14 @@ mod tests {
...
-#[test]
-fn test_map_multithread() {
-    let result: Vec<usize> = Executor::multi_thread(3, "search-test")
-        .map(|i| Ok(i * 2), 0..10)
-        .unwrap();
-    assert_eq!(result.len(), 10);
-    for i in 0..10 {
-        assert_eq!(result[i], i * 2);
+    #[test]
+    fn test_map_multithread() {
```
fulmicoton, line 127: oops thanks for that

petr-tik:
For me, the errors went away when I updated the minor version of matches.

I have the same version of rustc as before. Not sure how else rustc could
fix it. Would be great to get someone else's input.

fulmicoton:
> For me, the errors went away when I updated the minor version of matches. I have the same version of rustc as before. Not sure how else rustc could fix it. Would be great to get someone else's input.


Actually I'm ok with the minor here.

Cargo.toml
fulmicoton: ok with fixing the patch here. Can you move it as dev-dependency?

petr-tik: Will do

src/core/pool.rs
petr-tik: I have been thinking about how to test a blocking call. I have a commented out test below, trying to follow your logic, but I haven't found a good way of testing it

petr-tik:
I would be happy to rebase my commits, if it helps git log

fulmicoton:
@petr-tik  I can't review your code. It includes commits from the main branch... Did you rebase master into your branch or something like that?

petr-tik:
I merged upstream into my branch, since the head moved on. 

Also, I now realise that the PR is against the dependabot branch.

Do you prefer that I rebase with force or open a new PR?

fulmicoton:
Please open a new PR

Issue #492, opened 2019-02-06
Add groups of operations


fulmicoton:
**Is your feature request related to a problem? Please describe.**
People may want to group operations atomically.
One example from @elbow-jason is maintaining a unique index. 
Updating a doc consists of a `delete + insert`. In a scenario where, for some reason, the `IndexWriter` is shared between thread with a `Arc<RwLock>`, an update happening on two different thread could end up executed as `delete delete insert insert`. The final state being two docs with the same primary id.

Another use case would be to enforce a set of documents to belong to the same segment in order to enable some collector optimization. (For instance, docs with a parent/child relationship)

**Describe the solution you'd like**

Add a `UserOperation` enum in `src/indexer/operation.rs` that would contain both Add and Delete without any opstamp.

Add a method named `IndexWrite.run(&[Operation])` to run a list of operations in a grouped fashion.
These operations must get contiguous opstamp. (So the stamper must be modified to make it possible to allocate a range of ids).

The deletes are queued normally, but the adds need to happen on the same segment for the second use case. Right now the work is dispatched to the `SegmentWriter` using a channel of AddOperation.
The item should become `Vec<AddOperation>`. 

`IndexWriter.add_document` and `IndexWriter.delete_term` remain unchanged from the user point of view. They should probably rely on the `.run(&[singleton])`  internally however.

elbow-jason:
Closing as this was added in #493

PR #493, opened 2019-02-06
Add grouped operations

elbow-jason:
This PR addresses #492.

### todos

- [X] add `UserOperation` enum in `src/indexer/operation.rs`
- [X] add a method named `IndexWrite.run(&[Operation])`
- [X] change `AddOperation` channel to instead take `Vec<AddOperation>`
- [x] Add correct and meaningful tests

Tests seem to be passing, but I am not super confident about the changes I made WRT the heap_size check and the `delete_cursor.skip_to` call.


src/indexer/stamper.rs
```diff
@@ -60,12 +61,43 @@ impl Stamper {
     pub fn stamp(&self) -> u64 {
         self.0.fetch_add(1u64, Ordering::SeqCst) as u64
     }
+
+    /// Given a desired count `n`, `stamps` returns an iterator that
+    /// will supply `n` number of u64 stamps.
+    pub fn stamps(&self, n: u64) -> MultiStamp {
```
fulmicoton: Could we simply return a `std::ops::Range`?

elbow-jason: Absolutely! I had not considered `Range` here. I'll make the changes tonight.

src/indexer/index_writer.rs
```diff
@@ -643,12 +647,45 @@ impl IndexWriter {
...
+                    let delete_operation = DeleteOperation {
+                        opstamp: opstamp,
+                        term: term.clone(),
+                    };
+                    self.delete_queue.push(delete_operation);
+                }
+                UserOperation::Add(doc) => {
+                    let add_operation = AddOperation {
+                        opstamp: opstamp,
+                        document: doc.clone(),
```
fulmicoton: Ah... Since we borrow `UserOperation` we end up needing to clone docs. Should we take a `Vec<UserOperation>` as argument then?

src/indexer/index_writer.rs
```diff
@@ -643,12 +647,45 @@ impl IndexWriter {
...
         let add_operation = AddOperation { opstamp, document };
-        let send_result = self.document_sender.send(add_operation);
+        let send_result = self.document_sender.send(vec![add_operation]);
         if let Err(e) = send_result {
             panic!("Failed to index document. Sending to indexing channel failed. This probably means all of the indexing threads have panicked. {:?}", e);
         }
         opstamp
     }
+
+    /// Runs a group of document operations.
```
fulmicoton: Document what happens if someone runs an empty `array`  of operations.

src/indexer/index_writer.rs
```diff
@@ -643,12 +647,45 @@ impl IndexWriter {
...
+    pub fn run(&mut self, user_operations: &[UserOperation]) -> Option<u64> {
+        let mut stamps = self.stamper.stamps(user_operations.len() as u64);
```
fulmicoton: Opstamp are free, I suggest allocation `user_operations.len() + 1`  operations and give the last one to the "batch".

That way you can change the `Option<u64>`  to `u64`. (Even empty operations return an opstamp).
There might be value (maybe?) for the batch and the last operation to have a different opstamp too.

I also allocate opstamp to operations like `commit` etc. I suspect it might help people not having to think too much about the `n` vs `n+1` problem. 

If you choose this path, you can set last opstamp to the last opstamp of the range directly and `.zip(...)` the iterators.

src/indexer/index_writer.rs
fulmicoton: (or `Vec<>` if we end up using `Vec`)

src/indexer/stamper.rs
fulmicoton: Cool

src/indexer/index_writer.rs
fulmicoton: We want at least 2 very simple unit test. One for when empty operations. One for non-empty.

src/indexer/stamper.rs
```diff
@@ -78,5 +110,28 @@ mod test {
 
         assert_eq!(stamper.stamp(), 10u64);
         assert_eq!(stamper_clone.stamp(), 11u64);
+
```
fulmicoton: Once you use Range you can simply this unit test and `assert_eq!` on the range itself, and remove the other test.

src/indexer/index_writer.rs
elbow-jason: I think using `Vec<UserOperation>` is a better alternative. It is unlikely that the `UserOperation`s that are passed to `IndexWriter.run` will be needed by the caller after. I'll make the changes.

src/indexer/index_writer.rs
elbow-jason: `There might be value (maybe?) for the batch and the last operation to have a different opstamp too.`
I think so. This "extra" opstamp could give delimitation to a group of operations. A sort of forensics? 

I'll make the changes.

src/indexer/index_writer.rs
elbow-jason: I added some tests, but they are only asserting that the opstamps returned by `run` are correct. There are no other assertions.

fulmicoton: That's fine

src/indexer/index_writer.rs
```diff
@@ -643,17 +648,86 @@ impl IndexWriter {
...
+        let stamps = Range {
+            start: start,
+            end: last_opstamp,
+        };
+        (last_opstamp, stamps)
+    }
+
+    /// Runs a group of document operations ensuring that the operations are
+    /// assigned contigous u64 opstamps and that add operations of the same
+    /// group are flushed into the same segment.
```
fulmicoton, line 677: great comment.

src/indexer/index_writer.rs
```diff
@@ -643,17 +648,86 @@ impl IndexWriter {
...
+    /// Each operation of the given `user_operations` will receive an in-order,
+    /// contiguous u64 opstamp. The entire batch itself is also given an
+    /// opstamp that is 1 greater than the last given operation. This 
+    /// `batch_opstamp` is the return value of `run`. An empty group of
+    /// `user_operations`, an empty `Vec<UserOperation>`, still receives
+    /// a valid opstamp even though no changes were _actually_ made to the index.
+    /// 
+    /// Like adds and deletes (see `IndexWriter.add_document` and
+    /// `IndexWriter.delete_term`), the changes made by calling `run` will be
+    /// visible to readers only after calling `commit()`.
```
fulmicoton, line 690: Perfect!

src/indexer/index_writer.rs
```diff
@@ -662,6 +736,33 @@ mod tests {
     use Index;
     use Term;
 
+    #[test]
+    fn test_operations_group() {
+        let mut schema_builder = schema::Schema::builder();
+        let text_field = schema_builder.add_text_field("text", schema::TEXT);
+        let index = Index::create_in_ram(schema_builder.build());
+        let mut index_writer = index.writer(3_000_000).unwrap();
```
fulmicoton: by default, tantivy uses several indexing threads. Can you force it to 1 in these unit tests?... Or 2?

As is, the unit test will fail on CPUs with a lot of cores, becasue the amount of allocated memory per thread will be lower than the lowest allowed amount.

elbow-jason:
By changing `add_document` and `delete_term` to use `run` and having `run` return the `opstamp` of the batch, I think `add_document` and `delete_term` will be returning 1 more than they used to. Should I remove the +1? 
Something like:
```rust 
pub fn add_document(&mut self, document: Document) -> u64 {
        self.run(vec![UserOperation::Add(document)]) - 1
    }
```

fulmicoton:
One minor comment on unit test that needs to be addressed. Good work!

src/indexer/index_writer.rs
elbow-jason: :-)

src/indexer/index_writer.rs
elbow-jason: I do not know how to force it to use fewer threads. How would this be done?

src/indexer/index_writer.rs
```diff
@@ -409,8 +410,12 @@ impl IndexWriter {
                     // this is a valid guarantee as the
                     // peeked document now belongs to
                     // our local iterator.
-                    if let Some(operation) = document_iterator.peek() {
-                        delete_cursor.skip_to(operation.opstamp);
+                    if let Some(operations) = document_iterator.peek() {
+                        if let Some(last) = operations.last() {
+                            delete_cursor.skip_to(last.opstamp);
```
fulmicoton: No we want to skip to the first opstamp. Not the last one.

elbow-jason: I was worried about this code. hehe. I do not actually understand what we are doing in this block.
Does this just move the cursor to align with the about-to-be-written operation(s)?

fulmicoton: It might be nice to add a unit test to check that this is correct.

Basically, batch the following operations :
- one delete for `doc!(field=>"a")`
- one add for `doc!(field=>"a")`
- one add for `doc!(field=>"b")`
- one delete for `doc!(field=>"b")`

Assert that after commit there is one doc with "a" and 0 doc with "b"

elbow-jason: So that previous operations can be deleted after this?

fulmicoton:
> By changing `add_document` and `delete_term` to use `run` and having `run` return the `opstamp` of the batch, I think `add_document` and `delete_term` will be returning 1 more than they used to. Should I remove the +1?

Let's not reuse `run` then and keep the code as is.

src/indexer/index_writer.rs
fulmicoton: > So that previous operations can be deleted after this?

I don't understand the question.

fulmicoton: > I was worried about this code. hehe. I do not actually understand what we are doing in this block.
> Does this just move the cursor to align with the about-to-be-written operation(s)?

Deletes are very tricky. They need to affect all existing segments, as well as segments currently being built. The way it works is as follows : there is a queue containing delete operations.

That queue is "owned" by different object and each of them can be pointing at a different place in that queue, depending on how up to date they are.

Segment being written are not impacted by deletes that happened before their creation, hence the `.skip_to`.

They are however impacted by deletes that happen after, and the segment writer keeps track of the opstamp that created each document in the segment, so that when we apply the deletes to the brand new segment , we can actually check that the delete operation happened before the document was added, and that we should indeed mark the document as deleted.

elbow-jason: Sorry! My question probably makes no sense because I don't know what is going on in the `delete_cursor`. I'll look at the code and hopefully my questions will be answered or at least I will know what questions to ask next.

src/indexer/index_writer.rs
fulmicoton: there is a method called something like ".write_with_num_threads(1, 3_000_000)" or something like that. You can check other unit tests. Most of them should be calling it.

src/indexer/index_writer.rs
fulmicoton: Yes it is hard. I wrote an explanation right above.

elbow-jason: Excellent explanation! So, if I understand the concept correctly, `skip_to` moves the delete cursor forward "in time" to the "edge" of the about-to-be-written docs, right?

fulmicoton: Precisely yes.

fulmicoton:
> Tests seem to be passing, but I am not super confident about the changes I made WRT the heap_size check and the `delete_cursor.skip_to` call.

What did you mean by your change on `heap_size`?

elbow-jason:
> What did you mean by your change on `heap_size`?

I move [this code](https://github.com/tantivy-search/tantivy/blob/b16b2a6b9af1273ab58e2d47498b3b5583f612ea/src/indexer/index_writer.rs#L282-L288) to check for mem usage after each batch of documents instead of after each document.

I don't know where I got `heap_size` from.

fulmicoton:
Great job thanks you!

PR #495, opened 2019-02-12
MsQueue to channel

petr-tik:
Implements own Queue using crossbeam-channel. Relies on blocking behaviour of the pop() method

Related to #485 and #491

### DIFF
--- a/src/core/pool.rs
+++ b/src/core/pool.rs
@@ -107,7 +107,7 @@ impl<T> Pool<T> {
     pub fn acquire(&self) -> LeasedItem<T> {
         let generation = self.generation();
         loop {
-            let gen_item = self.queue.pop();
+            let gen_item = self.queue.pop().unwrap();
             if gen_item.generation >= generation {
                 return LeasedItem {
                     gen_item: Some(gen_item),
--- a/src/core/pool.rs
+++ b/src/core/pool.rs
@@ -10,7 +10,7 @@ pub struct GenerationItem<T> {
     item: T,
 }
 
-/// Queue to replace the Michael-Scott Queue, now deprecated in crossbeam
+/// Queue implementation for the Object Pool below
 /// Uses the unbounded Linked-List type queue from crossbeam-channel
 /// Splits the Queue into sender and receiver
 struct Queue<T> {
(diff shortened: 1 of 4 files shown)


src/core/pool.rs
```diff
@@ -146,4 +177,45 @@ mod tests {
...
+        let q = Queue::new();
+        let elem = 5;
+        q.push(elem);
+        let res = q.pop();
+        assert_eq!(res.unwrap(), elem);
+    }
+
+    #[test]
+    fn test_queue_pushes_and_pops_order() {
+        let elements = vec![1, 2, 3, 4];
```
fulmicoton: There is no need to test for order actually.

src/core/pool.rs
```diff
@@ -146,4 +177,45 @@ mod tests {
...
+    //         let elements_for_pool = vec![1, 2];
+    //         // Clone vector, so we can calculate its length later
+    //         pool.publish_new_generation(elements_for_pool.clone());
+
+    //         let mut threads = vec![];
+    //         let sleep_dur = time::Duration::from_millis(10);
+    //         // spawn one more thread than there are elements in the pool
+    //         for _thread_idx in 0..elements_for_pool.len() + 1 {
+    //             threads.push(thread::spawn(move || {
+    //                 pool.acquire();
```
fulmicoton: The point of this test is to see if the pool behaves correctly when the pool is exhausted.

Right now, your leased item is dropped  right away (before the call to thread::sleep()).
Please fix as follows:

`let _leased_searcher = pool.acquired();`

fulmicoton (changes requested):

petr-tik:
What about RecvError and dealing with Result from .send()?

fulmicoton:
> What about RecvError and dealing with Result from .send()?

According to doc, they only exist to express that the respective other side of the pipe was dropped.
By design it cannot happen in that pool, so it's ok for you to `panic` on errors.

Just use `.expect()` in both cases.

src/core/pool.rs
```diff
@@ -9,21 +10,50 @@ pub struct GenerationItem<T> {
     item: T,
 }
 
+/// Queue to replace the Michael-Scott Queue, now deprecated in crossbeam
```
fulmicoton: no need for this comment. People don't need to know the history of this queue.

src/core/pool.rs
```diff
@@ -9,21 +10,50 @@ pub struct GenerationItem<T> {
...
+        Queue {
+            sender: s,
+            receiver: r,
+        }
+    }
+
+    /// Sender trait returns a Result type, which is ignored.
+    /// The Result is not handled at the moment
+    fn push(&self, elem: T) {
+        self.sender.send(elem);
```
fulmicoton: This generates a warning. Make sure you remove all warning at compilation.

petr-tik:
Rebased onto master and _force_-pushed (sorry). 

thanks for your comments, I addressed them to the best of my ability.

fulmicoton:
Thanks!

PR #503, opened 2019-02-23
Remove note about panicking in get_field docs

mauri870:
Since get_field rely on calling get on the underlying InnerSchema HashMap
it shouldn't fail if the field was not found, it simply returns None.


fulmicoton:
@mauri870  Absolutely yes. I think it used to not return an option, and I forgot to update the doc when I changed the implementaiton.

PR #504, opened 2019-02-24
prove bug with facets

jannickj:


fulmicoton:
Tracked in #505. Tested added as part of #508

PR #511, opened 2019-02-28
Derive `Debug` for `Schema` and `SchemaInner`

torkleyy:


fulmicoton:
Wouldn't it be more helpful to rely on a prettified json serialization?

torkleyy:
Yeah, probably. I mainly added to have some representation at all.

PR #512, opened 2019-03-04
WIP: compiling to wasm

jonfk:
I started an experiment to compile tantivy to wasm. These are some of the features I had to adjust to get it to compile. 

The index writing doesn't work because it seems to be using std lib features that aren't supported by wasm target yet, such as `std::thread`s. I can see from https://github.com/fulmicoton/tantivy-wasm that tantivy seems to have worked with wasm at some point with `StaticDirectory` which confirms that at least the index writing never worked in wasm.

Although there is no `StaticDirectory` in the current master, I don't see the `Directory` trait having changed since when it existed, so I could add it back. But I am a little stumped at how the `DATA: &'static [u8]` was generated. Would appreciate any pointers

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,7 +72,7 @@ overflow-checks = true
 [features]
 # by default no-fail is disabled. We manually enable it when running test.
 default = ["mmap", "no_fail"]
-mmap = ["atomicwrites", "fs2", "memmap"]
+mmap = ["atomicwrites", "fs2", "memmap", "fst-regex/mmap"]
 lz4-compression = ["lz4"]
 no_fail = ["fail/no_fail"]
 unstable = [] # useful for benches.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,8 +17,7 @@ byteorder = "1.0"
 lazy_static = "1"
 regex = "1.0"
 tantivy-fst = "0.1"
-memmap = { version="0.7", optional = true }
-fst-regex = { version = "0.2", features = [], default-features = false }
+memmap = {version = "0.7", optional=true}
 lz4 = {version="1.20", optional=true}
 snap = {version="0.2"}
 atomicwrites = {version="0.2.2", optional=true}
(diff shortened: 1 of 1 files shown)


fulmicoton:
> The index writing doesn't work because it seems to be using std lib features that aren't supported by wasm target yet, such as std::threads. I can see from https://github.com/fulmicoton/tantivy-wasm that tantivy seems to have worked with wasm at some point with StaticDirectory which confirms that at least the index writing never worked in wasm.

Yes this is correct.

So for my small experiment, I just:
- indexed things using a non-wasm program
- merged all of the segments together.
- wrote a small program to put the resulting segment in the static directory format. 
- included the resulting index file in the WASM program using `include_bytes`

(Of course this is not a good solution )

Cargo.toml
```diff
@@ -72,10 +72,11 @@ overflow-checks = true
 [features]
 # by default no-fail is disabled. We manually enable it when running test.
 default = ["mmap", "no_fail"]
-mmap = ["atomicwrites", "fs2"]
+mmap = ["atomicwrites", "fs2", "memmap", "fst-regex/mmap"]
```
fulmicoton: why do we need `fst-regex/mmap` ?

petr-tik:
Hey @jonfk - thanks for the PR. 

I have been hacking away on a WIP branch in my own fork. I followed Paul's earlier work on it, but haven't tracked upstream for a while. Have a look and let me know, if any of it is useful. 

https://github.com/petr-tik/tantivy/commits/compiling_for_wasm

fulmicoton:
Thanks @jonk, I made some change your PR and merged it.

jonfk:
Thanks @fulmicoton for the quick work. 

@petr-tik I will take a look at your branch. It seems like you are using a similar method to what @fulmicoton was using to generate the data to be served to the wasm. If we could add a serialization and deserialization scheme for the index so that we could have a better way to serve the index to the wasm.

fulmicoton:
> If we could add a serialization and deserialization scheme for the index so that we could have a better way to serve the index to the wasm.

Can you be as explicit as possible about what you mean by serialization and deserialization scheme for the index?

Also do you need to index in web assembly or are you trying to serve a static index?

jonfk:
@fulmicoton Sorry for being vague about a serialization/deserialization scheme, it's just because I am not very knowledgeable about the inner workings of tantivy. 

What I meant is some kind of defined format to transport the index. It could be json, protobuf, etc... which would allow some kind of client (here wasm) that could serve as the read only querying side of the index. 

I don't know if that would be possible without breaking changes over time, which would really be a disadvantage of defining such a format.

My idea would be to implement a read-only Directory for wasm that would use indexeddb as the backing store to serve search queries.

PR #521, opened 2019-03-26
Fix non english stemmers

takistakis:
The non english stemmers were not working because `transform` was explicitly
using `Algorithm::English`

I think that it was not caught on testing, because the english stemmer happened
to work correctly for the spanish phrase "Hola, feliz contribuyente!".

So, I replaced the test with a greek phrase.

I also didn't understand why the `stemmer_algorithm` was wrapped in an `Arc`.

I removed it for simplicity, but if it was there for a reason, I could add it
back.


fulmicoton:
Ouch... thanks.

> I also didn't understand why the stemmer_algorithm was w.rapped in an Arc.

Maybe because we cannot ensure it will remain a small `enum`?
Probably no reason.

PR #524, opened 2019-04-11
Enables clearing the index

### LINKED ISSUE #510: Add a method to clear the index

I'd like to be able to rebuild the index.

It would be convenient to just call `index.clear()` and re-submit all the documents.

**Describe alternatives you've considered**

Alternatively, one could just drop all the `tantivy` resources and delete the directory.

petr-tik:
Closes #510


petr-tik:
@torkleyy it would be great to get your thoughts on the right semantics of the `clear` method. 

I have added an example, trying to guess how you would rebuild an index - let me know

examples/clear_and_rebuild_index.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    let mut index_writer = index.writer_with_num_threads(1, 50_000_000)?;
+    index_writer.add_document(doc!(title => "The Old Man and the Sea"));
+    index_writer.add_document(doc!(title => "The modern Promotheus"));
+    index_writer.commit()?;
+
+    // clears the index
+    index_writer.clear()?;
+    // have to commit, otherwise deleted terms remain available
+    index_writer.commit()?;
```
petr-tik: @fulmicoton @torkleyy 

Should clear have an implicit/hidden commit at the end of it or should we make users commit manually? 

The current implementation makes the user do it, otherwise old entries show up.

I am leaning towards hiding/doing the commit for the caller to help them fall into the [pit of success](https://blog.codinghorror.com/falling-into-the-pit-of-success/) instead of forgetting to do it and finding old files.

fulmicoton: I don't think we need an example for that one. We could have a code sample in the rustdoc though.

fulmicoton: The benefit of having an extra commit step is that people is :
- you can write something where no user will ever see the state where your index was empty.
- you can write code that removes all segments, adds new documents (typically a full reindexing) coming from an external source. (e.g. a KV store like Cassandra for instance). If this external source fails, you can abort the process without having lost everything.

I think requiring a commit would be nice.

src/indexer/index_writer.rs
```diff
@@ -461,6 +461,15 @@ impl IndexWriter {
         self.segment_updater.garbage_collect_files()
     }
 
+    /// Clear the index
+    /// Enables users to rebuild the index,
+    /// by clearing and resubmitting necessary documents
+    /// Garbage collects unused files after removing segments
+    pub fn clear(&mut self) -> Result<()> {
+        self.segment_updater.remove_all_segments();
```
fulmicoton, line 504: Does garbage collection do anything if you have not committed?

src/indexer/index_writer.rs
```diff
@@ -461,6 +461,15 @@ impl IndexWriter {
         self.segment_updater.garbage_collect_files()
     }
 
+    /// Clear the index
+    /// Enables users to rebuild the index,
+    /// by clearing and resubmitting necessary documents
+    /// Garbage collects unused files after removing segments
```
fulmicoton: Clear -> Clears

Add a line about the fact that this operation, like all others requires `commiting`.

> Garbage collects unused files after removing segments

I would write something like 
"This operation removes all existing segments from the index.
The files associated to these segments will be garbage collected upon `.commit()`."

src/indexer/index_writer.rs
```diff
@@ -1048,4 +1057,21 @@ mod tests {
...
+        // writing the segment
+        let mut index_writer = index.writer_with_num_threads(4, 12_000_000).unwrap();
+        {
+            // create 8 segments with 100 tiny docs
+            for _doc in 0..100 {
+                index_writer.add_document(doc!(text_field => "a"));
+            }
+        }
+        let res = index_writer.clear();
+        assert!(res.is_ok());
```
fulmicoton, line 1179: This does not test all much here.

src/indexer/index_writer.rs
```diff
@@ -1048,4 +1057,21 @@ mod tests {
         assert_eq!(num_docs_containing("b"), 0);
         fail::cfg("RAMDirectory::atomic_write", "off").unwrap();
     }
+
+    #[test]
```
fulmicoton, line 1103: missing unit test.


#  test_add_then_clear.
What happens if within the same commit :
```
1) I add a doc
2) I clear the index
3) I commit
```
I would expect the resulting index to contain no docs. (I don't think this is the case right now).
Check that the opstamps are ever increasing (add opstamp < clear opstamp < commit opstamp) 

# test_clear_then_add

What happens if within a commit :
```
1) I clear the index
2 ) I add a doc
3) I commit
```
I would expect the document to be here.

# test_clear_and_rollback

```
1) add docs
2) commit
3) I clear the index
4) I rolback
```
I would expect the index to contain original docs.

fulmicoton: `test_add_then_clear` might be tricky to implement. 

The most efficient way to do it might be to `rollback`.
For tracing purpose, I'd prefer if you pulled all of the code except logging into another private method of IndexWriter,
Something like `rollback_impl()` and call `rollback_impl()` at the beginning of `.clear()` and in `.rollback()`.

So `fn rollback(&mut self)` would look something like 

```rust
fn rollback(&mut self) {
     info!("Rolling back to opstamp {}", self.committed_opstamp);
     self.rollback_impl();
     Ok(())
}
```

One thing that sucks with this implementation, is that we reset the opstamp counter too. We might need a bit of refactoring to do things properly.

src/indexer/index_writer.rs
```diff
@@ -461,6 +461,15 @@ impl IndexWriter {
         self.segment_updater.garbage_collect_files()
     }
 
+    /// Clear the index
+    /// Enables users to rebuild the index,
+    /// by clearing and resubmitting necessary documents
+    /// Garbage collects unused files after removing segments
+    pub fn clear(&mut self) -> Result<()> {
```
fulmicoton: Return an `opstamp` associated to this operation.

petr-tik: Will do. I want to clarify something, before I spend any time writing regexes and sending a chunky commit that touches many files. 

I was thinking of type aliasing u64 as TStamp or OpStamp
```rust
pub type Opstamp = u64;
```

I think it will improve readability instead of seeing methods like 
https://github.com/tantivy-search/tantivy/blob/28323bbb4395278da46e80c3bf1e0fae4783eeef/src/indexer/index_writer.rs#L629-L634

to get 

```rust
pub fn delete_term(&mut self, term: Term) -> Opstamp 
```

I know IDEs and lookups can help, but I thought aliasing is an easy readability improvement that I can make in this commit.

Thoughts?

torkleyy:
@petr-tik The semantics look good to me, thanks!

src/indexer/index_writer.rs
fulmicoton: I'm ok with introducing a type alias for Opstamp

fulmicoton:
@petr-tik please ping me when you want me to review your code again!

examples/clear_and_rebuild_index.rs
```diff
@@ -0,0 +1,63 @@
...
+    let index = Index::create_in_ram(schema.clone());
+
+    let mut index_writer = index.writer_with_num_threads(1, 50_000_000)?;
+    index_writer.add_document(doc!(title => "The Old Man and the Sea"));
+    index_writer.add_document(doc!(title => "The modern Promotheus"));
+    index_writer.commit()?;
+
+    // clears the index
+    index_writer.clear()?;
+    // have to commit, otherwise deleted terms remain available
```
fulmicoton: I don't understand the comment about deleted_terms.

src/indexer/index_writer.rs
fulmicoton: The typo is still here.

petr-tik:
I think I fixed it for the case of  `cargo test -- --test-threads 1`. 

However, `test_add_then_delete_all_documents` fails _sometimes_ when I run it with --test-threads 2. 

Adding `reader.reload().unwrap()` makes tests pass. I am deep into the weeds of multi-threading and `cargo test` execution model. Would appreciate a review and any advice

thanks

petr-tik:
I would be happy to squash merge, as too many commits here are WIP and merges. 

Keen to get your review, when you have time

src/indexer/index_writer.rs
fulmicoton: This comment was not addressed.
GC has no effect here.

src/indexer/index_writer.rs
```diff
@@ -462,6 +461,54 @@ impl IndexWriter {
...
+    ///     Ok(())
+    /// }
+    /// ```
+    pub fn delete_all_documents(&mut self) -> Result<Opstamp> {
+        // Delete segments
+        self.segment_updater.remove_all_segments();
+        // Garbage collect
+        self.garbage_collect_files()?;
+        // Return new stamp - reverted stamp
+        self.stamper.revert(self.committed_opstamp);
```
fulmicoton: this does not sound like a good idea. the opstamp is great to take track of where we are along a log.

fulmicoton, line 508: I need to think of what is the right behavior here. Do we want this stamp setting to happen in the `rollback` function.

Shouldn't it stay incremental and should rollback and clear index get their own opstamp?

src/indexer/index_writer.rs
```diff
@@ -1049,4 +1099,151 @@ mod tests {
...
+        // ls -t target/debug/deps/ | rg ^tantivy | head -1
+        // rust-gdb $MOST_RECENT
+        // $gdb break src/indexer/stamper.rs:tantivy::indexer::stamper::Stamper::stamp
+        // $gdb run --test-threads 1 --test test_add_then_delete_all_documents
+        let mut schema_builder = schema::Schema::builder();
+        let text_field = schema_builder.add_text_field("text", schema::TEXT);
+        let index = Index::create_in_ram(schema_builder.build());
+        let reader = index
+            .reader_builder()
+            .reload_policy(ReloadPolicy::OnCommit)
```
fulmicoton: Let's make it `ReloadPolicy::Manual`. You do the reload anyway manually. `OnCommit` introduce a race condition we want to avoid in unit tests.

src/indexer/index_writer.rs
```diff
@@ -1049,4 +1099,151 @@ mod tests {
...
+        assert!(first_commit_tstamp > add_tstamp);
+
+        // delete_all_documents the index
+        let clear_tstamp = index_writer.delete_all_documents().unwrap();
+        assert_eq!(clear_tstamp, add_tstamp);
+
+        // commit the clear command - now documents aren't available
+        let second_commit = index_writer.commit();
+        assert!(second_commit.is_ok());
+        let _second_commit_tstamp = second_commit.unwrap();
```
fulmicoton: does this need to start by a "_" ?  assert! are not disabled in release.

src/indexer/segment_manager.rs
```diff
@@ -118,6 +118,29 @@ impl SegmentManager {
             });
     }
 
+    pub(crate) fn remove_all_segments(&self) {
+        let mut registers_lock = self.write();
+        registers_lock
+            .committed
+            .segment_entries()
```
fulmicoton: `registers_lock.committed.clear()`  does the job here I believe.

src/indexer/segment_manager.rs
```diff
@@ -118,6 +118,29 @@ impl SegmentManager {
...
+            .iter()
+            .for_each(|segment| {
+                registers_lock
+                    .committed
+                    .remove_segment(&segment.segment_id())
+            });
+
+        registers_lock
```
fulmicoton: same as above.

src/indexer/segment_updater.rs
```diff
@@ -42,9 +42,9 @@ use Result;
 /// Save the index meta file.
 /// This operation is atomic :
 /// Either
-//  - it fails, in which case an error is returned,
+///  - it fails, in which case an error is returned,
 /// and the `meta.json` remains untouched,
-/// - it success, and `meta.json` is written
+/// - it succeeds, and `meta.json` is written
```
fulmicoton, line 47: thanks for all of the typo fixing by the way :)

fulmicoton:
@petr-tik good work. feel free to squash.

I added a bunch of minor comments (code could be shorter here and there) and one major comment for which I don't really have an answer. We can merge as is and address it in another issue.

src/indexer/index_writer.rs
fulmicoton: Please ignore this comment. It is very valid but it requires a bit more reflexion. I created #564 where we can host this discussion.

petr-tik:
I wanted to squash commit from the github UI, but pushing https://github.com/tantivy-search/tantivy/pull/524/commits/0cc26a6907b5c82ebebf4e68fc5ca050d5cfcdff has greyed the `Merge pull request` button for me. 

Please approve, so I can squash and merge

src/indexer/index_writer.rs
fulmicoton: The GC call is useless here @petr-tik  . Can you remove it?

fulmicoton:
I imported created another branch added a commit and merged.  I am not sure why github does not detect this is already merged but closing...

PR #525, opened 2019-04-13
add ascii folding support

drusellers:
This pull request provides a rough first pass at ascii folding as a port from the Lucene code base.

ASCII Folding takes a a non-ASCII character such as `á` and converts it into its ASCII equivalent `a`.

This is a no dependency model, but as I'm writing this I found this crate which is taking the same basic approach.

https://github.com/mpaltun/asciifolding/blob/master/src/lib.rs

I'd be happy to replace my implementation with this.

I'm also open to other approaches or ideas as to implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+Tantivy 0.10.0
+=====================
+- Added an ASCII folding filter (@drusellers)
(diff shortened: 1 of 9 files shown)


fulmicoton:
Merged in master

PR #531, opened 2019-04-19
Delete files from target/ dir to avoid caching them on CI

petr-tik:
idea from here https://github.com/rust-lang/cargo/issues/5885#issuecomment-432723546


petr-tik:
I noticed that travis spends nearly 3 minutes in every CI run caching files in the target dir. 

If we can avoid doing that, CI should run faster

Issue #528, opened 2019-04-18
Tokenizer and accentued language


julienfastre:
**Is your feature request related to a problem? Please describe.**

I am dealing with tantivy and an accentued language (i.e. french). I would like to provide searching without taking accent into account (searching with forgetting an accent should match the world indexed with an accent).

It is a very common mistake, in french, to forget accent when typing a word.

**Describe the solution you'd like**

I think it would useful to provide a filter "tantivy::tokenizer::UnaccentFilter" which return the token without any accent:

```rust
use tantivy::tokenizer::*;

let tokenizer = NGramTokenizer
    .filter(LowerCaser)
    .filter(UnaccentFilter);
```
The `UnaccentFilter` will do some job like this:

```rust
étude => etude
accentué => accentue
âäàéè => aaaee
// etc.
```

julienfastre:
I am currently implenting such filter in my project. I can try to implement this if this feature is accepted.

fulmicoton:
@drusellers  already opened a PR to handle this.

julienfastre:
Good news. I did a research on issues to avoid duplicates, but not in PR. 

I would be happy to give help if needed.

fulmicoton:
Can you review the code? https://github.com/tantivy-search/tantivy/pull/525

julienfastre:
I am very new to Rust, and I do not feel comfident (it is my first experience with rust but I enjoy It!). 

May I add some tests to the PR to check the code ?

What @drusellers thinks about that ? May I help for something ?

drusellers:
I would love any tests you might have.

drusellers:
@julienfastre I added the 

```
accentué => accentue
âäàéè => aaaee
```

to my unit tests. Are there any other edge cases you can think of?

See unit tests here: https://github.com/tantivy-search/tantivy/blob/dds/ascii-folding/src/tokenizer/ascii_folding_filter.rs#L1569

julienfastre:
Hi @drusellers. I was thinking of reproducing the same tests of the Lucene implementation, here: https://github.com/apache/lucene-solr/blob/master/lucene/analysis/common/src/test/org/apache/lucene/analysis/miscellaneous/TestASCIIFoldingFilter.java

At least, the function:

- testLatin1Accents ;
- testUnmodifiedLetters ;
- testEmptyTerm (seems not to be covered in the actual tests) ;
- and maybe testAllFoldings ?

[...]

PR #533, opened 2019-04-23
Add tests for ascii folding

julienfastre:
Add some tests for ascii folding. 

Those tests are inspired by Lucene : https://github.com/apache/lucene-solr/blob/master/lucene/analysis/common/src/test/org/apache/lucene/analysis/miscellaneous/TestASCIIFoldingFilter.java

This PR add the following tests to the feature `ascii_folding` from @drusellers. It should apply on PR https://github.com/tantivy-search/tantivy/pull/525

- test_latin1_characters: a quick test for most commonly used characters in European languages ;
- test_unmodified_letters: test that characters which should not be modified are kept ;
- test_all_foldings: test all foldings

Will help to issue #528.

Thanks for your feedback :smile:

### DIFF
--- a/src/tokenizer/ascii_folding_filter.rs
+++ b/src/tokenizer/ascii_folding_filter.rs
@@ -1560,6 +1560,7 @@ mod tests {
     use super::to_ascii;
     use tokenizer::AsciiFoldingFilter;
     use tokenizer::SimpleTokenizer;
+    use tokenizer::RawTokenizer;
     use tokenizer::TokenStream;
     use tokenizer::Tokenizer;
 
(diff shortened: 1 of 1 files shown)


src/tokenizer/ascii_folding_filter.rs
```diff
@@ -1509,7 +1509,7 @@ fn fold_non_ascii_char(c: char) -> Option<&'static str> {
         '\u{FF0A}' // ＊  [FULLWIDTH ASTERISK]
         => Some("*"),
         '\u{FF0C}' // ，  [FULLWIDTH COMMA]
-        => Some("),"),
```
julienfastre, line 1513: @drusellers I had to change the folding here. Seems correct for you ?

julienfastre:
I think this PR is ready for an integration on the branch of @drusellers, PR https://github.com/tantivy-search/tantivy/pull/525

@drusellers could you review ?

PR #536, opened 2019-04-24
Tidied up the Stamper module and upgraded to Rust 1.34

petr-tik:
Added stamper.revert method to be used for rollback - rolling back to a previous
commit in case of deleting all documents or rolling operations back should reset
the stamper as well

Added type alias for Opstamp - helps code readibility instead of seeing u64
returned by functions.

Moved to AtomicU64 on stable rust (since 1.34) - where possible use standard
library interfaces.


src/core/index_meta.rs
```diff
@@ -1,4 +1,5 @@
 use core::SegmentMeta;
+use indexer::Opstamp;
```
fulmicoton: I think we can put `Opstamp` in the toplevel namespace.
(ie: `tantivy::Opstamp`)

src/indexer/stamper.rs
```diff
@@ -2,31 +2,40 @@ use std::ops::Range;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
 
+pub type Opstamp = u64;
```
fulmicoton: We want some rustdoc on `Opstamp`.

src/lib.rs
```diff
@@ -254,6 +254,14 @@ pub mod merge_policy {
 /// as they are added in the segment.
 pub type DocId = u32;
 
+/// A u64 assigned to every operation incrementally
+///
+/// Used to sequence operations in the correct order during
+/// asynchronous processing of adding documents, merging segments or rollback
+/// Modelled by an AtomicU64 or a u64 under a RW-lock
+/// on platforms without atomic integer support.
+pub type Opstamp = u64;
```
petr-tik, line 263: @fulmicoton Thoughts on the docstring?

petr-tik:
Please review when you can. Feel free to change the docstring and squash everything into 1 commit for merging. 

I recommend using the commit message from 8e509213634ccbf82c3380e1b660ec1ffb735301

PR #551, opened 2019-05-22
Remove mut in add_document and delete_term. Made stamper ordering rel…

fulmicoton:
Would this make your life easier?


hntd187 (approved):
Why the change of ordering? Doesn't relaxed guarantee there is an order, not one that is enforced? If so, would that potentially break something with out of order stamping?

fulmicoton:
@hntd187 you are right. I think this is safe, but there is no need to take that risk.

PR #553, opened 2019-05-27
Make IndexMeta public

hntd187:
Without IndexMeta being public it leads to some awkward code like 
https://github.com/toshi-search/Toshi/blob/master/src/handlers/summary.rs#L37 

Maybe there is a better way to do this, but making IndexMeta public certainly makes this much easier.


fulmicoton:
SGTM. Merging.

PR #603, opened 2019-07-25
add support for float

### LINKED ISSUE #327: Implement float field
The fastfield is fairly obvious.

Please investigate and report the way `FloatPoint` work.

Making it an indexed field on the other hand is not obvious at all, and requires a notion of equality.
Lucene seems to have deprecated its first implementation.

The term dictionary is lexicographically byte ordered, so supporting range queries will require a specific encoding.

trinity-1686a:
as for i64, they are mapped to u64 for indexing, but using a different formula.
query parser doesn't work yet
fix #327

### DIFF
--- a/src/schema/term.rs
+++ b/src/schema/term.rs
@@ -19,15 +19,27 @@ where
     B: AsRef<[u8]>;
 
 impl Term {
-    /// Builds a term given a field, and a u64-value
+    /// Builds a term given a field, and a i64-value
     ///
-    /// Assuming the term has a field id of 1, and a u64 value of 3234,
+    /// Assuming the term has a field id of 1, and a i64 value of 3234,
     /// the Term will have 8 bytes.
     ///
     /// The first four byte are dedicated to storing the field id as a u64.
     /// The 4 following bytes are encoding the u64 value.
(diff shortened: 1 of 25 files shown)


src/schema/term.rs
```diff
@@ -19,9 +19,9 @@ where
     B: AsRef<[u8]>;
 
 impl Term {
-    /// Builds a term given a field, and a u64-value
+    /// Builds a term given a field, and a i64-value
```
fulmicoton, line 22: thanks for catching that.

fulmicoton (commented):
Everything looks awesome. I have no comments.

I'm not sure how it works but I tried to add a commit fixing the TODO you added in value.rs's  `Value::text`.

Can you also add something the CHANGELOG for tantivy 0.11?

trinity-1686a:
> I'm not sure how it works but I tried to add a commit fixing the TODO you added in value.rs's `Value::text`.

you have write access to the branch so it works as one would expect.

There are 2 others TODO, one about an error message/actual behavior missmatch, and one regarding `src/fastfield/readers.rs:FastFieldReaders::u64_lenient`, should it support f64 too?

PR #605, opened 2019-07-26
add checksum check in ManagedDirectory

### LINKED ISSUE #400: Add a checksum to all data files.
This can be done in using wrappers over the Write and Read of the ManagedDirectory.

The write would add a footer to written files.

Probably use CRC32c or Adler. Both seem to have good pure rust implementation.

The checksum would not be used on open, but the index would have `validate_checksum()` method that could return an Error pointing to a specific corrupted file.

trinity-1686a:
`.managed.json` is not checksummed as it does not write itself through ManagedDirectory, but meta.json does (which but an little bit of binary at the end of an otherwise ascii file).

fix #400


src/directory/managed_directory.rs
```diff
@@ -207,26 +210,99 @@ impl ManagedDirectory {
...
+            hasher: Some(Hasher::new()),
+            writer
+        }
+    }
+}
+
+impl<W: Write> Drop for CrcProxy<W> {
+    fn drop(&mut self) {
+        // this should probably be logged
+        let _ = self.writer.write_all(&self.hasher.take().unwrap().finalize().to_be_bytes());
```
fulmicoton: IO in `drop` looks cool but it cannot throw.

Let's assume the fs fucked up on these last 4 bytes. (granted that's not a common event).
We want to detect the problem as soon as possible.
Also we rely on the idea that the underlying `io::Write` flushes on `Drop` which is not enforced by the trait I believe.

`.flush()` looks like a nicer place to put that in, but I assume you were put off by the possibility that tantivy (or someone), may call `.flush` more than once.

Having our own `Write` trait with a `.terminate(self) -> io::Result<()>` method would be great, but that would not be boxable...

I'm not sure the right place where to put this actually... 


What do you think?

src/directory/managed_directory.rs
```diff
@@ -207,26 +210,99 @@ impl ManagedDirectory {
...
-        self.directory.open_write(path)
+        Ok(io::BufWriter::new(Box::new(
+            CrcProxy::new(self.directory.open_write(path)?.into_inner().map_err(|_|()).expect("buffer is empty"))
+        )))
     }
 
     fn atomic_write(&mut self, path: &Path, data: &[u8]) -> io::Result<()> {
         self.register_file_as_managed(path)?;
-        self.directory.atomic_write(path, data)
+        let mut with_crc = Vec::with_capacity(data.len() + CRC_SIZE);
```
fulmicoton: I don't think we want checksum for `atomic_write`.
I really like `meta.json` and `managed.json` to be human readable. I think it demystifies a lot of tantivy and suspect it can help operation.

src/directory/managed_directory.rs
```diff
@@ -207,26 +210,99 @@ impl ManagedDirectory {
...
+
+    /// List files for which checksum does not match content
+    pub fn list_damaged(&self) -> result::Result<HashSet<PathBuf>, OpenReadError> {
+        let mut hashset = HashSet::new();
+        let meta_informations_rlock = self
+            .meta_informations
+            .read()
+            .expect("Managed directory rlock poisoned in list damaged.");
+
+        println!("list_of_segments={:?}", meta_informations_rlock.managed_paths);
```
fulmicoton: println remaining.

src/directory/managed_directory.rs
```diff
@@ -207,26 +210,99 @@ impl ManagedDirectory {
...
+        Ok(data[len..]==crc)
+    }
+
+    /// List files for which checksum does not match content
+    pub fn list_damaged(&self) -> result::Result<HashSet<PathBuf>, OpenReadError> {
+        let mut hashset = HashSet::new();
+        let meta_informations_rlock = self
+            .meta_informations
+            .read()
+            .expect("Managed directory rlock poisoned in list damaged.");
```
fulmicoton: I'm ok with unwrapping rlock... under certain conditions.

So the key idea is that you want to make sure that the lock cannot be poisoned... AND make it trivial 
for the reader that it cannot be poisoned. 

Here the conditions are not met because you hold the lock, while you perform a lot of IO, some of them 
depending on user defined code (if there is a custom directory involved for instance).

Can you clone `managed_paths` right after acquiring the lock and release it right away?

trinity-1686a: I did made sure to not panic myself, but did not consider user code being run here. I'll do the clone

src/directory/managed_directory.rs
trinity-1686a: woops

src/directory/mod.rs
```diff
@@ -24,18 +24,29 @@ pub use self::ram_directory::RAMDirectory;
...
+
+/// Trait used to indicate when no more write need to be done on a writer
+pub trait TerminatingWrite: Write {
+    /// Indicate that the writer will no longer be used
+    fn terminate(mut self) -> io::Result<()> where Self: Sized {
+        self.flush()
+    }
+}
+
+impl<W: TerminatingWrite + ?Sized> TerminatingWrite for Box<W> {}
```
trinity-1686a: afterthought: this implementation is wrong, it won't call inner's terminate, but the default one defined just above.
It can't call inner's implementation because of `?Sized` (require otherwise Box is hardly usable at all), but removing the need for Sized on `terminate` mean it can't take by value, and the borrow check can't help us maintain the contract that a TerminatingWrite can't be used after a call to terminate

src/directory/managed_directory.rs
```diff
@@ -207,26 +216,209 @@ impl ManagedDirectory {
...
+    pub fn to_bytes(&self) -> Vec<u8> {
+        match self {
+            Footer::V0(f) => f.to_bytes()
+        }
+    }
+
+    pub fn from_bytes(data: &[u8]) -> Self {
+        let len = data.len();
+        assert!(len >= 2);
+        let size = u16::from_be_bytes([data[len - 2], data[len - 1]]);
```
fulmicoton: The rest of the code base use little endian everywhere.
I'd prefer if you used byteorder, like the rest of the code too. https://docs.rs/byteorder/1.1.0/byteorder/trait.ByteOrder.html#tymethod.read_u16

src/directory/managed_directory.rs
```diff
@@ -207,26 +216,209 @@ impl ManagedDirectory {
...
+
+    pub fn tantivy_version(&self) -> (u8, u8, u8) {
+        match self {
+            Footer::V0(f) => f.tantivy_version,
+        }
+    }
+
+    pub fn crc(&self) -> u32 {
+        match self {
+            Footer::V0(f) => f.crc,
```
fulmicoton: meaningful variable name please.

src/directory/managed_directory.rs
```diff
@@ -207,26 +216,209 @@ impl ManagedDirectory {
...
+        self.writer.write_all(&footer)?;
+        self.writer.flush()?;//should we assuse calling terminate on inner already flush?
+        self.writer.terminate()
+    }
 }
 
 impl Directory for ManagedDirectory {
     fn open_read(&self, path: &Path) -> result::Result<ReadOnlySource, OpenReadError> {
-        self.directory.open_read(path)
+        let ros = self.directory.open_read(path)?;
```
fulmicoton: cryptic variable name

src/directory/managed_directory.rs
```diff
@@ -28,6 +30,13 @@ fn is_managed(path: &Path) -> bool {
         .unwrap_or(true)
 }
 
+/// Returns true iff the file require a footer.
+fn require_footer(path: &Path) -> bool {
+    is_managed(path) &
```
fulmicoton: `is_managed` is very misleading. It seems to imply it is part of the files marked as managed, as in "part of the files listed in `.managed.json`.

src/directory/managed_directory.rs
trinity-1686a: By adding some indirection it would be possible to have it boxable, but it does not really look too good either https://gist.github.com/fdb-hiroshima/73eca3fd7e4ff47b5c92fad05cbb3a83

src/directory/mod.rs
```diff
@@ -24,18 +24,48 @@ pub use self::ram_directory::RAMDirectory;
...
-use std::io::{BufWriter, Write};
+use std::io::{self, BufWriter, Write};
 
 #[cfg(feature = "mmap")]
 pub use self::mmap_directory::MmapDirectory;
 
 pub use self::managed_directory::ManagedDirectory;
 
+/// Struct used to prevent from calling [`terminate_ref`](trait.TerminatingWrite#method.terminate_ref) directly
+pub struct AntiCallToken(());
```
fulmicoton, line 36: interesting pattern.

src/directory/managed_directory.rs
```diff
@@ -207,19 +210,198 @@ impl ManagedDirectory {
...
+        assert!(len >= 2);
+        let size = LittleEndian::read_u16(&data[len - 2..]);
+        assert!(
+            len >= size as usize,
+            "len({}) is smaller than size({})",
+            len,
+            size
+        );
+        let footer = &data[len - size as usize..];
+        let index_version = footer[0];
```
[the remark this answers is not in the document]
fulmicoton: yes.

src/directory/managed_directory.rs
```diff
@@ -207,19 +210,198 @@ impl ManagedDirectory {
...
+
+impl V0 {
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut res = vec![0; 10];
+        res[0] = 0;
+        res[1] = self.tantivy_version.0;
+        res[2] = self.tantivy_version.1;
+        res[3] = self.tantivy_version.2;
+        LittleEndian::write_u32(&mut res[4..8], self.crc);
+        LittleEndian::write_u16(&mut res[8..], Self::size());
```
fulmicoton: u32 for footer len please

src/directory/managed_directory.rs
```diff
@@ -207,19 +210,198 @@ impl ManagedDirectory {
...
+        Self {
+            tantivy_version: (
+                env!("CARGO_PKG_VERSION_MAJOR").parse().unwrap(),
+                env!("CARGO_PKG_VERSION_MINOR").parse().unwrap(),
+                env!("CARGO_PKG_VERSION_PATCH").parse().unwrap(),
+            ),
+            crc,
+        }
+    }
+    pub fn size() -> u16 {
```
fulmicoton: Do we actually use this? It is likely to not be constant in the future.

If it does not take `self`, you can make it a `const` and use it. Or make it private

src/directory/managed_directory.rs
```diff
@@ -207,19 +210,198 @@ impl ManagedDirectory {
...
+
+#[derive(Debug, Copy, Clone)]
+pub struct V0 {
+    pub tantivy_version: (u8, u8, u8),
+    pub crc: u32,
+}
+
+impl V0 {
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut res = vec![0; 10];
```
fulmicoton: can we have unit tests for serialize / deserialization of V0?

src/directory/managed_directory.rs
```diff
@@ -207,19 +210,198 @@ impl ManagedDirectory {
...
+    }
+}
+
+impl<W: TerminatingWrite> TerminatingWrite for FooterProxy<W> {
+    fn terminate_ref(&mut self, _: AntiCallToken) -> io::Result<()> {
+        let crc = self.hasher.take().unwrap().finalize();
+        let footer = Footer::V0(V0::from_crc(crc)).to_bytes();
+        let mut writer = self.writer.take().unwrap();
+        writer.write_all(&footer)?;
+        writer.flush()?; //should we assume calling terminate on inner already flush?
```
fulmicoton: I'm ok with having flush in `terminate`.

fulmicoton: You can remove this comment and comment explicitely in the `TerminatingWrite` that `.terminate_ref()` should flush.

src/directory/managed_directory.rs
```diff
@@ -207,19 +210,198 @@ impl ManagedDirectory {
...
+            self.directory
+                .open_write(path)?
+                .into_inner()
+                .map_err(|_| ())
+                .expect("buffer should be empty"),
+        ))))
     }
 
+    // TODO: to be correct, this require an non enforced contract, atomic writes must only be
+    // read back as atomics, and normal write must only be read by normal read
```
fulmicoton: I'm happy with this. Can you remove this comment and document this contract in `Directory` if it is not already there?

fulmicoton:
@fdb-hiroshima I am very happy with all of the choice you made.

I added a couple of nitpicks. No need to address all of them but I think there could be more unit tests especially for serialization/deserialization.

Could you also add your contribution in the CHANGELOG?

src/directory/footer.rs
```diff
@@ -0,0 +1,213 @@
...
+            ));
+        }
+        let footer = &data[len - size as usize..];
+        let meta_len = LittleEndian::read_u32(&footer[size - 20..]) as usize;
+        let tantivy_major = LittleEndian::read_u32(&footer[size - 16..]);
+        let tantivy_minor = LittleEndian::read_u32(&footer[size - 12..]);
+        let tantivy_patch = LittleEndian::read_u32(&footer[size - 8..]);
+        Ok(Footer {
+            tantivy_version: (tantivy_major, tantivy_minor, tantivy_patch),
+            meta: String::from_utf8_lossy(&footer[size - meta_len - 20..size - 20]).into_owned(),
```
fulmicoton, line 79: well thought!

Issue #602, opened 2019-07-25
Remove watching directory introduced in 0.9


fulmicoton:
Tantivy 0.9 introduced a funcationality to watch and autoreload an index when a change in meta.json has been detected.

It has caused errors to different users on various platforms. 

Let's just remove it and "poll" `meta.json`, or only enable the functionality when
the writer and the reader are in the same process.

halvorboe:
Could try to tackle this. 

A few questions: 
1. Should we fix it together with https://github.com/tantivy-search/tantivy/issues/702?
2. Is this related to https://github.com/tantivy-search/tantivy/issues/494 as well?
3. Are there any good reasons to use multiple processes?

halvorboe:
Tests for watching are also failing on my machine:
Run 1:
```
failures:

---- directory::tests::mmap_directory_tests::test_watch stdout ----
thread 'directory::tests::mmap_directory_tests::test_watch' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `3`', src/directory/tests.rs:219:9

---- core::index::tests::mmap_specific::test_index_on_commit_reload_policy_different_directories stdout ----
thread 'core::index::tests::mmap_specific::test_index_on_commit_reload_policy_different_directories' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `2`', src/core/index.rs:554:9


failures:
    core::index::tests::mmap_specific::test_index_on_commit_reload_policy_different_directories
    directory::tests::mmap_directory_tests::test_watch

test result: FAILED. 384 passed; 2 failed; 2 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '--lib'
```
Run 2:
```
failures:

---- directory::tests::mmap_directory_tests::test_watch stdout ----
thread 'directory::tests::mmap_directory_tests::test_watch' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `3`', src/directory/tests.rs:219:9

---- core::index::tests::mmap_specific::test_index_on_commit_reload_policy_mmap stdout ----
thread 'core::index::tests::mmap_specific::test_index_on_commit_reload_policy_mmap' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `2`', src/core/index.rs:554:9

---- core::index::tests::mmap_specific::test_index_on_commit_reload_policy_different_directories stdout ----
[...]

PR #612, opened 2019-08-03
Replace deprecated tempdir/atomicwrites dependencies

kornelski:
The `tempdir` crate is dead.

### DIFF
--- a/src/directory/mmap_directory.rs
+++ b/src/directory/mmap_directory.rs
@@ -19,7 +19,6 @@ use crate::directory::WatchCallback;
 use crate::directory::WatchCallbackList;
 use crate::directory::WatchHandle;
 use crate::directory::WritePtr;
-use atomicwrites;
 use memmap::Mmap;
 use std::collections::HashMap;
 use std::convert::From;
@@ -37,6 +36,7 @@ use std::sync::RwLock;
 use std::sync::Weak;
 use std::thread;
 use tempfile::TempDir;
+use tempfile::NamedTempFile;
 
 /// Create a default io error given a string.
 pub(crate) fn make_io_err(msg: String) -> io::Error {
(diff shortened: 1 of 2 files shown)


kornelski:
@fulmicoton what do you think about the failure of atomic write test? I'm not sure if that's due to difference of implementation of atomicwrite vs tempfile: https://github.com/Stebalien/tempfile/pull/106 — or just a bug elsewhere?

fulmicoton:
@kornelski I don't understand. It looks like you replaced `AtomicWrite` by named tempfile. Are Tempfile supposed to offer the same guarantee to begin with?

src/directory/mmap_directory.rs
```diff
@@ -506,8 +506,9 @@ impl Directory for MmapDirectory {
     fn atomic_write(&mut self, path: &Path, data: &[u8]) -> io::Result<()> {
         debug!("Atomic Write {:?}", path);
         let full_path = self.resolve_path(path);
-        let meta_file = atomicwrites::AtomicFile::new(full_path, atomicwrites::AllowOverwrite);
-        meta_file.write(|f| f.write_all(data))?;
+        let mut temp = NamedTempFile::new_in(full_path.parent().expect("has filename"))?;
```
fulmicoton, line 509: This looks like a bug to me.

kornelski:
https://lib.rs/atomicwrites readme says:

> Alternatives
>
> tempfile has a persist method doing the same thing.

And atomicwrites is implemented using the `tempdir` crate:

https://github.com/untitaker/rust-atomicwrites/blob/a0db3c6ad80b834c50ee26fda51afd8c3d3f24d0/src/lib.rs#L126-L133

but `tempdir` is deprecated, obsoleted by `tempfile`.

So effectively both do exactly the same thing - make a named temp file, write to it, then rename over original. The only difference I could see is that atomicwrites sets `MOVEFILE_WRITE_THROUGH` on windows, and tempfile doesn't.

fulmicoton:
@kornelski Ah I see! The unit test failing is related to the way we watch the directory. (using the notify crate). It has some buggy behavior (starting to watch triggered some error on some plateforms), so I actually intend to deprecate it.

I think we can merge your branch once this is done. This is tracked by #602.

fulmicoton:
@kornelski thanks for the research and the analysis. This is of great help!

fulmicoton:
@kornelski should we close this then?

kornelski:
This removes atomicwrites, which removes tempdir from dependencies, so I still think it's worth doing. 

As far as I understand it's waiting for #602 and/or `directory::tests::test_mmap_directory` to be changed.

PR #630, opened 2019-08-14
Add a doctest to BooleanQuery

### LINKED ISSUE #446: Improv Doc for BooleanQuery.
Quoting @fdb-hiroshima

> The only issue I had with it was about BooleanQuery, I had a hard time understanding how to build theme. I was looking for some kind of function, and took a lot of time to figure out it implement From. Maybe it could just simply be mentioned in the documentation of BooleanQuery::new_multiterms_query, "to create a BooleanQuery from Box<Query> use From<Vec<(Occur, Box)>> instead" or something like that

Maybe having a `new(...)` constructor actually would not hurt, actually.

petr-tik:
Closes #446 - add a doctest explaining and showing that a BooleanQuery is the building block for other query types. 

This is a draft, since I am not sure if this closes the ticket completely. The first comment suggests adding a `::new()` constructor. Doesn't the `::from` constructor already satisfy our requirements? 

I can add a type alias explaining that (Occur, Box<dyn Query>) = Clause thus changing the type signature to 
Vec<Clause>

Mark a function that is only used in tests to be compiled for tests only. If we expect it to be used externally (it's marked pub), maybe we should add an integration test to simulate that behaviour.

Fix doc-comments in a couple of related files

### DIFF
--- a/src/query/boolean_query/boolean_query.rs
+++ b/src/query/boolean_query/boolean_query.rs
@@ -9,7 +9,8 @@ use crate::Result;
 use crate::Searcher;
 use std::collections::BTreeSet;
 
-/// The boolean query combines a set of queries
+/// The boolean query returns a set of documents
+/// that matches the Boolean combination of constituent subqueries.
 ///
 /// The documents matched by the boolean query are
 /// those which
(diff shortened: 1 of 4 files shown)


src/query/boolean_query/boolean_query.rs
```diff
@@ -62,6 +122,7 @@ impl Query for BooleanQuery {
 impl BooleanQuery {
     /// Helper method to create a boolean query matching a given list of terms.
     /// The resulting query is a disjunction of the terms.
+    #[cfg(test)]
```
fulmicoton: I suspect this is useful for users too.

src/query/occur.rs
```diff
@@ -1,5 +1,5 @@
 /// Defines whether a term in a query must be present,
-/// should be present or must not be present.
+/// should be present or must be not present.
```
fulmicoton: nice catch.

src/query/regex_query.rs
```diff
@@ -6,7 +6,7 @@ use crate::Searcher;
 use std::clone::Clone;
 use tantivy_fst::Regex;
 
-// A Regex Query matches all of the documents
+/// A Regex Query matches all of the documents
```
fulmicoton: nice

src/query/boolean_query/boolean_query.rs
```diff
@@ -19,6 +20,65 @@ use std::collections::BTreeSet;
 /// `MustNot` occurence.
 /// * match at least one of the subqueries that is not
 /// a `MustNot` occurence.
+///
+/// ```rust
+///#[macro_use]
+///extern crate tantivy;
```
fulmicoton: Can you use rust 2018 style?

src/query/boolean_query/boolean_query.rs
```diff
@@ -19,6 +20,65 @@ use std::collections::BTreeSet;
...
+///        index_writer.add_document(doc!(
+///            title => "The Diary of a Young Girl",
+///        ));
+///        index_writer.commit().unwrap();
+///    }
+///
+///    let reader = index.reader()?;
+///    let searcher = reader.searcher();
+///
+///    let query_parser = QueryParser::for_index(&index, vec![title]);
```
fulmicoton: This looks like a doctest for query parser than for `BooleanQuery`.

In my experience, users have a hard time understanding how to compose subqueries into another one using  `BooleanQuery`.

For instance, having an app with a search field and a facet system that makes it possible to drill down.

That requires composign the output of the query parser with the filtering facet `TermQuery`.

trinity-1686a:
> This is a draft, since I am not sure if this closes the ticket completely. The first comment suggests adding a ::new() constructor. Doesn't the ::from constructor already satisfy our requirements?

::from is doing exactly what I would expect from a ::new, it's just that I don't always check traits and at that time I got stuck a few hours trying to figure out how to actually build a BooleanQuery from it's sub queries (I was also discovering the whole crate, which is quiet big, and I had no idea where to search). My original comment was about pointing to that ::from so that people don't miss it as I did, so just a word about it or an example in doc-test

petr-tik:
so I am not sure about this PR. I think doc-tests need to strike a good balance between answering users questions with relevant detail and verbosity. I see this on the verbose end of the spectrum at the moment. It's bound to use other tantivy concepts since BooleanQuery is an interface to build your own query from primitives, but I would prefer to make it more concise.

src/query/boolean_query/boolean_query.rs
```diff
@@ -19,6 +20,105 @@ use std::collections::BTreeSet;
...
+///    let count4 = searcher.search(&query4, &Count)?;
+///    assert_eq!(count4, 4);
+///
+///    // ("title:diary OR \"dairy cow\"") AND body:found
+///    let body_query: Box<dyn Query> = Box::new(TermQuery::new(
+///        Term::from_field_text(body, "found"),
+///        IndexRecordOption::Basic,
+///    ));
+///    let query5 = BooleanQuery::from(vec![(Occur::Must, body_query),
+///                                         (Occur::Must, Box::new(query4))]);
```
petr-tik: Here, I have had to Box query4 up to stop rustc from showing a type-error

```
error[E0308]: mismatched types
  --> src/query/boolean_query/boolean_query.rs:116:55
   |
92 |                                         (Occur::Must, query4)]);
   |                                                       ^^^^^^ expected struct `std::boxed::Box`, found struct `tantivy::query::BooleanQuery`
   |
   = note: expected type `std::boxed::Box<dyn tantivy::query::Query>`
              found type `tantivy::query::BooleanQuery`
```

I think I should expand the from method to deal with dynamic types, but I am not sure if it breaks any promises/interfaces and if it's a good API to give to everyone

fulmicoton: By dynamic type you mean dynamic trait?
The query needs ownership anyway, so we woudl have to box_clone. I'd rather be explicit.

Any solution that allows not to own the query will require introducing a lifetime or Generics or both.
I don't think we want that.

fulmicoton:
The branch is marked as WIP. (I did not know this could be done).
Let me know when you think it is ready so I can merge it

fulmicoton (approved):

PR #632, opened 2019-08-15
Create segment id from string, expose inventory.

Frando:
Hi,
we're starting to use tantivy as the search engine core for [sonar](https://github.com/Frando/sonar/), which will end up being integrated with a [Dat](https://dat.foundation) based p2p stack. We're looking into replicating the actual tantivy index over p2p as well - as its basically an append-only log of segments, it should be very much feasable.

As a very first step this PR allows to create segments from predefined info by allowing to create a segment id from string and exposing the segment inventory, which allows to add new segments then.

### DIFF
--- a/src/core/segment_id.rs
+++ b/src/core/segment_id.rs
@@ -46,6 +46,12 @@ impl SegmentId {
         SegmentId(create_uuid())
     }
 
+    /// Creates a new segment id from a uuid string.
+    pub fn from_uuid_string(uuid_string: &str) -> Result<SegmentId, uuid::parser::ParseError> {
+        let uuid = Uuid::parse_str(uuid_string)?;
+        Ok(SegmentId(uuid))
+    }
+
     /// Returns a shorter identifier of the segment.
     ///
     /// We are using UUID4, so only 6 bits are fixed,
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -216,7 +216,7 @@ impl Index {
         Index::open(mmap_directory)
     }
 
-    pub(crate) fn inventory(&self) -> &SegmentMetaInventory {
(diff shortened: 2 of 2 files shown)


fulmicoton:
In principle I'm ok with this change, but could you link the code using this new public API?
I want to make sure we expose the right thing in the right place.

Frando:
See https://github.com/Frando/sonar/blob/master/src-rust/index.rs#L255
and basic tests in [rust](https://github.com/Frando/sonar/blob/master/src-rust/index.rs#L322) and [Node.js](https://github.com/Frando/sonar/blob/master/test/segments.js).

Right now the segment files are copied into the target directory manually, then the new segment is added and then the new meta.json is saved manually. The latter certainly is a workaround but works well for me for now. If this whole path turns out to be working well, I'd consider working on a PR that would formalize an API to add segments to an index.

I'm very much open to suggestions if there's a better path!

src/core/segment_id.rs
```diff
@@ -46,6 +46,12 @@ impl SegmentId {
         SegmentId(create_uuid())
     }
 
+    /// Creates a new segment id from a uuid string.
+    pub fn from_uuid_string(uuid_string: &str) -> Result<SegmentId, uuid::parser::ParseError> {
```
fulmicoton, line 50: I want to avoid emitting symbols from other libraries (here the `uuid` crate)  to avoid a dependency difficulty in clients.
I created another branch with this part of the problem.

fulmicoton:
I was uncomfortable with exposing an external library symbol, as well as the `Inventory` instances.
I think the following PR should address your use case.
https://github.com/tantivy-search/tantivy/pull/636/files
https://github.com/tantivy-search/tantivy/pull/637/files

fulmicoton:
Merging #636 and #637 and closing this PR.

PR #641, opened 2019-08-22
RegexQuery performance: make it possible to cache Regexes - remastered by fulmicoton  (Closes #639)

### LINKED ISSUE #639: RegexQuery performance
Hi,

I'm trying to implement prefix matching. My idea was to use a RegexQuery for
this: given a user input of "foo", build an automaton that matches "foo.+" and
use that to search the fst.

(Aside: is there a way already to do this in tantivy and I've missed it?)

While implementing this, I stumbled upon a limitation of the current
RegexQuery: the RegexQuery struct owns a String with the given regex pattern,
and builds an automaton whenever it's needed in the `specialized_weight` method.

This is a problem as building the automaton is an expensive operation, and I have
to search for the same prefix over N fields.

What I did to address on a branch of mine:

- change RegexQuery to own a fully built Regex instead of a string pattern

- change Regex and the inner DFA to derive `Clone`, as the `specialized_weight`
  method needs to return an owned Regex. This is assuming that a) making Regex
  and DFA derive Clone is OK, and b) that cloning a Regex is cheap compared to
  recompiling from a pattern

What do you think of this? Are my assumptions correct?

If there isn't a simpler way to do prefix matching that I have completely
missed, and if you agree on the approach, I'd be happy to make PRs to
tantivy-fst and to tantivy for this.

fulmicoton:
@brainlock can you review this?

### DIFF
--- a/src/query/automaton_weight.rs
+++ b/src/query/automaton_weight.rs
@@ -8,15 +8,13 @@ use crate::termdict::{TermDictionary, TermStreamer};
 use crate::DocId;
 use crate::TantivyError;
 use crate::{Result, SkipResult};
+use std::sync::Arc;
 use tantivy_fst::Automaton;
 
 /// A weight struct for Fuzzy Term and Regex Queries
-pub struct AutomatonWeight<A>
-where
-    A: Automaton + Send + Sync + 'static,
-{
+pub struct AutomatonWeight<A> {
     field: Field,
-    automaton: A,
+    automaton: Arc<A>,
 }
 
 impl<A> AutomatonWeight<A>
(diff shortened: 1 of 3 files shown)


src/query/automaton_weight.rs
```diff
@@ -8,28 +8,30 @@ use crate::termdict::{TermDictionary, TermStreamer};
 use crate::DocId;
 use crate::TantivyError;
 use crate::{Result, SkipResult};
+use std::sync::Arc;
 use tantivy_fst::Automaton;
 
 /// A weight struct for Fuzzy Term and Regex Queries
-pub struct AutomatonWeight<A>
-where
-    A: Automaton + Send + Sync + 'static,
```
brainlock, line 16: Just curious: did you intentionally remove the `A: Automaton + Send + Sync + 'static` constraint here? What's the implication of this? (I see that the impl block still has a constraint)

brainlock (approved):
👍 perfect this fixes my issue

src/query/automaton_weight.rs
fulmicoton: It has more or less no effects. As far as I can tell, both schools exists in rust.
A lot of people like to never put any boundary on that kind of struct.
I personally like to put it sometimes just to express the intent of the struct.

Here I removed it because it is long... but I could have kept `A: Automaton` for instance.

brainlock: I see. Thanks

Issue #656, opened 2019-09-22
Memory leak


fulmicoton:
As reported by email.

Quoting

the thread is not stoped  when Index dropped .because watcher_wrapper_clone.inner._watcher is not dropped  so  code can't run Err(_e) break;

![2F0CCD18@8B556C17 5349835D](https://user-images.githubusercontent.com/1021506/65383252-a4495380-dd4d-11e9-84b7-120b98de510f.jpg)

petr-tik:
Did they give a test case to reproduce or is it findable in our unit tests? 

I've run the following to use the leak sanitizer in LLVM, but to be honest, I don't really know where to go from here.

```bash
$ RUSTC_WRAPPER= RUSTFLAGS="-Z sanitizer=leak" cargo +nightly test --target x86_64-unknown-linux-gnu mmap_directory
   Compiling tantivy v0.11.0 (/home/petr_tik/Coding/tantivy)
    Finished dev [unoptimized + debuginfo] target(s) in 3m 08s
     Running target/x86_64-unknown-linux-gnu/debug/deps/tantivy-d173cbeea1b61f7b

running 6 tests
test directory::mmap_directory::tests::test_open_non_existant_path ... ok
test directory::mmap_directory::tests::test_watch_wrapper ... ok
test directory::mmap_directory::tests::test_open_empty ... ok
test directory::mmap_directory::tests::test_cache ... ok
test directory::tests::test_mmap_directory ... ok
test directory::mmap_directory::tests::test_mmap_released ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 341 filtered out


=================================================================
==21426==ERROR: LeakSanitizer: detected memory leaks

Indirect leak of 144 byte(s) in 2 object(s) allocated from:
    #0 0x55a3a96ee195  (/home/petr_tik/Coding/tantivy/target/x86_64-unknown-linux-gnu/debug/deps/tantivy-d173cbeea1b61f7b+0x4cb195)
    #1 0x55a3a95048c4  (/home/petr_tik/Coding/tantivy/target/x86_64-unknown-linux-gnu/debug/deps/tantivy-d173cbeea1b61f7b+0x2e18c4)
    #2 0x55a3aa7c8cb7  (/home/petr_tik/Coding/tantivy/target/x86_64-unknown-linux-gnu/debug/deps/tantivy-d173cbeea1b61f7b+0x15a5cb7)

[...]

PR #657, opened 2019-09-26
Address #656

fulmicoton:
Broke the reference loop to make sure that the watch_router can
be dropped, and the thread exits.


qiongtubao:
Is there any progress in this merge request?

petr-tik:
Looks like we can close this PR, since you merged changes to master

PR #667, opened 2019-10-19
Fixed #664

### LINKED ISSUE #664: Documentation Error:: tantivy::schema::Field is not an u8
**Describe the bug**

https://github.com/tantivy-search/tantivy/blob/master/src/schema/field.rs#L10

The docs say that this is an u8, but the implementation is a u32, thus the docs are wrong.

petr-tik:


src/schema/field.rs
```diff
@@ -3,12 +3,11 @@ use std::io;
...
 
-/// `Field` is actually a `u8` identifying a `Field`
-/// The schema is in charge of holding mapping between field names
-/// to `Field` objects.
+/// `Field` is represented by an unsigned 32-bit integer type
+/// The schema holds the mapping between field names and `Field` objects.
 ///
-/// Because the field id is a `u8`, tantivy can only have at most `255` fields.
-/// Value 255 is reserved.
+/// Because the field id is a `u32`, tantivy can have at most `4,294,967,295` fields.
```
fulmicoton: You can remove these last 2 lines (l.8-9). 
A lot of things will break before someone tries to have 4B fields.

src/schema/field.rs
```diff
@@ -3,12 +3,11 @@ use std::io;
...
+/// The last value is reserved.
```
fulmicoton: You can remove this line too

petr-tik (commented):
addressed your code review and force pushed (prefer not polluting the git history with such commits)

fulmicoton:
Thanks!!!!

PR #669, opened 2019-10-21
Added handling of pre-tokenized text fields (#642).

### LINKED ISSUE #642: External Tokenizers/Pre-tokenized text
It'd be nice to have a easy way to load pre-tokenized text with (optional?) offset data.

e.g. solr has [PreAnalyzedField](https://lucene.apache.org/solr/guide/8_1/working-with-external-files-and-processes.html#WorkingwithExternalFilesandProcesses-ThePreAnalyzedFieldType).

Would also be nice if it was accessible from the python API, since it's easy to get really good tokenization in python.

EDIT: the problem with this is queries.
solr let's you specify an analyzer chain or (determined at query time) give it a token stream for queries: https://github.com/apache/lucene-solr/blob/master/solr/core/src/test/org/apache/solr/schema/PreAnalyzedFieldTest.java

But that's pretty unsatisfying, since you either can't use the solr query parsers as well (you need to implement your own query parser to determine what parts of the query to analyze) or get sub-optimal results.

Maybe this is a bad idea, and the right thing is just to register tokenizers from python?

kkoziara:
The commit adds handling of pre-tokenized text. The Document object can now contain a struct with original text and vector of tokens, which will be directly indexed. Loading from JSON is also handled.
Closes #642

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ Tantivy 0.11.0
 - TopDocs collector: ensure stable sorting on equal score. #671 (@brainlock)
 - Added handling of pre-tokenized text fields (#642), which will enable users to
   load tokens created outside tantivy. See usage in examples/pre_tokenized_text. (@kkoziara)
- 
+- Fix crash when committing multiple times with deleted documents. #681 (@brainlock)
+
 ## How to update?
 
 - `Box<dyn BoxableTokenizer>` has been replaced by a `BoxedTokenizer` struct.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Tantivy 0.11.0
 - Avoid rebuilding Regex automaton whenever a regex query is reused. #639 (@brainlock)
 - Add footer with some metadata to index files. #605 (@fdb-hiroshima)
 - TopDocs collector: ensure stable sorting on equal score. #671 (@brainlock)
+- Added handling of pre-tokenized text fields (#642), which will enable users to
+  load tokens created outside tantivy. See usage in examples/pre_tokenized_text.
  
 ## How to update?
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Tantivy 0.11.0
 - Add footer with some metadata to index files. #605 (@fdb-hiroshima)
 - TopDocs collector: ensure stable sorting on equal score. #671 (@brainlock)
 - Added handling of pre-tokenized text fields (#642), which will enable users to
-  load tokens created outside tantivy. See usage in examples/pre_tokenized_text.
+  load tokens created outside tantivy. See usage in examples/pre_tokenized_text. (@kkoziara)
  
 ## How to update?
 
--- a/examples/pre_tokenized_text.rs
+++ b/examples/pre_tokenized_text.rs
@@ -115,12 +115,27 @@ fn main() -> tantivy::Result<()> {
         .search(&query, &(TopDocs::with_limit(2), Count))
         .unwrap();
 
-    println!("Docs counts: {}", count);
+    assert_eq!(count, 2);
 
     for (_score, doc_address) in top_docs {
         let retrieved_doc = searcher.doc(doc_address)?;
         println!("Document: {}", schema.to_json(&retrieved_doc));
     }
+
+    // In contrary to the previous query, when we search for the "man" term we
+    // should get no results, as it's not one of the indexed tokens. SimpleTokenizer
+    // only splits text on whitespace / interpunction.
+
+    let query = TermQuery::new(
+        Term::from_field_text(title, "nan"),
+        IndexRecordOption::Basic,
+    );
+
+    let (top_docs, count) = searcher
+        .search(&query, &(TopDocs::with_limit(2), Count))
+        .unwrap();
+
+    assert_eq!(count, 0);
 
     Ok(())
 }
--- a/examples/pre_tokenized_text.rs
+++ b/examples/pre_tokenized_text.rs
@@ -1,0 +1,126 @@
+// # Pre-tokenized text example
+//
+// This example shows how to use pre-tokenized text. Sometimes yout might
+// want to index and search through text which is already split into
+// tokens by some external tool.
+//
+// In this example we will:
+// - use tantivy tokenizer to create tokens and load them directly into tantivy,
+// - import tokenized text straight from json,
+// - perform a search on documents with pre-tokenized text
+
+use tantivy::tokenizer::{SimpleTokenizer, Token, TokenStream, TokenizedString, Tokenizer};
+
+use tantivy::collector::{Count, TopDocs};
+use tantivy::query::TermQuery;
+use tantivy::schema::*;
+use tantivy::{doc, Index, ReloadPolicy};
+use tempfile::TempDir;
+
+fn tokenize_it(text: &str) -> Vec<Token> {
+    let mut ts = SimpleTokenizer.token_stream(text);
+    let mut tokens = vec![];
+    while ts.advance() {
+        tokens.push(ts.token().clone());
+    }
+    tokens
+}
+
+fn main() -> tantivy::Result<()> {
+    let index_path = TempDir::new()?;
+
+    let mut schema_builder = Schema::builder();
+
+    // now we add `TOKENIZED` `TextOptions` to mark field as pre-tokenized
+    // in addition the title will be also stored, so we can see it in
+    // returned results
+    schema_builder.add_text_field("title", TEXT | STORED | TOKENIZED);
+    schema_builder.add_text_field("body", TEXT | TOKENIZED);
+
+    let schema = schema_builder.build();
+
+    let index = Index::create_in_dir(&index_path, schema.clone())?;
+
+    let mut index_writer = index.writer(50_000_000)?;
+
+    // We can create a document manually, by setting the fields
+    // one by one in a Document object.
+    let title = schema.get_field("title").unwrap();
+    let body = schema.get_field("body").unwrap();
+
+    let title_text = "The Old Man and the Sea";
+    let body_text = "He was an old man who fished alone in a skiff in the Gulf Stream";
+
+    // Content of our first document
+    // We create `TokenizedString` which contains original text and vector of tokens
+    let title_tok = TokenizedString {
+        text: String::from(title_text),
+        tokens: tokenize_it(title_text),
+    };
+
+    println!(
+        "Original text: \"{}\" and tokens: {:?}",
+        title_tok.text, title_tok.tokens
+    );
+
+    let body_tok = TokenizedString {
+        text: String::from(body_text),
+        tokens: tokenize_it(body_text),
+    };
+
+    // Now lets create a document and add our `TokenizedString` using
+    // `add_tokenized_text` method of `Document`
+    let mut old_man_doc = Document::default();
+    old_man_doc.add_tokenized_text(title, &title_tok);
+    old_man_doc.add_tokenized_text(body, &body_tok);
+
+    // ... now let's just add it to the IndexWriter
+    index_writer.add_document(old_man_doc);
+
+    // `Document` can be obtained directly from JSON:
+    let short_man_json = r#"{
+        "title":[{
+            "text":"The Old Man",
+            "tokens":[
+                {"offset_from":0,"offset_to":3,"position":0,"text":"The","position_length":1},
+                {"offset_from":4,"offset_to":7,"position":1,"text":"Old","position_length":1},
+                {"offset_from":8,"offset_to":11,"position":2,"text":"Man","position_length":1}
+            ]
+        }]
+    }"#;
+
+    let short_man_doc = schema.parse_document(&short_man_json)?;
+
+    index_writer.add_document(short_man_doc);
+
+    // Let's commit changes
+    index_writer.commit()?;
+
+    // ... and now is the time to query our index
+
+    let reader = index
+        .reader_builder()
+        .reload_policy(ReloadPolicy::OnCommit)
+        .try_into()?;
+
+    let searcher = reader.searcher();
+
+    // We want to get documents with token "Man", we will use TermQuery to do it
+    let query = TermQuery::new(
+        Term::from_field_text(title, "Man"),
+        IndexRecordOption::Basic,
+    );
+
+    let (top_docs, count) = searcher
+        .search(&query, &(TopDocs::with_limit(2), Count))
+        .unwrap();
+
+    println!("Docs counts: {}", count);
+
+    for (_score, doc_address) in top_docs {
+        let retrieved_doc = searcher.doc(doc_address)?;
+        println!("Document: {}", schema.to_json(&retrieved_doc));
+    }
+
+    Ok(())
+}
--- a/examples/pre_tokenized_text.rs
+++ b/examples/pre_tokenized_text.rs
@@ -9,7 +9,7 @@
 // - import tokenized text straight from json,
 // - perform a search on documents with pre-tokenized text
 
-use tantivy::tokenizer::{SimpleTokenizer, Token, TokenStream, TokenizedString, Tokenizer};
+use tantivy::tokenizer::{PreTokenizedString, SimpleTokenizer, Token, TokenStream, Tokenizer};
 
 use tantivy::collector::{Count, TopDocs};
 use tantivy::query::TermQuery;
@@ -17,11 +17,11 @@ use tantivy::schema::*;
 use tantivy::{doc, Index, ReloadPolicy};
 use tempfile::TempDir;
 
-fn tokenize_it(text: &str) -> Vec<Token> {
-    let mut ts = SimpleTokenizer.token_stream(text);
+fn pre_tokenize_text(text: &str) -> Vec<Token> {
+    let mut token_stream = SimpleTokenizer.token_stream(text);
     let mut tokens = vec![];
-    while ts.advance() {
-        tokens.push(ts.token().clone());
+    while token_stream.advance() {
+        tokens.push(token_stream.token().clone());
     }
     tokens
 }
@@ -31,11 +31,8 @@ fn main() -> tantivy::Result<()> {
 
     let mut schema_builder = Schema::builder();
 
-    // now we add `TOKENIZED` `TextOptions` to mark field as pre-tokenized
-    // in addition the title will be also stored, so we can see it in
-    // returned results
-    schema_builder.add_text_field("title", TEXT | STORED | TOKENIZED);
-    schema_builder.add_text_field("body", TEXT | TOKENIZED);
+    schema_builder.add_text_field("title", TEXT | STORED);
+    schema_builder.add_text_field("body", TEXT);
 
     let schema = schema_builder.build();
 
@@ -52,10 +49,10 @@ fn main() -> tantivy::Result<()> {
     let body_text = "He was an old man who fished alone in a skiff in the Gulf Stream";
 
     // Content of our first document
-    // We create `TokenizedString` which contains original text and vector of tokens
-    let title_tok = TokenizedString {
+    // We create `PreTokenizedString` which contains original text and vector of tokens
+    let title_tok = PreTokenizedString {
         text: String::from(title_text),
-        tokens: tokenize_it(title_text),
+        tokens: pre_tokenize_text(title_text),
     };
 
     println!(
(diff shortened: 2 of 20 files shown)


examples/pre_tokenized_text.rs
```diff
@@ -0,0 +1,126 @@
...
+        .reload_policy(ReloadPolicy::OnCommit)
+        .try_into()?;
+
+    let searcher = reader.searcher();
+
+    // We want to get documents with token "Man", we will use TermQuery to do it
+    let query = TermQuery::new(
+        Term::from_field_text(title, "Man"),
+        IndexRecordOption::Basic,
+    );
```
petr-tik, line 111: It will make the example that much more user-friendly if we add a query that shows what kind of words yield no results - in this case looking for "man" in lowercase should show 0

```rust
    let query = TermQuery::new(
        Term::from_field_text(title, "man"),
        IndexRecordOption::Basic,
    );
```

src/tokenizer/tokenized_string.rs
```diff
@@ -0,0 +1,188 @@
...
+/// TokenStream implementation which wraps TokenizedString
+pub struct TokenizedStream {
+    tokenized_string: TokenizedString,
+    current_token: i64,
+}
+
+impl From<&TokenizedString> for TokenizedStream {
+    fn from(s: &TokenizedString) -> TokenizedStream {
+        TokenizedStream {
+            tokenized_string: s.clone(),
```
petr-tik: Is the clone() strictly necessary here?

petr-tik:
@kkoziara - awesome PR - thanks for your work! Feel free to add more info in the commit message and an entry to the changelog showcasing a) your feature, b) what it will enable others to do (load text tokenized outside tantivy)

My comments are minor and a couple of questions. 

Also, I noticed that you didn't add a method to access the TokStr in the `impl Value` trait in 
https://github.com/tantivy-search/tantivy/blob/75d367680335a7ecb5847628366f2b80f2ce2cfd/src/schema/value.rs#L124-L177

Is that because you don't expect people to access it?

examples/pre_tokenized_text.rs
kkoziara: I totally agree. I will add this query with some comment.

src/tokenizer/tokenized_string.rs
kkoziara: Probably it isn't. from should take ownership of s. Is this what you had in mind?

kkoziara:
@petr-tik thanks for the comments.  I will put some info in the changelog and I will add the method to access `TokStr` to the `Value` implementation (I just missed this). I noticed that there are two conventions there: for `fn text` it returns `None` if the variant is incorrect but for other variants it panics. Is one of these two behaviours preferred?

petr-tik:
the last commit introduces more clones, which goes against our general conversation about minimising memcpy above. 

I don't see this as a huge issue at the moment though. 

> I noticed that there are two conventions there: for `fn text` it returns `None` if the variant is incorrect but for other variants it panics. Is one of these two behaviours preferred?

I am not sure about the preferred convention  wrt Value::TokStr, best ask @fulmicoton here or over gitter.

examples/pre_tokenized_text.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    // In contrary to the previous query, when we search for the "man" term we
+    // should get no results, as it's not one of the indexed tokens. SimpleTokenizer
+    // only splits text on whitespace / interpunction.
+
+    let query = TermQuery::new(
+        Term::from_field_text(title, "nan"),
+        IndexRecordOption::Basic,
+    );
```
petr-tik, line 131: typo for "man", right?

examples/pre_tokenized_text.rs
```diff
@@ -0,0 +1,141 @@
...
+    assert_eq!(count, 2);
+
+    for (_score, doc_address) in top_docs {
+        let retrieved_doc = searcher.doc(doc_address)?;
+        println!("Document: {}", schema.to_json(&retrieved_doc));
+    }
+
+    // In contrary to the previous query, when we search for the "man" term we
+    // should get no results, as it's not one of the indexed tokens. SimpleTokenizer
+    // only splits text on whitespace / interpunction.
```
petr-tik: you can make the explanation more explicit like this:

> Using Pre-TokenizedString means the tokens are stored as is avoiding stemming and lowercasing, which preserves full words in their original form

interpunction - punctuation?

examples/pre_tokenized_text.rs
fulmicoton: ... and put comments about what is the purpose of this `TermQuery` etc.

src/schema/text_options.rs
```diff
@@ -33,13 +34,25 @@ impl TextOptions {
         self.indexing = Some(indexing);
         self
     }
+
+    /// Returns true if the text is already tokenized in the form of TokenString
+    pub fn is_tokenized(&self) -> bool {
```
fulmicoton: This is not only spot, but I suggest we use the word `pretokenized` instead of `tokenized` everywhere to be more explicit.

kkoziara: You mean changing only TextOptions.is_tokenized -> TextOptions.is_pretokenized, TextOptions.set_tokenized -> TextOptions.set_pretokenized and Value.tokenized_text -> Value.pretokenized_text or also TokenizedString -> PretokenizedString, Value::TokStr -> Value::PretokStr ?

examples/pre_tokenized_text.rs
kkoziara: Yes

kkoziara:
@fulmicoton, should `Value.tokenized_text` return None if called on variant different than `TokStr` or should it panic?

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ Tantivy 0.11.0
 - API change around `Box<BoxableTokenizer>`. See detail in #629
 - Avoid rebuilding Regex automaton whenever a regex query is reused. #639 (@brainlock)
 - Add footer with some metadata to index files. #605 (@fdb-hiroshima)
+- Added handling of pre-tokenized text fields (#642), which will enable users to
+  load tokens created outside tantivy. See usage in examples/pre_tokenized_text.
```
fulmicoton: Please put your name or your github handle here.

examples/pre_tokenized_text.rs
```diff
@@ -0,0 +1,141 @@
...
+use tantivy::tokenizer::{SimpleTokenizer, Token, TokenStream, TokenizedString, Tokenizer};
+
+use tantivy::collector::{Count, TopDocs};
+use tantivy::query::TermQuery;
+use tantivy::schema::*;
+use tantivy::{doc, Index, ReloadPolicy};
+use tempfile::TempDir;
+
+fn tokenize_it(text: &str) -> Vec<Token> {
+    let mut ts = SimpleTokenizer.token_stream(text);
```
fulmicoton: `token_stream`.

Examples are the last place we want to use abbreviated variable names.

examples/pre_tokenized_text.rs
```diff
@@ -0,0 +1,141 @@
...
+
+use tantivy::tokenizer::{SimpleTokenizer, Token, TokenStream, TokenizedString, Tokenizer};
+
+use tantivy::collector::{Count, TopDocs};
+use tantivy::query::TermQuery;
+use tantivy::schema::*;
+use tantivy::{doc, Index, ReloadPolicy};
+use tempfile::TempDir;
+
+fn tokenize_it(text: &str) -> Vec<Token> {
```
fulmicoton: This is not an iterator. Let's not call that it.

rename to `pretokenize_text`.

examples/pre_tokenized_text.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+    tokens
+}
+
+fn main() -> tantivy::Result<()> {
+    let index_path = TempDir::new()?;
+
+    let mut schema_builder = Schema::builder();
+
+    // now we add `TOKENIZED` `TextOptions` to mark field as pre-tokenized
```
fulmicoton: PRETOKENIZED

examples/pre_tokenized_text.rs
```diff
@@ -0,0 +1,141 @@
...
+    // Now lets create a document and add our `TokenizedString` using
+    // `add_tokenized_text` method of `Document`
+    let mut old_man_doc = Document::default();
+    old_man_doc.add_tokenized_text(title, &title_tok);
+    old_man_doc.add_tokenized_text(body, &body_tok);
+
+    // ... now let's just add it to the IndexWriter
+    index_writer.add_document(old_man_doc);
+
+    // `Document` can be obtained directly from JSON:
```
fulmicoton: Pretokenized text can also be fed as JSON as follows ...

src/core/index_meta.rs
```diff
@@ -285,6 +285,6 @@ mod tests {
             payload: None,
         };
         let json = serde_json::ser::to_string(&index_metas).expect("serialization failed");
-        assert_eq!(json, r#"{"segments":[],"schema":[{"name":"text","type":"text","options":{"indexing":{"record":"position","tokenizer":"default"},"stored":false}}],"opstamp":0}"#);
+        assert_eq!(json, r#"{"segments":[],"schema":[{"name":"text","type":"text","options":{"indexing":{"record":"position","tokenizer":"default"},"stored":false,"tokenized":false}}],"opstamp":0}"#);
```
fulmicoton: pretokenized

src/schema/document.rs
```diff
@@ -78,6 +79,12 @@ impl Document {
         self.add(FieldValue::new(field, value));
     }
 
+    /// Add a text field with tokens.
```
fulmicoton: Add a `pre-tokenized` text.

Ideally for that kind of stuff, we should have specific `Pre-tokenized text` feature page, and a link to that whereever needed.

src/schema/schema.rs
```diff
@@ -450,7 +451,8 @@ mod tests {
         "record": "basic",
         "tokenizer": "raw"
       },
-      "stored": false
+      "stored": false,
+      "tokenized": false
```
fulmicoton: pretokenized

src/schema/text_options.rs
fulmicoton: Everyyyyywhere :)

src/schema/text_options.rs
```diff
@@ -109,6 +123,14 @@ pub const TEXT: TextOptions = TextOptions {
         record: IndexRecordOption::WithFreqsAndPositions,
     }),
     stored: false,
+    tokenized: false,
+};
+
+/// The field is already tokenized, should come as TokenizedString
+pub const TOKENIZED: TextOptions = TextOptions {
```
fulmicoton: PRETOKENIZED

src/schema/text_options.rs
```diff
@@ -9,6 +9,7 @@ use std::ops::BitOr;
 pub struct TextOptions {
     indexing: Option<TextFieldIndexing>,
     stored: bool,
+    tokenized: bool,
```
fulmicoton: Here goes the annoying comment, that will require extra work... :)

The problems with putting this on `TextOptions` is that it makes it possible to represent nonsensical TextOptions.
For instance, what is the point of a text field that is not indexed but is pretokenized?

For this reasons, the field tokenizer is defined in `TextFieldIndexing`. 
Here are alternative options that are better, and we can discuss
- Declare the concept of pretokenized as a tokenizer (I did not investigate the implication in the rest of the code)
- Remove all change in the schema. If the document carries tokens, then use those and ignore the configure tokenizer  <----- I like this solution better.

src/tokenizer/tokenized_string.rs
```diff
@@ -0,0 +1,188 @@
+use crate::tokenizer::{Token, TokenStream, TokenStreamChain};
+use std::cmp::Ordering;
+
+/// Struct representing tokenized text
```
fulmicoton: pretokenized

fulmicoton:
@kkoziara Very impressive change!

I put a lot of comments, but I really want you to read the one in text_options.rs first because it makes a lot of the other comment obsolete.

I think a nicer way to handle the API is to remove any implication with the schema. Transparently, if the user feeds in token information in the `Document`, tantivy should just ignore the tokenizer that is set in the schema and use your code.

What do you think? @petr-tik @kkoziara ?

I note that you decided to entirely avoid the problem of "how do we tokenize in the query parser". This is indeed a tricky question. I am ok with shipping this change without answering it.

src/schema/text_options.rs
kkoziara: If you are concerned mainly with the nonsensical TextOptions I think we could simply make pretokenized field being always indexed as well i.e. `PRETOKENIZED` `TextOptions` would simply be `TEXT` with `pretokenized` set to `true`.
Making API simpler is a good argument, but there are some things which we should consider. 1) Taking away indication from the schema that the field is pre-tokenized will make us guess the `Value` type( pretokenized vs text) from JSON when parsing and 2) this allows for fields with mixed` PretokenizedString` and `String` values.
I think both of these are not really a problem (but maybe I miss something), so we can go with eliminating the flag from `TextOptions`.

kkoziara:
@fulmicoton I answered the questions regarding the schema in the comment above. In general I think we can remove the flag from `TextOptions`.
I think the problem of "how do we tokenize in the query parser" can be somewhat offloaded to the users now. They know what kind of tokens will be indexed and might use a custom tokenizer that allows them to simply specify the queries. If the tokens do not contain whitespaces they can use existing `SimpleTokenizer` and concatenate tokens in the query with the whitespace. Next step could be adding to tantivy a tokenizer which splits on custom regex or tokenizer which understands escaped whitespaces.

src/schema/text_options.rs
fulmicoton: > If you are concerned mainly with the nonsensical TextOptions I think we could simply make pretokenized field being always indexed as well i.e. PRETOKENIZED TextOptions would simply be TEXT with pretokenized set to true.

I won't discuss the first solution, since we both seem to agree the second one (removing pretokenized from the schema) is better. Let's go for this then.

> Taking away indication from the schema that the field is pre-tokenized will make us guess the Value type( pretokenized vs text) 

Yes. An alternative solution would be to have a Text Value type that optionally have a `pretokenized: Option<TokenizedText>`. Either way the deserialization/serialization logic needs to be a bit smarter.

petr-tik:
> I think the problem of "how do we tokenize in the query parser" can be somewhat offloaded to the users now. They know what kind of tokens will be indexed and might use a custom tokenizer that allows them to simply specify the queries. If the tokens do not contain whitespaces they can use existing `SimpleTokenizer` and concatenate tokens in the query with the whitespace. 

I think this PR doesn't need to grow any more and we can deal with changes in the query parser in another ticket. Please open it with a good example of a query and how it should be parsed. 

> Next step could be adding to tantivy a tokenizer which splits on custom regex or tokenizer which understands escaped whitespaces.

Not sure if this is related to the above or is a separate feature request. Either way, please open a ticket for this too (or add to the ticket for the one above).

fulmicoton:
@petr-tik  @kkoziara Sorry for the long radio silence.
When using Pretokenized fields, do we want the tokenized version of the field to be serialized in the DocStore too?

src/indexer/segment_writer.rs
```diff
@@ -158,26 +159,45 @@ impl SegmentWriter {
...
+                    let mut offsets = vec![];
+                    let mut total_offset = 0;
+
+                    for field_value in field_values {
+                        match field_value.value() {
+                            Value::PreTokStr(tok_str) => {
+                                offsets.push(total_offset);
+                                total_offset += match tok_str.tokens.last() {
+                                    Some(token) => token.offset_to,
+                                    None => 0,
```
fulmicoton: Minor nitpick.

I think the following code is clearer:

```rust
if let Some(last_token) = tok_str.tokens.last() {
     total_offset += last_token.offset_to;
}
```

src/tokenizer/tokenized_string.rs
```diff
@@ -0,0 +1,188 @@
...
+        tok_strings: &'a [&'a PreTokenizedString],
+    ) -> Box<dyn TokenStream + 'a> {
+        if tok_strings.len() == 1 {
+            Box::new(PreTokenizedStream::from((*tok_strings[0]).clone()))
+        } else {
+            let mut offsets = vec![];
+            let mut total_offset = 0;
+            for &tok_string in tok_strings {
+                offsets.push(total_offset);
+                let offset = match tok_string.tokens.last() {
```
fulmicoton:
```

if let Some(last_token) = token_strings {
    total_offset += token.offset_to;
}

File: src/tokenizer/tokenized_string.rs [hunk 19]
```diff
@@ -0,0 +1,188 @@
...
+                    Some(token) => token.offset_to,
+                    None => 0,
+                };
+                total_offset += offset;
+            }
+            let token_streams: Vec<_> = tok_strings
+                .iter()
+                .map(|tok_string| PreTokenizedStream::from((*tok_string).clone()))
+                .collect();
```
fulmicoton: is clone required here?

src/tokenizer/tokenized_string.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+impl TokenStream for PreTokenizedStream {
+    fn advance(&mut self) -> bool {
+        self.current_token += 1;
+        self.current_token < self.tokenized_string.tokens.len() as i64
+    }
+
+    fn token(&self) -> &Token {
+        if self.current_token < 0 {
```
fulmicoton: this is equivalent to 

```rust
assert!(self.current_token < 0, "TokenStream not initialized. You should call advance() at least once.")
```

src/tokenizer/tokenized_string.rs
```diff
@@ -0,0 +1,188 @@
...
+            panic!("TokenStream not initialized. You should call advance() at least once.")
+        }
+        &self.tokenized_string.tokens[self.current_token as usize]
+    }
+
+    fn token_mut(&mut self) -> &mut Token {
+        if self.current_token < 0 {
+            panic!("TokenStream not initialized. You should call advance() at least once.")
```
fulmicoton: same as above.

src/tokenizer/tokenized_string.rs
```diff
@@ -0,0 +1,188 @@
...
+                position: 4,
+                text: String::from("a"),
+                position_length: 1,
+            },
+        ];
+        let mut i = 0;
+        while tok_stream.advance() {
+            assert!(*tok_stream.token() == expected_tokens[i]);
+            i += 1;
+        }
```
fulmicoton, line 186: A token_stream emitting no tokens at all would pass this test.
There are different ways to fix that. For instance...

```rust
for expected_token in expected_tokens {
    assert!(token_stream.advance());
    assert_eq!(token_stream().token(), &expected_token);
}
assert!(!token_stream.advance());
```

fulmicoton:
@kkoziara Incorporated minor changed in #687 Can you have a look at them. I also sent an invite to the tantivy-search team.

@kkoziara @petr-tik I am not sure we want to keep the pre-tokenized stuff in the store... What do you guys think?

kkoziara:
@fulmicoton I pushed the fixes for tests and refactored code a bit. Thanks for pointing out the problems. I didn't get the invite to the team though. 
As for the pre-tokenized text in store I also had doubts. I thought that it might be a good idea to store the document in the exact form it was provided by the user e.g for reindexing purposes. The easiest way to achieve this was to add proper serialization for pre-tokenized text values and I went this way. I guess we could ignore the tokens totally after they were indexed by serializing only the text from pre-tokenized text field, and treating it as string field from this moment on. It would be a simple solution and would mostly require removing some code. What do you think?

petr-tik:
Hey @kkoziara - thanks for pinging again, sorry about the delay - I was travelling to a very remote part of England with barely any internet. My excitement about this change has only grown and I want to land it. 

I believe some of fulmicoton's review comments are still unaddressed, please include them in your next commit. 

> As for the pre-tokenized text in store I also had doubts. I thought that it might be a good idea to store the document in the exact form it was provided by the user e.g for reindexing purposes. The easiest way to achieve this was to add proper serialization for pre-tokenized text values and I went this way. 

The (de)serialisation logic is one of the components I am least proficient in here, so I will defer to Paul on this. It feels potentially wasteful to store pre-tokenised text, if the only way to query it is via a query parser that will go through the same tokenisation as the pre-tokenised text. 

Last I checked, we don't support re-indexing from inside tantivy - users needs to kick off a full indexing pipeline again every time they add a field (or we break index compatibility across versions). 

It's great that you are thinking about re-indexing for this feature but I don't think you should worry about solving a problem the rest of the library doesn't support. 

A todo or easily-greppable comment reminding "future us" what you end up doing with your reasons for doing so will be useful for future work about re-indexing.

fulmicoton (approved):

fulmicoton:
Reindexing is a VERY important feature.... But it is not something for tantivy. I could see that in a toshi or a Bayard. In that case they should probably use an external kv store.

I think we can merge this PR and add a separate ticket to replace pretokenized field by a text field when building the stored doc.

Issue #671, opened 2019-10-23
Unstable ordering of TopCollector results with the same score


brainlock:
Hi!

We're using a `TopCollector` to fetch results sorted by score, and also to implement pagination in a simple way. Given an `offset` and a `limit` by the client code, we always fetch the top `offset+limit` results, then return the last `limit` ones. For example, on page 1 we fetch the top 10 documents and return them, on page 2 we fetch the top 20 documents and return the last 10 of those, etc. This is good enough for us as we don't have many pages of results in practice.

First question: is this a completely nonsensical way to do pagination? Is there an obviously better way that I've overlooked? If so, please do tell me and ignore the rest of this issue :)

Provided that this is a legitimate usage of `TopCollector`, we ran into an edge case that could potentially break pagination.

The problem is that the BinaryHeap that backs TopCollector doesn't guarantee any ordering when returning its sorted contents (`into_sorted_vec()`), for items whose ordering evaluates to Equal.

This could cause some items to never be included in a search result, if they are fetched after the max capacity is reached when querying for page 1 (for example TopCollector(10)), and if they are sorted in a position < 10 when querying for page 2 (TopCollector(20)).

In the end it really boils down to the following test not passing:

```rust

    #[test]
    fn test_top_collector_stable_ordering_for_equal_feature() {
        let mut top_collector_limit_2 = TopSegmentCollector::new(0, 2);
        top_collector_limit_2.collect(4, 3.14);
        top_collector_limit_2.collect(5, 3.14);
        top_collector_limit_2.collect(6, 3.14);

        let mut top_collector_limit_3 = TopSegmentCollector::new(0, 3);
        top_collector_limit_3.collect(4, 3.14);
        top_collector_limit_3.collect(5, 3.14);
        top_collector_limit_3.collect(6, 3.14);

        assert_eq!(
            top_collector_limit_2.harvest(),
            top_collector_limit_3.harvest()[..2].to_vec(),
        );
    }
```

In order to be able to use TopCollector to do pagination in this way, I think we need the results to be stable when items have Ordering::Equal (the specific order of course doesn't matter, it just needs to be the same when querying with different limits).

Any suggestion?

Thank you!

Alberto

----------------

# Solution 

The binary heap as used today by construction picks the first doc encountered in case of a tie.
This behavior is a side-effect of the code, and is consistent at the leaf level and when merging the different leaf so-called "fruits".

[...]

PR #675, opened 2019-10-24
TopDocs: ensure stable sorting on equal score

brainlock:
See gh-671

### DIFF
--- a/src/collector/top_collector.rs
+++ b/src/collector/top_collector.rs
@@ -37,12 +37,12 @@ impl<T: PartialOrd, D: PartialOrd> Ord for ComparableDoc<T, D> {
         let by_feature = other
             .feature
             .partial_cmp(&self.feature)
-            .unwrap_or_else(|| Ordering::Equal);
+            .unwrap_or(Ordering::Equal);
 
         let lazy_by_doc_address = || {
             self.doc
                 .partial_cmp(&other.doc)
-                .unwrap_or_else(|| Ordering::Equal)
+                .unwrap_or(Ordering::Equal)
         };
 
         // In case of a tie on the feature, we sort by ascending
--- a/src/collector/top_collector.rs
+++ b/src/collector/top_collector.rs
@@ -254,3 +254,69 @@ mod tests {
         );
     }
 }
+
+#[cfg(all(test, feature = "unstable"))]
+mod bench {
+    use super::TopSegmentCollector;
+    use test::Bencher;
+
+    #[bench]
+    fn bench_top_segment_collector_collect_not_at_capacity(b: &mut Bencher) {
+        let mut top_collector = TopSegmentCollector::new(0, 400);
+
+        b.iter(|| {
+            for i in 0..100 {
+                top_collector.collect(i, 0.8);
+            }
+        });
+    }
+
+    #[bench]
+    fn bench_top_segment_collector_collect_at_capacity(b: &mut Bencher) {
+        let mut top_collector = TopSegmentCollector::new(0, 100);
+
+        for i in 0..100 {
+            top_collector.collect(i, 0.8);
+        }
+
+        b.iter(|| {
+            for i in 0..100 {
+                top_collector.collect(i, 0.8);
+            }
+        });
+    }
+
+    #[bench]
+    fn bench_top_segment_collector_collect_and_harvest_many_ties(b: &mut Bencher) {
+        b.iter(|| {
+            let mut top_collector = TopSegmentCollector::new(0, 100);
+
+            for i in 0..100 {
+                top_collector.collect(i, 0.8);
+            }
+
+            // it would be nice to be able to do the setup N times but still
+            // measure only harvest(). We can't since harvest() consumes
+            // the top_collector.
+            top_collector.harvest()
+        });
+    }
+
+    #[bench]
+    fn bench_top_segment_collector_collect_and_harvest_no_tie(b: &mut Bencher) {
+        b.iter(|| {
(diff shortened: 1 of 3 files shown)


brainlock:
If you look at the second commit, if we add the constraint that D is also PartialOrd in ComparableDoc, we can move the second ordering comparison directly in its implementation. Let me know whether you prefer solution 1 or 2, and I'll rebase.

src/collector/top_collector.rs
```diff
@@ -175,6 +180,26 @@ impl<T: PartialOrd + Clone> TopSegmentCollector<T> {
     }
 }
 
+fn stable_sort<T, D>(a: &ComparableDoc<T, D>, b: &ComparableDoc<T, D>) -> Ordering
```
fulmicoton: it's not a `stable_sort` so it should not be called stable sort.

fulmicoton: I usually name that kind of argument `left` and `right`. A lot of ppl use `lhs`, and `rhs` (as left hand side, and right hand side). Let's avoid `a` and `b` anyway.

src/collector/top_collector.rs
```diff
@@ -25,29 +24,41 @@ struct ComparableDoc<T, D> {
...
+impl<T: PartialOrd, D: PartialOrd> Ord for ComparableDoc<T, D> {
     #[inline]
     fn cmp(&self, other: &Self) -> Ordering {
-        other
+        // Reversed to make BinaryHeap work as a min-heap
+        let by_feature = other
             .feature
             .partial_cmp(&self.feature)
-            .unwrap_or_else(|| Ordering::Equal)
+            .unwrap_or_else(|| Ordering::Equal);
```
fulmicoton: `unwrap_or` is better here.

fulmicoton: Alternatively you could write it :

```rust

....
   .unwrap_or_else(lazy_by_doc_address)

```

src/collector/top_collector.rs
```diff
@@ -175,6 +180,26 @@ impl<T: PartialOrd + Clone> TopSegmentCollector<T> {
...
+    D: PartialOrd,
+{
+    // Reversed since our ComparableDoc reverses the ordering (to
+    // make BinaryHeap work as a min-heap).
+    let by_feature = b
+        .feature
+        .partial_cmp(&a.feature)
+        .unwrap_or_else(|| Ordering::Equal);
+
+    let lazy_by_doc_address = || a.doc.partial_cmp(&b.doc).unwrap_or_else(|| Ordering::Equal);
```
fulmicoton: `unwrap_or(Ordering::Equal)`..

The closure version is only useful if you do computation in it.

src/collector/top_collector.rs
```diff
@@ -175,6 +180,26 @@ impl<T: PartialOrd + Clone> TopSegmentCollector<T> {
...
+where
+    T: PartialOrd,
+    D: PartialOrd,
+{
+    // Reversed since our ComparableDoc reverses the ordering (to
+    // make BinaryHeap work as a min-heap).
+    let by_feature = b
+        .feature
+        .partial_cmp(&a.feature)
+        .unwrap_or_else(|| Ordering::Equal);
```
fulmicoton: no need for `_or_else`. See below.

fulmicoton:
I added a couple of comments. 

The first version is less likely to induce a performance regression...
But the second version is actually easier to proofread and more robust to changes...

Let's check that there are no perf regression with version 2 and go with version 2. Do you know how to run the search benchmark?

fulmicoton:
Can you add your change in the changelog ?
And don't forget your username... Boasting is our only salary in opensource ;)

fulmicoton:
I don't think there is a `cargo bench` for the TopCollector. (I think there used to be one but I removed it).

I'm ok with `cargo bench` if you want such a bench. (that would be very useful for the future)
Otherwise, the search-benchmark-game.

brainlock:
I ended up doing both, the search-benchmark-game felt like a useful "integration test".

The search-benchmark-game doesn't show any significant regression:
[search-benchmark-game-pr-675-against-master.tar.gz](https://github.com/tantivy-search/tantivy/files/3772334/search-benchmark-game-pr-675-against-master.tar.gz)
The reference run is the merge base of this PR.

Running the new benchmarks doesn't show any regression. Of the four, the one that does take longer in the PR (v2, so the latest commit) is the one for the cases where there are many ties on the score (`bench_top_segment_collector_collect_and_harvest_many_ties`), but that was expected as in that case more comparisons have to be done.

src/collector/top_collector.rs
brainlock: (this code doesn't exist anymore if we go for V2)

brainlock: (this code doesn't exist anymore if we go for V2)

src/collector/top_collector.rs
brainlock: fixed

src/collector/top_collector.rs
brainlock: fixed

src/collector/top_collector.rs
brainlock: fixed

brainlock:
@fulmicoton If you think the current diff is good to go, I'll rebase this into one commit.

fulmicoton:
@brainlock Awesome thank you !

brainlock:
Oops I wanted to rebase the ugly tmps and fixups away :/ Oh well too late :-)

Thanks for merging the pr!

PR #679, opened 2019-10-30
Use `slice::iter` instead of `into_iter` to avoid future breakage

TethysSvensson:
Use `slice::iter` instead of `into_iter` to avoid future breakage

`an_array.into_iter()` currently just works because of the autoref
feature, which then calls `<[T] as IntoIterator>::into_iter`. But
in the future, arrays will implement `IntoIterator`, too. In order
to avoid problems in the future, the call is replaced by `iter()`
which is shorter and more explicit.

A crater run showed that your crate is affected by a potential future 
change. See https://github.com/rust-lang/rust/pull/65819 for more information.

### DIFF
--- a/src/directory/managed_directory.rs
+++ b/src/directory/managed_directory.rs
@@ -327,8 +327,7 @@ mod tests_mmap_specific {
                 .unwrap();
             assert!(managed_directory.exists(test_path1));
             assert!(managed_directory.exists(test_path2));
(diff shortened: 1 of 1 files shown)


fulmicoton:
@TethysSvensson This does not compile on ARM. Check the travis-ci output.

fulmicoton:
Thanks!

PR #682, opened 2019-10-31
fix crash on commit with deleted documents

### LINKED ISSUE #681: Crash on commit with deleted documents
The following sequence of events:

    - start with an empty index
    - add and then delete a document
    - add at least 32 documents
    - commit
    - commit again

leads to a crash on the second commit operation

```
thread 'segment_updater0' panicked at 'index out of bounds: the len is 4 but the index is 4', src/fastfield/delete.rs:65:21
```

Repro code:
```rust
        let mut schema_builder = schema::Schema::builder();
        let text_field = schema_builder.add_text_field("text", schema::TEXT);
        let index = Index::create_from_tempdir(schema_builder.build()).unwrap();

        let mut index_writer = index.writer(3_000_000).unwrap();

        // there must be one deleted document in the segment
        index_writer.add_document(doc!(text_field=>"b"));
        index_writer.delete_term(Term::from_field_text(text_field, "b"));

        // we need enough data to trigger the bug (at least 32 documents)
        for _ in 0..32 {
            index_writer.add_document(doc!(text_field=>"c"));
        }

        // a delete here would fix up the internal data structure that is being
        // accessed out of bounds (would make it big enough)

        // first commit: fine
        index_writer.commit().unwrap();

        // second commit: crash
        // in the first commit, there was no "previous deleted bitset" to check
        index_writer.commit().unwrap();
```

PR with a fix incoming.

brainlock:
The following sequence of events:

- start with an empty index
- add and then delete a document
- add at least 32 documents
- commit
- commit again

leads to a crash on the second commit operation

```
thread 'segment_updater0' panicked at 'index out of bounds: the len is 4 but the index is 4', src/fastfield/delete.rs:65:21
```

This happens because the `DeleteBitSet` used to track the deleted
documents in a segment is sized to be big enough to represent the
max doc id added, but the `is_deleted()` method can be used with
arbitrarily big indexes, leading to the access out of bounds.

The ultimate culprit is the `advance_deletes` function, which is
doing:
```
        for doc in 0u32..max_doc {
            if segment_reader.is_deleted(doc) {
                delete_bitset.insert(doc as usize);
            }
        }
```

`max_doc` here is bigger than the maximum index of the delete bitset
because we added enough documents.

This patch fixes the crash by making `DeleteBitSet::is_deleted(doc)`
method safe to call with any index.

Fixes gh-681

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Tantivy 0.11.0
 - Avoid rebuilding Regex automaton whenever a regex query is reused. #639 (@brainlock)
 - Add footer with some metadata to index files. #605 (@fdb-hiroshima)
 - TopDocs collector: ensure stable sorting on equal score. #671 (@brainlock)
+- Fix crash when committing multiple times with deleted documents. #681 (@brainlock)
  
 ## How to update?
 
--- a/src/fastfield/delete.rs
+++ b/src/fastfield/delete.rs
@@ -61,10 +61,16 @@ impl DeleteBitSet {
     /// Returns true iff the document has been marked as deleted.
     #[inline(always)]
     pub fn is_deleted(&self, doc: DocId) -> bool {
-        let byte_offset = doc / 8u32;
-        let b: u8 = (*self.data)[byte_offset as usize];
-        let shift = (doc & 7u32) as u8;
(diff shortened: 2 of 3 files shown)


fulmicoton:
Fixed in #683

PR #685, opened 2019-11-04
Replace `scoped_pool::Pool` with `rayon::ThreadPool`

### LINKED ISSUE #652: Redesign the core/executor
As mentioned in https://github.com/tantivy-search/tantivy/issues/638#issuecomment-528652433, our core/executor relies on the [scoped_pool](https://github.com/reem/rust-scoped-pool/) crate, which looks abandoned over the last 2 years. 

The Executor struct relies on a ThreadPool from scoped_pool but the comments mention Rayon's ThreadPool. 

Using development activity as an indicator, Rayon looks like a better long-term dependency to use when building a core executor than scoped_pool. 

Rayon provides [a scope method](https://doc.servo.org/rayon/struct.ThreadPool.html#method.scope) to create a scope inside a ThreadPool. However, it requires the FnOnce to exhibit the Send trait. 

This discussion [outlines pros and cons](https://github.com/rayon-rs/rayon/issues/562) of this approach and how other people suggest working around it. 

If rayon doesn't work, we can roll our own on top of Crossbeam

kardeiz:
Resolves #652.

This was a pretty straightforward swap. However, `rayon::ThreadPoolBuilder::build` can fail, which lead to a small signature change for the public `Index::set_multithread_executor`.

There is a test failure (`directory::tests::test_mmap_directory`), but this is already failing on `master` and is unrelated to this change.

Please feel free to close if someone else was working on this...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,10 +50,10 @@ owned-read = "0.4"
 failure = "0.1"
 htmlescape = "0.3.1"
 fail = "0.3"
-scoped-pool = "1.0"
 murmurhash32 = "0.2"
 chrono = "0.4"
 smallvec = "0.6"
+rayon = "1"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.3"
@@ -63,6 +63,10 @@ rand = "0.7"
 maplit = "1"
 matches = "0.1.8"
 time = "0.1.42"
+
+[dev-dependencies.fail]
+version = "0.3"
+features = ["failpoints"]
 
 [profile.release]
 opt-level = 3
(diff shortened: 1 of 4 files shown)


petr-tik:
Hey @kardeiz - thanks for your awesome PR and welcome to tantivy. 

You are the first one to pro-actively pick up this ticket and send a good-looking PR. The logic looks good to me.

I think it will help us understand the performance of this change, if you run and show us the [benchmark suite](https://github.com/tantivy-search/search-benchmark-game/) against current master and your branch. 

I don't expect it to make a huge difference but I think it's better to be safe than sorry. 

Let me know, if you find it difficult to set up and run the benchmark suite locally. If you manage to, feel free to post the results as plain-text tables in a comment on this thread. 

Thanks again

fulmicoton:
Good Work. No need to run the bench. They don't even use multithreading.

fulmicoton:
(Just waiting for CI to push the button. )

PR #694, opened 2019-11-09
Add index compatibility check in footer

### LINKED ISSUE #662: Return an error if the foot version is incompatible with tantivy's version
For instance, we could do that with a method on the `VersionedFooter`.

```rust
pub fn is_compatible(&self) -> Result<()> {
   // ...
}
```

Also... The current data model for version might be too simple. 
For instance, using a compilation flag, one can decide to use LZ4 or Snappy for the docstore compression. This should be part of the footer version.

petr-tik:
Fixes #662 

Replace the `20` magic constant with COMMON_FOOTER_SIZE

Add a docstring showing how footer is serialised and ascii art
Add a test for footer length checking

### DIFF
--- a/src/directory/footer.rs
+++ b/src/directory/footer.rs
@@ -7,6 +7,8 @@ use std::io::Write;
 
 const COMMON_FOOTER_SIZE: usize = 4 * 5;
 
+type CrcHashU32 = u32;
+
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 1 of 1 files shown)


petr-tik:
I sent you a couple of questions about it on gitter, but this is what I have so far, including adding some docstrings and a test

fulmicoton:
Merged in a different PR.

PR #698, opened 2019-11-10
Prevent tokens from being stored in the document store.

kkoziara:
Commit adds prepare_for_store method to Document, which changes all
PreTokenizedString values into String values. The method is called
before adding document to the document store to prevent tokens from
being saved there. Commit also adds small changes to comments in
pre_tokenized_text example.
This follows the discussion in #669 PR.

### DIFF
--- a/src/schema/document.rs
+++ b/src/schema/document.rs
@@ -155,6 +155,21 @@ impl Document {
             .find(|field_value| field_value.field() == field)
             .map(FieldValue::value)
     }
+
+    /// Prepares Document for being stored in the document store
+    ///
+    /// Method transforms PreTokenizedString values into String
+    /// values.
+    pub fn prepare_for_store(&mut self) {
+        for i in 0..self.field_values.len() {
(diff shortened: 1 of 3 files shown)


src/schema/document.rs
```diff
@@ -155,6 +155,23 @@ impl Document {
...
             .map(FieldValue::value)
     }
+
+    /// Prepares Document for being stored in the document store
+    ///
+    /// Method transforms PreTokenizedString values into String
+    /// values.
+    pub fn prepare_for_store(&mut self) {
+        let mut i = 0;
+        while i < self.field_values.len() {
```
fulmicoton: why not a for-loop?

kkoziara: No particular reason. I'll change to `for` if that's the preferred way.

fulmicoton: Yes please use a for loop whenever possible. It makes it easier to proofread.

fulmicoton:
Merged in a different PR with an extra commit fixing the loop.

Issue #692, opened 2019-11-08
Out of bounds slice access


kornelski:
I've seen this a couple of times:

```
thread 'segment_updater0' panicked at 'slice index starts at 18446744073633252107 but ends at 27', src/libcore/slice/mod.rs:2591:5
stack backtrace:
  12: rust_begin_unwind
             at src/libstd/panicking.rs:307
  13: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
  14: core::slice::slice_index_order_fail
             at src/libcore/slice/mod.rs:2591
  15: <core::ops::range::Range<usize> as core::slice::SliceIndex<[T]>>::index
             at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libcore/slice/mod.rs:2766
  16: core::slice::<impl core::ops::index::Index<I> for [T]>::index
             at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libcore/slice/mod.rs:2568
  17: tantivy::directory::footer::Footer::from_bytes
             at /root/.cargo/git/checkouts/tantivy-cd4bdf03c5df22c3/6a8a855/src/directory/footer.rs:79
  18: tantivy::directory::footer::Footer::extract_footer
             at /root/.cargo/git/checkouts/tantivy-cd4bdf03c5df22c3/6a8a855/src/directory/footer.rs:85
  19: <tantivy::directory::managed_directory::ManagedDirectory as tantivy::directory::directory::Directory>::open_read
             at /root/.cargo/git/checkouts/tantivy-cd4bdf03c5df22c3/6a8a855/src/directory/managed_directory.rs:250
  20: tantivy::core::segment::Segment::open_read
             at /root/.cargo/git/checkouts/tantivy-cd4bdf03c5df22c3/6a8a855/src/core/segment.rs:80
  21: tantivy::core::segment_reader::SegmentReader::open
             at /root/.cargo/git/checkouts/tantivy-cd4bdf03c5df22c3/6a8a855/src/core/segment_reader.rs:149
  22: tantivy::indexer::index_writer::advance_deletes
             at /root/.cargo/git/checkouts/tantivy-cd4bdf03c5df22c3/6a8a855/src/indexer/index_writer.rs:142
  23: tantivy::indexer::segment_updater::SegmentUpdater::purge_deletes
             at /root/.cargo/git/checkouts/tantivy-cd4bdf03c5df22c3/6a8a855/src/indexer/segment_updater.rs:233
[...]

PR #699, opened 2019-11-10
Implementing is_compatible

petr-tik:
Opening this now to get feedback -  please don't merge yet. 

I have only found Footer::extract_footer used outside the module, so decided to hide Footer::from_bytes from public consumption

Add a new error type for IncompatibleIndex
add a prototypical call to footer.is_compatible() in ManagedDirectory::open_read to make sure we error before reading it further


petr-tik:
2 questions:

1. Is this compatibility logic correct? 
2. The [stacktrace](https://github.com/tantivy-search/tantivy/issues/692#issue-520006546) from #692 shows ManagedDirectory::open_read, which we currently deal with.

However, the local lib_search library [uses](https://gitlab.com/crates.rs/crates.rs/blob/master/search_index/src/lib_search_index.rs#L49) Index::open_in_dir, which calls into MmapDirectory, which currently doesn't create or check footers.

Therefore, this fix won't help people to check the compatibility of indexes they open through Index::open_in_dir. 

I don't think it's a great idea to surface the footer interface all the way to the user and i welcome any advice/recommendation on where to fit Footer::is_compatible in MmapDirectory::open_read.

petr-tik:
ManagedDirectory::wrap uses Directory::atomic_read, which looks like a better place to do footer compatibility checking (or is it both?).

src/directory/footer.rs
```diff
@@ -102,6 +105,27 @@ impl Footer {
     pub fn size(&self) -> usize {
         self.versioned_footer.size() as usize + self.meta.len() + COMMON_FOOTER_SIZE
     }
+
+    /// Compares the version found in the footer with the version of tantivy running
+    /// Indexes are incompatible when the major and minor versions are different between the
+    /// current library and serialised index are different
```
fulmicoton: This is incorrect. We now have two versions in the Footer. 

Tantivy's version is here for debugging purpose. What is important is the index version. Right now only 0 is available.

src/lib.rs
```diff
@@ -171,8 +171,9 @@ pub use crate::postings::Postings;
 pub use crate::reader::LeasedItem;
 pub use crate::schema::{Document, Term};
 
-/// Expose the current version of tantivy, as well
-/// whether it was compiled with the simd compression.
+/// Expose the current version of tantivy as found in Cargo.toml during compilation.
+/// eg. "0.11.0"
+/// as well whether it was compiled with the simd compression.
```
petr-tik, line 176: @fulmicoton Does this method give any meaningful info about simd compression in tantivy? Does this refer to snappy vs lz4 compression methods? 

If it doesn't - should it?

fulmicoton: Ah no the SIMD copmression bit is irrelevant today. There used to be a different codec, but now tantivy uses a scalar fixture when the instruction set is not available. All platform use the same codec.

PR #713, opened 2019-11-23
Added max segment size to consider and option to consider deletes.

### LINKED ISSUE #706: Add a maximum segment size for merge policies, Deletes
This ticket is about fixing the heuristic of the `LogMergePolicy`.
There is actually two different problems in this ticket.

# Problem 1

The relative benefit of merging segments diminishes as the segment gets bigger and bigger.
We should add a configurable `max_segment_size` expressed in number of documents to limit the size of two segments that can be considered for merge.

# Problem 2

The current merge policy only takes in account the size (in number of documents) to merge two segments. We do not take in account the number of deleted documents.

This can lead to the following problem. A very large segment is created, and the index consists in 
this large segment and a lot of small segments.

Until the small segments get merged up to the point when `min_merge_size` large segment are created, the large segment never gets merged.
This can take a very long time (e.g. months). In the meanwhile, deletes also happen and this mammoth segment may actually contain a very large ratio of deletes.

The heuristic should be changed to take in account the ratio of deleted file to avoid this problem.

hntd187:
This seems like a reasonable start to this.  Would close #706 , but I'd imagine we want to do some more work here.


src/indexer/log_merge_policy.rs
```diff
@@ -6,26 +6,40 @@ use std::f64;
 const DEFAULT_LEVEL_LOG_SIZE: f64 = 0.75;
 const DEFAULT_MIN_LAYER_SIZE: u32 = 10_000;
 const DEFAULT_MIN_MERGE_SIZE: usize = 8;
+const DEFAULT_MAX_MERGE_SIZE: usize = 5000;
 
 /// `LogMergePolicy` tries tries to merge segments that have a similar number of
```
petr-tik: Since you are working on this file, please delete the duplicate "tries" from this docstring

src/indexer/log_merge_policy.rs
```diff
@@ -6,26 +6,40 @@ use std::f64;
 const DEFAULT_LEVEL_LOG_SIZE: f64 = 0.75;
 const DEFAULT_MIN_LAYER_SIZE: u32 = 10_000;
 const DEFAULT_MIN_MERGE_SIZE: usize = 8;
+const DEFAULT_MAX_MERGE_SIZE: usize = 5000;
```
fulmicoton: That default is way too low. Let's make it `10_000_000` by default maybe?

src/indexer/log_merge_policy.rs
```diff
@@ -6,26 +6,40 @@ use std::f64;
...
 }
 
 impl LogMergePolicy {
     fn clip_min_size(&self, size: u32) -> u32 {
         cmp::max(self.min_layer_size, size)
     }
 
+    fn doc_count(&self, seg: &SegmentMeta) -> u32 {
+        if self.consider_deletes {
+            seg.num_docs() + seg.num_deleted_docs()
```
fulmicoton: that's `seg.max_doc()`... I am not sure this is a good heuristic here.
Can you remove the part of your PR that handles the deletes (and keep the `max_doc` thing) ?

I think the delete heuristic might require more discussion.

fulmicoton (changes requested):
Can you remove the bits of this PR that takes in account the number of deleted docs. I think the current heuristics is not good.

src/indexer/log_merge_policy.rs
hntd187: Sounds good to me

src/indexer/log_merge_policy.rs
hntd187: Sounds good, removed it

fulmicoton:
Closing this PR, See https://github.com/tantivy-search/tantivy/pull/765.
@halvorboe  addressed the comment in the CR and opened a new CR. 
I attributed the change to both of you in the CHANGELOG though.

PR #716, opened 2019-11-25
Updated rust-stemmers version to 1.2

chrihunstad:
New version including support for Norwegian language


fulmicoton (commented):
I'm ok with upgrading but please set the version to 1.2.

fulmicoton:
I'm ok with upgrading but please set the version to 1.2.

fulmicoton (changes requested):
I'm ok with upgrading but please set the version to 1.2.

PR #724, opened 2019-12-12
Make error handling richer in Footer::is_compatible

petr-tik:
Change return type to Result<bool, TantivyError>, instead of bool

Add an Incompatibility enum that enriches the IncompatibleIndex error variant
with information, which then allows us to generate a developer-friendly hint how
to upgrade library version or switch feature flags for a different compression
algorithm

Set up infra for TantivyError::IncompatibleIndex ->
OpenReadError::IncompatibleIndex -> TantivyError::IncompatibleIndex, so we can
ergonomically pass it through ManagedDirectory::open_* methods

Make strings identifying the compression algorithm both 4-characters ("snap" and
"lz4 ") long to make serde easier


petr-tik:
I have pulled your changes (thanks for fixing the test fail) and added my stuff on top. I haven't added tests on this - want to get your opinions on the PR in general before I defensively cover it with tests. 

Disregard the other PR - I "git push"ed and it also pushed my very WIP commits from a couple of weeks ago.

src/directory/footer.rs
```diff
@@ -86,16 +89,41 @@ impl Footer {
 
     /// Confirms that the index will be read correctly by this version of tantivy
     /// Has to be called after `extract_footer` to make sure it's not accessing uninitialised memory
-    pub fn is_compatible(&self) -> bool {
-        let version = &*crate::VERSION;
+    pub fn is_compatible(&self) -> Result<bool, TantivyError> {
```
fulmicoton: I don't understand the dichotomy between incompatibility with false and incompatibiliy expressed by an error.
What is the point?

petr-tik: I thought adding an error type makes it easier to surface it the error type to the top, where bool will fail without any information on how to fix it. 

When it's true, it is redundant - yes. We can replace bool with () in that case. 

In other cases (apart from the one of UnknownVersion), we can show an error and a specific fix to the user. 

They could have accidentally forgotten that they made an index with lz4 and are now using a library that was compiled with snappy. Seeing "false" gives them little actionable information on how to fix it.

fulmicoton: This is way too complicated for no reason at all.

Let's have it return `Result<(), Incompatibility>` or `bool`.
I don't think `TantivyError` is needed, did I miss something?

Also what is the point of the intermediary `IncompatibleIndex` ?

src/directory/footer.rs
```diff
@@ -120,7 +148,7 @@ impl BinarySerializable for VersionedFooter {
             VersionedFooter::V1 { crc32, compression } => {
                 // Serializes a valid `VersionedFooter` or panics if the version is unknown
                 // [   version    |   crc_hash  | compression_mode ]
-                // [    0..4      |     4..8    |  variable ]
```
fulmicoton: What was wrong with the variable compression mode?

src/directory/footer.rs
petr-tik: Returning an Ok(Incompatibility::CompressionMismatch) feels semantically weird to me. 

Index incompatiblity is an error after which we cannot do anything else with the index. 

> Also what is the point of the intermediary `IncompatibleIndex` ?

I don't understand your usage of the word intermediary - can you please explain

src/directory/footer.rs
petr-tik: I thought it would be simpler, if we fix the compression_mode as a 4-char string. 

In the future, when compile-time enums are supported (and if we add more compression modes), we can replace the compression string with an enum, which will (de)serialise to a u32 value.

src/directory/footer.rs
fulmicoton: | Returning an Ok(Incompatibility::CompressionMismatch) feels semantically weird to me.

I think it would be clearer.

My bad I think I misunderstood the nature of `IncompatibleIndex`. I think you can use `Incompatibility` as an error directly though.

petr-tik: If I understand your suggestion, the caller of is_compatible will unwrap the always Ok(incompatibility) and pattern match against different variants.  

Surely, we then lose the user friendly errors defined in TantivyError:: incompatible?

src/directory/error.rs
```diff
@@ -175,6 +176,15 @@ impl From<IOError> for OpenReadError {
     }
 }
 
+impl From<crate::error::TantivyError> for OpenReadError {
```
fulmicoton: can we remove this now?

src/directory/error.rs
```diff
@@ -186,8 +196,8 @@ impl fmt::Display for OpenReadError {
                 "an io error occurred while opening a file for reading: '{}'",
                 err
             ),
-            OpenReadError::IncompatibleIndex(ref footer) => {
-                write!(f, "Incompatible index format: {:?}", footer)
+            OpenReadError::IncompatibleIndex(ref _incompatibility) => {
+                unreachable!("This should never print - only the top-level TantivyError::IncompatibleIndex should")
```
fulmicoton: why is that so?

src/directory/footer.rs
fulmicoton: Err(Incompatibility) you mean? Why do we lose the user friendly errors?

src/directory/footer.rs
fulmicoton: I don't think it is simpler. You introduced an implicit contract about the compression string. It needs to be exactly 4 chars and the code is not simpler.

src/directory/footer.rs
```diff
@@ -86,16 +89,41 @@ impl Footer {
...
+    pub fn is_compatible(&self) -> Result<bool, TantivyError> {
+        let library = &*crate::VERSION;
         match &self.versioned_footer {
             VersionedFooter::V1 {
                 crc32: _,
                 compression,
             } => {
-                return compression == &version.store_compression;
+                // V1 means should be equal with 1
+                if library.index_format_version != 1u32 {
```
fulmicoton: this should never happen... i am ok with not testing this.

src/directory/footer.rs
```diff
@@ -86,16 +89,41 @@ impl Footer {
...
+                        library: library.clone(),
+                        index: self.version.clone(),
+                    }));
+                }
+
+                Ok(true)
             }
-            VersionedFooter::UnknownVersion { version: _ } => false,
+            // TODO - discuss if we should add another error type
+            VersionedFooter::UnknownVersion { version: _ } => return Ok(false),
```
fulmicoton: and return an `Err(IncompatibleIndex)` here instead.

petr-tik: If I am going to change the return type to 
`Result<(), Incompatibility>`, I cannot return `Err(IncompatibleIndex)` but I can return `Err(Incompatibility)`

src/directory/managed_directory.rs
```diff
@@ -267,9 +264,7 @@ impl Directory for ManagedDirectory {
         let read_only_source = self.directory.open_read(path)?;
         let (footer, reader) = Footer::extract_footer(read_only_source)
             .map_err(|err| IOError::with_path(path.to_path_buf(), err))?;
-        if !footer.is_compatible() {
-            return Err(OpenReadError::IncompatibleIndex(footer));
-        }
+        footer.is_compatible().unwrap();
```
petr-tik: Since I deleted the `From<TantivyError> for OpenReadError` shim, returning the Incompatibility wrapped inside the Err() from is_compatible makes it impossible to propagate it further. 

What do you want to return here?

PR #726, opened 2019-12-13
Make u64_lenient() handle f64 fast fields too

caio:
Without this, we get a panic during merge since the merger will get a `None` where it expects something.

Prior to this patch, you can reproduce the panic with:

```rust
    use tantivy::{
        self,
        schema::{SchemaBuilder, FAST},
        Document, Index, Result,
    };

    #[test]
    fn pass() -> Result<()> {
        let mut builder = SchemaBuilder::new();
        let field = builder.add_f64_field("f64", FAST);
        let index = Index::create_in_ram(builder.build());

        let mut writer = index.writer_with_num_threads(1, 50_000_000)?;

        for i in 0..1000 {
            let mut doc = Document::new();
            doc.add_f64(field, 0.42);
            writer.add_document(doc);

            if i % 5 == 0 {
                writer.commit()?;
            }
        }

        writer.commit()?;

        Ok(())
    }
```

I couldn't simply add this as a test since `cargo test pass` gets stuck for me when trying it (merger thread dies, main thread waits forever I'd guess).

### DIFF
--- a/src/indexer/merger.rs
+++ b/src/indexer/merger.rs
@@ -1496,7 +1496,12 @@ mod tests {
     #[test]
     fn merges_f64_fast_fields_correctly() -> crate::Result<()> {
         let mut builder = schema::SchemaBuilder::new();
+
+        let fast_multi = IntOptions::default().set_fast(Cardinality::MultiValues);
+
         let field = builder.add_f64_field("f64", schema::FAST);
+        let multi_field = builder.add_f64_field("f64s", fast_multi);
+
         let index = Index::create_in_ram(builder.build());
 
         let mut writer = index.writer_with_num_threads(1, 3_000_000)?;
@@ -1508,7 +1513,11 @@ mod tests {
 
         for i in 0..100 {
             let mut doc = Document::new();
-            doc.add_u64(field, 42);
+            doc.add_f64(field, 42.0);
+
+            doc.add_f64(multi_field, 0.24);
+            doc.add_f64(multi_field, 0.27);
+
             writer.add_document(doc);
 
             if i % 5 == 0 {
--- a/src/indexer/merger.rs
+++ b/src/indexer/merger.rs
@@ -1492,4 +1492,37 @@ mod tests {
             assert_eq!(&vals, &[20]);
         }
     }
+
+    #[test]
+    fn merges_f64_fast_fields_correctly() -> crate::Result<()> {
+        let mut builder = schema::SchemaBuilder::new();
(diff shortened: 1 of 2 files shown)


fulmicoton:
Nice catch. That is indeed terrible! Can you add a simple unit test to your PR?

caio:
Sure thing!  There's nothing public on IndexMerger (besides `open`) and it doesn't look like it'll be trivial to manually build it, so can you give me a little bit of a direction of where to place the test? A "take a look at test X that does something similar" should be more than enough to get me started.

fulmicoton:
I'm ok with something similar with the unit test you wrote. Create 2 or more segment within a loop and wait for merging threads.

You can simply assert that the number of segment is equal to one at the end. 
just put a comment explaining you are testing that merging f64 fields works as expected with a referenec to this issue.

caio:
Done! My luck is so great that my test code was causing the creation of `7` segments, which didn't trigger the merge into a single segment since the default threshold is 8 :rofl: 

Hold up on the merging even if this looks good though: It looks like `u64s_lenient` (so, for the multi valued float fast fields) will need a patch as well and I'm hoping to pile it on this PR.

Let me know what you prefer wrt commits, I can squash everything together once we're happy if that's what you like.

caio:
Now I think we're good!

src/indexer/merger.rs
```diff
@@ -1492,4 +1492,46 @@ mod tests {
...
+        let mut builder = schema::SchemaBuilder::new();
+
+        let fast_multi = IntOptions::default().set_fast(Cardinality::MultiValues);
+
+        let field = builder.add_f64_field("f64", schema::FAST);
+        let multi_field = builder.add_f64_field("f64s", fast_multi);
+
+        let index = Index::create_in_ram(builder.build());
+
+        let mut writer = index.writer_with_num_threads(1, 3_000_000)?;
```
fulmicoton, line 1507: cool

src/indexer/merger.rs
```diff
@@ -1492,4 +1492,46 @@ mod tests {
...
+
+        // Make sure we'll attempt to merge every created segment
+        let mut policy = crate::indexer::LogMergePolicy::default();
+        policy.set_min_merge_size(2);
+        writer.set_merge_policy(Box::new(policy));
+
```
fulmicoton, line 1513: exactly :)

fulmicoton:
Thank you @caio !!!

caio:
Thank you! Tantivy is awesome and I hope I'll be able to come back and help if I can soon

PR #737, opened 2019-12-20
[RFC] directory: Convert the ReadOnlySource to provide a Read + Seek based …

poljar:
This PR implements what was briefly discussed on Gitter, a `ReadOnlySource` implementation that uses a `Read + Seek` based API.

Please do note that this is for now a initial implementation, a proof of concept. Plenty of places do questionable things or are a bit sloppy and need to be improved but, it can serve for us as a starting point for discussion. That being said it does work correctly with the current `Directory` implementations and passes all tests.

The motivation for a `Read + Seek` based implementation is [this](https://github.com/matrix-org/seshat/blob/master/src/encrypted_dir.rs#L564) `Directory` implementation that transparently encrypts/decrypts files that Tantivy reads/writes. This could be quite important for client applications that provide full text search over sensitive data.

I am sure that there are more use-cases for such an `Read + Seek` based API, and making the `Directory` more flexible is a welcome improvement.

A downside of this API is that allocating memory buffers and copying data is necessary instead of the direct access that we get with the slice based approach.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,6 @@ tantivy-query-grammar = { version="0.11", path="./query-grammar" }
 bitpacking = {version="0.8", default-features = false, features=["bitpacker4x"]}
 census = "0.4"
 fnv = "1.0.6"
-owned-read = "0.4"
 failure = "0.1"
 htmlescape = "0.3.1"
 fail = "0.3"
--- a/src/common/composite_file.rs
+++ b/src/common/composite_file.rs
@@ -1,6 +1,7 @@
 use crate::common::BinarySerializable;
 use crate::common::CountingWriter;
 use crate::common::VInt;
+use crate::common::HasLen;
 use crate::directory::ReadOnlySource;
 use crate::directory::{TerminatingWrite, WritePtr};
 use crate::schema::Field;
@@ -8,7 +9,7 @@ use crate::space_usage::FieldUsage;
 use crate::space_usage::PerFieldSpaceUsage;
 use std::collections::HashMap;
 use std::io::Write;
-use std::io::{self, Read};
+use std::io::{self, Read, Cursor};
 
 #[derive(Eq, PartialEq, Hash, Copy, Ord, PartialOrd, Clone, Debug)]
 pub struct FileAddr {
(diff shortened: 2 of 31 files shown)


fulmicoton:
I am afraid it would take too much time for me to go through this CR and get all changes required for it to be mergeable.

I am however very interested in eventually moving away from the current `ReadOnlySource`: 
- to be WASI friendly
- to allow for a distant (HDFS for instance) directory.

Could you comment on your specific use case ? This would help me reprioritize this change.

poljar:
The main motivation for this, as already stated, is a `Directory` implementation that transparently encrypts and decrypts the files Tantivy writes and reads.

The directory implementation can be found [here](https://github.com/matrix-org/seshat/blob/master/src/encrypted_dir.rs#L86). If there is interest for it, it could be upstreamed as well once it's done.

The problem with the current `ReadOnlySource` for the encrypted directory implemention is that it requires direct access to an array, or a type that has the following trait bound `Deref<Target = [u8]>`. This requirement probably comes from the memmap implementation that Tantivy uses from [here](https://docs.rs/memmap/0.7.0/memmap/struct.Mmap.html#deref-methods). Vectors of course share the same `Deref` trait implementation, and are used for the `RamDirectory` implementation.

This forces us to decrypt and load all the file content into memory in the `open_read()` implementation found [here](https://github.com/matrix-org/seshat/blob/master/src/encrypted_dir.rs#L86).

If you are on the other hand asking for the motivation for an encrypted index, we are developing a secure chat protocol with first class support for history. Naturally users want to have the ability to search their chat history, in public rooms that are not end to end encrypted the server provides such functionality. On the other hand in end to end encrypted rooms the server doesn't have the ability to read and index chat messages, thus the indexing needs to be done client side.

Since the chat messages are presumably sensitive information we are storing them encrypted and would like to store the index encrypted as well. This would avoid any kind of message leakage in the case of device theft for example.

I understand that this PR isn't in any shape of form mergeable as is, it isn't marked as an RFC for nothing. It is meant as a proof of concept for a `Read` + `Seek` based `ReadOnlySource` implementation.

If there is interest for Tantivy to converge to such a more flexible `ReadOnlySource` implementation we can split up the PR into smaller chunks that gradually get away from the `Deref<Target = [u8]>` requirement of its data source.

fulmicoton:
@poljar thanks for the thorough reply.

I simply had forgotten that conversation. Encryption is a valid use case...
It will happen most likely be made possible one day in the future.
As you probably notice in your refactoring, the most problematic part is the term
dictionary.

> If there is interest for Tantivy to converge to such a more flexible ReadOnlySource implementation we can split up the PR into smaller chunks that gradually get away from the Deref<Target = [u8]> requirement of its data source.

Still, this will require a lot of review and input from my side and I don't have the bandwidth to help with this atm. I hope we can make this happen in 2020.

fulmicoton:
@poljar Are you still using tantivy on matrix? I am currently working on a branch that should allow to write such a dictionary efficiently.

poljar:
> @poljar Are you still using tantivy on matrix? I am currently working on a branch that should allow to write such a dictionary efficiently.

Oh that's nice to hear. Yes, Tantivy is shipped with Element-desktop and we're taking the hit of keeping the index in memory. The average index size should be around 20-30MB, not horrible, but would be nice to get rid of that.

PR #742, opened 2019-12-25
Make SchemaBuilder::add_field() public

mosuka:
Make SchemaBuilder::add_field() public so that it can add a FieldEntry directly.
Use Case:
- For example, it can easily add user-specified schema to system default schema. see test code in the PR.


fulmicoton:
CI is failing due to `cargo fmt`. Can you fix that?

mosuka:
Sorry, I fixed that.

mosuka:
Thanks!

PR #744, opened 2019-12-29
Expose UserOperation as a public type.

knutwalker:
In order to make `IndexWriter::run` callable from outside of the create,
the `UserOperation` type needs to be publicly available.
Since the `indexer` module is private, we just export the `UserOperation`
type directly.


fulmicoton:
Thanks @knutwalker !

PR #745, opened 2019-12-29
Add atomic mutex implementation for ARM

### LINKED ISSUE #743: Cannot cross compile for ARMv5TE
I'm trying to compile the basic tantivy example for armv5te arch, and found the following error:

```
 $ CC=arm-linux-gnueabi-gcc-7 cargo build --target=armv5te-unknown-linux-musleabi
...
   Compiling tantivy v0.11.3
error[E0432]: unresolved import `std::sync::atomic::AtomicU64`
 --> /home/davide/.cargo/registry/src/github.com-1ecc6299db9ec823/tantivy-0.11.3/src/indexer/stamper.rs:3:25
  |
3 | use std::sync::atomic::{AtomicU64, Ordering};
  |                         ^^^^^^^^^
  |                         |
  |                         no `AtomicU64` in `sync::atomic`
  |                         help: a similar name exists in the module: `AtomicU32`
```

According to [this](https://doc.rust-lang.org/std/sync/atomic/#portability), atomics should work on linux armv5te builds (using kernel helpers), but apparently only AtomicU32 is available.

This issue actually prevents me to use tantivy on my armv5te based NAS.
Maybe stamper could be implemented in a more portable way, leveraging the max atomic size of the target platform: this will obviously limit its capabilities, but it should be fine to run.

davide-romanini:
Based on previous version [here](https://github.com/tantivy-search/tantivy/blob/da4691383956c27baeb118cb990c16dba1cf5ad1/src/indexer/stamper.rs).

This PR actually forces the mutex version on all arm devices. I tried to use the experimental `#[cfg(target_has_atomic = "64")]`, that seems to be more appropriate for this scenario, but failed.

With this change, I successfully compiled and run the basic example on my armv5te based Zyxel NSA325v2.

Fixes #743.

### DIFF
--- a/src/indexer/stamper.rs
+++ b/src/indexer/stamper.rs
@@ -1,13 +1,13 @@
-use std::ops::Range;
-use std::sync::Arc;
-use std::sync::atomic::Ordering;
 use crate::Opstamp;
+use std::ops::Range;
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
 
 #[cfg(not(target_arch = "arm"))]
 mod atomic_impl {
 
-    use std::sync::atomic::{AtomicU64, Ordering};
     use crate::Opstamp;
+    use std::sync::atomic::{AtomicU64, Ordering};
 
     #[derive(Default)]
     pub struct AtomicU64Wrapper(AtomicU64);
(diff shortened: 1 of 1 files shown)


fulmicoton:
LGTM , but can you run `cargo fmt` ?

fulmicoton:
It seems you missed one line in your rustfmt run. I manually ran it on your branch and merge than into master. Closing your PR.

Issue #547, opened 2019-05-17
Specify field importance via weight


kkonevets:
**Is your feature request related to a problem? Please describe.**

Sometimes it is desirable to give more importance to one field over another. For example,  query matching title of a document is more important than query matching body of a document.

**Describe the solution you'd like**

It would be nice to make queries like `title:(milk shake)^3 || body:(milk shake)`
This means that `title` field is 3 times more important than `body` field. This query syntax was taken from Solr.

kornelski:
If you add field weights, please also give it a programmatic API. I'd like to set weights of fields regardless of what query user makes (not necessarily override user-provided weight, but definitely provide defaults for all fields)

petr-tik:
I'm starting exploring this and wanted to double check a couple of things. 

## query boosting?
that we only need field boosting, not query boosting. Reusing the example above, does anyone need the ability to write:

"title:(oreo milk shake)^3 || title:(milk shake)". 

Obviously, they are different and require different implementations, hence I'm asking upfront. 

## negative/fractional boosts
Is it useful and valid to de-prioritise certain fields with a negative or fractional weight?

## composition of boost factors
How should boost factors compose? eg. If you have a field that got a boost of 5 at schema build-time and an individual query comes in like "(field: less important phrase)^0.2" - this should be back at 1 - default boost factor or am I missing anything? 

# interface

Currently I'm thinking about adding a boost method to the Field trait, which will allow both:
Schema::builder().add_u64_field().boost_by(3)
title:(milk shake)^3 - where the query parser will call boost_by with 3 parsed from the query.

kornelski:
I use weights in 0..=1 range, so fractional weights are useful. If it only accepts integers I'd probably use something like `boost_by(10000 * frac)`, and that would be fine as long as boosting all fields by numbers much larger than 3 doesn't cause any weirdly skewed results or overflows.

kornelski:
I don't have immediate use for boosts in the query, but I wonder if it was useful for processing of synonyms, e.g. if I preprocessed "oreo milk shake" query to something like:

"title:(oreo milk shake)^3 || title:(cookie milk drink)"

but that's a half-baked idea. It can probably be flipped around to boosting preprocessed fields instead:

"title:(oreo milk shake) || title_lower_boost:(cookie milk drink)"

audunhalland:
[...]

PR #753, opened 2020-01-20
Low level Query boost_by

audunhalland:
My naive attempt at adding "low-level" query boosting.

Adds `boost_by(self, score: f32) -> Self` method to `TermQuery`, `PhraseQuery`, `RegexQuery`, `FuzzyQuery`. This style does not break API compatibility for the various queries. Some internal function signatures have changed.

#547

### DIFF
--- a/src/query/automaton_weight.rs
+++ b/src/query/automaton_weight.rs
@@ -15,6 +15,7 @@ use tantivy_fst::Automaton;
 pub struct AutomatonWeight<A> {
     field: Field,
     automaton: Arc<A>,
+    boost: f32,
 }
 
 impl<A> AutomatonWeight<A>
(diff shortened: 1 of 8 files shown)


fulmicoton:
We need boosting to be enforced either by the `Query` trait or the `Weight` trait or it will be a nightmare.

@petr-tik I forgot what we discussed regarding the weight. Do we want it to sit in the `Query` or do we want it to be passed as an extra argument in the `.weight()` method.

Right now my feeling is `.weight(..)` might be better, but I don't remember what we discussed last time.

audunhalland:
You mean you have to support boosting also the compound queries (`BooleanQuery`)? Is boosting that query itself different from individually boosting each of its leaf queries?

PR #756, opened 2020-01-24
Move from `failure` to `thiserror`

Hirevo:
This PR removes the dependency on [**`failure`**](https://github.com/rust-lang-nursery/failure) and replaces it with [**`thiserror`**](https://github.com/dtolnay/thiserror).  

The upside of this change is that Tantivy's error types now implement the `std::error::Error` trait directly, making them more easily usable with the `?` in a function returning a `Result<_, Box<dyn std::error::Error>>`, for example.  
The error messages are exactly the same and defined in a similar manner (just `#[error("...")]` rather than `#[fail(display = "...")]`).  

Also, it is possible to remove a lot of manual `From` implementations with the `#[from]` annotation from `thiserror` (and also have `Error::source` point to the inner error automatically), but I have not done it yet here, but if there is interest into doing this migration, then I can gladly help with it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ bitpacking = {version="0.8", default-features = false, features=["bitpacker4x"]}
 census = "0.4"
 fnv = "1.0.6"
 owned-read = "0.4"
-failure = "0.1"
+thiserror = "1.0"
 htmlescape = "0.3.1"
 fail = "0.4"
 murmurhash32 = "0.2"
--- a/src/directory/error.rs
+++ b/src/directory/error.rs
@@ -5,18 +5,16 @@ use std::io;
 use std::path::PathBuf;
 
 /// Error while trying to acquire a directory lock.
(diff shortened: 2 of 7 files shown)


fulmicoton:
Generally speaking, I advise against sending that kind of PR to tantivy (or any other project for that matter). This is a controversial change, that has a lot of implication and needs to be discussed.

I would appreciate if you would open an issue instead, rather than open a PR.
Please open an issue so that we can gahter the discussion there.

Hirevo:
Yes, looking back at it, I agree that it is very sudden for a breaking change like this.  
The issue #760 is now opened to discuss it there instead.

Hirevo:
With the deprecation of the `failure` crate, I have updated this PR to the latest master revision of tantivy to ease the process of migrating, if you decide to adopt this going forward (most notably for the 1.0 release).

fulmicoton:
Merged as a different branch

PR #757, opened 2020-01-24
Backward iteration for termdict range

halvorboe:
Depending on https://github.com/tantivy-search/fst/pull/12.

Solves https://github.com/tantivy-search/tantivy/issues/689.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,7 @@
-Tantivy 0.12.1
-=======================
-- Added backward iteration for `TermDictionary` stream. (@halvorboe)
-
 Tantivy 0.12.0
 ======================
 - Removing static dispatch in tokenizers for simplicity. (#762)
+- Added backward iteration for `TermDictionary` stream. (@halvorboe)
 
 ## How to update?
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+Tantivy 0.11.4
+=======================
+- Added backwards iteration of `TermStreamer` range (@halvorboe)
+
 Tantivy 0.11.3
 =======================
 - Fixed DateTime as a fast field (#735)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 Tantivy 0.12.1
 =======================
-- Added backward iteration for ` TermDictionary` stream. (@halvorboe)
+- Added backward iteration for `TermDictionary` stream. (@halvorboe)
 
 Tantivy 0.12.0
 ======================
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tantivy"
-version = "0.12.1"
+version = "0.12.0"
 authors = ["Paul Masurel <paul.masurel@gmail.com>"]
 license = "MIT"
 categories = ["database-implementations", "data-structures"]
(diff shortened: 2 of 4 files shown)


fulmicoton:
Make sure CI passes. Here the issue is that you need to format your code apparently
 `cargo fmt --all`.

halvorboe:
Fixed it now. Still need to merge the PR in fst and update the dependency.

fulmicoton:
I don't understand the change from backward to backwards.
From what I understand backward is the US en and backwards is british en.
https://www.quickanddirtytips.com/education/grammar/backward-versus-backwards

Also, assuming the change of name is a good idea, this breaks API and requires to inc the version to `0.3.0`, not `0.2.1`.

fulmicoton:
See comment in the tantivy-fst PR conversaiton tab.

src/termdict/mod.rs
```diff
@@ -352,41 +352,91 @@ mod tests {
         let source = ReadOnlySource::from(buffer);
         let term_dictionary: TermDictionary = TermDictionary::from_source(&source);
 
-        let value_list = |mut streamer: TermStreamer<'_>| {
+        let value_list = |mut streamer: TermStreamer<'_>, backwards: bool| {
```
fulmicoton: I'd remove the `backwards: bool`  arg here and compute both backward and forward.
Reverse the backward one, and assert_eq! that it is equal to forward, and then return the result.

Note that it is not considered a good practies in the industry to make non atomic unit test. I don't think it is necessary to observe this practise in a project like tantivy.

halvorboe:
My bad. Googled it after making the change. Turns out backward is "as correct" as backwards - just looked really wrong to me. Reverted the change.

Cargo.toml
```diff
@@ -18,7 +18,7 @@ byteorder = "1.0"
 crc32fast = "1.2.0"
 once_cell = "1.0"
 regex ={version = "1.3.0", default-features = false, features = ["std"]}
-tantivy-fst = "0.2"
+# tantivy-fst = "0.2"
```
fulmicoton: I merged and published tantivy-fst = 0.2.1.

Can you update this and remove `[dependencies.tantivy-fst]`

fulmicoton:
LGTM

Can you:
- update `Cargo.toml` to use tantivy-fst = 0.2.1 directly
- add your change in the CHANGELOG, (and don't forget to mention your name!)

Cargo.toml
fulmicoton: I merged and published tantivy-fst = 0.2.1.

Can you update this and remove [dependencies.tantivy-fst]

fulmicoton (changes requested):
LGTM

Can you:

update Cargo.toml to use tantivy-fst = 0.2.1 directly
add your change in the CHANGELOG, (and don't forget to mention your name!)

Cargo.toml
halvorboe: Done :D

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+Tantivy 0.12.1
```
fulmicoton: It will part of 0.12.0.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tantivy"
-version = "0.11.3"
+version = "0.12.1"
```
fulmicoton: 0.12.0

fulmicoton (changes requested):

fulmicoton:
Awesome thank you @halvorboe!

PR #759, opened 2020-01-25
Facet path string

halvorboe:
Fix for https://github.com/tantivy-search/tantivy/issues/580.

### DIFF
--- a/src/schema/facet.rs
+++ b/src/schema/facet.rs
@@ -122,6 +122,11 @@ impl Facet {
     pub fn to_path(&self) -> Vec<&str> {
         self.encoded_str().split(|c| c == FACET_SEP_CHAR).collect()
     }
+
+    /// Extract the path as a string from the `Facet`.
+    pub fn to_path_string(&self) -> String {
+        "//".to_owned() + &self.to_path().join("/")
+    }
 }
 
 impl Borrow<str> for Facet {
@@ -264,5 +269,19 @@ mod tests {
         let v = ["first", "second", "third\\/not_fourth"];
         let facet = Facet::from_path(v.iter());
         assert_eq!(facet.to_path(), v);
+    }
+
+    #[test]
+    fn test_to_path_string() {
+        let v = ["first", "second", "third\\/not_fourth"];
+        let facet = Facet::from_path(v.iter());
+        assert_eq!(facet.to_path_string(), "//first/second/third\\/not_fourth");
+    }
+
+    #[test]
+    fn test_to_path_string_empty() {
+        let v: Vec<&str> = vec![];
+        let facet = Facet::from_path(v.iter());
+        assert_eq!(facet.to_path_string(), "//");
     }
 }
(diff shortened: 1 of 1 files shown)


fulmicoton:
@halvorboe your impl is incorrect, but this is probably due to confusing existing unit test.

The existing implementation in display is correct. Could you simply copy paste the logic from fmt::Display and call your method in fmt::Display. No need to care about the extra allocs.

Explanation:
There are only two representations of Facets:

The text one is //toto/titi.
The different step of the path may contain / or \ in which case it needs to be escaped by \

The "binary" one is 0u8 separated. The steps are not escaped.

src/schema/facet.rs
```diff
@@ -122,6 +122,11 @@ impl Facet {
     pub fn to_path(&self) -> Vec<&str> {
         self.encoded_str().split(|c| c == FACET_SEP_CHAR).collect()
     }
+
+    /// Extract the path as a string from the `Facet`.
+    pub fn to_path_string(&self) -> String {
+        "//".to_owned() + &self.to_path().join("/")
```
fulmicoton: Escaping is missing here. See comment in the Conversation tab of this PR.

src/schema/facet.rs
```diff
@@ -122,6 +122,11 @@ impl Facet {
     pub fn to_path(&self) -> Vec<&str> {
         self.encoded_str().split(|c| c == FACET_SEP_CHAR).collect()
     }
+
+    /// Extract the path as a string from the `Facet`.
+    pub fn to_path_string(&self) -> String {
+        format!("/{}", self.to_string())
```
fulmicoton: The initial double slash is not a bad idea, but this is not the way `Facet` have been designed.

We want `.to_path_string()` to be the opposite of `Facet::from(&str)`.

src/schema/facet.rs
```diff
@@ -265,4 +270,21 @@ mod tests {
...
         assert_eq!(facet.to_path(), v);
     }
+
+    #[test]
+    fn test_to_path_string() {
+        let v = ["first", "second", "third/not_fourth"];
+        let facet = Facet::from_path(v.iter());
+        assert_eq!(
+            facet.to_path_string(),
+            String::from("//first/second/third\\/not_fourth")
```
fulmicoton: `/first/second/third\\/not_fourth`

src/schema/facet.rs
```diff
@@ -265,4 +270,21 @@ mod tests {
...
+        );
+    }
+
+    #[test]
+    fn test_to_path_string_empty() {
+        let v: Vec<&str> = vec![];
+        let facet = Facet::from_path(v.iter());
+        assert_eq!(facet.to_path_string(), "//");
```
fulmicoton: `/`

src/schema/facet.rs
```diff
@@ -122,6 +122,11 @@ impl Facet {
     pub fn to_path(&self) -> Vec<&str> {
         self.encoded_str().split(|c| c == FACET_SEP_CHAR).collect()
     }
+
+    /// Extract the path as a string from the `Facet`.
```
fulmicoton: /// This function is the inverse of `Facet::from(&str)`.

src/schema/facet.rs
halvorboe: My bad, was due to the first comment you made on this issue. Can see what the original issue said now.

PR #765, opened 2020-01-29
Maximum size of segment to be considered for merge

### LINKED ISSUE #706: Add a maximum segment size for merge policies, Deletes

halvorboe:
Replicated changes from #713. Would close #706.

### DIFF
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -897,7 +897,7 @@ mod tests {
         let index_writer = index.writer(3_000_000).unwrap();
         assert_eq!(
             format!("{:?}", index_writer.get_merge_policy()),
-            "LogMergePolicy { min_merge_size: 8, min_layer_size: 10000, \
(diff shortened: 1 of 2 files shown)


fulmicoton:
thanks!

PR #767, opened 2020-01-29
Optimize TermDictionary::empty by lazy data source

### LINKED ISSUE #766: TermDictionary::empty() is too slow
**Describe the bug**
- What did you do?
Search for infrequent terms.
- What happened?
The search was too slow.
- What was expected?
That the search was fast.

0.11.3

**To Reproduce**
Create an index in RAM, and index a few documents with rather unique terms in each of them.
A RAM index will (based on my knowledge) create one segment per document.
Search for term unique to one document, or a term not present in the data set. The more infrequent the term is, the slower the search will be. I've tracked this down to slow `SegmentReader::inverted_index(term)` which in turn calls `TermDictionary::empty()`, where the term is not present in the segment. `TermDictionary::empty()` takes _5ms_ on my machine in debug mode. Multiply this with the number of segments by the number of non-matched terms in the full query, and a compound-query search in my very small RAM index easily takes 200-300ms.

Note that I haven't verified that this is the _only_ reason for the observed slowness.

audunhalland:
Fixes #766 .

My test queries now all complete in about 1ms, whereas before they would easily take 200-300ms.

### DIFF
--- a/src/termdict/termdict.rs
+++ b/src/termdict/termdict.rs
@@ -4,8 +4,8 @@ use crate::common::BinarySerializable;
 use crate::common::CountingWriter;
 use crate::directory::ReadOnlySource;
 use crate::postings::TermInfo;
-use crate::schema::FieldType;
 use crate::termdict::TermOrdinal;
+use once_cell::sync::Lazy;
 use std::io::{self, Write};
 use tantivy_fst;
 use tantivy_fst::raw::Fst;
@@ -29,7 +29,7 @@ where
     W: Write,
 {
     /// Creates a new `TermDictionaryBuilder`
-    pub fn create(w: W, _field_type: &FieldType) -> io::Result<Self> {
+    pub fn create(w: W) -> io::Result<Self> {
         let fst_builder = tantivy_fst::MapBuilder::new(w).map_err(convert_fst_error)?;
         Ok(TermDictionaryBuilder {
             fst_builder,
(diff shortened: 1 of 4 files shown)


src/termdict/termdict.rs
```diff
@@ -122,14 +130,8 @@ impl TermDictionary {
     }
 
     /// Creates an empty term dictionary which contains no terms.
-    pub fn empty(field_type: &FieldType) -> Self {
```
audunhalland, line 125: There was in fact no dependence on knowing anything about `FieldType`. So, to show that the same data source can be used for all types of `TermDictionary`, I have removed this argument.

audunhalland:
The downside to `Lazy`-style fixes is of course that behaviour becomes more difficult to measure, because the search engine needs to "warm up" first.

src/termdict/termdict.rs
```diff
@@ -29,7 +29,7 @@ where
     W: Write,
 {
     /// Creates a new `TermDictionaryBuilder`
-    pub fn create(w: W, _field_type: &FieldType) -> io::Result<Self> {
```
fulmicoton, line 32: I wonder what I was thinking here.

src/termdict/termdict.rs
fulmicoton: yes. no clue what this was about

PR #772, opened 2020-02-03
[WIP]  Alternative take on boosted queries

fulmicoton:


audunhalland:
Currently `BoostQuery` isn't `pub use`, so I can't test this branch in my own project yet

fulmicoton:
@audunhalland Now it is pub.

src/query/query.rs
```diff
@@ -104,4 +105,35 @@ impl QueryClone for Box<dyn Query> {
     }
 }
 
+pub struct BoostQuery {
```
audunhalland: Probably this needs a `::new` function

fulmicoton:
Added. `QueryParser` support too.

PR #785, opened 2020-03-03
Add Lindera to README.md

mosuka:
Could you add [Lindera](https://github.com/lindera-morphology/lindera-tantivy) to the third party tokenizer listed in README.md?


fulmicoton:
Can you put lindera in first place?

mosuka:
Thanks.
I put Lindera at the beginning of a Japanese tokenizer. How about this?

PR #793, opened 2020-03-13
Implemented PollWatcher

halvorboe:
Initial attempt at implementing the same behavior as notify.

TODO:
- [ ] Write tests for the actual watcher. Currently using the tests for the directory.
- [x] Refactor the code and remove notify.
- [x] Make poll rate constant/configureable.


halvorboe:
Found a cross platform bug here. On linux the resolution of the modification time is 1s, which makes it pretty much useless. Will looks into it.

src/directory/poll_watcher.rs
```diff
@@ -0,0 +1,73 @@
...
+
+impl PollWatcher {
+    pub fn new(path: &Path) -> Result<Self, OpenDirectoryError> {
+        let watcher_router: Arc<WatchCallbackList> = Default::default();
+        let watcher_router_clone = watcher_router.clone();
+        let meta_path = path.to_owned().join(*META_FILEPATH);
+        thread::Builder::new()
+            .name("meta-file-watch-thread".to_string())
+            .spawn(move || {
+                let mut current_hash: u64 = Self::hash(&meta_path).unwrap_or(0);
```
fulmicoton: what is the point of this hash?

src/directory/poll_watcher.rs
```diff
@@ -0,0 +1,73 @@
...
+                let mut current_meta_time: u128 = Self::last_update(&meta_path).unwrap_or(0);
+                loop {
+                    let new_hash: u64 = Self::hash(&meta_path).unwrap_or(0);
+                    let new_meta_time: u128 = Self::last_update(&meta_path).unwrap_or(0);
+                    // println!("hash: {} {}\ntime: {} {}\n---", new_hash, current_hash, new_meta_time, current_meta_time);
+                    if (new_hash != current_hash) || (new_meta_time != current_meta_time) {
```
fulmicoton: how is it possible to have a different hash? you are not checking the content of the file, you are checking the hash of the path right now.

src/directory/poll_watcher.rs
```diff
@@ -0,0 +1,73 @@
...
+use std::path::Path;
+use std::sync::Arc;
+use std::thread;
+use std::time::{Duration, UNIX_EPOCH};
+use std::fs;
+use std::collections::hash_map::DefaultHasher;
+use std::hash::{Hash, Hasher};
+
+
+const POLL_INTERVAL_MS: u64 = 10;
```
fulmicoton: `500ms` would be a more sensible default I believe.

fulmicoton: can we make it configurable? that would be nice espeically for unit tests.

fulmicoton (changes requested):
- make the polling interval configurable and increase the current default
- revisit the hash logic . I suspect it is broken.

src/directory/poll_watcher.rs
halvorboe: The problem is: Linux has a 1s granularity for the modified time on files. This means that if 10 updates happen in a second only the first one is caught. There are two possible solutions to this that I can see:
1. Use a hash of the file (which is what I was experimenting with).
2. Delay the update event util the next second to make sure that any update with the same modification time is in the refresh.

src/directory/poll_watcher.rs
halvorboe: I will try to make it configureable.

src/directory/poll_watcher.rs
halvorboe: The hash function is the function on line 57. It does give a different hash for different file contents.

src/directory/poll_watcher.rs
fulmicoton: I don't really understand the logic 

if `hashchange || mtime changed`. Is it purely for defensive reasons against collisions?
If so, the file is super small. I'd rather avoid the extra syscall and extra logic and spend a bit of CPU on a hash with negligeable collision probability md5, sha256 or whatnot.

src/directory/poll_watcher.rs
fulmicoton: Ah I see. `Self::hash` vs `calculate_hash`. Can you have a name that express discrimate the two semantics?

`Self::hash` could be renamed `hash_file_content` for instance.

src/directory/poll_watcher.rs
```diff
@@ -0,0 +1,73 @@
...
+    s.finish()
+}
+
+impl PollWatcher {
+    pub fn new(path: &Path) -> Result<Self, OpenDirectoryError> {
+        let watcher_router: Arc<WatchCallbackList> = Default::default();
+        let watcher_router_clone = watcher_router.clone();
+        let meta_path = path.to_owned().join(*META_FILEPATH);
+        thread::Builder::new()
+            .name("meta-file-watch-thread".to_string())
```
fulmicoton, line 50: Thanks for recalling to name the thread. This really helps debugging.

src/directory/poll_watcher.rs
```diff
@@ -0,0 +1,73 @@
...
+            .spawn(move || {
+                let mut current_hash: u64 = Self::hash(&meta_path).unwrap_or(0);
+                let mut current_meta_time: u128 = Self::last_update(&meta_path).unwrap_or(0);
+                loop {
+                    let new_hash: u64 = Self::hash(&meta_path).unwrap_or(0);
```
fulmicoton: So if we receive an io::Error here, it will send the event once, and then never send an event again as long as an io::Error is returned when reading the file.

I think it is worth at least logging something here.

src/directory/poll_watcher.rs
```diff
@@ -0,0 +1,73 @@
...
+    t.hash(&mut s);
+    s.finish()
+}
+
+impl PollWatcher {
+    pub fn new(path: &Path) -> Result<Self, OpenDirectoryError> {
+        let watcher_router: Arc<WatchCallbackList> = Default::default();
+        let watcher_router_clone = watcher_router.clone();
+        let meta_path = path.to_owned().join(*META_FILEPATH);
+        thread::Builder::new()
```
fulmicoton, line 49: When does this thread terminates?

src/directory/poll_watcher.rs
```diff
@@ -0,0 +1,73 @@
...
+            .name("meta-file-watch-thread".to_string())
+            .spawn(move || {
+                let mut current_hash: u64 = Self::hash(&meta_path).unwrap_or(0);
+                let mut current_meta_time: u128 = Self::last_update(&meta_path).unwrap_or(0);
+                loop {
+                    let new_hash: u64 = Self::hash(&meta_path).unwrap_or(0);
+                    let new_meta_time: u128 = Self::last_update(&meta_path).unwrap_or(0);
+                    // println!("hash: {} {}\ntime: {} {}\n---", new_hash, current_hash, new_meta_time, current_meta_time);
+                    if (new_hash != current_hash) || (new_meta_time != current_meta_time) {
+                        // println!("update...");
```
fulmicoton: remove

src/directory/poll_watcher.rs
```diff
@@ -0,0 +1,73 @@
...
+                        let _ = watcher_router_clone.broadcast();
+                    }
+                    thread::sleep(Duration::from_millis(POLL_INTERVAL_MS));
+                }
+            })?;
+        Ok(Self { watcher_router })
+    }
+
+    fn hash(path: &Path) -> Result<u64, io::Error> {
+        let data = fs::read_to_string(path)?;
```
fulmicoton: read to string decodes utf-8. The meta.json file is guaranteed to be utf-8, but we don't need that.

src/directory/poll_watcher.rs
fulmicoton: If you can test what happens on a spurious `io::Error` using fail-rs it would be super cool.It is a bit of work, so if you can't just add a todo on this file :)

src/directory/poll_watcher.rs
halvorboe: I have prototypes of SHA256 and MD5 as well. Just wanted to see if it could be implemented without adding any dependencies. If it runs every 500ms the performance should not be that different anyways. 

The reason why I'm testing hashchange or time is that I'm having a lot of strange bugs when running the code on different platforms. Tests pass almost every time on MacOS and inside docker on my computer. In the CI they don't pass. 

Will experiment more in my playground and benchmark the different solutions.

fulmicoton:
Implemented in #924.

PR #797, opened 2020-03-19
Fix prefix option for FuzzyTermQuery

Peachball:
The object created from `FuzzyTermQuery::new_prefix` before had it's prefix field set, but didn't use it anywhere.


fulmicoton:
Hi @Peachball thank you very much for the bug fix. Can you edit the CHANGELOG.md as well?

You will have to add a new version. If this bug is blocking for you I can publish a hotfix as a `0.12.1` , otherwise `0.13.0` .

Peachball:
Sure - I've updated the changelog with version `0.13.0`.

Issue #806, opened 2020-04-13
Make TweakScore and CustomScore mutable at the segment level.


fulmicoton:
Quoting @robyoung

```
Basically, I have a recipe search where the user provides a list of available ingredients. I would like to tweak the score so that recipes that can be made with no additional ingredients score highest.
I have tried to do this with TopDocs::tweak_score with no luck. Partly because I can't seem to access facet fields in here because FacetReader::facet_ords requires mutability. And partly because I can't figure out how to get the facet ords from the query.
```

fulmicoton:
Do we need it Sync too!?

fulmicoton:
Do we need facet_reader to be mut too?

robyoung:
I think this issue should be closed now.

PR #807, opened 2020-04-13
Make TweakScore and CustomScore mutable at the segment level

robyoung:
Addresses issue #806

I have also added a test that shows using `tweak_score` with a facet. This is the problem I was originally describing when I raised this on Gitter. I'm not sure if this kind of test is wanted or whether this is the correct place for it. Would love some guidance on that.

### DIFF
--- a/src/collector/custom_score_top_collector.rs
+++ b/src/collector/custom_score_top_collector.rs
@@ -28,7 +28,7 @@ where
 /// It is the segment local version of the [`CustomScorer`](./trait.CustomScorer.html).
 pub trait CustomSegmentScorer<TScore>: 'static {
     /// Computes the score of a specific `doc`.
-    fn score(&self, doc: DocId) -> TScore;
+    fn score(&mut self, doc: DocId) -> TScore;
 }
 
 /// `CustomScorer` makes it possible to define any kind of score.
@@ -117,9 +117,9 @@ where
 
 impl<F, TScore> CustomSegmentScorer<TScore> for F
 where
-    F: 'static + Sync + Send + Fn(DocId) -> TScore,
+    F: 'static + FnMut(DocId) -> TScore,
 {
-    fn score(&self, doc: DocId) -> TScore {
+    fn score(&mut self, doc: DocId) -> TScore {
         (self)(doc)
     }
 }
--- a/src/collector/tweak_score_top_collector.rs
+++ b/src/collector/tweak_score_top_collector.rs
@@ -29,7 +29,7 @@ where
 /// It is the segment local version of the [`ScoreTweaker`](./trait.ScoreTweaker.html).
 pub trait ScoreSegmentTweaker<TScore>: 'static {
     /// Tweak the given `score` for the document `doc`.
-    fn score(&self, doc: DocId, score: Score) -> TScore;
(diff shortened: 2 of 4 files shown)


robyoung:
Is there anything else I can do to this PR to make it easier to review?

fulmicoton:
@robyoung nothing. I am usually much more reactive, but due to COVID life has been a little hectic recently. Please wait a little bit longer!

src/collector/custom_score_top_collector.rs
```diff
@@ -117,9 +117,9 @@ where
 
 impl<F, TScore> CustomSegmentScorer<TScore> for F
 where
-    F: 'static + Sync + Send + Fn(DocId) -> TScore,
+    F: 'static + Sync + Send + FnMut(DocId) -> TScore,
```
fulmicoton: I haven't tried it but can we remove `Sync` here too?

src/collector/tweak_score_top_collector.rs
```diff
@@ -121,9 +121,112 @@ where
...
+                    let query_ords: HashSet<u64> = facets
+                        .iter()
+                        .filter_map(|key| facet_dict.term_ord(key.encoded_str()))
+                        .collect();
+
+                    let mut facet_ords_buffer: Vec<u64> = Vec::with_capacity(20);
+
+                    move |doc: DocId, original_score: Score| {
+                        ingredient_reader.facet_ords(doc, &mut facet_ords_buffer);
+                        let missing_ingredients = facet_ords_buffer
```
fulmicoton: That seems very suboptimal... What about just iter().filter().count()?

src/collector/tweak_score_top_collector.rs
```diff
@@ -121,9 +121,112 @@ where
...
+#[cfg(test)]
+mod tests {
+    use crate::collector::TopDocs;
+    use crate::query::BooleanQuery;
+    use crate::schema::*;
+    use crate::{DocId, Index, Score, SegmentReader};
+    use std::collections::HashSet;
+
+    #[test]
+    fn tweak_score_with_facets() {
```
fulmicoton: This looks more like an example than a test. Could you move this to an example?

src/collector/custom_score_top_collector.rs
fulmicoton: ... and possibly Send.

src/collector/tweak_score_top_collector.rs
robyoung: Absolutely! Great idea, I knew it felt wrong here.

robyoung:
No worries at all, I totally understand. I have an 11 month old, COVID makes life hard. Thank you for the review, I'll take a look at addressing them later today.

src/collector/custom_score_top_collector.rs
robyoung: I don't really understand the ramifications of this but the tests pass.

src/collector/tweak_score_top_collector.rs
robyoung: Good point, this was here for debugging but when I was figuring things out but doesn't need to be any more.

fulmicoton:
Awesome you even thought about changing the custom scorer! merging.

PR #808, opened 2020-04-13
Switch fst dependency to git

robyoung:
This allows the package to be built without first cloning the
tantivy-search/fst repo into the expected place. This should fix CI.


petr-tik (approved):
LGTM - simple way to keep CI green. Please mention #803 in the commit message, so merging automatically closes the issue

fulmicoton:
thanks!

robyoung:
ace, thanks, is that ok or does it need to be in a specific format?

PR #814, opened 2020-04-19
Two phase docsets

ghost:
Two phase iteration over docsets. The idea is to be able to avoid using positions when not all required terms are present. An example query:
+lowfreqterm +"highfreqterm1 highfreqterm2"
For now this is mostly the groundwork in scorer.rs and an initial twophasedocset.rs.
To accept this a benchmark should show improvements for queries like above. I intend to follow Lucene, with an optional TwoPhaseApproximiation from any Scorer, to be used later for intersections.
As Rust is quite new to me, please holler when things appear to be improvable.
cargo build warns about missing docs, I'll add those later.

### DIFF
--- a/src/core/segment_reader.rs
+++ b/src/core/segment_reader.rs
@@ -295,8 +295,8 @@ impl SegmentReader {
     }
 
     /// Returns an iterator that will iterate over the alive document ids
-    pub fn doc_ids_alive(&self) -> SegmentReaderAliveDocsIterator<'_> {
-        SegmentReaderAliveDocsIterator::new(&self)
+    pub fn doc_ids_alive<'a>(&'a self) -> impl Iterator<Item = DocId> + 'a {
+        (0u32..self.max_doc).filter(move |doc| !self.is_deleted(*doc))
     }
 
     /// Summarize total space usage of this segment.
@@ -321,52 +321,6 @@ impl SegmentReader {
 impl fmt::Debug for SegmentReader {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "SegmentReader({:?})", self.segment_id)
-    }
-}
-
-/// Implements the iterator trait to allow easy iteration
-/// over non-deleted ("alive") DocIds in a SegmentReader
-pub struct SegmentReaderAliveDocsIterator<'a> {
-    reader: &'a SegmentReader,
-    max_doc: DocId,
-    current: DocId,
-}
-
-impl<'a> SegmentReaderAliveDocsIterator<'a> {
-    pub fn new(reader: &'a SegmentReader) -> SegmentReaderAliveDocsIterator<'a> {
-        SegmentReaderAliveDocsIterator {
-            reader,
-            max_doc: reader.max_doc(),
-            current: 0,
-        }
-    }
-}
-
-impl<'a> Iterator for SegmentReaderAliveDocsIterator<'a> {
-    type Item = DocId;
-
-    fn next(&mut self) -> Option<Self::Item> {
-        // TODO: Use TinySet (like in BitSetDocSet) to speed this process up
-        if self.current >= self.max_doc {
-            return None;
-        }
-
-        // find the next alive doc id
-        while self.reader.is_deleted(self.current) {
-            self.current += 1;
-
-            if self.current >= self.max_doc {
-                return None;
-            }
-        }
-
-        // capture the current alive DocId
-        let result = Some(self.current);
-
-        // move down the chain
-        self.current += 1;
-
-        result
     }
 }
 


fulmicoton:
Two phase docset would be a great addition. Please let me know when this is ready for review!

Due to the coronavirus, I might not be as reactive as usual, so don't be shy about pingping me several time if I don't react quickly.

ghost:
Health is more important.
I am progressing quite slowly, learning rust and the tantivy code on scorer and docset along the way.
In case I get really stuck I'll try and formulate some questions on the code in the twophase branch in my tantivy fork.
I'll try and merge master into that branch regularly.
I normally do microcommits, so an occasional squash might be better once things start compiling.
Meanwhile there are also some small doc improvements and cargo fmt results on other code, please feel free to pick those commits anytime.

ghost:
I am still trying to find a working design for this in Rust. Please let me know what you think of the ideas below.

When using two phases, some DocSets will be shared between these phases, and on different levels in the query structure. This leads to an acyclic graph, roughly like this:
Scorers > DocSets (as now), and
Scorers > TwoPhase... > DocSets (as approximations, i.e. for 1st phase) 
where some of these DocSets are shared.

Since DocSets are mutable, this leads to interior mutability for these DocSets:

Rc < RefCell < dyn DocSet > >

possibly with Vec<> also somewhere in there.

(Added 16 May 2020: DocSet and Scorer are traits, and Scorer implements DocSet, so the graph of
structs (objects?) implementing these will contain fewer structs than indicated above).



In the current code on the twophase branch in my tantivy clone, I only have added Box<> things which still leads to lifetime error messages from the compiler, more or less as expected for interior mutability.
The current code also has:
impl DocSet for &PhraseScorer
this should probably be reverted to remove the & before PhraseScorer.
There also is:
pub trait TwoPhaseDocSet: DocSet {} 
but once the DocSet is shared I'd expect that this extension of DocSet will need to be replaced by a  method that provides the DocSet for the approximation.

ghost:
In order to show to the intended direction I set an interim goal to let PhraseScorer implement TwoPhase, and to let a simple phrase query be scored from this.
The current twophase branch shows how far I got: it compiles with a warning but nothing more.

There is a Scorer that also implements a method to return a TwoPhase trait object.
This TwoPhase object has a matches() method that indicates whether the approximation provided by the Scorer is an actual match. The TwoPhase object also has a match_cost() method the estimates the number of operations for matches().

For a phrase score the TwoPhase trait is implemented by PhraseTwoPhase, and that needs some
reference to its PhraseScorer to implemented matches() by phrase_scorer.phrase_match().
I could not get find a working lifetime for that simple reference to PhraseScorer.
This is what I expected, see the comment in PhraseScorer.two_phase().
To fix the lifetime problem, the reference should probably be an Rc<RefCell<...>> to the PhraseScorer.
However, to implement that all current Box dyn Scorer may need to be changed to Rc<RefCell<...>>, and that would involve quite a bit of code . Iirc there are about 38 occurrences for the regexp Box.*Scorer in the tantivy code.

Some questions:
I used a trait TwoPhase and an object implementing that to indicate that the Scorer implements two phase query evaluation.

Would it be better to try and use generics so there would be only the Scorer object which implements two phase when the generics indicate this? Any idea how that could be done?

I could also just try and use Rc<RefCell<dyn Scorer>> instead of Box<dyn Scorer>. I have no idea about the runtime cost of the reference counting, but I don't expect this to be really noticeable.

Once this works for a single PhraseQuery the next goal is the example query:
 +lowfreqterm +"highfreqterm1 highfreqterm2"
The approximating DocSet for that is represented by this intersection:
 +lowfreqterm +highfreqterm1 +highfreqterm2
so the boolean scorer implementing this will need this intersection as approximation,
and it will have the same matches() as the phrase query.

Btw. I had a look at boolean_weight.rs to see how to create intersections and unions.
The excluded scorers there use a generic TScoreCombiner. As the excluded scores are
always ignored, wouldn't it be better to always use a DoNothingCombiner there?
When so, I'll gladly open a PR.

ghost:
The branch as of yesterday, b49dddf, shows a design decision that will be hard to change:
Implement two phase scoring by having Scorer.two_phase() return an Option(TwoPhase) object.
When this does not return None, the Scorer itself becomes an approximation, and the TwoPhase object has a matches() method that indicates whether the approximation is an actual match.

The main reason for this is that there is no check in Rust to determine whether a Scorer also implements TwoPhase, so a separate object is needed. If Rust can do this somehow, please let me know.

One consequence is that the TwoPhase object needs  reference to its Scorer, and since there is already a reference to that Scorer, the reference will have to be shared by an Rc < RefCell< ... > >..
Some code showing this that for PhraseScorer is in test_phrase_count() in phrase_weight.rs on the above branch. This test passes, but other tests still fail.

ghost:
Restarted from master of 5 June 2020, and now I am stuck in an attempt to change the return type of weight.scorer from Box < dyn Scorer > to Rc < RefCell < dyn Scorer>>.
I marked the commit twophase-20200607.

The problem for this is that the downcast:: < TermScorer> method is not available on dyn Scorer, even though Scorer extends downcast_rs::DownCast. 
Normally in such cases the compiler adds a suggestion to use a trait implementing the method, in this case I would expect that trait to be downcast_rs::DownCast. There is no such suggestion, and adding use downcast_rs::DownCast; results in an error message that the trait is not used.

Any idea on how to get this working?

When I disable all the code using the downcast to TermScorer, the tests pass, except for test_boolean_termonly_intersection, as expected, but I prefer to keep the TermScorer specialisations.
The disabling of the TermScorer downcast is on the twophase branch of today, two commits after twophase-20200607.

PR #825, opened 2020-05-15
Minor doc and test improvements around fuzzy querying

robyoung:
There were a couple of things that confused me when reading up on fuzzy prefix querying.


fulmicoton:
Thanks!

PR #826, opened 2020-05-15
Add offset to TopDocsCollector

### LINKED ISSUE #822: introduce an offset to the topdocs collector to help pagination
A lot of users a confused on how to do pagination

robyoung:
Add an offset to `TopDocsCollector` and `TopDocs` to make it clearer how to handle pagination.

I feel like I may be missing something with this though as it seemed a bit too easy.

Closes #822

### DIFF
--- a/src/collector/top_collector.rs
+++ b/src/collector/top_collector.rs
@@ -70,23 +70,23 @@ where
     /// # Panics
     /// The method panics if limit is 0
     pub fn with_limit(limit: usize) -> TopCollector<T> {
-        Self::with_limit_and_offset(limit, 0)
-    }
-
-    /// Creates a top collector, with a number of documents equal to "limit" and
-    /// skipping the first "offset" documents.
-    ///
-    /// # Panics
-    /// The method panics if limit is 0
-    pub fn with_limit_and_offset(limit: usize, offset: usize) -> TopCollector<T> {
         if limit < 1 {
             panic!("Limit must be strictly greater than 0.");
         }
-        TopCollector {
+        Self {
             limit,
-            offset,
+            offset: 0,
             _marker: PhantomData,
         }
+    }
+
+    /// Skip the first "offset" documents when collecting.
+    ///
+    /// This is equivalent to `OFFSET` in MySQL or PostgreSQL and `start` in
+    /// Lucene's TopDocsCollector.
+    pub fn and_offset(mut self, offset: usize) -> TopCollector<T> {
+        self.offset = offset;
+        self
     }
 
     pub fn limit(&self) -> usize {
@@ -129,7 +129,10 @@ where
         segment_id: SegmentLocalId,
         _: &SegmentReader,
     ) -> crate::Result<TopSegmentCollector<F>> {
-        Ok(TopSegmentCollector::new(segment_id, self.limit + self.offset))
+        Ok(TopSegmentCollector::new(
+            segment_id,
+            self.limit + self.offset,
+        ))
     }
 }
 
@@ -203,7 +206,7 @@ impl<T: PartialOrd + Clone> TopSegmentCollector<T> {
 
 #[cfg(test)]
 mod tests {
-    use super::TopSegmentCollector;
+    use super::{TopCollector, TopSegmentCollector};
     use crate::DocAddress;
 
     #[test]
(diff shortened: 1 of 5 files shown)


src/collector/top_score_collector.rs
```diff
@@ -57,7 +57,7 @@ pub struct TopDocs(TopCollector<Score>);
 
 impl fmt::Debug for TopDocs {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "TopDocs({})", self.0.limit())
+        write!(f, "TopDocs({}, {})", self.0.limit(), self.0.offset())
```
fulmicoton: that will be confusing I think. Can we make it TopDocs(offset={},max_hits={}) or something like that?

src/collector/top_collector.rs
```diff
@@ -69,11 +70,21 @@ where
     /// # Panics
     /// The method panics if limit is 0
     pub fn with_limit(limit: usize) -> TopCollector<T> {
```
fulmicoton: Can we rename `limit` `max_hits`. Limit was not great to begin with... But it is even more confusing now that there is an offset.

src/collector/top_score_collector.rs
```diff
@@ -101,6 +101,43 @@ impl TopDocs {
...
+    /// let reader = index.reader().unwrap();
+    /// let searcher = reader.searcher();
+    ///
+    /// let query_parser = QueryParser::for_index(&index, vec![title]);
+    /// let query = query_parser.parse_query("diary").unwrap();
+    /// let top_docs = searcher.search(&query, &TopDocs::with_limit_and_offset(1, 1)).unwrap();
+    ///
+    /// assert_eq!(&top_docs[0], &(0.6099695, DocAddress(0, 3)));
+    /// ```
+    pub fn with_limit_and_offset(limit: usize, offset: usize) -> TopDocs {
```
fulmicoton: Maybe make it fluent pattern?

`TopDocs::with_max_hits(10).offset(40)` would give you hits from 40 to 50

src/collector/top_score_collector.rs
```diff
@@ -101,6 +101,43 @@ impl TopDocs {
...
+    /// index_writer.add_document(doc!(title => "A Dairy Cow"));
+    /// index_writer.add_document(doc!(title => "The Diary of a Young Girl"));
+    /// assert!(index_writer.commit().is_ok());
+    ///
+    /// let reader = index.reader().unwrap();
+    /// let searcher = reader.searcher();
+    ///
+    /// let query_parser = QueryParser::for_index(&index, vec![title]);
+    /// let query = query_parser.parse_query("diary").unwrap();
+    /// let top_docs = searcher.search(&query, &TopDocs::with_limit_and_offset(1, 1)).unwrap();
```
fulmicoton: Rust test are cool, but I would like a "normal unit test somewhere.".

In my experience, they compile faster, so I tend to run `cargo test --lib`.

Also, we need to cover more cases.
In particular, this test is pretty dangerous because both argument are the same (1 and 1) and we have positional argument.

The test would not catch if we interverted the two.

robyoung:
Thanks for the review. Your comments all sound great. I'll look at addressing them later today.

src/collector/top_collector.rs
robyoung: I'm assuming I should leave the `with_limit` and `limit` functions but just mark them deprecated. Does that sound ok?

robyoung: I've had a bit more of a think about this one I'm unsure about renaming this. I feel like `limit` and `offset` would be more universally recognisable to people and the semantics do match what they would expect. While it is not part of ISO SQL it is the naming used in the most common relational databases (see: [FETCH FIRST SQL clause](https://en.wikipedia.org/wiki/Select_(SQL)#FETCH_FIRST_clause)). Mixing limit and offset wording with something else may cause more confusion with people assuming the semantics of `max_hits` must be different from `limit` because `limit` was not used.

I've done a little investigating in other IR libraries and there is a little less consistency.
- In Lucene their [`TopDocsCollector`](https://lucene.apache.org/core/2_9_4/api/core/org/apache/lucene/search/TopDocsCollector.html) uses `start` and `howMany` (but also `numHits` in the `TopScoreDocCollector`).
- In Elasticsearch they use [`from` and `size`](https://www.elastic.co/guide/en/elasticsearch/reference/5.1/search-request-from-size.html#search-request-from-size).
- In Xapian they use [`pagesize` and `offset`](https://getting-started-with-xapian.readthedocs.io/en/latest/practical_example/searching/building.html) but I think that works better for them because of how their API works.

If you're definitely against using `limit` can I suggest we also use a different word for `offset`? Maybe `start` or `starting_from`.

src/collector/top_score_collector.rs
robyoung: I've used `and_offset` so that `offset` can still be used as an accessor.

src/collector/top_score_collector.rs
robyoung: Done

robyoung:
I have rebased on master and made some changes as a result.

fulmicoton:
Awesome thanks @robyoung!

fulmicoton:
Edited the CHANGELOG file accordingly, with your username.

PR #840, opened 2020-06-07
Boolean query: do not combine excluded scores.

ghost:


src/query/boolean_query/mod.rs
```diff
@@ -177,6 +177,40 @@ mod tests {
...
+                .searcher()
+                .search(boolean_query, &TEST_COLLECTOR_WITH_SCORE)
+                .unwrap()
+                .docs()
+                .iter()
+                .cloned()
+                .map(|doc| doc.1)
+                .collect::<Vec<DocId>>()
+        };
+        let boolean_query = BooleanQuery::from(vec![
```
fulmicoton: this test is not testing for this change does it?

Can you add an assert on the score?

ghost: Testing for the change is somewhat hard because it is testing for doing nothing, and indeed this test does not do that. I did verify that the DoNothing... score combiner is used by adding a println!("...") in its update() method and see that output for the test, but I left that println out of the PR.

I updated the test to compare the score value of a matching doc for an excluding query and a non excluding query.

fulmicoton: oh yes that makes sense. Thanks!

PR #844, opened 2020-06-30
Fix incorrect SimpleTokenizer link in documentation

sstangl:
Small patch to fix a warning when running `cargo doc`:

```
warning: `[tantivy::tokenizer::SimpleTokenizer]` cannot be resolved, ignoring it.
   --> src/query/query_parser/query_parser.rs:116:37
    |
116 | ///   tantivy's [`SimpleTokenizer`](tantivy::tokenizer::SimpleTokenizer), hence
    |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot be resolved, ignoring
    |
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`
```

Succeeds and links properly after the patch is applied with no warnings remaining.


fulmicoton:
thx!

PR #853, opened 2020-07-17
Assert nearly equals macro

fulmicoton:


drusellers:
TIL a decent way to unit test floats.

fulmicoton:
I prefer the macro to the non macro version because when it fails the failing file / line are more obvious.

PR #856, opened 2020-07-28
Block wand

fulmicoton:


ppodolsky:
Hi, @fulmicoton 
Do you need some testing? I could reindex my corpus and look at performance. If you are OK with it, just let me know when this PR would be somehow ready, I will easily tolerate minor bugs.

fulmicoton:
@ppodolsky That would be great. The PR is more or less ready. I'm running tests myself but this is a very quirky feature and the more testing the better.

Stuff you probably know. This only kicks in for union queries with a simple TopScore collector. (no count)

Stuff you probably do NOT know... 

Each segments blockwand information approximates the average fieldnorm on itself. This means that in theory it is possible to observe minor errors (a document that should have barely made it in the top K not making it in the top K) if you have more than one segment. If you have only one segment the results should always be exact.

Due to the non-associativity of float additions, the score computed by blockwand and the score computed by a normal union can differ a tiny bit. This means that in case of an almost tie (score are different but only due to the order of the TermScorer's score addition) the results can be different.

ppodolsky:
Okey, I'm gonna to test it then.
> This only kicks in for union queries with a simple TopScore collector.

BTW, is it possible to adapt blockwand for tweaked score collector? With tweaking function `f(original_score)` having range limited to [0.0; original_score] probably, if I correctly remind what is blockwand.

fulmicoton:
> BTW, is it possible to adapt blockwand for tweaked score collector? With tweaking function f(original_score) having range limited to [0.0; original_score] probably, if I correctly remind what is blockwand.


@ppodolsky Yes it should work ok. You just need the upperbound to be upperbounds... To get it to work with your code you will have to have your own customer collector though.

PR #860, opened 2020-08-11
Entirely removes usage of failure and replace it with thiserror

### LINKED ISSUE #760: Improve integration of errors with `std::error::Error`
Currently, Tantivy's error types implement `failure::Fail` but not `std::error::Error`.  
This makes them not directly usable with the `?` operator inside a function returning `Result<_, Box<dyn std::error::Error>>` or other catch-all-errors types like `anyhow::Error`.  

A solution for solving this is to directly implement `std::error::Error` for Tantivy's error types.  
We could these impls manually but there is a library called `thiserror` which provides a derive macro for generating an impls for both `std::error::Error` and `std::fmt::Display`, in a same manner as with `failure` (`#[error("error msg...")]` instead of `#[fail(display = "error msg...")]`).  
PR #756 demonstrate what the migration would look like.

Staying with `failure` is still an option, but forces the user to use `failure::Error` (and take a dependence on it) to collect Tantivy's errors as a `std::error::Error`.

saresend:
## What this PR does
This aims to fix #760 by removing the dependence on the now deprecated `failure` crate, and replacing it with `thiserror` error implementations instead. As an added bonus, this means that `TantivyErrors` now implement `std::error::Error`, meaning that downstream users aren't locked into using the `failure::Fail` trait, and can now use any error handling they like. 

## Test Plan 
* `./run-test.sh`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,13 +43,13 @@ bitpacking = {version="0.8", default-features = false, features=["bitpacker4x"]}
 census = "0.4"
 fnv = "1.0.6"
 owned-read = "0.4"
-failure = "0.1"
 htmlescape = "0.3.1"
 fail = "0.4"
 murmurhash32 = "0.2"
 chrono = "0.4"
 smallvec = "1.0"
 rayon = "1"
+thiserror = "1.0.20"
 
 [target.'cfg(windows)'.dependencies]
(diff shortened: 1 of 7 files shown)


saresend:
Realize this a duplicate of #756, closing

PR #866, opened 2020-08-21
Removing dependency to atomicwrites

fulmicoton:


raultang:
@fulmicoton @asafigan 

With this change, it is unable to create index if the data directory with OS system is in separated disk.
```
tempfile.into_temp_path().persist(full_path)?;
```
This line trying to create a hard link to `/tmp/` directory (`/tmp` directory is in system disk)
Error message:
PathPersistError { error: Os { code: 18, kind: Other, message: "Invalid cross-device link" }, path: "/tmp/.tmp6JJxbQ" }

fulmicoton:
@raultang Thanks for the report. Will revert soon.

asafigan:
I know that future versions of `atomicwrite` will use `tempfile` under the hood. So it could become an issue again in the future. We might want to figure out if `tempfile` supports this and how to work around it. Might still be better to use `atomicwrite` and fix any issues there rather than implementing it here.

asafigan:
I just read `tempfile`'s docs. If we use a builder to create the temporary file, we can specify the directory. I think this would solve the issue because we could make sure that the file is created on the correct disc and then there would be no link error.

asafigan:
Here is a link to the method in the documentation. https://docs.rs/tempfile/3.1.0/tempfile/struct.Builder.html#method.tempfile_in

PR #877, opened 2020-09-02
Support for indexed bytes fast field

raultang:
I am trying to implement string fast field, which equivalent to elasticsearch [keyword type](https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html),
This is probably also relate to https://github.com/tantivy-search/tantivy/issues/832

Base on my current understanding, bytes fast filed is similar to int fast filed, so not much special need to handle.
This change only include `SingleValue` which meet my needs right now, `MultiValues` is more complicated and need more time.


fulmicoton:
I am not sure this is the way elasticsearch. I am not sure, but aren't they encoding the term ordinal in the docvalue?

raultang:
> I am not sure this is the way elasticsearch. I am not sure, but aren't they encoding the term ordinal in the docvalue?

My understanding it also store the term direct in column store, later on doing aggregation etc can avoid ordinal lookup costs.

Anyway, this pull request is more relate to make the bytes fast field indexed, then it is searchable, string is just an example.

fulmicoton:
Sorry for the late answer, i'll start reviewing your code.

fulmicoton:
The PR will require more work. I pulled it as a feature branch in tantivy and invited you as a developer so we can iterate on this.

raultang:
Sure, that would be great.

fulmicoton:
@raultang Have you seen my email?

raultang:
@fulmicoton 
Sorry, just saw the email. And see detail in the email.

fulmicoton:
This has been reworked in a different PR.

PR #885, opened 2020-09-22
[tantivy] Add brotli codec for row storage

ppodolsky:
Hi, @fulmicoton 

What do you think about it? Until recently I had a Postgres golden copy of all my data that I had been indexing with Tantivy. Tantivy was used only for maintaining inverted index, without any `STORED` fields. Now I've faced with a very hostile environment of limited hard storage, so I would like to try to store all documents in Tantivy `store` and to compress them a little bit stronger than snappy/lz4.

There is a [great benchmark](https://quixdb.github.io/squash-benchmark/) I was looking at while considering different codecs. It looks that brotli could take a place of the slow and well-compressed codec in addition to fast and fat lz4.

### DIFF
--- a/src/store/compression_brotli.rs
+++ b/src/store/compression_brotli.rs
@@ -10,12 +10,12 @@ pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()>
     params.quality = 5;
 
     compressed.clear();
-    brotli::BrotliCompress(&mut Cursor::new(uncompressed), compressed, &params)?;
+    brotli::BrotliCompress(uncompressed, compressed, &params)?;
     Ok(())
 }
 
 pub fn decompress(compressed: &[u8], decompressed: &mut Vec<u8>) -> io::Result<()> {
     decompressed.clear();
-    brotli::BrotliDecompress(&mut Cursor::new(compressed), decompressed)?;
+    brotli::BrotliDecompress(compressed, decompressed)?;
     Ok(())
 }
--- a/src/store/compression_brotli.rs
+++ b/src/store/compression_brotli.rs
@@ -1,0 +1,21 @@
+use std::io::{self, Cursor};
+
+/// Name of the compression scheme used in the doc store.
+///
+/// This name is appended to the version string of tantivy.
+pub const COMPRESSION: &'static str = "brotli";
+
+pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
+    let mut params = brotli::enc::BrotliEncoderParams::default();
+    params.quality = 5;
+
+    compressed.clear();
+    brotli::BrotliCompress(&mut Cursor::new(uncompressed), compressed, &params)?;
+    Ok(())
+}
+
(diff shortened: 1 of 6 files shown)


fulmicoton:
I think this is a good change. Did you test it on your own data to check it was helping?

Let me have a longer look at this PR on Monday. Right now I am in vacation.

ppodolsky:
I did some measurments, but they are still not finalized.

All tests have been done with all fields are just `STORED`, without indexing. There were 100 000 documents. In the end I have called `wait_merging_threads`. Opt level 3.

```
Brotli 1:  31.51Mb 1.6  secs
Brotli 5:  22.07Mb 3.4  secs
Brotli 7:  22.03Mb 6.4  secs
Brotli 9:  22.00Mb 16.0 secs
Brotli 11: 18.25Mb 35.0 secs
Snappy:    33.70Mb 1.6  secs
```

The same data packed in protobuf, compressed in chunks of size 100 documents by brotli 5 took 21 Mb. Not sure what is the reason, probably this difference is due to presence of skip lists and field type in Tantivy binary format (#887). Or due to slightly different chunkization of data as I split documents by their numbers, not sizes.

I haven't measured decompression yet, but it is told that it was faster than snappy.

src/store/compression_brotli.rs
```diff
@@ -0,0 +1,21 @@
...
+///
+/// This name is appended to the version string of tantivy.
+pub const COMPRESSION: &'static str = "brotli";
+
+pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
+    let mut params = brotli::enc::BrotliEncoderParams::default();
+    params.quality = 5;
+
+    compressed.clear();
+    brotli::BrotliCompress(&mut Cursor::new(uncompressed), compressed, &params)?;
```
fulmicoton: probably not need for cursor here. see my comment in the other PR.

src/store/mod.rs
```diff
@@ -39,18 +39,28 @@ mod writer;
 pub use self::reader::StoreReader;
 pub use self::writer::StoreWriter;
 
+#[cfg(all(feature = "lz4", feature = "brotli"))]
+compile_error!("feature `lz4` or `brotli` shouldn't be enabled both.");
```
fulmicoton: ... "must not be enabled together."

- I am not a native speaker, but I think together is slightly more correct english -unsure I make a load of grammar mistakes-  
- rfc2119 describes what should be our definition of MAY, SHOULD, MUST, SHOULD NOT, MUST NOT in documetnation

fulmicoton (changes requested):
We probably need to make the compression dynamic. For one thing, it will make it easier to unit test >.<

No need to do anything in this PR however.

ppodolsky:
If you wish you can create an issue and assign me, I will try to find a time to look at dynamic codec selection as I need it too and fix other notes on the way.

fulmicoton:
@ppodolsky Can you address the couple of comments in the CR?

fulmicoton (changes requested):
Can you remove the use of Cursor and fix the comment?

ppodolsky:
Done, fixed some other comments in code on the way.

PR #899, opened 2020-10-02
[tantivy] Faster traversing of the whole segment store

ppodolsky:
@fulmicoton 
Another small proposal for more comfortable iteration over the whole .store file. It's non-intrusive change and I tried to be compatible with the code style, hope I succeded.

PR is implementing iterator interface for StoreReader, it allowes to iterate over all documents without looking at skiplists and potentially faster.

This is a self-sufficient feature. But in the future I would like to implement possibilty to enable/disable indexing dynamically, at least for stored fields and this feature is a required precondition.

### DIFF
--- a/src/store/reader.rs
+++ b/src/store/reader.rs
@@ -48,18 +48,18 @@ impl StoreReader {
         self.data.as_slice()
     }
 
-    fn compressed_block(&self, addr: usize) -> &[u8] {
+    fn compressed_block(&self, addr: usize) -> (usize, &[u8]) {
         let total_buffer = self.data.as_slice();
         let mut buffer = &total_buffer[addr..];
         let block_len = u32::deserialize(&mut buffer).expect("") as usize;
-        &buffer[..block_len]
+        (std::mem::size_of::<u32>() + block_len, &buffer[..block_len])
     }
 
     fn read_block(&self, block_offset: usize) -> io::Result<()> {
         if block_offset != *self.current_block_offset.borrow() {
             let mut current_block_mut = self.current_block.borrow_mut();
             current_block_mut.clear();
-            let compressed_block = self.compressed_block(block_offset);
+            let (_, compressed_block) = self.compressed_block(block_offset);
             decompress(compressed_block, &mut current_block_mut)?;
             *self.current_block_offset.borrow_mut() = block_offset;
         }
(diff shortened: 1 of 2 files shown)


src/store/reader.rs
```diff
@@ -93,6 +93,70 @@ impl StoreReader {
...
 
+pub struct StoreReaderIterator<'a> {
+    block_offset: usize,
+    document_offset: usize,
+    store_reader: &'a StoreReader,
+    uncompressed_block: Vec<u8>,
+}
+
+impl<'a> StoreReaderIterator<'a> {
+    pub fn has_more_documents(&self) -> bool {
```
fulmicoton, line 104: rename `has_more_documents_in_current_block`

fulmicoton, line 104: Should be private

src/store/reader.rs
```diff
@@ -93,6 +93,70 @@ impl StoreReader {
...
+        self.document_offset < self.uncompressed_block.len()
+    }
+
+    pub fn has_more_blocks(&self) -> bool {
```
fulmicoton, line 108: should be private

src/store/reader.rs
```diff
@@ -93,6 +93,70 @@ impl StoreReader {
...
+        decompress(&compressed_block, &mut self.uncompressed_block)?;
+
+        self.block_offset += block_size;
+        self.document_offset = 0;
+        Ok(())
+    }
+}
+
+impl<'a> IntoIterator for &'a StoreReader {
+    type Item = Document;
```
fulmicoton, line 125: I am working on a different PR to be more friendly with `io::Read` based Directory. Panicking on data corruption is a choice tantivy makes, but panicking on an `io::Error` is not something we can do (sorry you could not know that).

For this reason, `Iterator<Item=Document>` is not the right API for this reason. 
`Iterator<Item=crate::Result<Document>>` could be ok. (something similar to the `LineIterator` in the standard library).

Do you have an actual usage for this? Would `Iterator<Item=crate::Result<(DocId, Document)>>` be more useful?

src/store/reader.rs
```diff
@@ -93,6 +93,70 @@ impl StoreReader {
...
+
+impl<'a> Iterator for StoreReaderIterator<'a> {
+    type Item = Document;
+
+    fn next(&mut self) -> Option<Document> {
+        if !self.has_more_documents() {
+            if !self.has_more_blocks() {
+                return None;
+            }
+            self.read_block().unwrap();
```
fulmicoton, line 150: Return the result, once Item has been changed.

src/store/reader.rs
```diff
@@ -93,6 +93,70 @@ impl StoreReader {
...
+        }
+        let mut cursor = io::Cursor::new(&self.uncompressed_block[self.document_offset..]);
```
fulmicoton, line 152: I was a rust rookie when I started tantivy. I suspect you copied this code from somewhere else.
io::Cursor is not useful here since `&[u8]` implements read.

src/store/reader.rs
```diff
@@ -93,6 +93,70 @@ impl StoreReader {
...
+        self.uncompressed_block.clear();
+        decompress(&compressed_block, &mut self.uncompressed_block)?;
+
+        self.block_offset += block_size;
+        self.document_offset = 0;
+        Ok(())
+    }
+}
+
+impl<'a> IntoIterator for &'a StoreReader {
```
fulmicoton, line 124: This Iterator does not take in account deletes.

If you have a need for this, we can keep it, but let's make it explicit. In that case let's remove the `IntoIterator` implementation, and add a method `iter_docs_including_deleted()` with some documentation explaining this behavior.

We can also have an `iter_docs(&DeleteBitset)` but not that the block decompression would have to be made lazy if we want to avoid decompressing a block for which all docs have been deleted in that case.

fulmicoton (changes requested):
General comment: Is this something you need? Can you describe your actual use case?

Please open feature request so that we can discuss features before sending a PR.
One undiscussed feature of tantivy is that the codebase is rather small, making it possible for user to "adopt it" and rapidly own the codebase.
It means I need to be selective when it comes to features that are added or not.

ppodolsky:
Got it, I will create issues. For me it is easier sometimes to discuss (small) features having code, especially when I'm not familiar with full sources. Implementing helps to understand how mad your proposal is. 

As you see, I'm rookie too. Moreover, I've started to learn Rust months ago because of Tantivy. I needed a small inverted index and was chosing between Lucene, [Trinity](https://github.com/phaistos-networks/Trinity) and accidently found Tantivy. Maintainability and possibilty to understand sources bought me :)
Thank you for the review, I'll fix it now.

ppodolsky:
Let me explain the whole picture to make my goals clearer. It is an interesting and special use case.

There is a relatively large and open corpus of textual data. Every row in the corpus has a key and a global timestamp - time of modification.
1. Some users want to download the entire corpus and iterate over records to insert them into their own databases.
2. Some users want to download it and use as a key-value storage.
3. Some users want to download it and do a full-text search over text.
4. There must be an effective leader-follower replication. It means that
5. There must be a way to produce and apply differential updates for data
6.  It must be a space and time effective solution.

My current design is to pack this corpus into Tantivy .store format and provide it together with the software. Software includes a server with embedded Tantivy and scripts. To satisfy 2 and 3 I would like to give users a possibilty to enable indexing when they have .store files.
1 can be done by using iteration implemented in this PR.
4 and partically 5 are implemented by two ways - effective one and no so effective one. Effective one is to download prepared .store files consisting of new and modified rows and putting them into index. Not effective but more widespreaded way is to ask another server to return all rows starting from the certain time.

So, the plot is following - a single (actually not single, but it is not important) master search server and a trusted federation of replicas that could sync from master, from each other and can be started up by downloading .store and enabling indexing. Other users are roaming around and leaching .store files to work with them on their own or to reseed.

src/store/reader.rs
ppodolsky: Imo the method must take deletion into account for the logical consistency. I'm totally missed this point. So, should I leave the iterator interface and add there DeleteBitset checking?

fulmicoton:
@ppodolsky a similar feature was merged by @PSeitz

PR #902, opened 2020-10-04
[tantivy] Fix for schema deserialization error

ppodolsky:
There is a [feature in serde_yaml](https://github.com/dtolnay/serde-yaml/issues/94) that it cannot deal with borrowed strings. It makes impossible to deserialize yaml to Schema due to error `invalid type: string \"field_name\", expected a borrowed string`
This extra allocation is a tol-tol solution, but it's a far more easier way then rewriting whole serde_yaml and I hope deserializing schema is not a hot path.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,7 @@ rand = "0.7"
 maplit = "1"
 matches = "0.1.8"
 proptest = "0.10"
+serde_yaml = "0.8.13"
 
 [dev-dependencies.fail]
 version = "0.4"
--- a/src/schema/field_entry.rs
+++ b/src/schema/field_entry.rs
@@ -221,8 +221,8 @@ impl<'de> Deserialize<'de> for FieldEntry {
                             if ty.is_some() {
                                 return Err(de::Error::duplicate_field("type"));
                             }
-                            let type_string = map.next_value()?;
-                            match type_string {
+                            let type_string = map.next_value::<String>()?;
+                            match type_string.as_str() {
                                 "hierarchical_facet" => {
                                     field_type = Some(FieldType::HierarchicalFacet);
                                 }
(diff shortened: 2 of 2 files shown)


fulmicoton:
I don't understand anything about this bug.

Tantivy does not use Yaml. If you really need yaml serialization that's a feature request you want and you will need to explain why this is useful to you.

ppodolsky:
There was a one line test in the PR that did not work:
`serde_yaml::from_str::<FieldEntry>(field_content).unwrap();`
or similar
`serde_yaml::from_str::<Schema>(schema_content).unwrap();`

>  If you really need yaml serialization that's a feature request

You already have yaml serialization due to using `serde` crate :) There is just a small trouble with deserialization.

ppodolsky:
About the reasons - we are storing schemas in files, not in the code. Yaml is much easier for human to read and edit:
```
[{
  "name": "title",
  "type": "text",
  "options": {
    "indexing": {
      "record": "position",
      "tokenizer": "default"
    },
    "stored": false
  }
}, {
  "name": "body",
  "type": "text",
  "options": {
    "indexing": {
      "record": "position",
      "tokenizer": "default"
    },
    "stored": false
}]
```
vs
```
- name: text
  type: text
  options:
    indexing:
      record: position
      tokenizer: default
    stored: true
- name: body
  type: text
  options:
    indexing:
      record: position
      tokenizer: default
    stored: true
```
You can easily be mistaken with these curvy brackets while editing file manually

fulmicoton:
I see... And I had missed the fact that the dependency was just a dev-dependency for the unit test! I think that is ok then.
I may or may not ditch the unit test in the future when I get annoyed by the extra compilation time.

ppodolsky:
> I may or may not ditch the unit test in the future when I get annoyed by the extra compilation time.

Sure, the test is mostly for illustrating the issue.

fulmicoton:
Yes it is great!

PR #924, opened 2020-11-06
Implement FileWatcher

fulmicoton:


src/directory/file_watcher.rs
```diff
@@ -0,0 +1,136 @@
...
+        let state = self.state.clone();
+
+        thread::Builder::new()
+            .name("thread-tantivy-meta-file-watcher".to_string())
+            .spawn(move || {
+                let mut current_checksum = None;
+
+                while state.load(Ordering::Relaxed) == 1 {
+                    if let Ok(checksum) = FileWatcher::compute_checksum(&path) {
+                        if current_checksum.is_none() || current_checksum.unwrap() != checksum {
```
fulmicoton: i prefer `if let Some(current_checksum)= current_checksum_opt` to make the it visible that this code cannot panic.

src/directory/file_watcher.rs
```diff
@@ -0,0 +1,136 @@
...
+                            info!("Meta file {:?} was modified", path);
+                            current_checksum = Some(checksum);
+                            futures::executor::block_on(callbacks.broadcast());
+                        }
+                    }
+
+                    thread::sleep(if cfg!(test) {
+                        Duration::from_millis(1)
+                    } else {
+                        DEFAULT_POLLING_INTERVAL
```
fulmicoton: should we change the default value in test instead?

fulmicoton: the value of `DEFAULT_POLLING_INTERVAL` i mean

PR #926, opened 2020-11-10
Modified `Directory::exists` API to return `Result<bool, OpenReadError>`

guilload:
`OpenReadError` does not implement `PartialEq` so I added a bunch of ugly `unwrap` calls in the unit tests.

### DIFF
--- a/src/directory/ram_directory.rs
+++ b/src/directory/ram_directory.rs
@@ -177,8 +177,15 @@ impl Directory for RAMDirectory {
         self.fs.write().unwrap().delete(path)
     }
 
-    fn exists(&self, path: &Path) -> bool {
-        self.fs.read().unwrap().exists(path)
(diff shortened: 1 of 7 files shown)


src/directory/ram_directory.rs
```diff
@@ -177,8 +177,15 @@ impl Directory for RAMDirectory {
...
     }
 
-    fn exists(&self, path: &Path) -> bool {
-        self.fs.read().unwrap().exists(path)
+    fn exists(&self, path: &Path) -> Result<bool, OpenReadError> {
+        Ok(self
+            .fs
+            .read()
+            .map_err(|e| OpenReadError::IOError {
+                io_error: io::Error::new(io::ErrorKind::Other, e.to_string()),
```
guilload, line 185: Is there a better way of wrapping `e`?

PR #932, opened 2020-11-12
Fixing unit test.

fulmicoton:


src/directory/file_watcher.rs
```diff
@@ -104,35 +108,37 @@ mod tests {
         let (tx, rx) = crossbeam::channel::unbounded();
         let timeout = Duration::from_millis(100);
 
-        {
```
guilload, line 107: This is no longer testing that the background thread stops when the watcher is dropped.

src/directory/file_watcher.rs
```diff
@@ -104,35 +108,37 @@ mod tests {
...
+        assert!(rx.recv_timeout(timeout).is_err());
 
-            fs::write(&tmp_file, b"bar").unwrap();
-            assert_eq!(rx.recv_timeout(timeout), Ok(2));
-        }
+        atomic_write(&tmp_file, b"bar")?;
+        assert_eq!(rx.recv_timeout(timeout), Ok(2));
 
-        fs::write(&tmp_file, b"qux").unwrap();
+        mem::drop(watcher);
```
guilload, line 133: Now we're testing that when we drop the handle, the callback list becomes empty, and broadcast is no-op. I guess we should write two tests. I can work on that.

src/directory/file_watcher.rs
fulmicoton: very good point

guilload (approved):

PR #937, opened 2020-11-14
Cache store reader blocks in an LRU fashion

guilload:
Cache store reader blocks in an LRU fashion using the `lru` crate. I put a `Mutex` in front of the cache because `LruCache::get` is a mutable method. I introduced the `cache_hits` variable for testing purposes, but we can get rid of it. What's a reasonable default size for the cache capacity?  Approximately how many "lorem ipsum" documents need to be inserted in the store to span multiple blocks.

### DIFF
--- a/src/store/reader.rs
+++ b/src/store/reader.rs
@@ -6,18 +6,27 @@ use crate::directory::{FileSlice, OwnedBytes};
 use crate::schema::Document;
 use crate::space_usage::StoreSpaceUsage;
 use crate::DocId;
-use std::cell::RefCell;
+use lru::LruCache;
 use std::io;
 use std::mem::size_of;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex};
+
+const LRU_CACHE_CAPACITY: usize = 100;
+
+type Block = Arc<Vec<u8>>;
+
+type BlockCache = Arc<Mutex<LruCache<usize, Block>>>;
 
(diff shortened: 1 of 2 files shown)


src/store/reader.rs
```diff
@@ -6,17 +6,25 @@ use crate::directory::{FileSlice, OwnedBytes};
...
 use crate::space_usage::StoreSpaceUsage;
 use crate::DocId;
-use std::cell::RefCell;
+use lru::LruCache;
 use std::io;
 use std::mem::size_of;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex};
+
+static LRU_CACHE_CAPACITY: usize = 100;
```
fulmicoton: const?

src/store/reader.rs
```diff
@@ -107,3 +123,43 @@ fn split_file(data: FileSlice) -> io::Result<(FileSlice, FileSlice, DocId)> {
...
+mod tests {
+    use super::*;
+    use crate::{directory::RAMDirectory, store::tests::write_lorem_ipsum_store, Directory};
+    use std::path::Path;
+
+    #[test]
+    fn test_store_lru_cache() -> crate::Result<()> {
+        let directory = RAMDirectory::create();
+        let path = Path::new("store");
+        let store_wrt = directory.open_write(path)?;
```
fulmicoton: I think you can write into a Vec<u8>, and wrap it into an OwnedBytes...
I believe OwnedBytes impl FileHandle.

Issue #892, opened 2020-09-30
Add a FilterCollector wrapper


fulmicoton:
A `FilterCollector` could take a field (configured as a fast field), a predicate function over that field value, and another collector.

The filtercollector would then push docs to the underlying collector iff the fastvalue associated with the field matches the predicate.

fulmicoton:
@barrotsteindev would you be interested in picking this ticket?

barrotsteindev:
@fulmicoton Yes please

barrotsteindev:
Should we close this issue? @fulmicoton

PR #942, opened 2020-11-21
added initial implementation for filter_collector

barrotsteindev:
Initial implementation of #892

### DIFF
--- a/src/collector/filter_collector_wrapper.rs
+++ b/src/collector/filter_collector_wrapper.rs
@@ -41,7 +41,7 @@ use crate::{Score, SegmentReader, TantivyError};
 ///
 /// let query_parser = QueryParser::for_index(&index, vec![title]);
 /// let query = query_parser.parse_query("diary").unwrap();
-/// let no_filter_collector = FilterCollector::new(price, &|value| value > 20_120u64, TopDocs::with_limit(1));
+/// let no_filter_collector = FilterCollector::new(price, &|value| value > 20_120u64, TopDocs::with_limit(2));
 /// let top_docs = searcher.search(&query, &no_filter_collector).unwrap();
 ///
 /// assert_eq!(top_docs.len(), 1);
--- a/src/collector/filter_collector_wrapper.rs
+++ b/src/collector/filter_collector_wrapper.rs
@@ -9,8 +9,6 @@
 
 // ---
 // Importing tantivy...
-use std::marker::PhantomData;
-
 use crate::collector::{Collector, SegmentCollector};
 use crate::fastfield::FastFieldReader;
 use crate::schema::Field;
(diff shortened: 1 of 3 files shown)


src/collector/filter_collector_wrapper.rs
```diff
@@ -0,0 +1,154 @@
...
+/// assert_eq!(top_docs.len(), 2);
+/// assert_eq!(top_docs[0].1, DocAddress(0, 1));
+/// assert_eq!(top_docs[1].1, DocAddress(0, 3));
+///
+/// let filter_all_collector = FilterCollector::new(price, &|value| false, TopDocs::with_limit(2));
+/// let filtered_top_docs = searcher.search(&query, &filter_all_collector).unwrap();
+///
+/// assert_eq!(filtered_top_docs.len(), 0);
+/// ```
+pub struct FilterCollector<TCollector, TSegmentCollector> {
```
fulmicoton: I don't think TSegmentCollector and the PhantomData is necessary here. You can just refer to it as TCollector::Child everywhere.

src/collector/filter_collector_wrapper.rs
```diff
@@ -0,0 +1,154 @@
...
+    field: Field,
+    collector: TCollector,
+    predicate: &'static (dyn Fn(u64) -> bool + Send + Sync),
+    phantom: PhantomData<TSegmentCollector>,
+}
+
+impl<TCollector, TSegmentCollector> FilterCollector<TCollector, TSegmentCollector>
+where
+    TCollector: Collector<Child = TSegmentCollector> + Send + Sync,
+    TSegmentCollector: SegmentCollector + Send + Sync,
```
fulmicoton: `+ Send + Sync ` is not necessary for the TSegmentCollector I think

src/collector/filter_collector_wrapper.rs
```diff
@@ -0,0 +1,154 @@
...
+///
+/// let filter_all_collector = FilterCollector::new(price, &|value| false, TopDocs::with_limit(2));
+/// let filtered_top_docs = searcher.search(&query, &filter_all_collector).unwrap();
+///
+/// assert_eq!(filtered_top_docs.len(), 0);
+/// ```
+pub struct FilterCollector<TCollector, TSegmentCollector> {
+    field: Field,
+    collector: TCollector,
+    predicate: &'static (dyn Fn(u64) -> bool + Send + Sync),
```
fulmicoton: The predicate could be in the generics to avoid the static dispatch

src/collector/filter_collector_wrapper.rs
```diff
@@ -0,0 +1,154 @@
...
+/// index_writer.add_document(doc!(title => "A Dairy Cow", price => 21_240u64));
+/// index_writer.add_document(doc!(title => "The Diary of a Young Girl", price => 20_120u64));
+/// assert!(index_writer.commit().is_ok());
+///
+/// let reader = index.reader().unwrap();
+/// let searcher = reader.searcher();
+///
+/// let query_parser = QueryParser::for_index(&index, vec![title]);
+/// let query = query_parser.parse_query("diary").unwrap();
+/// let no_filter_collector = FilterCollector::new(price, &|value| true, TopDocs::with_limit(2));
```
fulmicoton: let's tweak the example to make it more compelling and put a limit on the price.

e.g. `value < 25_000`

src/collector/filter_collector_wrapper.rs
fulmicoton: I modified it directly.

src/collector/filter_collector_wrapper.rs
fulmicoton: You can pick the change here.
https://github.com/tantivy-search/tantivy/commit/4e93c681c0c4ac0b83fc85f4fee09bbe27fb15c8

src/collector/filter_collector_wrapper.rs
fulmicoton: https://github.com/tantivy-search/tantivy/commit/4e93c681c0c4ac0b83fc85f4fee09bbe27fb15c8

fulmicoton (changes requested):
Great job! 

I put a couple of comments.
Can you fix these and add a line in the CHANGELOG to log your change? And do not forget to add your username there.

src/collector/filter_collector_wrapper.rs
```diff
@@ -0,0 +1,154 @@
...
+    predicate: &'static (dyn Fn(u64) -> bool + Send + Sync),
+    phantom: PhantomData<TSegmentCollector>,
+}
+
+impl<TCollector, TSegmentCollector> FilterCollector<TCollector, TSegmentCollector>
+where
+    TCollector: Collector<Child = TSegmentCollector> + Send + Sync,
+    TSegmentCollector: SegmentCollector + Send + Sync,
+{
+    pub fn new(
```
fulmicoton, line 70: please some rustdoc here.

src/collector/filter_collector_wrapper.rs
barrotsteindev: @fulmicoton 
Do you mean that a `TPredicate` generic type should be added, or that the predicate should also be boxed?

The only reason I added the `'static` lifetime is because SegmentCollector uses it.
Do you think the predicates should be cloned for each SegmentCollector?
Or maybe an Arc would be better?
But this means we might pay a small performance hit because the SegmentCollector would have to dereference the predicate for each doc.

barrotsteindev: I implemented it for now as a reference, except now the reference is of type `&'static TPredicate`.

barrotsteindev:
Rebased on `main`

fulmicoton:
LGTM

PR #944, opened 2020-11-25
No filelen problem.

fulmicoton:


src/directory/file_slice.rs
```diff
@@ -57,15 +57,18 @@ pub struct FileSlice {
...
+        let num_bytes = file_handle.len();
+        FileSlice::new_with_num_bytes(file_handle, num_bytes)
+    }
+
+    /// Wraps a FileHandle.
+
+    pub fn new_with_num_bytes(file_handle: Box<dyn FileHandle>, num_bytes: usize) -> Self {
         FileSlice {
-            data: Arc::new(Box::new(data)),
+            data: Arc::new(file_handle),
```
guilload: I don't know if that buys us anything but we could define `FileSlice` as:
```rust
pub struct FileSlice {
    data: Arc<dyn FileHandle>,
...
```

and write:
```rust
data: Arc::from(file_handle),
```
here

fulmicoton: excellent point!

guilload (approved):

PR #945, opened 2020-11-25
Replace some `Arc<Box<dyn...` with `Arc<dyn...`

guilload:


src/directory/file_slice.rs
```diff
@@ -2,10 +2,11 @@ use stable_deref_trait::StableDeref;
 
 use crate::common::HasLen;
 use crate::directory::OwnedBytes;
-use std::sync::Arc;
+use std::sync::{Arc, Weak};
 use std::{io, ops::Deref};
 
-pub type BoxedData = Box<dyn Deref<Target = [u8]> + Send + Sync + 'static>;
+pub type BoxedData = Arc<dyn Deref<Target = [u8]> + Send + Sync + 'static>;
```
guilload: Let me know if you want to change the names of `BoxedData` and `WeakBoxedData` for something else.

fulmicoton: `ArcBytes` and `WeakArcBytes` maybe?

src/indexer/delete_queue.rs
```diff
@@ -53,7 +53,7 @@ impl DeleteQueue {
             return block;
         }
         let block = Arc::new(Block {
-            operations: Arc::default(),
+            operations: Arc::new([DeleteOperation::default(); 0]),
```
fulmicoton: Can you change it to `[]` ?
I like that your version is explicity about the content of the array, but I have no clue on whether it DeleteOption::default() is effectively called or not at all.

PR #948, opened 2020-11-30
Fix perf regression in the benchmark for the Count collector.

fulmicoton:
In order to reduce IO, we introduced a way to instanciate a dummy
constant FieldnormReader which worked by allocating a buffer with
as many bytes as there are docs in the segments.

This allocation is not a negligible by any mean.

This PR works by offering two implementation for the
FieldnormReader.
The const field norm reader simply returns the same value all of the
time, while the array based one does the same as the current one.

### DIFF
--- a/src/fieldnorm/reader.rs
+++ b/src/fieldnorm/reader.rs
@@ -61,16 +61,38 @@ impl FieldNormReaders {
 /// precompute computationally expensive functions of the fieldnorm
 /// in a very short array.
 #[derive(Clone)]
-pub struct FieldNormReader {
-    data: OwnedBytes,
+pub struct FieldNormReader(ReaderImplEnum);
+
+impl From<ReaderImplEnum> for FieldNormReader {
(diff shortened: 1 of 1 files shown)


src/fieldnorm/reader.rs
```diff
@@ -61,16 +61,38 @@ impl FieldNormReaders {
 /// precompute computationally expensive functions of the fieldnorm
 /// in a very short array.
 #[derive(Clone)]
-pub struct FieldNormReader {
-    data: OwnedBytes,
+pub struct FieldNormReader(ImplEnum);
```
guilload: nit suggestion: `InnerImpl` or `ReaderImpl`?

guilload (approved):
LGTM :+1:

src/fieldnorm/reader.rs
fulmicoton: renamed

PR #950, opened 2020-12-01
Fix clippy warning

guilload:
Shall I add clippy to CI?


fulmicoton:
Yeah that sounds like a good idea.

PR #952, opened 2020-12-09
Encode blockwand on a single byte.

fulmicoton:


src/postings/skip.rs
```diff
@@ -1,32 +1,50 @@
...
+fn decode_block_wand_max_tf(max_tf_code: u8) -> u32 {
+    if max_tf_code == u8::MAX {
+        u32::MAX
+    } else {
+        max_tf_code as u32
+    }
+}
+
+#[inline(always)]
+fn read_u32(data: &[u8]) -> u32 {
```
guilload, line 24:
```rust
use std::convert::TryInto;
...
fn read_u32(data: &[u8]) -> u32 {
    u32::from_le_bytes(data[..4].try_into().unwrap())
}
```
?

guilload (approved):
Made two minor comments. LGTM.

src/postings/skip.rs
fulmicoton: Yes

PR #953, opened 2020-12-09
WIP generic filter collector

fulmicoton:


src/collector/filter_collector_wrapper.rs
```diff
@@ -20,99 +22,139 @@ use crate::{Score, SegmentReader, TantivyError};
...
+        let field_entry = schema.get_field_entry(self.field);
+        if !field_entry.is_fast() {
+            return Err(TantivyError::SchemaError(format!(
+                "Field {:?} is not a fast field.",
+                field_entry.name()
+            )));
+        }
+        let requested_type = TPredicateValue::to_type();
+        let field_schema_type = field_entry.field_type().value_type();
+        if requested_type != field_schema_type {
```
barrotsteindev, line 116: Here is the run time check @fulmicoton

fulmicoton:
@barrotsteindev LGTM. Can you add the change in CHANGELOG.md and remember to put your username.

fulmicoton:
@barrotsteindev also can you fix the unit test?

barrotsteindev:
Haha yeah I'm on it

barrotsteindev:
Done @fulmicoton :)

barrotsteindev:
Linking to issue #892

barrotsteindev:
@fulmicoton , after looking further into `u64_lenient`, does it mean we currently only support numeric based (e.g. Date, i64 u64) fast fields?

fulmicoton:
@barrotsteindev f64 are supported as well. 
They are probably not well bitpacked, but the principle is always the same.

Under the hood fast field are bitpacked u64 fields.
A fast monotonic mapping make the conversion between u64 <-> the actual type.

PR #954, opened 2020-12-09
add order_by method

jakeszhong:
This pull request is to order by score or fast fields ascending, you can check the example in `order_by_search.rs`.


fulmicoton:
I am scared by the performance implication to having a dynamic number of fields here. That's  a looooot of allocation you are doing here (one per doc matching).
Also your implementation would have been shorter if you relied on `.tweak_score`.

Can you change your code into two features
- handle sorting by any type of fast field
- add `.desc()` to the TopScore.*Collector, TweakedScore.*Collector, and the CustomScore.*Collector.

For the second point, the ideal solution is probably some kind of simplifying refactoring here.
The ideal solution here is probably to replace these three top score collectors into one single top score collector that would look like the TweakedScoreCollector and add .desc() to it.

The tricky part is to find out how to handle the fact that the score might not require the score originally.

fulmicoton:
If you still want to work on this... I suggest you send a rough implementation to get feedback as early as possible

Issue #936, opened 2020-11-14
Right way to use only TF score for ranking


snakeztc:
I am trying to use a custom scoring function instead of the default BM25. For example, I wish to turn off IDF and only use the numerator in the BM25 score. What's the best way to achieve that? Would I need to write some custom struct that replaces BM25?

fulmicoton:
Hello @snakeztc. Custom similarity has been requested a couple of times, but so far it was more motivated by a will to contribute than an actual use case. 

What is your actual use case? 
(I think we can try to deliver the feature relatively fast once if we establish that it is needed.)

snakeztc:
What I am looking for is a way to customize the score of each term in a document. Different from TF-IDF or BM25, which compute the score of a term in a document using a fixed formula, recent neural ranking method, like [DeepCT](http://www.cs.cmu.edu/~zhuyund/papers/TheWebConf_2020_Dai.pdf) or [SPARTA](https://arxiv.org/abs/2009.13013) use ML models to predict the score of a term. 

Therefore, what I try to achieve is to set the score of a term in a document via inputs, instead of let the index compute it. I have created a temporal fix in my [own repo](https://github.com/snakeztc/tantivy-ztc/blob/7772b73bab452b07cc98108554f5c7680855a7fd/tantivy-0.13.2/examples/scored_text.rs#L40) (not very elegant maybe). It's done by changing PreTokStr data structure. 

It would be great if you can support it relatively fast in the main repo. Thanks! @fulmicoton

fulmicoton:
Sorry I need to read about DeepCT in detail before we can iterator on this PR. It might take a couple of weeks.

snakeztc:
Alright. it essentially sets the score(doc, term) by a model, instead of depending on BM25 or TF-IDF. Then it's used as a Inverted Index with posting list as normal.

shikhar:
I believe this is Lucene's approach: https://issues.apache.org/jira/browse/LUCENE-7854

Issue #815, opened 2020-04-19
(Approximate) Nearest Neighbour / Vector Similarity Search


bloodbare:
**Is your feature request related to a problem? Please describe.**
Nowaday search can be based by BM25 and semantic search. Using cosine similarity with a provided vector at index time we could ponderate BM25 and vector distance to score. I would say that vector is provided at index time and at search time. One document to many indexs by sentence and one query with a vector.

**Describe the solution you'd like**
Integrate Faiss (on any other vector search library) onto tantivy so the merge is done at tantivy at search time.

**[Optional] describe alternatives you've considered**
Right now we have an external vector and tantivy and we merge at middleware but may be an interesting feature to discuss about.

I'm just evaluating if it makes sense.

fulmicoton:
Using embedding for semantic search is a hot topic.

I might be a bit premature for tantivy right now (I do not know many successful implementation in the industry), but this is something we might want to revisit later.

If you do not have a very specific use case, let's keep this ticket open for the moment and revisit it later.

bloodbare:
My initial idea was (from the API point of view) to provide a new field type that can index on a vector index (like https://docs.rs/crate/hnsw/0.2.0/source/README.md) on the same "transaction" and on search beeing able to search by vector similarity. How the vector is computed from the text should be outside of the indexer (imho)

acertain:
Another approach (for text) is https://github.com/AdeDZY/DeepCT, which uses ML to weigh terms in documents & queries, and then uses bm25 with those weights.

https://microsoft.github.io/msmarco/ is good to keep an eye on, the current full ranking entries seem to be inverted index with some ml for query expansion or indexing + language model re-ranking.

fulmicoton:
@acertain this sounds super interesting !

ansjsun:
I am trying to do these. Let them work together, not one contains the other， But I found that more transformation may be on faiss.

snakeztc:
I am very interested in this too and would love to work together. I have research experience in both DeepCT and vector-based search.

fulmicoton:
@snakeztc Could you lead the development of such a feature? Also DeepCT and dense vector nearest neighbor are very different problem. Which one do you need?

This is a very valuable feature, but it adds up a lot of complexity so the condition for it to be merged are:
- a) it should be behind a feature flag.
- b) it should not be half baked. 
- c) it should be well tested
- d) it should have an actual user.

I can help with the design/code review provided we are shooting for eventually shipping this.

[...]

PR #955, opened 2020-12-10
add score mode into Inverted Index

snakeztc:
Add score mode into BM25 index to support ML based inverted index, e.g. Deep CT or SPARTA.


snakeztc:
Please take look at this solution to issue  #815 #936

PR #962, opened 2021-01-04
Remove dynamic dispatch, reformulate TokenStream as Iterators

troublescooter:
This PR needs a `Clone` implementation in the rust-stemmers crate.

This changes the `TokenFilter` and `Tokenizer` traits to an associated types API and removes the `TokenStream` trait formulating what depended on it as iterators.

The main idea is to take the struct `TextAnalyzer` which contained the dynamically dispatched types and move the dynamic dispatch a layer up, now hiding the statically dispatched types behind `dyn TextAnalyzerT` (placeholder name).

Can you benchmark these changes?


troublescooter:
[rust-stemmers Clone derive PR](https://github.com/CurrySoftware/rust-stemmers/pull/7)

fulmicoton:
Thanks. Give me a bit of time to review your work.

troublescooter:
Usually `fn builder() -> AnalyzerBuilder` would be an associated function of the struct obtained by doing `AnalyzerBuilder.build()` but this doesn't work well here because what is being built is a type containing generic type parameters and not a value, so it would need type annotations which are ugly, so `fn analyzer_builder` is a normal function. Just in case you're wondering why the builder pattern looks so abused.

fulmicoton:
I imported your branch in tantivy-search, and added a benchmark.

I also imported your branch in tantivy-search, with the benchmark.
You can check it here. https://github.com/tantivy-search/tantivy/pull/967

Tokenizing alice in wonderlong is around 40% slower on your branch than in main.
main - 1.4ms
after-your-change: 2ms

Feel free to investigate. I did not profile, but I was expecting a larger regression.
In main, there are way less allocation, since I reuse keep copying the data in the same Token's String buffer.

Because you effectively emit the tokens, you need to[ allocate a new buffer for every token](https://github.com/tantivy-search/tantivy/blob/troublescooter-static/src/tokenizer/simple_tokenizer.rs#L44).

fulmicoton:
About the removal of the internal static dispatch. tantivy used to do something like that.
https://github.com/tantivy-search/tantivy/commit/811fd0cb9e4e17a49599300c4ff2e4c7716eaa2a#diff-54c918ab79693b0b4c4573f4cb7d8c1e21e21e913b38100fa26ed09989589c21

I modified the code to dynamic dispatch for all of the tokenfilters to simplify the code and avoid gigantic binaries. I never actually timed it though.

PR #996, opened 2021-03-23
Make facet indexation and storage optional

### LINKED ISSUE #986: Facet being stored should be optional
Right now facet field are always stored. It should be optional.
Arguably, indexed should be optional too.

lpouget:
Added a `FacetOptions` for `HierarchicalFacet` which add indexed and stored flags to it.
Propagate change and update tests accordingly
Added a test to ensure that a not indexed flag was taken care of.
Added on `Value` implementation the `path()` function to return the stored facet.

Fixes #986

### DIFF
--- a/src/core/segment_reader.rs
+++ b/src/core/segment_reader.rs
@@ -108,19 +108,22 @@ impl SegmentReader {
     /// Accessor to the `FacetReader` associated to a given `Field`.
     pub fn facet_reader(&self, field: Field) -> crate::Result<FacetReader> {
         let field_entry = self.schema.get_field_entry(field);
-        if field_entry.field_type() != &FieldType::HierarchicalFacet {
-            return Err(crate::TantivyError::InvalidArgument(format!(
+
+        match field_entry.field_type() {
+            FieldType::HierarchicalFacet(_) => {
+                let term_ords_reader = self.fast_fields().u64s(field)?;
+                let termdict = self
+                    .termdict_composite
+                    .open_read(field)
+                    .map(TermDictionary::open)
+                    .unwrap_or_else(|| Ok(TermDictionary::empty()))?;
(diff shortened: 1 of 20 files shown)


src/schema/field_entry.rs
```diff
@@ -132,7 +133,7 @@ impl FieldEntry {
             | FieldType::Date(ref options) => options.is_stored(),
             FieldType::Str(ref options) => options.is_stored(),
             // TODO make stored hierarchical facet optional
```
fmassot: I think you can now remove this comment :)

fmassot (commented):
This PR seems good, I still need to understand the tests test_facet_not_populated_for_any_docs / test_facet_not_indexed_for_all_docs, not sure what is the goal here

lpouget:
> This PR seems good, I still need to understand the tests test_facet_not_populated_for_any_docs / test_facet_not_indexed_for_all_docs, not sure what is the goal here

The goal here was to be able to ensure presence of the doc in the store even if it's not indexed (and by extension check it's not indexed).

But maybe it's not the best place to do this.

src/core/segment_reader.rs
```diff
@@ -108,19 +108,22 @@ impl SegmentReader {
     /// Accessor to the `FacetReader` associated to a given `Field`.
     pub fn facet_reader(&self, field: Field) -> crate::Result<FacetReader> {
         let field_entry = self.schema.get_field_entry(field);
-        if field_entry.field_type() != &FieldType::HierarchicalFacet {
-            return Err(crate::TantivyError::InvalidArgument(format!(
+
```
fmassot, line 111: +1

fmassot (commented):
@fulmicoton I'm ok with the PR. I'm just a little bit annoyed by the verbosity of tests but it's probably better to handle that in another issue.

Issue #563, opened 2019-06-10
score fuzzy term by distance


buster:
When playing around with tantivy, i came up with a query including multiple fuzzy term queries.
Unvortunately the results don't make much sense (e.g. imagine a text field "the quick brown fox" and some other completely different text fields and then do a boolean query with multiple fuzzy terms. The results are completely unusable, because even an exact search term results in some completely different term scored the same).
I think it comes down to this: The single fuzzy term query doesn't make a difference between exact matches or some levenshtein distance.
The test should find the second added document first (with a higher score), but doesn't.

```
#[cfg(test)]
mod test {
    use tantivy::query::FuzzyTermQuery;
    use tantivy::collector::TopDocs;
    use tantivy::schema::Schema;
    use tantivy::schema::{TEXT, STORED};
    use tantivy::Index;
    use tantivy::Term;

    #[test]
    pub fn test_fuzzy_term() {
        let mut schema_builder = Schema::builder();
        let country_field = schema_builder.add_text_field("country", TEXT | STORED);
        let schema = schema_builder.build();
        let index = Index::create_in_ram(schema);
        {
            let mut index_writer = index.writer_with_num_threads(1, 10_000_000).unwrap();
            index_writer.add_document(doc!(
                country_field => "WENN ROT WIE RUBIN",
            ));
            index_writer.add_document(doc!(
                country_field => "WENN ROT WIE ROBIN",
            ));
            index_writer.commit().unwrap();
        }
        let reader = index.reader().unwrap();
        let searcher = reader.searcher();
        {
            let term = Term::from_field_text(country_field, "robin");

            let fuzzy_query = FuzzyTermQuery::new(term, 2, true);
            let top_docs = searcher
                .search(&fuzzy_query, &TopDocs::with_limit(100))
                .unwrap();
            assert_eq!(top_docs.len(), 2, "Expected 2 documents");
            let (score, adr) = top_docs[0];
            let document = index.schema().to_named_doc(&searcher.doc(adr).expect("document"));
            let json = index.schema().to_json(&searcher.doc(adr).expect("document"));
            println!("{}", json);
            println!("{:?}", top_docs);
            assert_eq!(document.0.get("country").expect("second document")[0].text().unwrap(), "WENN ROT WIE ROBIN");
        }
    }
}
```

Somehow the distance needs to be considered in the scoring, else the fuzzy query doesn't make much sense.

Or is there some other way to achieve this, i didn't see?

[...]

PR #998, opened 2021-03-23
Use Levenshtein distance to score documents in fuzzy term queries

lerouxrgd:
Fix for #563

### DIFF
--- a/src/query/automaton_weight.rs
+++ b/src/query/automaton_weight.rs
@@ -4,7 +4,7 @@ use crate::query::ConstScorer;
 use crate::query::{BitSetDocSet, Explanation};
 use crate::query::{Scorer, Weight};
 use crate::schema::{Field, IndexRecordOption};
-use crate::termdict::{TermDictionary, TermStreamer};
+use crate::termdict::{TermDictionary, TermWithStateStreamer};
 use crate::TantivyError;
 use crate::{DocId, Score};
 use std::io;
@@ -33,9 +33,9 @@ where
     fn automaton_stream<'a>(
         &'a self,
         term_dict: &'a TermDictionary,
-    ) -> io::Result<TermStreamer<'a, &'a A>> {
+    ) -> io::Result<TermWithStateStreamer<'a, &'a A>> {
         let automaton: &A = &*self.automaton;
-        let term_stream_builder = term_dict.search(automaton);
+        let term_stream_builder = term_dict.search_with_state(automaton);
         term_stream_builder.into_stream()
     }
 }
@@ -51,8 +51,7 @@ where
         let inverted_index = reader.inverted_index(self.field)?;
         let term_dict = inverted_index.terms();
         let mut term_stream = self.automaton_stream(term_dict)?;
-        while term_stream.advance() {
-            let term_info = term_stream.value();
+        while let Some((_term, term_info, _state)) = term_stream.next() {
             let mut block_segment_postings = inverted_index
                 .read_block_postings_from_terminfo(term_info, IndexRecordOption::Basic)?;
             loop {
--- a/src/query/automaton_weight.rs
+++ b/src/query/automaton_weight.rs
@@ -1,12 +1,14 @@
-use crate::common::BitSet;
 use crate::core::SegmentReader;
-use crate::query::ConstScorer;
-use crate::query::{BitSetDocSet, Explanation};
+use crate::query::fuzzy_query::DFAWrapper;
+use crate::query::score_combiner::SumCombiner;
+use crate::query::Explanation;
+use crate::query::{ConstScorer, Union};
 use crate::query::{Scorer, Weight};
 use crate::schema::{Field, IndexRecordOption};
 use crate::termdict::{TermDictionary, TermWithStateStreamer};
 use crate::TantivyError;
 use crate::{DocId, Score};
+use std::any::{Any, TypeId};
 use std::io;
 use std::sync::Arc;
 use tantivy_fst::Automaton;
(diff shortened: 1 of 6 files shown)


src/query/automaton_weight.rs
```diff
@@ -46,35 +48,27 @@ where
...
+            let segment_postings =
+                inverted_index.read_postings_from_terminfo(term_info, IndexRecordOption::Basic)?;
+            let scorer = ConstScorer::new(segment_postings, boost * score);
+            scorers.push(scorer);
         }
-        let doc_bitset = BitSetDocSet::from(doc_bitset);
-        let const_scorer = ConstScorer::new(doc_bitset, boost);
-        Ok(Box::new(const_scorer))
+
+        let scorer = Union::<_, SumCombiner>::from(scorers);
```
maufl, line 64: Will this sum up the score of all `ConstScorer`s and report the sum as score for each document?

lerouxrgd:
In [this](https://github.com/tantivy-search/tantivy/issues/563#issuecomment-805177242) comment @maufl  suggested an alternative specialization based implementation to address the `A::State` type issue. I can update this PR with it if needed.

src/query/automaton_weight.rs
lerouxrgd: As I understand it, yes. So if a doc contains `japan` and `japin` and we search for `japon`, then its final score will be 0.5 + 0.5 = 1.0

maufl: Maybe I did not understand this method correctly then. I though it would return a scorer for multiple different documents, and in this case sum the score of different documents.

src/query/automaton_weight.rs
```diff
@@ -46,35 +48,27 @@ where
...
-                }
-                block_segment_postings.advance();
-            }
+
+        let mut scorers = vec![];
+        while let Some((_term, term_info, state)) = term_stream.next() {
+            let score = automaton_score(self.automaton.as_ref(), state);
+            let segment_postings =
+                inverted_index.read_postings_from_terminfo(term_info, IndexRecordOption::Basic)?;
+            let scorer = ConstScorer::new(segment_postings, boost * score);
```
fulmicoton, line 60: This is impractical.
It may end up being a list of way too many scorers.

You should probably use a TAAT (term at a time) strategy here.
Let's say you define you know how to compute the score associate to one term match, as function of its doc_freq, term_freq and most importantly the distance (note that the doc_freq alone is not necessarily the best indication...
If I am at levenshtein 1 of two words... The one with the highest document frequency is probably the one that I was shooting for).

You probably want the score for the fuzzytermquery to be the max of the term score of all of the term that were found in the document. Not the sum.

So you could keep the bitset that was there before.
In addition it, you could open a `Vec<Score>` with a len equal to max doc and intialize it to 0.

As you go throught he terms, you then update the score of the doc that match.

score[doc] = max(score[doc], termscore(lev_distance, term_frequency, doc_frequency, ...))

Once the Bitset and the Vector of score has been populated, you can return a Scorer that iterates through the bitset and returns the computed scores. You can for instance implement a wrapper around the BitSetDocSet...

```
struct BitSetDocSetWithPrecomputedScore {
   underlying_bitset: BitSetDocSet
   scores: Vec<Score>
}

impl DocSet {
....
}

impl Scorer {
   fn score(&self) -> Score {
     let doc = self.doc();
      self.scores[doc as usize]
   }
}
```

src/termdict/fst_termdict/streamer.rs
```diff
@@ -149,3 +149,153 @@ where
         }
     }
 }
+
+/// `TermWithStateStreamerBuilder` is a helper object used to define
+/// a range of terms that should be streamed.
+pub struct TermWithStateStreamerBuilder<'a, A = AlwaysMatch>
```
fulmicoton, line 155: I think it is ok to just have a TermWithStateStreamer and avoid the code duplication.

Most of the time the state is small and "Copy".

src/query/automaton_weight.rs
fulmicoton: The AutomatonWeight is used by more than the fuzzy term Query. 

For the regex query for isntance, this extra vec with scores is too big an overhead. Please find a solution to avoid the extra cost.

src/query/automaton_weight.rs
```diff
@@ -83,6 +77,25 @@ where
...
+{
+    if TypeId::of::<DFAWrapper>() == automaton.type_id() && TypeId::of::<u32>() == state.type_id() {
+        let dfa = automaton as *const A as *const DFAWrapper;
+        let dfa = unsafe { &*dfa };
+
+        let id = &state as *const A::State as *const u32;
+        let id = unsafe { *id };
+
+        let dist = dfa.0.distance(id).to_u8() as f32;
+        1.0 / (1.0 + dist)
```
fulmicoton, line 93: Can we document this score somewhere in the doc of the FuzzyTermScorer?

Ideally I think this should be implemented as an array `[f32; 5]` that is passed by the user. (5 is sufficient. We do not handle any levenshtein distance above 4.)

I'm ok with `[1f32, 0.5f32, 0.3333f32, ...]` as a default.

src/query/automaton_weight.rs
lerouxrgd: My understanding was that the `doc_id` were not particularly between `0` and `max_doc`, so I didn't consider using a score vector like `scores[doc as usize]`, was I wrong or should it be a `HashMap` instead of a `Vec` ?

As for the scoring formula, I see how to get `doc_freq` from `term_info` for the current segment but I still need to sum those for across all segments, right ? And as we have only access to a specific `SegmentReader` here, how do we get the others without a `Searcher` ?

Moreover,  to get the `term_freq` I still need the `SegmentPostings` so I don't see how to conciliate that with the `Bitset`/`BitSetDocSet` previous approach. All this made me use the simple `1 / (1 + dist)` scoring instead.

Finally, to avoid impacting regex queries I think we can get back to the previous implementation for the generic case, and just have a specialized `impl Weigth for Automaton<DFAWrapper>` as @maufl suggested.

src/termdict/fst_termdict/streamer.rs
lerouxrgd: So is it OK if I remove the extra `TermWithState*` structs I added, and instead add the state to all the `TermStreamer*` structs as @maufl also suggested ?

src/query/automaton_weight.rs
lerouxrgd: Just to be sure, isn't max distance levenshtein 2 ? In this [range](https://github.com/tantivy-search/tantivy/blob/main/src/query/fuzzy_query.rs#L38)

lerouxrgd: Or the transposition below multiplies it by 2 again ?

fulmicoton: Ah yes you are right!
The limit of 4 is within the levenshtein automata library. I reduced it in tantivy.

src/termdict/fst_termdict/streamer.rs
fulmicoton: Yes

src/query/automaton_weight.rs
fulmicoton: > My understanding was that the doc_id were not particularly between 0 and max_doc

You are guaranteed that doc_id < max_doc.

fulmicoton: The scoring part seems very difficult. I commented on the issue.

I'd like us to take a step back and think longer on what is the spec we want.

liamwarfield: Here's my two stabs at the score problem.

I think it might be useful to approach the question of "how do you adjust scores in fuzzy searching" by asking "how likely was it that the initial query term was really an incorrect version of another term". If you knew the probability of an error you could weight the score of each fuzzy term `score_adj=score*P(T|Q)` where `P(T|Q)` is "the probability that term T is being searched for given query Q". Just multiplying `P(T|Q)` has issues (floating point error, few high probability events tend to dominate, etc) so `-log(P(T_i))` or `I(T_i)` (the Shanon information) should be used instead. This leaves us with `score_ajusted=score*(1/I(T_i))` (I'm not entirely sure this is correct). This leaves the question of how do you get `P(T_i)` or `I(T_i)`. [Long](https://en.wikipedia.org/wiki/LogSumExp) story [short](https://samsartor.dev/search-engine-1/), I(T_i) can approximated by levenshtein distance. This leaves use at `score_adj=score*(1/(α + dist))` where α is I(Q) (the information given that an initial query with `levenshtein distance=0` is correct). It just so happens that `score_adj=score*1 / (1 + dist)` is valid here! `α=1` Implies that the probability a Lev dist of 0 is 10%, `α=.04575749056` would correspond to a 90% probability.

Another approach would be to try and fix the [scoring function](https://en.wikipedia.org/wiki/Okapi_BM25#The_ranking_function) itself. BM25 uses inverse document frequency to compensate for words that rarely show up in documents. One interpretation is that IDF function is a stand in for the amount of information a document containing a term conveys. As explained above LV dist can represent the information that a search being incorrect conveys. To incorperate Lev dist into BM25, it might be useful to investigate the effect of subtracting LV dist from the LDF:

`score(D,Q) = Sum_i_n( (IDF(q_i) - LevDist(q_i)) * (f(q_i,D)*(k + 1))/(f(q_i, D) +k*( 1- b + b(|D|/avgdl)))`

This would leave the score of dist=0 terms unchanged and would lower scores of results with higher Lev dist.

LifeIsStrange:
This is not a topic I am well versed in but is this PR using beyong a regular Levenshtein distance ranking, or a more elaborate [Levenshtein automaton](https://en.wikipedia.org/wiki/Levenshtein_automaton) ?, if the latter, then this is [the strategy used by Lucene](https://blog.mikemccandless.com/2011/03/lucenes-fuzzyquery-is-100-times-faster.html)

fulmicoton:
it is using a levenshtein automaton.

LifeIsStrange:
Nice to hear!
Also [this](https://hal.archives-ouvertes.fr/hal-01360482/document) might be a useful analysis?
see also [a faster alternative to levenshtein automaton?](https://arxiv.org/abs/1008.1191)

joepio:
If I understand correctly from [this comment](https://github.com/quickwit-inc/tantivy/issues/563#issuecomment-810073006), this PR should be considered discarded, right?

PR #1001, opened 2021-03-29
Add a normed options on fields

### LINKED ISSUE #922: Make it possible to index a field without fieldnorms
Right now indexed fields come with a fieldnorm. This fieldnorm is used for BM25 scoring.

Bm25 is not always relevant. This ticket is about adding an indexing option for which fieldnorm data is not retained.

lpouget:
Make fieldnorm indexation optional:

* for all types except text: added a NORMED options
* for text field:
  * if STRING, field has not fieldnorm retained
  * if TEXT, field has fieldnorm computed

### DIFF
--- a/src/fieldnorm/mod.rs
+++ b/src/fieldnorm/mod.rs
@@ -26,3 +26,97 @@ pub use self::serializer::FieldNormsSerializer;
 pub use self::writer::FieldNormsWriter;
 
 use self::code::{fieldnorm_to_id, id_to_fieldnorm};
+
+#[cfg(test)]
+mod tests {
+    use crate::common::CompositeFile;
+    use crate::fieldnorm::FieldNormReader;
+    use crate::fieldnorm::FieldNormsSerializer;
+    use crate::fieldnorm::FieldNormsWriter;
+    use crate::{
+        directory::{Directory, RAMDirectory, WritePtr},
+        schema::{STRING, TEXT},
+    };
+    use once_cell::sync::Lazy;
+    use std::{
+        panic::{catch_unwind, AssertUnwindSafe},
+        path::Path,
+    };
+
+    use crate::schema::{Field, Schema, STORED};
+
+    pub static SCHEMA: Lazy<Schema> = Lazy::new(|| {
+        let mut schema_builder = Schema::builder();
+        schema_builder.add_text_field("field", STORED);
+        schema_builder.add_text_field("txt_field", TEXT);
+        schema_builder.add_text_field("str_field", STRING);
+        schema_builder.build()
+    });
+
+    pub static FIELD: Lazy<Field> = Lazy::new(|| SCHEMA.get_field("field").unwrap());
+    pub static TXT_FIELD: Lazy<Field> = Lazy::new(|| SCHEMA.get_field("txt_field").unwrap());
+    pub static STR_FIELD: Lazy<Field> = Lazy::new(|| SCHEMA.get_field("str_field").unwrap());
+
+    #[ignore]
+    #[test]
+    pub fn test_fieldnorm_bug() -> crate::Result<()> {
+        let path = Path::new("test");
+        let directory: RAMDirectory = RAMDirectory::create();
+        {
+            let write: WritePtr = directory.open_write(Path::new("test"))?;
+            let serializer = FieldNormsSerializer::from_write(write)?;
+            let mut fieldnorm_writers = FieldNormsWriter::for_schema(&SCHEMA);
+            fieldnorm_writers.fill_up_to_max_doc(1u32);
+            fieldnorm_writers.record(0u32, *TXT_FIELD, 5);
+            fieldnorm_writers.record(1u32, *TXT_FIELD, 3);
+            fieldnorm_writers.serialize(serializer)?;
+        }
+        let file = directory.open_read(&path)?;
+        {
+            let fields_composite = CompositeFile::open(&file)?;
+            assert!(fields_composite.open_read(*FIELD).is_none());
+            assert!(fields_composite.open_read(*TXT_FIELD).is_none());
+            assert!(fields_composite.open_read(*STR_FIELD).is_none());
+            let data = fields_composite.open_read(*TXT_FIELD).unwrap();
+            let fieldnorm_reader = FieldNormReader::open(data)?;
+            assert_eq!(fieldnorm_reader.fieldnorm(0u32), 5u32);
+            assert_eq!(fieldnorm_reader.fieldnorm(1u32), 3u32);
+        }
+        Ok(())
+    }
+
+    #[test]
+    pub fn test_fieldnorm() -> crate::Result<()> {
+        let path = Path::new("test");
+        let directory: RAMDirectory = RAMDirectory::create();
+        {
+            let write: WritePtr = directory.open_write(Path::new("test"))?;
+            let serializer = FieldNormsSerializer::from_write(write)?;
+            let mut fieldnorm_writers = FieldNormsWriter::for_schema(&SCHEMA);
+            fieldnorm_writers.fill_up_to_max_doc(1u32);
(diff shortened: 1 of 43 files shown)


fulmicoton:
@lpouget  Will review next week. Sorry for having taken so much time!

src/schema/flags.rs
```diff
@@ -30,6 +30,18 @@ pub const INDEXED: SchemaFlagList<IndexedFlag, ()> = SchemaFlagList {
     tail: (),
 };
 
+#[derive(Clone)]
+pub struct NormedFlag;
+/// Flag to mark the field as indexed.
+///
+/// The `INDEXED` flag can only be used when building `IntOptions` (`u64`, `i64` and `f64` fields)
```
fulmicoton: The comment seems incorrect.

src/fieldnorm/mod.rs
```diff
@@ -26,3 +26,97 @@ pub use self::serializer::FieldNormsSerializer;
...
+    #[ignore]
+    #[test]
+    pub fn test_fieldnorm_bug() -> crate::Result<()> {
+        let path = Path::new("test");
+        let directory: RAMDirectory = RAMDirectory::create();
+        {
+            let write: WritePtr = directory.open_write(Path::new("test"))?;
+            let serializer = FieldNormsSerializer::from_write(write)?;
+            let mut fieldnorm_writers = FieldNormsWriter::for_schema(&SCHEMA);
+            fieldnorm_writers.fill_up_to_max_doc(1u32);
```
fulmicoton: this line is actually not required.

src/fieldnorm/mod.rs
```diff
@@ -26,3 +26,97 @@ pub use self::serializer::FieldNormsSerializer;
...
+        schema_builder.build()
+    });
+
+    pub static FIELD: Lazy<Field> = Lazy::new(|| SCHEMA.get_field("field").unwrap());
+    pub static TXT_FIELD: Lazy<Field> = Lazy::new(|| SCHEMA.get_field("txt_field").unwrap());
+    pub static STR_FIELD: Lazy<Field> = Lazy::new(|| SCHEMA.get_field("str_field").unwrap());
+
+    #[ignore]
+    #[test]
+    pub fn test_fieldnorm_bug() -> crate::Result<()> {
```
fulmicoton: That's an interesting use of ignore, but let's not do that.

Let's keep the test but let's give it the right values.

fulmicoton: (and change its name of course)

src/fieldnorm/mod.rs
```diff
@@ -26,3 +26,97 @@ pub use self::serializer::FieldNormsSerializer;
...
+        }
+        Ok(())
+    }
+
+    #[test]
+    pub fn test_fail_fieldnorm_cannot_registered_twice() {
+        let mut fieldnorm_writers = FieldNormsWriter::for_schema(&SCHEMA);
+        fieldnorm_writers.fill_up_to_max_doc(1u32);
+        fieldnorm_writers.record(1u32, *TXT_FIELD, 5);
+        let result = catch_unwind(AssertUnwindSafe(|| {
```
fulmicoton: You could use the `#[should_panic]` attribute.

src/schema/flags.rs
```diff
@@ -30,6 +30,18 @@ pub const INDEXED: SchemaFlagList<IndexedFlag, ()> = SchemaFlagList {
...
+/// Of course, text fields can also be indexed... But this is expressed by using either the
+/// `STRING` (untokenized) or `TEXT` (tokenized with the english tokenizer) flags.
+pub const NORMED: SchemaFlagList<NormedFlag, ()> = SchemaFlagList {
```
fulmicoton: I think we do not need a flag for normed. It is a rather advanced attribute.

On the other hand, we probably want INDEXED to imply NORMED for backward compatibility

fulmicoton: In other words, people who want to disable NORMED won't be able to use this syntactic sugar and will have to configure the field manually which is fine.

fulmicoton (changes requested):
See comments inline.

fmassot:
I decided to finish it as I think @lpouget has a baby to take care :)
I thought the issue was easy but... I needed to touch many parts of the code and potentially I missed some stuff.
I think the PR is almost ready, I need to add some specific tests when with fields without field norms.

@fulmicoton @PSeitz When you have the time (it's not urgent at all), you can have a look.

src/indexer/merger.rs
```diff
@@ -41,17 +41,29 @@ use tantivy_bitpacker::minmax;
...
+                .field_type()
+                .has_fieldnorms()
+            {
+                let fieldnorms_reader = reader.get_fieldnorms_reader(field)?;
+                for doc in reader.doc_ids_alive() {
+                    let fieldnorm_id = fieldnorms_reader.fieldnorm_id(doc);
+                    count[fieldnorm_id as usize] += 1;
+                }
+            } else {
+                let segment_num_tokens = reader.inverted_index(field)?.total_num_tokens();
```
PSeitz: This estimation will work well for id fields, but we can probably do better for zipfs distribution.

src/indexer/merger.rs
```diff
@@ -851,10 +863,11 @@ impl IndexMerger {
             segment_map[*old_doc_id as usize] = Some(new_doc_id as DocId);
         }
 
-        // The total number of tokens will only be exact when there has been no deletes.
+        // The total number of tokens will only be exact when there has been no deletes
+        // and if the field has a norm.
```
PSeitz: norm is an approximation, not exact

src/indexer/segment_writer.rs
```diff
@@ -242,7 +242,9 @@ impl SegmentWriter {
                         )
                     };
 
-                    self.fieldnorms_writer.record(doc_id, field, num_tokens);
+                    if field_entry.has_fieldnorms() {
+                        self.fieldnorms_writer.record(doc_id, field, num_tokens);
```
PSeitz: I think fieldnorms_writer should be optional and the check simply if the writer exists

src/schema/bytes_options.rs
```diff
@@ -56,13 +71,25 @@ impl BytesOptions {
...
+impl Default for BytesOptions {
+    fn default() -> BytesOptions {
+        BytesOptions {
+            indexed: false,
+            fieldnorms: false,
+            fast: false,
+            stored: false,
+        }
+    }
+}
```
PSeitz: why not use the derive? seem to be the same values

src/schema/text_options.rs
```diff
@@ -69,6 +71,17 @@ impl TextFieldIndexing {
         &self.tokenizer
     }
 
+    /// Sets fieldnorms TODO
```
PSeitz:
```suggestion
    /// Sets fieldnorms
```

src/schema/text_options.rs
```diff
@@ -69,6 +71,17 @@ impl TextFieldIndexing {
...
+    pub fn set_fieldnorms(mut self, fieldnorms: bool) -> TextFieldIndexing {
+        self.fieldnorms = fieldnorms;
+        self
+    }
+
+    /// Sets fieldnorms TODO
```
PSeitz:
```suggestion
    /// Returns true iff fieldnorms are stored.
```

src/schema/bytes_options.rs
fmassot: yes!

src/indexer/merger.rs
fulmicoton: Yes but this is not fieldnorm here.

fulmicoton:
@lpouget I will take over that PR and merge it. I hope this is ok with you.

src/indexer/segment_writer.rs
fulmicoton: Agreed

fulmicoton:
This PR was merged as branch issue/922b

PR #1006, opened 2021-04-06
Implements merge split

evanxg852000:
https://github.com/tantivy-search/tantivy/issues/1005

Implements merge split from different dirs into one single index.
I found that `Schema` already implements `PartialEq` so we are currently making use of it.
only question remaining is: Is this the impl we want?


fulmicoton:
The impl of PartialEq seems correct

Issue #357, opened 2018-07-30
Facet Collector is deterministic but in presence of a tie, choses the top facet using a non-trivial rule. 


fulmicoton:
low priority bug. (yeah those are rare)

When collection top K facets and in presence of a tie, we rely on the fact that facets are appended sorted by their "facet path" to solve ties elegantly : the first element of a tie with the tie count wins.

The problems with this method is that if the tie happens within the first k elements, `(10, /facet/a)` vs `(10, /facet/b)`, 10 ends being the min count,   and we happen an element with higher count `(11, /facet/c)`, the logic that decides whether  ` /facet/a` or ` /facet/b` will be removed from the top K, while deterministic, might remove either `/facet/a` or `/facet/b`. From the user perspective, we would like to keep `/facet/a` to keep the well-defined logic :
we return the top 10 facets, and in case of tie we keep the first according to the lexicographical order.

hardikpnsp:
I am a bit new to Rust, would like to contribute to this issue. seems like I can change the implementation of `Ord` trait of `Hit` in order to support order by count and in case of a tie - order by facet.

PR #1007, opened 2021-04-07
Facet topk lexicographic tie break

hardikpnsp:
## Summary

- top_k in FacetCollector, when choosing between facets with equal counts will break tie in lexicographical order of facet. 
- changes the implementation of the `Ord` trait of `Hit` in order to support order by count and in case of a tie, order by facet.

## Checks

- [x] all tests pass
- [x] new tests for changed functionality added  

closes (#357)

### DIFF
--- a/src/collector/facet_collector.rs
+++ b/src/collector/facet_collector.rs
@@ -657,6 +657,45 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn test_facet_collector_topk_tie_break() {
+        let mut schema_builder = Schema::builder();
+        let facet_field = schema_builder.add_facet_field("facet", INDEXED);
+        let schema = schema_builder.build();
+        let index = Index::create_in_ram(schema);
+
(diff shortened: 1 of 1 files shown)


src/collector/facet_collector.rs
```diff
@@ -37,7 +37,10 @@ impl<'a> PartialOrd<Hit<'a>> for Hit<'a> {
 
 impl<'a> Ord for Hit<'a> {
     fn cmp(&self, other: &Self) -> Ordering {
-        other.count.cmp(&self.count)
+        match other.count.cmp(&self.count) {
```
guilload: You can chain orderings using [`Ordering::then`](https://doc.rust-lang.org/std/cmp/enum.Ordering.html#method.then).

hardikpnsp: oh cool! Thanks for the feedback. I knew there had to be a better way of doing this. will make changes in accordance with this.

fulmicoton:
LGTM.

PR #1010, opened 2021-04-16
add lz4 block format compressor as default docstore compressor

### LINKED ISSUE #1009: Add lz4_flex block format as store compressor
The lz4 block format should be sufficient for tantivy, due to the block size of 16_384 bytes.
lz4_flex is pure rust so this will improve compile times.

PSeitz:
add lz4 block compressor using lz4_flex, add lz4-block-compression feature flag
add snappy-compression feature flag for snap compressor, make snap crate optional
set lz4-block-compression as default feature flag

Closes #1009

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,9 +20,10 @@ once_cell = "1"
 regex ={version = "1", default-features = false, features = ["std"]}
 tantivy-fst = "0.3"
 memmap = {version = "0.7", optional=true}
+lz4_flex = { version = "0.7", default-features = false, features = ["checked-decode"], optional=true }
 lz4 = {version="1", optional=true}
 brotli = {version="3.3.0", optional=true}
-snap = "1"
+snap = {version="1.0", optional=true}
 tempfile = {version="3", optional=true}
 log = "0.4"
 serde = {version="1", features=["derive"]}
@@ -74,10 +75,12 @@ debug-assertions = true
 overflow-checks = true
 
 [features]
-default = ["mmap"]
+default = ["mmap", "lz4-block-compression" ]
 mmap = ["fs2", "tempfile", "memmap"]
 brotli-compression = ["brotli"]
 lz4-compression = ["lz4"]
+lz4-block-compression = ["lz4_flex"]
+snappy-compression = ["snap"]
 failpoints = ["fail/failpoints"]
 unstable = [] # useful for benches.
 wasm-bindgen = ["uuid/wasm-bindgen"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ memmap = {version = "0.7", optional=true}
 lz4_flex = { version = "0.7", default-features = false, features = ["checked-decode"], optional=true }
 lz4 = {version="1", optional=true}
 brotli = {version="3.3.0", optional=true}
-snap = {version="1.0", optional=true}
+snap = {version="1", optional=true}
 tempfile = {version="3", optional=true}
 log = "0.4"
 serde = {version="1", features=["derive"]}
--- a/src/store/compression_lz4_block.rs
+++ b/src/store/compression_lz4_block.rs
@@ -12,11 +12,8 @@ pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()>
 
     compressed.extend_from_slice(&[0, 0, 0, 0]);
     compress_into(uncompressed, compressed);
-    let size = uncompressed.len() as u32;
-    compressed[0] = size as u8;
-    compressed[1] = (size >> 8) as u8;
-    compressed[2] = (size >> 16) as u8;
-    compressed[3] = (size >> 24) as u8;
+    let num_bytes = uncompressed.len() as u32;
+    compressed[0..4].copy_from_slice(&num_bytes.to_le_bytes());
     Ok(())
 }
 
@@ -24,9 +21,12 @@ pub fn decompress(compressed: &[u8], decompressed: &mut Vec<u8>) -> io::Result<(
     decompressed.clear();
     //next lz4_flex version will support slice as input parameter.
     //this will make the usage much less ugly
-    let size = compressed.get(..4).ok_or(io::ErrorKind::InvalidData)?;
-    let size: &[u8; 4] = size.try_into().unwrap();
-    let uncompressed_size = u32::from_le_bytes(*size) as usize;
+    let uncompressed_size_bytes: &[u8; 4] = compressed
+        .get(..4)
+        .ok_or(io::ErrorKind::InvalidData)?
+        .try_into()
+        .unwrap();
+    let uncompressed_size = u32::from_le_bytes(*uncompressed_size_bytes) as usize;
     // reserve more than required, because blocked writes may write out of bounds, will be improved
     // with lz4_flex 1.0
     decompressed.reserve(uncompressed_size + 4 + 24);
--- a/src/store/compression_lz4_block.rs
+++ b/src/store/compression_lz4_block.rs
@@ -33,6 +33,7 @@ pub fn decompress(compressed: &[u8], decompressed: &mut Vec<u8>) -> io::Result<(
     unsafe {
         decompressed.set_len(uncompressed_size);
     }
-    decompress_into(&compressed[4..], decompressed).map_err(|_err| io::ErrorKind::InvalidData)?;
+    decompress_into(&compressed[4..], decompressed)
+        .map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err.to_string()))?;
     Ok(())
 }
--- a/src/store/compression_lz4_block.rs
+++ b/src/store/compression_lz4_block.rs
@@ -1,0 +1,38 @@
+use std::io::{self};
+
+use core::convert::TryInto;
+use lz4_flex::{compress_into, decompress_into};
+/// Name of the compression scheme used in the doc store.
+///
+/// This name is appended to the version string of tantivy.
+pub const COMPRESSION: &str = "lz4_block";
+
+pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
+    compressed.clear();
+
+    compressed.extend_from_slice(&[0, 0, 0, 0]);
+    compress_into(uncompressed, compressed);
+    let size = uncompressed.len() as u32;
+    compressed[0] = size as u8;
+    compressed[1] = (size >> 8) as u8;
+    compressed[2] = (size >> 16) as u8;
+    compressed[3] = (size >> 24) as u8;
+    Ok(())
+}
+
+pub fn decompress(compressed: &[u8], decompressed: &mut Vec<u8>) -> io::Result<()> {
+    decompressed.clear();
+    //next lz4_flex version will support slice as input parameter.
+    //this will make the usage much less ugly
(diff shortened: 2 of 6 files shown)


PSeitz:
Not sure if 

1. adding lz4-block-compression ff is the best way to go, or if it should behave as fallback as before with the snap crate. downside of removing the ff would be, that it would be always included even if unused, but compile overhead should be not too big.
2. appveyor should be configured to run different sets of feature flags, e.g. for each compressor

Cargo.toml
```diff
@@ -20,9 +20,10 @@ once_cell = "1"
 regex ={version = "1", default-features = false, features = ["std"]}
 tantivy-fst = "0.3"
 memmap = {version = "0.7", optional=true}
+lz4_flex = { version = "0.7", default-features = false, features = ["checked-decode"], optional=true }
 lz4 = {version="1", optional=true}
 brotli = {version="3.3.0", optional=true}
-snap = "1"
+snap = {version="1.0", optional=true}
```
fulmicoton: version=1

We are a library, so we do not want to impose a fixed version. By stopping at the major we ensure compatibility (assuming the maintainer follows SemVer), we allow users to rely on more recent versions of snap, and we reduce the chance that they might end up with two+ versions of snap in their binary application.

Before 1., the x in version 0.x.y, is treated as a major release which is why you see a lot of of 0.x in this. 

In applications on the other hand, we rely on Cargo.lock to fix versions.

Cargo.toml
```diff
@@ -74,10 +75,12 @@ debug-assertions = true
 overflow-checks = true
 
 [features]
-default = ["mmap"]
+default = ["mmap", "lz4-block-compression" ]
```
fulmicoton, line 78: oh yeah that makes more sense.

src/store/compression_lz4_block.rs
```diff
@@ -0,0 +1,39 @@
...
+/// This name is appended to the version string of tantivy.
+pub const COMPRESSION: &str = "lz4_block";
+
+pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
+    compressed.clear();
+
+    compressed.extend_from_slice(&[0, 0, 0, 0]);
+    compress_into(uncompressed, compressed);
+    let size = uncompressed.len() as u32;
+    compressed[0] = size as u8;
```
fulmicoton: compress[0..4].copy_from_slice(num_bytes.as_ref())

src/store/compression_lz4_block.rs
```diff
@@ -0,0 +1,39 @@
...
+///
+/// This name is appended to the version string of tantivy.
+pub const COMPRESSION: &str = "lz4_block";
+
+pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
+    compressed.clear();
+
+    compressed.extend_from_slice(&[0, 0, 0, 0]);
+    compress_into(uncompressed, compressed);
+    let size = uncompressed.len() as u32;
```
fulmicoton: `num_uncompressed_bytes` would be more explicit.

I am not fond of size generally speaking because it is often ambiguous on whether we are talking of bytes, elements etc. Here a more important ambiguity for the reader is are we talking about compressed bytes or uncompressed bytes.

src/store/compression_lz4_block.rs
fulmicoton: (if you are wondering the generated code is much better too)

src/store/compression_lz4_block.rs
```diff
@@ -0,0 +1,39 @@
...
+    compressed[2] = (size >> 16) as u8;
+    compressed[3] = (size >> 24) as u8;
+    Ok(())
+}
+
+pub fn decompress(compressed: &[u8], decompressed: &mut Vec<u8>) -> io::Result<()> {
+    decompressed.clear();
+    //next lz4_flex version will support slice as input parameter.
+    //this will make the usage much less ugly
+    let size = compressed.get(..4).ok_or(io::ErrorKind::InvalidData)?;
```
fulmicoton: no shadowing please. I know it is allowed by rust and clippy but it sucks. 
Here the name does not match the type either.

num_uncompressed_bytes_arr = ...

src/store/mod.rs
```diff
@@ -39,9 +39,40 @@ mod writer;
 pub use self::reader::StoreReader;
 pub use self::writer::StoreWriter;
 
+// compile_error doesn't scale very well, enum like feature flags would be great to have in Rust
```
fulmicoton, line 42: I think we will handle the compression scheme dynamically before that.

src/store/compression_lz4_block.rs
```diff
@@ -0,0 +1,39 @@
...
+    let size: &[u8; 4] = size.try_into().unwrap();
+    let uncompressed_size = u32::from_le_bytes(*size) as usize;
+    // reserve more than required, because blocked writes may write out of bounds, will be improved
+    // with lz4_flex 1.0
+    decompressed.reserve(uncompressed_size + 4 + 24);
+    unsafe {
+        decompressed.set_len(uncompressed_size);
+    }
+    decompress_into(&compressed[4..], decompressed)
```
fulmicoton, line 36: I am extremely confused here.
It seems like decompress_into is expecting a `&mut Vec`.
Aren't we supposed to pass it a clear one? What is the point of set_len?

src/store/compression_lz4_block.rs
```diff
@@ -0,0 +1,39 @@
...
+
+pub fn decompress(compressed: &[u8], decompressed: &mut Vec<u8>) -> io::Result<()> {
+    decompressed.clear();
+    //next lz4_flex version will support slice as input parameter.
+    //this will make the usage much less ugly
+    let size = compressed.get(..4).ok_or(io::ErrorKind::InvalidData)?;
+    let size: &[u8; 4] = size.try_into().unwrap();
+    let uncompressed_size = u32::from_le_bytes(*size) as usize;
+    // reserve more than required, because blocked writes may write out of bounds, will be improved
+    // with lz4_flex 1.0
```
fulmicoton, line 31: It might be a good idea to document that in lz4_flex.

src/store/compression_lz4_block.rs
fulmicoton: Ok I look into lz4_flex to understand... I think I understood the gist of it.

The API has the same name but is very different between safe and unsafe, which got me confused.

src/store/compression_lz4_block.rs
```diff
@@ -0,0 +1,39 @@
...
+    decompressed.reserve(uncompressed_size + 4 + 24);
+    unsafe {
+        decompressed.set_len(uncompressed_size);
+    }
+    decompress_into(&compressed[4..], decompressed)
+        .map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err.to_string()))?;
```
fulmicoton, line 37: should we add an assert on the size of decompressed?

Cargo.toml
PSeitz: That makes sense. I was a little confused, about brotli being specific.

src/store/compression_lz4_block.rs
PSeitz: I agree the naming is bad, generally I also prefer explicit names to increase readability

src/store/compression_lz4_block.rs
PSeitz: Generally I don't agree, I've seen a lot of bugs in my last project because there was no shadowing allowed (most notably outer db connection used in inner scope)

fulmicoton: so the shadowing was used to drop stuff?

fulmicoton: Interesting... I'll be happy to reconsider if such a case comes up shadowing in a scope to precisely "hide" something from outerscope.

PSeitz: No, it was used as shadowing for a inner scope, roughly like this

```
let my db_con = get_new_connection()
// do stuff with db_con
{
    let my db_con = get_new_connection()
    // do stuff in separate transaction with inner db con
    // if outer db connection was used sometimes, really bad things did happen like a transaction deadlock
    db_con.commit();
}
// continue on outer connection
db_con.commit();
```

src/store/compression_lz4_block.rs
PSeitz: sorry for that, I just applied my knowledge here to avoid a copy. the next lz4_flex version with support for slices should improve this

src/store/compression_lz4_block.rs
fulmicoton: Yes sorry I understood after and edited my message accordingly.

src/store/compression_lz4_block.rs
PSeitz: with support for slices this will be fixed

src/store/compression_lz4_block.rs
fulmicoton: slice API will require you do special case the last part?

Or will the the API will look like :
- pass a slice that has decompressed_len + extra len and trim it done to decompressed_len on your own
or
- pass a slice that has decompressed_len but it needs to be from a vec with an extra capacity.

An alternative could be 
decompress(compress: &[u8], decompressed: &mut Vec<u8>, len_upperbound: usize);

src/store/compression_lz4_block.rs
PSeitz: Yes, but the actual number of bytes decompressed is not yet returned, but will be with the next version.

src/store/compression_lz4_block.rs
PSeitz: The slice api will accept a slice in the correct length for uncompressed data. The blocked writes will be disabled for the last bytes, so that there will be no out of bounds writes. It's also an requirement for the frame API or blocks next to each other could accidentally overwrite data 
|Block 1| Block 2 | ..

src/store/compression_lz4_block.rs
fulmicoton: but it is resized isn't it? Otherwise I believe we have a bug.

src/store/compression_lz4_block.rs
```diff
@@ -0,0 +1,39 @@
...
+    let uncompressed_size = u32::from_le_bytes(*uncompressed_size_bytes) as usize;
+    // reserve more than required, because blocked writes may write out of bounds, will be improved
+    // with lz4_flex 1.0
+    decompressed.reserve(uncompressed_size + 4 + 24);
+    unsafe {
+        decompressed.set_len(uncompressed_size);
+    }
+    decompress_into(&compressed[4..], decompressed)
+        .map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err.to_string()))?;
+    Ok(())
```
fulmicoton, line 38: What is the size of `decompressed` when we exit this function?

src/store/compression_lz4_block.rs
PSeitz: The unsafe version doesn't change the size of the vec. It's a bad API and this function shouldn't be public because the safe version behaves differently (push changes the size of the vector), but this will be fixed in the next version

src/store/compression_lz4_block.rs
fulmicoton: Ok!

src/store/compression_lz4_block.rs
fulmicoton: Ah my bad I thought the set_len was also uncompressed_size + the extra padding.

PR #1012, opened 2021-04-17
Add support for date range queries

### LINKED ISSUE #516: Add Date Field Support for Range Queries
Once Date Fields are merged in to master #487,
support for range queries on Date fields should be added.

rihardsk:
Closes #516. Probably also #891 (but i'll let someone do that manually).

I've extended the query grammar to support rfc3999 dates in range queries, i.e., `some_date_field:[2015-08-02T18:54:42+02:30 TO 2021-04-13T19:46:26.266051969Z}` and `some_date_field: <= 2015-08-02T18:54:42-02:00` are now valid queries.

I noticed from the commented-out test case, that was already provided, that the originally intended syntax might have been different, e.g, `2015-08-02T18:54:42+02:30..2021-04-13T19:46:26.266051969Z` but i opted to extend the existing range syntax instead so that we get things like exclusive range bounds and flexible ranges (using `>`, `<=`, ...) _for free_. Hope that's ok.

### DIFF
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -1,11 +1,11 @@
 use super::user_input_ast::{UserInputAST, UserInputBound, UserInputLeaf, UserInputLiteral};
 use crate::Occur;
-use combine::error::StringStreamError;
 use combine::parser::char::{char, digit, letter, space, spaces, string};
 use combine::parser::Parser;
 use combine::{
     attempt, choice, eof, many, many1, one_of, optional, parser, satisfy, skip_many1, value,
 };
+use combine::{error::StringStreamError, parser::combinator::recognize};
 
 fn field<'a>() -> impl Parser<&'a str, Output = String> {
     (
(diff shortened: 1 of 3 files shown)


rihardsk:
Btw, i encountered some unexpected behavior when trying to change the date range test to exercise range bound exclusivity. Here's the commit where i modify the test https://github.com/rihardsk/tantivy/commit/7fada2e27dbfba918f724a55083302a9eaf97b1b, see the extended commit message for details.

This is the failing line https://github.com/rihardsk/tantivy/blob/7fada2e27dbfba918f724a55083302a9eaf97b1b/src/fastfield/multivalued/mod.rs#L130. I would expect the test not to fail, but, afaik, it's not something my changes have broken (the date range test passes after the changes). I've left out the referenced commit from the PR so that CI doesn't fail.

query-grammar/src/query_grammar.rs
```diff
@@ -35,6 +35,62 @@ fn word<'a>() -> impl Parser<&'a str, Output = String> {
...
+/// which invokes chrono::DateTime::parse_from_rfc3339 on the value to actually parse
+/// it (instead of merely extracting the datetime value as string as done here).
+fn date_time<'a>() -> impl Parser<&'a str, Output = String> {
+    let two_digits = || recognize::<String, _, _>((digit(), digit()));
+
+    // Parses a time zone
+    // -06:30
+    // Z
+    let time_zone = {
+        let utc = recognize::<String, _, _>(char('Z'));
```
fulmicoton, line 53: isn't it the same as char('Z')?

query-grammar/src/query_grammar.rs
```diff
@@ -35,6 +35,62 @@ fn word<'a>() -> impl Parser<&'a str, Output = String> {
...
 
+/// Parses a date time according to rfc3339
+/// 2015-08-02T18:54:42+02
+/// 2021-04-13T19:46:26.266051969+00:00
+///
+/// NOTE: also accepts 999999-99-99T99:99:99.266051969+99:99
+/// We delegate rejecting such invalid dates to the logical AST compuation code
+/// which invokes chrono::DateTime::parse_from_rfc3339 on the value to actually parse
+/// it (instead of merely extracting the datetime value as string as done here).
+fn date_time<'a>() -> impl Parser<&'a str, Output = String> {
```
fulmicoton, line 46: very readable code! thanks!

fulmicoton:
LGTM! 

I'll add a line in the CHANGELOG with your username if that's ok.
I'll investigate your bug report separately.

Issue #981, opened 2021-02-08
Rethink / Document the tantivy index format.


fulmicoton:

drusellers:
@fulmicoton since there is a pretty big rethink open, have you looked at how Vespa does things? I've been reading a lot about that engine with all of the ES circus going on.

fulmicoton:
@drusellers Yes I did have a look at Vespa. Mostly for the way they distribute things, but not their index format.

drusellers:
https://www.youtube.com/watch?v=SzZ_A9G6PMY

Checkout 41:00 - where they talk about read/write workloads

59:00 - a discussion about in place updates

ppodolsky:
It is further development of BMW algorithm. I just leave it here, think there is no any other better place for it. https://doi.org/10.1145/3077136.3080780

I'm gonna dive into in a while, had a look but although cannot say yet if it is applicable to Tantivy or is just another one pure academical research.

fulmicoton:
Assigning to @fmassot , but he will just lead the effort.  People who want to join the discussion is welcome.

PR #1015, opened 2021-04-19
Document index file formats - Issue/981

fmassot:
related to #981 

This is a work in progress and is there to have early feedbacks on the documentation structure.

### DIFF
--- a/doc/src/index-fileformats.md
+++ b/doc/src/index-fileformats.md
@@ -1,0 +1,75 @@
+# Tantivy index file formats
+This document defines the index file formats used in Tantivy. 
+
+## Definitions
+### Index and segments
+Tantivy indexes is composed of multiple sub-indexes called segments.
+
+
+## Index metadata json
+For a given index, Tantivy stores the following metadata in a json file `meta.json` :
+- the list of segments with its metadata id, max_doc, deletes ;
+- the index schema represented by the list of fields with name, type and option data ;
+- the opstamp
+
+```
+{
+  "segments": [
+    {
+      "segment_id": uuid string,
+      "max_doc": int|null,
+      "deletes": int|null
+    },
+    ...
+  ],
+  "schema": [
+    {
+      "name": string,
+      "type": string,
+      "options": {
+        "indexing": {
+          "record": string,
+          "tokenizer": string
+        },
+        "stored": bool
+      }
+    },
+ ],
+ ...
+ "opstamp": int
+}
+```
+
+
+## Summary of segment file extensions
+For a given segment, Tantivy stores a bunch of files whose name is set by segment uuid and whose extension defines the datastructure.
+
+| Name | Extension | Description |
+| --- | --- | --- |
+| Doc ids | `.idx` | Id |
+| Term dictionnary | `.term` | Terms |
+| Token positions | `.pos` | Token positions |
+| Token positions skip | `.posidx` | Token positions skip |
+| Docstore | `.store` | Docstore |
+| Fastfields | `.fast` | Fast fields |
+| Fieldnorm | `.fieldnorm` | Fieldnorm (BM25) |
+| Tombstone | `.del` | Deleted docs id |
+
+### Doc ids `.idx`
+
+### Term dictionnary `.term`
+
+### Token positions `.pos`
+
+### Token positions skip `.pos`
+
+### Doc store `.pos`
+
+### Fast fields `.pos`
+
+### Fieldnorm `.pos`
+
+### Delete documents `.del`
+
+
+
--- a/doc/src/index-fileformats.md
+++ b/doc/src/index-fileformats.md
@@ -10,7 +10,7 @@ Tantivy indexes is composed of multiple sub-indexes called segments.
 For a given index, Tantivy stores the following metadata in a json file `meta.json` :
 - the list of segments with its metadata id, max_doc, deletes ;
 - the index schema represented by the list of fields with name, type and option data ;
-- the opstamp
+- the opstamp or timestamp of the operation (commit).
 
 ```
 {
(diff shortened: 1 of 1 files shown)


doc/src/index-fileformats.md
```diff
@@ -0,0 +1,90 @@
+# Tantivy index file formats
+This document defines the index file formats used in Tantivy. 
+
+Tantivy has one metadata json file for each index and a few files for each segment, each file storing a specific datastructure.
+
+## Index metadata json
+For a given index, Tantivy stores the following metadata in a json file `meta.json` :
```
PSeitz, line 9: Linking to the code can also help to familiarize with the code base
https://github.com/tantivy-search/tantivy/blob/fef428a9c692a2c1cf3242642657695c71dbc0c1/src/core/index_meta.rs#L206-L216

Generally I think it would be nice to have this information in the rust docs and referenced here, so it's unlikely to be outdated

doc/src/index-fileformats.md
```diff
@@ -0,0 +1,90 @@
...
+- the list of segments with its metadata id, max_doc, deletes ;
+- the index schema represented by the list of fields with name, type and option data ;
```
PSeitz: misplaced semicolons

doc/src/index-fileformats.md
fmassot: yes, for the metadata file it's really relevant as the code is almost sufficient

fmassot: @PSeitz I would actually use a github link to avoid pointing to a specific version such as this link: https://docs.rs/tantivy/0.14.0/tantivy/

doc/src/index-fileformats.md
```diff
@@ -0,0 +1,119 @@
...
+    },
+ ],
+ ...
+ "opstamp": int
+}
+```
+
+
+## Endianness
+All integers and floats are serialiazed in litte indian order.
```
fulmicoton: That's actually incorrect. In terms, we use bigendian explicitly. That's because big endian lexicographical order is the natural order of integer.

fulmicoton: (and we use little endian everywhere else because modern CPUs are little endian)

doc/src/index-fileformats.md
```diff
@@ -0,0 +1,119 @@
...
+    - a file address composed by the field id in u32 and an index encoded as a variable integer
+
+
+
+## Summary of segment file extensions
+For a given segment, Tantivy stores a bunch of files whose name is set by segment uuid and whose extension defines the datastructure.
+
+| Name | Extension | Description |
+| --- | --- | --- |
+| Posting list | `.idx` | Sorted lists of document ids, associated to terms |
```
fulmicoton: the phrase in the description is a tad ambiguous, partly because of the comma I think. Possible rephrasing "list of sorted doc ids associated to each term"

doc/src/index-fileformats.md
```diff
@@ -0,0 +1,119 @@
...
+| Term dictionnary | `.term` | Dictionary associating `Term`s to an address into the `postings` file and the `positions` file |
+| Term positions | `.pos` | Positions of terms in each document |
+| Term positions file index | `.posidx` | Index to seek within the position file |
+| Document store | `.store` | Row-oriented, compressed storage of the documents |
+| Fast fields | `.fast` | Column-oriented random-access storage of fields |
+| Fieldnorm | `.fieldnorm` | Stores the sum  of the length (in terms) of each field for each document ? |
```
fulmicoton, line 52: it stores an approximation of that yes.

doc/src/index-fileformats.md
```diff
@@ -0,0 +1,119 @@
...
+| Tombstone | `.del` | Bitset describing which document of the segment is deleted  |
+
+### Posting list `.idx`
+[Ref](../../src/postings/serializer.rs)
+
+Posting list file is a composite file so it contains the footer data as described in composite file section.
```
fulmicoton: I think we can have a specific section for composite files. It is a used in many places.

doc/src/index-fileformats.md
```diff
@@ -0,0 +1,119 @@
...
+
+The posting list is composed of blocks of 128 documents where the following data is stored:
```
fulmicoton: It is probably better to present it as two separate structure I believe.

- A skip list with the last doc per block, the number of bits per block, and the block_wand_max
- The actual compressed data:
 `{bitpacked_docids, bitpacked_termfreq}^N/128` {`{vintencoded_docids, vintencoded_termfreq}%N/128`}

doc/src/index-fileformats.md
```diff
@@ -0,0 +1,119 @@
...
+- last doc id encoded stored as **u32**
+- number of bytes used to decompress bitpacked doc ids stored as **u8**
+- if field has frequency: 
+  - number of bytes used to decompress bitpacked term frequencies stored as **u8**
+  - fieldnorm_id as **u8**
+  - block_wand_max as **u8**
```
fulmicoton: that's optional (I think, not sure)

fulmicoton:
I like the lucene way of describing the file formats (it is not proper to lucene actually I have seen it elsewhere).
e.g.

https://lucene.apache.org/core/3_0_3/fileformats.html#Segments%20File

label ---> <description of the encoding>
<B>^N  means B repeated N times.

doc/src/index-fileformats.md
fmassot: Yes, I started a composite file section at the beginning

doc/src/index-fileformats.md
fmassot: fieldnorm_id and block_wand_max are always stored if we have frequency, if there's no bm25_weight, default values (0,0) are used

doc/src/index-fileformats.md
fmassot: nice, I was pretty sure I was missing something but did not remind where bigendian was used

fmassot:
> I like the lucene way of describing the file formats (it is not proper to lucene actually I have seen it elsewhere).
> e.g.
> 
> https://lucene.apache.org/core/3_0_3/fileformats.html#Segments%20File
> 
> label --->
> ^N means B repeated N times.

yes, I finally understand it and will use it, I really wanted to not keep the current format!

fmassot (commented):
nice of you to update the file formats!

doc/src/index-fileformats.md
```diff
@@ -39,19 +39,7 @@ For a given index, Tantivy stores the following metadata in a json file `meta.js
...
-
-The composite file footer can be divided into two main parts:
-- on the last 4 bytes, the `footer_len` is stored as an **u32**
-- on the last [`footer_len` - 4.. end - 4] bytes, the following data is stored:
-  - the number of fields stored as a variable int
-  - for each field:
-    - offset stored as variable integer
-    - a file address composed by the field id in u32 and an index encoded as a variable integer
-
+By default integers and floats are serialized with little indian order. In some specific cases, Tantivy uses big indian, the documentation will explicitely indicates it in this case.
```
evanxg852000: I think it's `endian` as opposed to `indian`

doc/src/index-fileformats.md
```diff
@@ -0,0 +1,161 @@
...
+By default integers and floats are serialized with little endian order. In some specific cases, Tantivy uses big indian, the documentation will explicitely indicates when big endian is used.
+
+
+### Composite file structure
+[Source](../../src/common/composite_file.rs)
+
+All segment files needs to store data for each field except for tombstone file (.del). In this case, a footer is added which stores for each field an offset that indicates the starting point (or file address) of its data.
+
+```
+Footer --> {{field_offset, field_file_address}^num_field, num_field, footer_len}
```
PSeitz, line 66: I think this is called Header, Footer would be at the end of the data

mocobeta:
Hi,
thanks for your effort!
Would it be possible to strictly separate "data structure (data type)" and its "description"? I mean, relatively recent Lucene format documentation is written as this.
https://lucene.apache.org/core/8_8_1/core/org/apache/lucene/codecs/lucene84/Lucene84PostingsFormat.html#Termdictionary

Jfyi, I wanted to show one bad example... old Lucene file format documentation is mixed up with various information; unfortunately it has become really difficult to understand with its growth.
https://lucene.apache.org/core/8_8_1/core/org/apache/lucene/codecs/lucene50/Lucene50TermVectorsFormat.html

PSeitz:
I think this is a great way of describing a format: https://github.com/mocobeta/lucene-postings-format, well done @mocobeta

doc/src/index-fileformats.md
```diff
@@ -0,0 +1,161 @@
+# Tantivy index file formats
+This document defines the index file formats used in Tantivy. 
+
+Tantivy has one metadata json file for each index and a bunch of files for each segment, each file storing a specific datastructure.
```
Celthi, line 4: Should there be a space between `data` and `structure`?

doc/src/index-fileformats.md
0xVitalii: 👌

PR #1016, opened 2021-04-19
Change to an exlusive bound in the date range test

rihardsk:
So that both, inclusive and exclusive, bounds are tested for date range queries.

This was left out from PR #1012 because the changes were buggy and made another date test fail. It's fixed now.


fulmicoton:
LGTM

PR #1019, opened 2021-04-21
fixed merge policy bug

evanxg852000:
While reviewing my code, @fulmicoton  pointed out this bug in LogMergePolicy.  
This fixes the issue and adds some test assertions to confirms the merge policy is returning the actual selected Segments.


fulmicoton:
Wait the bug was in tantivy originally >.<  haha

fulmicoton:
Thanks for the follow up Evance!

PR #1020, opened 2021-04-21
fixes doc  small typos

evanxg852000:
Please let's clarify this removal ` There is then a logarithmic mapping that` I could not understand it.


fulmicoton:
Thank you Evance!

PR #1021, opened 2021-04-21
add Index::builder, add index_settings to IndexMeta

### LINKED ISSUE #1018: Add builder API to create Index/IndexMeta
As prerequisite for #1014, we need an API layer above `Schema` to set global options like index sorting

PSeitz:
fixes #1018

### DIFF
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -101,7 +101,7 @@ impl IndexBuilder {
     /// The index will be allocated in anonymous memory.
     /// This should only be used for unit tests.
     pub fn create_in_ram(self) -> Result<Index, TantivyError> {
-        let ram_directory = RAMDirectory::create();
+        let ram_directory = RamDirectory::create();
         Ok(self
             .create(ram_directory)
             .expect("Creating a RAMDirectory should never fail"))
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -1,5 +1,4 @@
-use super::segment::Segment;
-use crate::core::Executor;
+use super::{segment::Segment, IndexSettings};
 use crate::core::IndexMeta;
 use crate::core::SegmentId;
 use crate::core::SegmentMeta;
@@ -22,6 +21,7 @@ use crate::schema::FieldType;
 use crate::schema::Schema;
 use crate::tokenizer::{TextAnalyzer, TokenizerManager};
 use crate::IndexWriter;
+use crate::{core::Executor, schema::SchemaBuilder};
 use std::collections::HashSet;
 use std::fmt;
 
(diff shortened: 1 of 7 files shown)


src/indexer/segment_updater.rs
```diff
@@ -46,6 +47,7 @@ const NUM_MERGE_THREADS: usize = 4;
 pub fn save_new_metas(schema: Schema, directory: &dyn Directory) -> crate::Result<()> {
     save_metas(
         &IndexMeta {
+            index_settings: IndexSettings::default(),
```
PSeitz: @fulmicoton 
Can we load the index metas from Index to access index_settings or does it not yet exist?

If it doesn't exist, we would need to either write them upfront or move IndexMeta into Index to have a place to store index settings on creation.

src/core/index.rs
```diff
@@ -55,6 +55,66 @@ fn load_metas(
...
+/// let body_field = schema_builder.add_text_field("body", TEXT);
+/// let schema = schema_builder.build();
+/// let settings = IndexSettings::default();
+/// let index = Index::builder().schema(schema).settings(settings).create_in_ram();
+///
+/// ```
+pub struct IndexBuilder {
+    index_meta: IndexMeta,
+}
+impl Default for IndexBuilder {
```
fulmicoton: do we use default?

src/indexer/segment_updater.rs
fulmicoton: I would give it the same status as schema.

pass it as an argument in `save_new_metas` and load it from IndexMeta and store it as a member of Index.

src/indexer/segment_updater.rs
```diff
@@ -204,6 +206,7 @@ pub fn merge_segments<Dir: Directory>(
     );
 
     let index_meta = IndexMeta {
+        index_settings: indices[0].load_metas()?.index_settings, // index_settings of all segments should be the same
```
fulmicoton, line 217: let's hold a IndexSettings member in Index (like we do with schema)

src/core/index.rs
PSeitz: removed

src/core/index.rs
```diff
@@ -55,17 +55,142 @@ fn load_metas(
...
+        self.schema = Some(schema);
+        self
+    }
+    /// Creates a new index using the `RAMDirectory`.
+    ///
+    /// The index will be allocated in anonymous memory.
+    /// This should only be used for unit tests.
+    pub fn create_in_ram(self) -> Result<Index, TantivyError> {
+        let ram_directory = RAMDirectory::create();
+        Ok(Index::create(
```
fulmicoton, line 105: So we you ended up not inversing the logic?

src/core/index.rs
```diff
@@ -147,36 +259,34 @@ impl Index {
...
-        let mmap_directory = MmapDirectory::create_from_tempdir()?;
-        Index::create(mmap_directory, schema)
+        IndexBuilder::new().schema(schema).create_from_tempdir()
     }
 
     /// Creates a new index given an implementation of the trait `Directory`.
     ///
     /// If a directory previously existed, it will be erased.
-    pub fn create<Dir: Directory>(dir: Dir, schema: Schema) -> crate::Result<Index> {
+    pub fn create<Dir: Directory>(
```
fulmicoton, line 265: can we move this guy as a private function of the Builder, and keep the original create for backward compatibility.

src/core/index.rs
```diff
@@ -147,36 +259,34 @@ impl Index {
...
+        save_new_metas(schema.clone(), settings.clone(), &directory)?;
+        let mut metas = IndexMeta::with_schema(schema);
+        metas.index_settings = settings;
+        let index = Index::open_from_metas(directory, &metas, SegmentMetaInventory::default());
         Ok(index)
     }
 
     /// Creates a new index given a directory and an `IndexMeta`.
-    fn create_from_metas(
+    fn open_from_metas(
```
fulmicoton, line 278: yes that's a more sensible name.

fulmicoton:
LGTM apart for minor comment

src/core/index.rs
PSeitz: yes that's better. done

fulmicoton: I do not see the change. Is it fixed?

PSeitz: Yes, the last commit: https://github.com/tantivy-search/tantivy/pull/1021/commits/b9b9e9e518178ca07ce79a2bbd0cde5ba45064ab

PR #1026, opened 2021-04-29
Sort Index/Docids By Field

### LINKED ISSUE #1014: Sort DocId by key
Today, DocId are only allocated using an incremental per-segment id.

In some case, carefully choosing the ordering of ids can improve performance.
For instance, 
- sorting by some a priori quality metric used in scoring (e.g. PageRank) make it possible to stop searching before havingn scored all of the matching results. (Note: that the result then obtained is not necessarily correct)
- sorting by a location field or a tenant id in a multitenant context, can also considerably improve the performance.
- sorting by a sort field like timestamps make it trivial to sort by this field.

It should be possible in tantivy to specify a field in the meta.json file as the ordering key.

- [x] specify and implement how this parameter should be set in the API. 
- [x] implement and test  the logic it for new segment
- [ ] implement and test it for merge
- [ ] understand lifecycle of temp data

# Open questions

- Do we require that field to be a fast field?
A: Generally no, but the first implementation will only support fast fields.

PSeitz:
closes #1014


src/core/index.rs
```diff
@@ -64,14 +64,14 @@ fn load_metas(
...
-/// use tantivy::{Index, IndexSettings};
+/// use tantivy::{Index, IndexSettings, IndexSortByField, Order};
 ///
 /// let mut schema_builder = Schema::builder();
 /// let id_field = schema_builder.add_text_field("id", STRING);
 /// let title_field = schema_builder.add_text_field("title", TEXT);
 /// let body_field = schema_builder.add_text_field("body", TEXT);
 /// let schema = schema_builder.build();
-/// let settings = IndexSettings::default();
+/// let settings = IndexSettings{sort_by_field: IndexSortByField{field:"title".to_string(), order:Order::Asc}};
```
PSeitz: this example is not possible currently and should be changed (or implemented)

src/core/index_meta.rs
```diff
@@ -193,9 +194,36 @@ impl InnerSegmentMeta {
...
+#[derive(Clone, Serialize)]
+pub struct IndexSettings {
+    /// Sorts the documents by information
+    /// provided in `IndexSortByField`
+    pub sort_by_field: IndexSortByField,
+}
+/// Settings to presort the documents in an index
+///
+/// Presorting documents can greatly performance
+/// in some scenarios, by appliying top n
```
fulmicoton: typo

src/core/segment.rs
```diff
@@ -97,5 +97,9 @@ pub trait SerializableSegment {
     ///
     /// # Returns
     /// The number of documents in the segment.
-    fn write(&self, serializer: SegmentSerializer) -> crate::Result<u32>;
+    fn write(
+        &self,
+        serializer: SegmentSerializer,
+        docid_map: Option<&DocidMapping>,
```
fulmicoton: Rename `DocIdMapping`

src/core/segment_component.rs
```diff
@@ -22,6 +22,8 @@ pub enum SegmentComponent {
     /// Accessing a document from the store is relatively slow, as it
     /// requires to decompress the entire block it belongs to.
     Store,
+    /// Temporary storage of the documents, before streamed to `Store`.
+    TempStore,
     /// Bitset describing which document of the segment is deleted.
     Delete,
 }
```
fulmicoton, line 29: You need to update `SegmentComponent::iterator()`. Without doing so, you will experience strange bugs.

src/fastfield/multivalued/writer.rs
```diff
@@ -94,7 +93,35 @@ impl MultiValuedFastFieldWriter {
...
+    /// With docid_map it accounts for the new mapping, returning values in the order of the
+    /// new docids.
+    fn get_ordered_values<'a: 'b, 'b>(
+        &'a self,
+        docid_map: Option<&'b DocidMapping>,
+    ) -> impl Iterator<Item = &'b [u64]> {
+        let docid_iter = if let Some(docid_map) = docid_map {
+            Box::new(docid_map.iter_old_docids().cloned()) as Box<dyn Iterator<Item = u32>>
+        } else {
+            Box::new(self.doc_index.iter().enumerate().map(|el| el.0 as u32))
```
fulmicoton, line 108: did you double check that doc_index was of len max_doc, and not `max_doc + 1`?

src/fastfield/multivalued/writer.rs
```diff
@@ -94,7 +93,35 @@ impl MultiValuedFastFieldWriter {
...
+                as Box<dyn Iterator<Item = u32>>
+        };
+        docid_iter.map(move |docid| self.get_values_for_docid(docid))
+    }
 
+    /// returns all values for a docids
+    fn get_values_for_docid(&self, docid: u32) -> &[u64] {
```
fulmicoton: nitpick `doc_id`

src/fastfield/multivalued/writer.rs
```diff
@@ -94,7 +93,35 @@ impl MultiValuedFastFieldWriter {
...
+        let start_pos = self.doc_index[docid as usize] as usize;
+        let end_pos = if docid as usize + 1 == self.doc_index.len() {
```
fulmicoton: i recently learn (was it in= your PR or maybe @evanxg852000 ?) that you can also do:

```
self.doc_index.get(docid as usize + 1).cloned().unwrap_or(self.vals.len())
```

src/fastfield/writer.rs
```diff
@@ -141,17 +142,18 @@ impl FastFieldsWriter {
         &self,
         serializer: &mut FastFieldSerializer,
         mapping: &HashMap<Field, FnvHashMap<UnorderedTermId, TermOrdinal>>,
+        docid_map: Option<&DocidMapping>,
     ) -> io::Result<()> {
         for field_writer in &self.single_value_writers {
-            field_writer.serialize(serializer)?;
+            field_writer.serialize(serializer, docid_map)?;
```
fulmicoton: nitpick doc_id_map

src/fastfield/writer.rs
```diff
@@ -141,17 +142,18 @@ impl FastFieldsWriter {
         &self,
         serializer: &mut FastFieldSerializer,
         mapping: &HashMap<Field, FnvHashMap<UnorderedTermId, TermOrdinal>>,
+        docid_map: Option<&DocidMapping>,
```
fulmicoton: DocIdMapping

src/fieldnorm/reader.rs
```diff
@@ -117,8 +117,8 @@ impl FieldNormReader {
     /// The fieldnorm is a value approximating the number
     /// of tokens in a given field of the `doc_id`.
     ///
-    /// It is imprecise, and always lower than the actual
-    /// number of tokens.
+    /// It is imprecise, and equal or lower than
```
fulmicoton, line 120: good catch...

src/indexer/index_sorter.rs
```diff
@@ -0,0 +1,426 @@
+use super::SegmentWriter;
+use crate::{DocId, IndexSettings, Order, TantivyError};
+use std::cmp::Reverse;
+
+/// Struct to provide mapping from old docid to new docid and vice versa
+pub struct DocidMapping {
```
fulmicoton: `DocIdMapping`

src/indexer/index_sorter.rs
```diff
@@ -0,0 +1,426 @@
...
+    new_docid_to_old: Vec<u32>,
```
fulmicoton: s/u32/DocId/g

src/indexer/index_sorter.rs
```diff
@@ -0,0 +1,426 @@
...
+    old_docid_to_new: Vec<u32>,
+}
+
+impl DocidMapping {
+    /// returns the new docid for the old docid
+    pub fn get_new_docid(&self, docid: DocId) -> DocId {
```
fulmicoton: nitpick `doc_id`

src/indexer/index_sorter.rs
```diff
@@ -0,0 +1,426 @@
...
+pub(crate) fn sort_index(
+    settings: IndexSettings,
+    segment_writer: &mut SegmentWriter,
+) -> crate::Result<DocidMapping> {
+    let docid_mapping = get_docid_mapping(settings, segment_writer)?;
+    Ok(docid_mapping)
+}
+
+// Generates a document mapping in the form of [index new docid] -> old docid
+fn get_docid_mapping(
```
fulmicoton: nitpick doc_id

src/indexer/index_sorter.rs
```diff
@@ -0,0 +1,426 @@
...
+
+    // create new docid to old docid index (used in fast_field_writers)
+    let data = fast_field.get_data();
+    let mut docid_and_data = data
+        .into_iter()
+        .enumerate()
+        .map(|el| (el.0 as DocId, el.1))
+        .collect::<Vec<_>>();
+    if settings.sort_by_field.order == Order::Desc {
+        docid_and_data.sort_unstable_by_key(|k| Reverse(k.1));
```
fulmicoton: i suspect stable sort might be useful here. Considering the benefit is tiny compared to the rest of the work that is done, can we get rid of the unstable part?

src/indexer/index_sorter.rs
```diff
@@ -0,0 +1,426 @@
...
+        .iter()
+        .enumerate()
+        .map(|el| (el.0 as u32, *el.1))
+        .collect::<Vec<_>>();
+    new_docid_and_old_docid.sort_unstable_by_key(|k| k.1);
+    let old_docid_to_new = new_docid_and_old_docid
+        .iter()
+        .map(|el| el.0)
+        .collect::<Vec<_>>();
+    let docid_map = DocidMapping {
```
fulmicoton: It is faster and simpler to just do:

```
let mut old_docid_to_new = vec![0; max_doc];
for i in 0..max_doc {
   old_docid_to_new[new_docid_to_old[i]] = i;
}
```

It also has the merit of expliciting the invariant.

src/indexer/index_sorter.rs
```diff
@@ -0,0 +1,426 @@
...
+    segment_writer: &mut SegmentWriter,
+) -> crate::Result<DocidMapping> {
+    let docid_mapping = get_docid_mapping(settings, segment_writer)?;
+    Ok(docid_mapping)
+}
+
+// Generates a document mapping in the form of [index new docid] -> old docid
+fn get_docid_mapping(
+    settings: IndexSettings,
+    segment_writer: &mut SegmentWriter,
```
fulmicoton: is mut required here?

src/indexer/index_sorter.rs
```diff
@@ -0,0 +1,426 @@
...
+    pub fn get_old_docid(&self, docid: DocId) -> DocId {
+        self.new_docid_to_old[docid as usize]
+    }
+    /// iterate over old docids in order of the new docids
+    pub fn iter_old_docids(&self) -> std::slice::Iter<'_, u32> {
+        self.new_docid_to_old.iter()
+    }
+}
+
+pub(crate) fn sort_index(
```
fulmicoton: shouldn;'t we remove this function and call get_doc_id_mapping directly?

src/indexer/index_sorter.rs
fulmicoton: nitpick: get_doc_id_mapping

src/indexer/segment_writer.rs
```diff
@@ -308,29 +317,62 @@ fn write(
...
+        &term_ord_map,
+        docid_map,
+    )?;
+    // finalize temp docstore and create version, which reflects the docid_map
+    if let Some(docid_map) = docid_map {
+        let store_write = serializer
+            .segment_mut()
+            .open_write(SegmentComponent::Store)?;
+        let mut store_writer = StoreWriter::new(store_write);
+        std::mem::swap(&mut serializer.store_writer, &mut store_writer);
```
fulmicoton: This is more a `mem::replace` than a `mem::swap` job here. I think clippy sends a warning (?)

src/postings/recorder.rs
```diff
@@ -113,12 +117,24 @@ impl Recorder for NothingRecorder {
...
+        docid_map: Option<&DocidMapping>,
     ) {
         let buffer = buffer_lender.lend_u8();
         self.stack.read_to_end(heap, buffer);
-        // TODO avoid reading twice.
-        for doc in VInt32Reader::new(&buffer[..]) {
-            serializer.write_doc(doc as u32, 0u32, &[][..]);
+        //TODO avoid reading twice.
+        if let Some(docid_map) = docid_map {
+            let mut docids = VInt32Reader::new(&buffer[..])
```
fulmicoton: We should probably double check that this allocation has no impact...
The iterator we have does not even helps with size hinting.

src/postings/recorder.rs
```diff
@@ -171,13 +187,27 @@ impl Recorder for TermFrequencyRecorder {
...
     ) {
         let buffer = buffer_lender.lend_u8();
         self.stack.read_to_end(heap, buffer);
         let mut u32_it = VInt32Reader::new(&buffer[..]);
-        while let Some(doc) = u32_it.next() {
-            let term_freq = u32_it.next().unwrap_or(self.current_tf);
-            serializer.write_doc(doc as u32, term_freq, &[][..]);
+        if let Some(docid_map) = docid_map {
+            let mut docid_and_tf = vec![];
+            while let Some(old_docid) = u32_it.next() {
```
fulmicoton, line 197: why not a for loop?

src/postings/recorder.rs
```diff
@@ -171,13 +187,27 @@ impl Recorder for TermFrequencyRecorder {
...
+                let term_freq = u32_it.next().unwrap_or(self.current_tf);
+                docid_and_tf.push((docid_map.get_new_docid(old_docid), term_freq));
+            }
+            docid_and_tf.sort_unstable_by_key(|&(docid, _)| docid);
```
fulmicoton: some comment as a above. I suspect the performance impact is high. especially considering all the resize.

src/postings/recorder.rs
```diff
@@ -225,10 +255,12 @@ impl Recorder for TfAndPositionRecorder {
         buffer_lender: &mut BufferLender,
         serializer: &mut FieldSerializer<'_>,
         heap: &MemoryArena,
+        docid_map: Option<&DocidMapping>,
     ) {
         let (buffer_u8, buffer_positions) = buffer_lender.lend_all();
         self.stack.read_to_end(heap, buffer_u8);
         let mut u32_it = VInt32Reader::new(&buffer_u8[..]);
+        let mut docid_and_positions = vec![];
```
fulmicoton: same comment as above... Can we check that the allocation is not too expensive

fulmicoton (changes requested):
Great job.
I added a bunch of comments.

Most notably.
- SegmentComponents::iterator need to be updated
- the algorithm for the reverse mapping is overkill

src/indexer/index_sorter.rs
PSeitz: yep, done

src/indexer/index_sorter.rs
PSeitz: ah yes that's much butter

src/indexer/index_sorter.rs
PSeitz: `get_field_writer` required &mut. I made some changes, it's non mut now

src/indexer/index_sorter.rs
PSeitz: yes, this is some residue when I started

src/core/segment_component.rs
PSeitz: I missed that, if there other such cases on enums, I think it makes sense to replace them with custom derive macros.

src/fastfield/multivalued/writer.rs
PSeitz: yes I used something similar recently.

src/indexer/segment_writer.rs
PSeitz: clippy doesn't complain, but yes this is a better version

src/postings/recorder.rs
PSeitz: I can count the number of docids, which are recorded in order to correctly resize, but I think it would be best to have here also a datastructure which allows random access to retrieve the data in order without any additional allocation. This would mean to replace the ExpUnrolledList with something closer to BlockedBitpacker, but I'm not yet sure about the exact datastructure.

PSeitz:
@fulmicoton Regarding the performance comments. To get an idea, I did measure the sort index feature on the webserver log dataset. The dataset is not ideal for all cases, since it does not contain a field with large text. The data looks like this

```json
{"date":1548129382,"ip":"5.78.198.52","url":"GET /settings/logo HTTP/1.1"}
{"date":1548129382,"ip":"66.249.66.194","url":"GET /m/browse/gas-heater/%D8%A8%D8%AE%D8%A7%D8%B1%DB%8C-%DA%AF%D8%A7%D8%B2%DB%8C HTTP/1.1"}
{"date":1548129382,"ip":"66.249.66.194","url":"GET /m/article/608/%D8%B9%D9%84%D8%AA-%D8%B5%D8%AF%D8%A7-%D8%AF%D8%A7%D8%AF%D9%86-%DA%AF%D8%B1%D8%AF%D9%86-%D9%88-%D8%AF%D8%B1%D9%85%D8%A7%D9%8
```

Config:
Date is used as fast field, url is recorded with positions.

```json
"schema": [
    {
      "name": "date",
      "type": "date",
      "options": {
        "indexed": false,
        "fast": "single",
        "stored": true
      }
    },
    {
      "name": "ip",
      "type": "text",
      "options": {
        "indexing": {
          "record": "basic",
          "tokenizer": "raw"
        },
        "stored": true
      }
    },
    {
      "name": "url",
      "type": "text",
      "options": {
        "indexing": {
          "record": "position",
          "tokenizer": "raw"
        },
        "stored": true
      }
    }
  ],
  "index_settings": {
    "sort_by_field": {
      "field":"date",
      "order":"Asc"
    }
  }
```

The performance overhead for sorting the index looks good. Indexing time with went from ~48,5 sec without sorting to ~54 sec with sorting the index. Almost all of the extra time is spend in recreating the doc store with 4,5 sec.

I removed the unnecessary ser/deserialization in the docstore and it is now down to 3,5 sec for doc store reordering and ~53 sec total.

A second dataset with larger text would be good to verify.

src/postings/recorder.rs
PSeitz: Ah I forgot to append to this comment. Since recorder data is quite sparse in most cases (zipfs), memory overhead should not be large for the intermediate data to reorder for most cases, expect peaks for very common words like "and" etc, which could include all docids and a lot of positions. ExpUnrolledList is a good choice for infrequent words, but for these very frequent words, the memory peak could be costly to reorder on some datasets. I didn't see any issues on the webserver data set, I'll test with the wiki dataset.

PSeitz:
I did measure on the Wikipedia dataset, sorted on a random generated number field.

```
{
  "segments": [],
  "schema": [
    {
      "name": "title",
      "type": "text",
      "options": {
        "indexing": {
          "record": "position",
          "tokenizer": "en_stem"
        },
        "stored": true
      }
    },
    {
      "name": "body",
      "type": "text",
      "options": {
        "indexing": {
          "record": "position",
          "tokenizer": "en_stem"
        },
        "stored": true
      }
    },
    {
      "name": "url",
      "type": "text",
      "options": {
        "indexing": {
          "record": "basic",
          "tokenizer": "en_stem"
        },
        "stored": false
      }
    },
    {
      "name": "number",
      "type": "u64",
      "options": {
        "indexed": false,
        "fast": "single",
        "stored": true
      }
    }
  ],
  "index_settings": {
    "sort_by_field": {
      "field":"number",
      "order":"Asc"
    }
  },
  "opstamp": 0
}
```
The peformance over head is low, ~390 sec instead of ~380 sec. The maximum memory allocated in the recorder remapping (inkluding positions) is 5MB.

src/postings/recorder.rs
fulmicoton: Sorry I meant it as opposed to using the borrow buffer trick. We only do one term at a time. We could keep using the same buffer everytime.

src/core/segment_component.rs
fulmicoton: I think this is the only case, so macros might be a tad overkill.

src/indexer/index_sorter.rs
```diff
@@ -0,0 +1,413 @@
+use super::SegmentWriter;
```
fulmicoton: Rename the file doc_id_mapping.rs

src/postings/recorder.rs
PSeitz: done. It's not possible in the other cases, but as mentioned in the comments below it should have no noticeable impact on indexing speed.

src/fastfield/readers.rs
```diff
@@ -119,7 +119,7 @@ impl FastFieldReaders {
 
     /// Returns the `u64` fast field reader reader associated to `field`.
     ///
-    /// If `field` is not a u64 fast field, this method returns `None`.
+    /// If `field` is not a u64 fast field, this method returns an Error.
```
fulmicoton, line 122: thanks

src/indexer/merger_sorted_index.rs
```diff
@@ -0,0 +1,385 @@
...
+        let merger: IndexMerger =
+            IndexMerger::open(index.schema(), index.settings().clone(), &segments[..])?;
+        b.iter(|| {
+            merger.generate_doc_id_mapping(&sort_by_field).unwrap();
+        });
+
+        Ok(())
+    }
+
+    //#[bench]
```
fulmicoton: remove?

src/indexer/merger_sorted_index.rs
```diff
@@ -0,0 +1,385 @@
+#[cfg(test)]
```
fulmicoton: rename merger_sorted_index_test maybe?

src/indexer/segment_updater.rs
```diff
@@ -171,6 +172,7 @@ pub fn merge_segments<Dir: Directory>(
     }
 
     let target_schema = indices[0].schema();
+    let target_setttings = indices[0].settings().clone();
```
fulmicoton: typo: target_settings

src/indexer/segment_writer.rs
```diff
@@ -103,11 +103,15 @@ impl SegmentWriter {
     /// be used afterwards.
     pub fn finalize(mut self) -> crate::Result<Vec<u64>> {
         self.fieldnorms_writer.fill_up_to_max_doc(self.max_doc);
-        let mapping = if let Some(settings) = self.segment_serializer.segment().index().settings() {
-            Some(sort_index(settings.clone(), &mut self)?)
-        } else {
-            None
-        };
+        let mapping = self
```
fulmicoton: Can you explicitly add the type here? Without editor hinting, it is like a riddle :D

src/postings/recorder.rs
```diff
@@ -278,7 +279,8 @@ impl Recorder for TfAndPositionRecorder {
             }
             if let Some(doc_id_map) = doc_id_map {
                 // this simple variant to remap may consume to much memory
-                doc_id_and_positions.push((doc_id_map.get_new_doc_id(doc), buffer_positions.to_vec()));
```
fulmicoton: Can you remind me why lending is not an option for doc_id_and_positions?

src/indexer/merger.rs
```diff
@@ -56,7 +58,37 @@ fn compute_total_num_tokens(readers: &[SegmentReader], field: Field) -> crate::R
...
 }
 
+/// `ReaderWithOrdinal` is used to be able to easier associate
+/// data with a `SegmentReader`. The ordinal is supposed to be
+/// used as an index access.
+///
+/// The ordinal identifies the position within `Merger` readers.
+#[derive(Clone, Copy)]
+pub struct SegmentReaderWithOrdinal<'a> {
+    reader: &'a SegmentReader,
```
fulmicoton: nitpick: The access is not helping, and once this struct is private, these two could be `pub` member without accessors.

src/postings/recorder.rs
PSeitz: lending of buffer_positions? this is just a temp vec which gets cleared on every loop

src/indexer/merger.rs
PSeitz: yes that's better

src/indexer/merger.rs
```diff
@@ -624,18 +845,39 @@ impl IndexMerger {
...
                         let term_freq = segment_postings.term_freq();
                         segment_postings.positions(&mut positions_buffer);
-
-                        let delta_positions = delta_computer.compute_delta(&positions_buffer);
-                        field_serializer.write_doc(remapped_doc_id, term_freq, delta_positions);
+                        // if doc_id_mapping exists, the docids are reordered, they are
+                        // not just stacked. The field serializer expects monotonically increasing
+                        // docids, so we collect and sort them first, before writing.
+                        //
+                        // I think this is not strictly necessary, it would be possible to
```
fulmicoton, line 856: That sounds very expensive. Did you test it on a dataset?

src/postings/recorder.rs
```diff
@@ -171,13 +188,27 @@ impl Recorder for TermFrequencyRecorder {
...
+        doc_id_map: Option<&DocIdMapping>,
     ) {
         let buffer = buffer_lender.lend_u8();
         self.stack.read_to_end(heap, buffer);
         let mut u32_it = VInt32Reader::new(&buffer[..]);
-        while let Some(doc) = u32_it.next() {
-            let term_freq = u32_it.next().unwrap_or(self.current_tf);
-            serializer.write_doc(doc as u32, term_freq, &[][..]);
+        if let Some(doc_id_map) = doc_id_map {
+            let mut doc_id_and_tf = vec![];
```
fulmicoton, line 197: why was lending not possible here?

src/postings/recorder.rs
```diff
@@ -225,10 +256,12 @@ impl Recorder for TfAndPositionRecorder {
         buffer_lender: &mut BufferLender,
         serializer: &mut FieldSerializer<'_>,
         heap: &MemoryArena,
+        doc_id_map: Option<&DocIdMapping>,
     ) {
         let (buffer_u8, buffer_positions) = buffer_lender.lend_all();
         self.stack.read_to_end(heap, buffer_u8);
         let mut u32_it = VInt32Reader::new(&buffer_u8[..]);
+        let mut doc_id_and_positions = vec![];
```
fulmicoton, line 264: why was lending not possible here

src/indexer/merger.rs
PSeitz: merging was 2x slower with sorting on a dataset I tested, but I have a change in the pipeline, which will increase merge speed in both cases :).
Currently the overhead of the sorting is the same as in `recorder.rs`

src/postings/recorder.rs
PSeitz: the bufferlender returns u32, but we need (u32, u32). I could extend buffer lender though.

src/indexer/merger.rs
PSeitz: 0,9% of the merge time is spend in the vector push of the posting, in the webserver dataset (tokenized url field with positions)

src/postings/recorder.rs
PSeitz: this is of type (u32, Vec<u32>), we can't lend the Vec<u32>, since it needs to be owned.

I should also note, that I couldn't find any recorder serialize methods in the perf profile.

src/postings/recorder.rs
fulmicoton: yes. Or push them two by two.

PR #1028, opened 2021-04-29
Support MoreLikeThisQuery

evanxg852000:
This add support for `MoreLikeThis Query`
Examples:
```rust
        let index = create_test_index();

        // search base 1st doc ignoring word `old`
        let query = tantivy::query::MoreLikeThisQuery::builder()
            .with_min_doc_frequency(1)
            .with_max_doc_frequency(10)
            .with_min_term_frequency(1)
            .with_min_word_length(2)
            .with_max_word_length(5)
            .with_boost_factor(1.0)
            .with_stop_words(vec!["old".to_string()])
            .with_document(DocAddress::new(0, 0));

        let top_docs = searcher.search(&query, &TopDocs::with_limit(5)).unwrap();
      
```
I avoided exposing this api as one function because it ends up having many parameters

Note: one possible semantic I am not sure of is using the raw timestamp for `FieldValue:Date` is this what we want.

https://github.com/tantivy-search/tantivy/issues/1011
https://github.com/tantivy-search/tantivy/issues/799

The logic is based on Lucene more-like-this.
Both [Lucene](https://github.com/apache/lucene/tree/main/lucene/queries/src/java/org/apache/lucene/queries/mlt) & [this](https://gist.github.com/kardeiz/d6bee7f84d0a015ed9a1bbd014b7b682) were used for this implementations.


src/query/mlt/query.rs
```diff
@@ -0,0 +1,284 @@
...
+///     .with_max_word_length(5)
+///     .with_boost_factor(1.0)
+///     .with_stop_words(vec!["for".to_string()])
+///     .with_document(DocAddress::new(2, 1));
+///
+/// ```
+#[derive(Debug, Clone)]
+pub struct MoreLikeThisQuery {
+    mlt: MoreLikeThis,
+    doc_address: Option<DocAddress>,
```
fulmicoton: we could use a small private enum called `TargetDocument` to replace the two option and the error handling.

src/query/mlt/query.rs
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+        if let Some(ref doc_fields) = self.doc_fields {
+            return self
+                .mlt
+                .query_with_document_fields(searcher, doc_fields)?
+                .weight(searcher, scoring_enabled);
+        }
+
+        Err(TantivyError::InvalidArgument("".to_string()))
```
fulmicoton: this is the logic that can be removed if we had an enum.

src/query/mlt/mlt.rs
```diff
@@ -0,0 +1,387 @@
...
+                        if self.is_noise_word(token.text.clone()) {
+                            let term = Term::from_field_text(field, &token.text);
+                            *term_frequencies.entry(term).or_insert(0) += 1;
+                        }
+                    });
+                }
+            }
+            FieldType::Str(text_options) => {
+                let mut token_streams: Vec<BoxTokenStream> = vec![];
+                let mut offsets = vec![];
```
fulmicoton: what are the offsets for!???

src/query/mlt/mlt.rs
```diff
@@ -0,0 +1,387 @@
...
+                let mut total_offset = 0;
```
fulmicoton: same question as above.

src/query/mlt/mlt.rs
```diff
@@ -0,0 +1,387 @@
...
+    fn create_score_term(
+        &self,
+        searcher: &Searcher,
+        per_field_term_frequencies: HashMap<Term, usize>,
+    ) -> Result<Vec<ScoreTerm>> {
+        let mut score_terms = BinaryHeap::new();
+        let num_docs = searcher
+            .segment_readers()
+            .iter()
+            .map(|x| x.num_docs() as u64)
```
fulmicoton: let's never user 1 letter variable name. segment_reader or at the very least reader.

src/query/mlt/mlt.rs
```diff
@@ -0,0 +1,387 @@
...
+            {
+                continue;
+            }
+
+            let doc_freq = searcher.doc_freq(&term)?;
+
+            // ignore terms with less than min_doc_frequency
+            if self
+                .min_doc_frequency
+                .map(|x| doc_freq < x)
```
fulmicoton: no x here.

fulmicoton: I recommend call self.min_doc_frequency -> self.min_doc_frequency_opt, and using min_doc_frequency as the temp variable name.

src/query/mlt/mlt.rs
```diff
@@ -0,0 +1,387 @@
...
+                .unwrap_or(false)
+            {
+                continue;
+            }
+
+            // ignore terms with more than max_doc_frequency
+            if self
+                .max_doc_frequency
+                .map(|x| doc_freq > x)
```
fulmicoton: one letter variable.

src/query/mlt/mlt.rs
```diff
@@ -0,0 +1,387 @@
...
+                .collect()
+        } else {
+            score_terms.into_sorted_vec()
+        };
+        Ok(score_terms)
+    }
+
+    /// Computes the similarity
+    fn idf(&self, doc_freq: u64, doc_count: u64) -> f32 {
+        let x = ((doc_count - doc_freq) as f32 + 0.5) / (doc_freq as f32 + 0.5);
```
fulmicoton: no sure if it is possible, but have  you check if we could reuse the logic in the BM25 object?

src/query/mlt/mlt.rs
```diff
@@ -0,0 +1,387 @@
...
+            }
+
+            // compute similarity & score
+            let idf = self.idf(doc_freq, num_docs);
+            let score = (term_frequency as f32) * idf;
+            score_terms.push(ScoreTerm::new(term, score));
+        }
+
+        // limit ourself to max_query terms. we need to sort so to avoid discarding important terms
+        let score_terms = if let Some(max_query_terms) = self.max_query_terms {
```
fulmicoton: needlessly convoluted...

It is ok to use truncate and mutate stuff if it makes your code more readable.

// score_terms assumed sorted
if let Some(max_query_terms) = self.max_query_terms_opt {
    score_terms.truncate(max_query_terms);
}

src/query/mlt/mlt.rs
```diff
@@ -0,0 +1,387 @@
...
+        return self.stop_words.contains(&word);
+    }
+
+    /// Couputes the score for each term while ignoring not useful terms
+    fn create_score_term(
+        &self,
+        searcher: &Searcher,
+        per_field_term_frequencies: HashMap<Term, usize>,
+    ) -> Result<Vec<ScoreTerm>> {
+        let mut score_terms = BinaryHeap::new();
```
fulmicoton: Why use a binary heap here?

It would have actually made sense if we were effectively using it to keep the Top-K, but we are not, so we might as well just use a Vec and sort it at the end. It would be clearer and more efficient.

The most efficient solution is to indeed use a binary heap, but then you need to keep its size below  max_query_terms

src/query/mlt/mlt.rs
evanxg852000: I will check

src/query/mlt/mlt.rs
evanxg852000: Thanks for the eyes. lot of issue with the borrow checker so i was thinking of another strategy. but resultved it. I think there still one comment not needed there. 
Just wondering whay `clippy` did not spot this

src/query/mlt/mlt.rs
evanxg852000: I really don't this this a good style. it's kind of redondant to embed the type knowledge in the variable name. That is fine for dynamically typed languages but here we already have that in the type (i.e Option<T>).
This is why `shadowing` was added IHMO but you also advised to not use shadowing. 
I also cannot call the temp variable `min_doc_frequency`. because it is also used in the context of `max_doc_frequency`.
In this case I think it's ok to shadow if the variable's scope does not  span multiple lines
```rust 
if self.min_doc_frequency
        .map(|min_doc_frequency| doc_freq < min_doc_frequency)
        .unwrap_or(false)
    {
        continue;
    }
```

src/query/mlt/mlt.rs
fulmicoton: (it might not be possible.. i did not investigate)

src/query/mlt/mlt.rs
fulmicoton: Let me start by saying I am ok with your solution.
Mine was merely a suggestion.

> I really don't this this a good style.

es I think my opinion on shadowing is not common in the rust world.

> it's kind of redondant to embed the type knowledge in the variable name.
> That is fine for dynamically typed languages but here we already have that in the type (i.e Option).

Well.. yes and no. It is not as if the type is often visible on the screen is it?
If it is inferred by the compiler, then it is visible nowhere. (well in your editor hint if you  are in a powerful editor)
If it is in your struct definition, it is 3 screens above what you are reading. 

The way I see it, redundancy is not good in programming for two reasons...
- it increases signal to noise ratio. Here it is so minor I think it does not apply much
- it increases the number of places to write when doing a modification. Were we to refactor the variable 
to an enum, we would have to rename the variable name.
I'm ok with the redundancy of adding _opt on that kind of variable.

> This is why shadowing was added IHMO but you also advised to not use shadowing.

Interestingly shadowing was added without being a conscious choice, and now it is just there. 
https://internals.rust-lang.org/t/history-of-shadowing-in-rust/6441. 
Most rust developers like that feature.

src/query/bm25.rs
```diff
@@ -9,7 +9,7 @@ use serde::Serialize;
 const K1: Score = 1.2;
 const B: Score = 0.75;
 
-fn idf(doc_freq: u64, doc_count: u64) -> Score {
+pub(crate) fn idf(doc_freq: u64, doc_count: u64) -> Score {
```
fulmicoton, line 12: cool

CHANGELOG.md
```diff
@@ -8,9 +8,10 @@ Tantivy 0.15.0
 - Bugfix consistent tie break handling in facet's topk (@hardikpnsp) #357
 - Date field support for range queries (@rihardsk) #516
 - Added lz4-flex as the default compression scheme in tantivy (@PSeitz) #1009
-- Renamed a lot of symbols to avoid all uppercasing on acronyms, as per new clippy recommendation. For instance, RAMDireotory -> RamDirectory. (@pmasurel)
+- Renamed a lot of symbols to avoid all uppercasing on acronyms, as per new clippy recommendation. For instance, RAMDirectory -> RamDirectory. (@pmasurel)
```
fulmicoton, line 11: thanks!

src/query/mlt/mlt.rs
```diff
@@ -0,0 +1,385 @@
...
+            }
+
+            // compute similarity & score
+            let idf = idf(doc_freq, num_docs);
+            let score = (*term_frequency as f32) * idf;
+            if let Some(limit) = self.max_query_terms {
+                if score_terms.len() > limit {
+                    // update the least significant term
+                    let least_significant_term_score = score_terms.peek().unwrap().0.score;
+                    if least_significant_term_score < score {
```
fulmicoton, line 367: yes

fulmicoton:
Good job!

PR #1030, opened 2021-04-29
add BlockedBitpacker

PSeitz:
move tantivy bitpacker to subcrate
remove byteorder dependency
use BlockedBitpacker in IntFastFieldWriter

### DIFF
--- a/bitpacker/src/blocked_bitpacker.rs
+++ b/bitpacker/src/blocked_bitpacker.rs
@@ -1,39 +1,50 @@
+use crate::BitUnpacker;
+
 use super::{bitpacker::BitPacker, compute_num_bits};
 
+const BLOCK_SIZE: usize = 128;
+
 #[derive(Debug, Clone)]
-struct BlockedBitpacker {
+pub struct BlockedBitpacker {
+    // bitpacked blocks
     compressed_blocks: Vec<u8>,
+    // uncompressed data, collected until blocksize
     cache: Vec<u64>,
     offset_and_bits: Vec<(u32, u8)>,
-    blocksize: u32,
+    num_elem_last_block: usize,
 }
 
 impl BlockedBitpacker {
-    fn new(blocksize: u32) -> Self {
+    pub fn new() -> Self {
         Self {
             compressed_blocks: vec![],
             cache: vec![],
             offset_and_bits: vec![],
-            blocksize,
+            num_elem_last_block: 0,
         }
     }
 
-    fn add(&mut self, el: u64) {
-        self.cache.push(el);
-        if self.cache.len() > self.blocksize as usize {
+    pub fn get_memory_usage(&self) -> usize {
+        self.compressed_blocks.capacity() + self.offset_and_bits.capacity() + self.cache.capacity()
+    }
+
+    pub fn add(&mut self, val: u64) {
+        self.cache.push(val);
+        if self.cache.len() == BLOCK_SIZE as usize {
             self.flush();
         }
     }
 
-    fn flush(&mut self) {
+    pub fn flush(&mut self) {
         if self.cache.is_empty() {
             return;
         }
         let mut bit_packer = BitPacker::new();
+        let offset = self.compressed_blocks.len() as u32;
         let num_bits_block = self
             .cache
             .iter()
-            .map(|el| compute_num_bits(*el))
+            .map(|val| compute_num_bits(*val))
             .max()
             .unwrap();
         for val in self.cache.iter() {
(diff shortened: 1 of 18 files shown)


PSeitz:
https://github.com/tantivy-search/tantivy/issues/1027

bitpacker/src/blocked_bitpacker.rs
```diff
@@ -0,0 +1,146 @@
...
+    // bitpacked blocks
+    compressed_blocks: Vec<u8>,
+    // uncompressed data, collected until BLOCK_SIZE
+    cache: Vec<u64>,
+    offset_and_bits: Vec<BlockedBitpackerEntryMetaData>,
+}
+
+#[derive(Debug, Clone, Default)]
+struct BlockedBitpackerEntryMetaData {
+    encoded: u64,
```
fulmicoton, line 27: i assume this is to avoid padding and avoid using repr(packed) ?

bitpacker/src/blocked_bitpacker.rs
```diff
@@ -0,0 +1,146 @@
...
+            );
+            unpacked
+        } else {
+            self.cache[pos_in_block]
+        }
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = u64> + '_ {
+        let bitpacked_elems = self.offset_and_bits.len() * BLOCK_SIZE;
+        let iter = (0..bitpacked_elems)
```
fulmicoton, line 139: maybe add a todo to optimize this?

bitpacker/src/blocked_bitpacker.rs
PSeitz: yes I should add some comments, we will never need the full 8byte for offset and repr(packed) would also be slower because of alignment

bitpacker/src/blocked_bitpacker.rs
```diff
@@ -0,0 +1,146 @@
...
+            self.flush();
+        }
+    }
+
+    pub fn flush(&mut self) {
+        if self.cache.is_empty() {
+            return;
+        }
+        let mut bit_packer = BitPacker::new();
+        let num_bits_block = self
```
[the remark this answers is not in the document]
PSeitz: I agree this should be added eventually, but this version is already better in the timestamp scenario

https://github.com/tantivy-search/tantivy/issues/1027#issuecomment-829860639

bitpacker/src/blocked_bitpacker.rs
PSeitz: yes, compression can also be further optimized

bitpacker/src/blocked_bitpacker.rs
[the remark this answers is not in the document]
PSeitz: I added base_value, memory usage is down 90% in the timestamp case

bitpacker/src/blocked_bitpacker.rs
```diff
@@ -0,0 +1,176 @@
...
+    assert_eq!(meta.num_bits(), 6);
+}
+
+impl BlockedBitpacker {
+    pub fn new() -> Self {
+        let mut compressed_blocks = vec![];
+        compressed_blocks.resize(8, 0);
+        Self {
+            compressed_blocks,
+            cache: vec![],
```
fulmicoton: it is more a buffer than a cache isn't it?

bitpacker/src/blocked_bitpacker.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    pub fn add(&mut self, val: u64) {
+        self.cache.push(val);
+        if self.cache.len() == BLOCK_SIZE as usize {
+            self.flush();
+        }
+    }
+
+    pub fn flush(&mut self) {
+        if self.cache.is_empty() {
```
fulmicoton: nitpick...

nowadays I like to avoid the unwrap using the following pattern instead of line 90 and line 86-88

```rust
let base_value = if let Some(min_value) = self.cache.iter().min() {
   min_value
} else {
   return;
};
```

bitpacker/src/blocked_bitpacker.rs
```diff
@@ -0,0 +1,176 @@
...
+            return;
+        }
+        let mut bit_packer = BitPacker::new();
+        let base_value = self.cache.iter().min().unwrap();
+        let num_bits_block = self
+            .cache
+            .iter()
+            .map(|val| compute_num_bits(*val - base_value))
+            .max()
```
fulmicoton: there is minmax in common. You can move it into bitpacker maybe?

bitpacker/src/blocked_bitpacker.rs
```diff
@@ -0,0 +1,176 @@
...
+        // mini_buffer before checking to flush
+        // (to be done in BitPacker)
+        for val in self.cache.iter() {
+            bit_packer
+                .write(
+                    *val - base_value,
+                    num_bits_block,
+                    &mut self.compressed_blocks,
+                )
+                .expect("cannot write bitpacking to output"); // write to im can't fail
```
fulmicoton: im means in memory?

bitpacker/src/blocked_bitpacker.rs
PSeitz: yes that's a better name

bitpacker/src/blocked_bitpacker.rs
PSeitz: I can do that, but there is still a unwrap() in line 96

bitpacker/src/blocked_bitpacker.rs
PSeitz: yes

bitpacker/src/blocked_bitpacker.rs
PSeitz: hmm I think the current version is easier to understand, since we don't use min here
```
let num_bits_block = self
    .buffer
    .iter()
    .map(|val| compute_num_bits(*val - base_value))
    .max()
    .unwrap();

```

```
let (_, num_bits_block) = minmax(
    self.buffer
        .iter()
        .map(|val| compute_num_bits(*val - base_value)),
).unwrap();

```

fulmicoton: i meant combine with the idea above of course

```
let (base_val, max_val) = if let Some((min_val, val)) = min_max() {
   (min_val, max_val)
} else {
   return
};
let num_bits = compute_num_bits(max_val - base_val);
```

PSeitz: nice, 10% faster

bitpacker/src/blocked_bitpacker.rs
```diff
@@ -0,0 +1,173 @@
...
+        self.buffer.push(val);
+        if self.buffer.len() == BLOCK_SIZE as usize {
+            self.flush();
+        }
+    }
+
+    pub fn flush(&mut self) {
+        if let Some((min_value, max_value)) = minmax(self.buffer.iter()) {
+            let mut bit_packer = BitPacker::new();
+            let num_bits_block = compute_num_bits(*max_value - min_value);
```
fulmicoton, line 88: I'm surprise this `*` compiles... what is it about?

bitpacker/src/blocked_bitpacker.rs
```diff
@@ -0,0 +1,173 @@
...
+            self.compressed_blocks
+                .resize(self.compressed_blocks.len() - 8, 0); // remove padding for bitpacker
+            let offset = self.compressed_blocks.len() as u64;
+            // todo performance: for some bit_width we
+            // can encode multiple vals into the
+            // mini_buffer before checking to flush
+            // (to be done in BitPacker)
+            for val in self.buffer.iter() {
+                bit_packer
+                    .write(
```
fulmicoton, line 100: i forgot, do we actually use write on a io::Write or is it always in memory?

bitpacker/src/blocked_bitpacker.rs
PSeitz: It's a residue from before. It seems usize implements sub on &usize

https://doc.rust-lang.org/std/primitive.usize.html#impl-Sub%3C%26%27_%20usize%3E

PR #1033, opened 2021-05-05
Capping IndexWriter Num thread

evanxg852000:
Added small changes for capping index writer number of thread to a maximum of 8.

[Issue Link](https://github.com/tantivy-search/tantivy/issues/1032)

### DIFF
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -432,7 +432,7 @@ impl Index {
 
     /// Creates a multithreaded writer
     ///
-    /// Tantivy will automatically define the number of threads to use, but 
+    /// Tantivy will automatically define the number of threads to use, but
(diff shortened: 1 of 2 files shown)


src/core/index.rs
```diff
@@ -441,7 +442,7 @@ impl Index {
     /// # Panics
     /// If the heap size per thread is too small, panics.
     pub fn writer(&self, overall_heap_size_in_bytes: usize) -> crate::Result<IndexWriter> {
-        let mut num_threads = num_cpus::get();
+        let mut num_threads = std::cmp::max(num_cpus::get(), MAX_NUM_THREAD);
```
fulmicoton: you meant min i believe.

evanxg852000: 😱

PR #1037, opened 2021-05-16
Don't use overly-minimal dependencies

kornelski:
I'm trying to get the crates ecosystem to work with `-Z minimal-versions` Cargo option, but many crates incorrectly specify required versions of their dependencies, including `tantivy`.

`tantivy` doesn't actually work with the versions of dependencies it specifies, and relies on them being implicitly upgraded to latest.

If actually old deps are used as specified in `Cargo.toml`, tantivy won't compile:

```
error[E0599]: no variant or associated item named `Norwegian` found for enum `Algorithm` in the current scope
  --> /Users/kornel/.cargo/registry/src/-10eda45d2c7a9533/tantivy-0.14.0/src/tokenizer/stemmer.rs:44:37
   |
44 |             Norwegian => Algorithm::Norwegian,
   |                                     ^^^^^^^^^ variant or associated item not found in `Algorithm`

error[E0277]: the trait bound `Algorithm: Clone` is not satisfied
  --> /Users/kornel/.cargo/registry/src/-10eda45d2c7a9533/tantivy-0.14.0/src/tokenizer/stemmer.rs:61:5
   |
61 |     stemmer_algorithm: Algorithm,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `Algorithm`
   |
```

I've upgraded all deps, because many of them have their own problems (especially in their older versions), and `tantivy` isn't tested in CI with minimal versions (it's impractical in the current state of the ecosystem), so it can't guarantee that old versions will remain to work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,53 +14,53 @@ edition = "2018"
 
 [dependencies]
 base64 = "0.13"
-byteorder = "1"
-crc32fast = "1"
-once_cell = "1"
-regex ={version = "1", default-features = false, features = ["std"]}
+byteorder = "1.4.3"
+crc32fast = "1.2.1"
+once_cell = "1.7.2"
+regex ={ version = "1.5.4", default-features = false, features = ["std"] }
 tantivy-fst = "0.3"
 memmap = {version = "0.7", optional=true}
-lz4_flex = { version = "0.7", default-features = false, features = ["checked-decode"], optional=true }
-lz4 = {version="1", optional=true}
-brotli = {version="3.3.0", optional=true}
-snap = {version="1", optional=true}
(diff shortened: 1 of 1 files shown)


fulmicoton:
I'm ok with this change, but can we make the caret explicit?
I stupidly thought it defaulted to the = operator.

byteorder = "^1.4.3"

kornelski:
It is a common surprise, but I don't think it can be fixed.

Adding `^` may just perpetuate the confusion by suggesting it was necessary to add and is different from a plain version number.

fulmicoton:
@kornelski that's one to see it indeed. Anyway thanks.

PR #1043, opened 2021-05-17
fix and refactor log merge policy, fixes #1035

### LINKED ISSUE #1035: LogMergePolicy set_max_merge_size not being respected
**Describe the bug**
- What did you do?
Configured the LogMergePolicy with max merge size and inserted 38M documents with a single final commit in the end.
```rust
let writer = index
    .writer(1024 * 1024 * 1024)
    .expect("failed to create writer");

let mut policy = tantivy::merge_policy::LogMergePolicy::default();
policy.set_max_merge_size(3_000_000);

writer.set_merge_policy(Box::new(policy));
```
- What happened?

Multiple 1M document segments merged into 9M document segments and two 9M document segments merged into a ~ 18M document segment
- What was expected?

The expected behavior was that no segment larger than 3M documents would be merged. As an aside, the desired behavior for my use case was to limit the maximum size of a segment to 3M documents in order to have enough segments to optimize parallelism.

If "master",  ideally give the specific sha1 revision.

Version 13.2

**To Reproduce**

If your bug is deterministic, can you give a minimal reproducing code?
Some bugs are not deterministic. Can you describe with precision in which context it happened?
If this is possible, can you share your code?

The time required to reindex all documents is not trivial so no attempt was made to reproduce the bug

PSeitz:
fixes a bug in log merge policy where an index was wrongly referenced by its index

### DIFF
--- a/src/indexer/log_merge_policy.rs
+++ b/src/indexer/log_merge_policy.rs
@@ -6,8 +6,8 @@ use std::f64;
 
 const DEFAULT_LEVEL_LOG_SIZE: f64 = 0.75;
 const DEFAULT_MIN_LAYER_SIZE: u32 = 10_000;
-const DEFAULT_MIN_MERGE_SIZE: usize = 8;
-const DEFAULT_MAX_MERGE_SIZE: usize = 10_000_000;
+const DEFAULT_MIN_NUM_SEGMENTS_IN_MERGE: usize = 8;
+const DEFAULT_MAX_DOCS_BEFORE_MERGE: usize = 10_000_000;
 
 /// `LogMergePolicy` tries to merge segments that have a similar number of
 /// documents.
@@ -91,8 +91,8 @@ impl MergePolicy for LogMergePolicy {
 impl Default for LogMergePolicy {
     fn default() -> LogMergePolicy {
         LogMergePolicy {
-            min_num_segments: DEFAULT_MIN_MERGE_SIZE,
-            max_docs_before_merge: DEFAULT_MAX_MERGE_SIZE,
+            min_num_segments: DEFAULT_MIN_NUM_SEGMENTS_IN_MERGE,
+            max_docs_before_merge: DEFAULT_MAX_DOCS_BEFORE_MERGE,
             min_layer_size: DEFAULT_MIN_LAYER_SIZE,
             level_log_size: DEFAULT_LEVEL_LOG_SIZE,
         }
--- a/src/indexer/log_merge_policy.rs
+++ b/src/indexer/log_merge_policy.rs
@@ -24,7 +24,7 @@ impl LogMergePolicy {
         cmp::max(self.min_layer_size, size)
     }
 
-    /// Set the minimum number of segments that may be merged together in a layer.
+    /// Set the minimum number of segments that may be merged together.
     pub fn set_min_num_segments(&mut self, min_num_segments: usize) {
         self.min_num_segments = min_num_segments;
     }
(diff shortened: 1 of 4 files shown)


src/indexer/log_merge_policy.rs
```diff
@@ -54,40 +54,35 @@ impl LogMergePolicy {
...
 impl MergePolicy for LogMergePolicy {
     fn compute_merge_candidates(&self, segments: &[SegmentMeta]) -> Vec<MergeCandidate> {
-        let mut size_sorted_tuples = segments
+        let mut size_sorted_segments = segments
             .iter()
-            .map(SegmentMeta::num_docs)
-            .enumerate()
-            .filter(|(_, s)| s <= &(self.max_merge_size as u32))
-            .collect::<Vec<(usize, u32)>>();
+            .filter(|s| s.num_docs() <= (self.max_merge_size as u32))
```
fulmicoton: s/s/segment_meta/g please

src/indexer/log_merge_policy.rs
```diff
@@ -54,40 +54,35 @@ impl LogMergePolicy {
...
-        if let Some(&(first_ind, first_score)) = size_sorted_log_tuples.first() {
-            let mut current_max_log_size = first_score;
-            let mut levels = vec![vec![first_ind]];
-            for &(ind, score) in (&size_sorted_log_tuples).iter().skip(1) {
-                if score < (current_max_log_size - self.level_log_size) {
-                    current_max_log_size = score;
+        if let Some(&(first_segment, log_size)) = sorted_segments_with_log_size.first() {
+            let mut current_max_log_size = log_size;
+            let mut levels = vec![vec![first_segment]];
+            for &(segment, segment_log_size) in (&sorted_segments_with_log_size).iter().skip(1) {
```
fulmicoton: isn't it the same without the parenthesis and the &?

fulmicoton:
```
for &(segment, segment_log_size) in sorted_segments_with_log_size.iter().skip(1) {
```

PSeitz: yes, I assume someone started with an iter on the ref &sorted_segments_with_log_size and added the skip later

src/indexer/log_merge_policy.rs
```diff
@@ -54,40 +54,35 @@ impl LogMergePolicy {
...
             levels
                 .iter()
                 .filter(|level| level.len() >= self.min_merge_size)
-                .map(|ind_vec| {
-                    MergeCandidate(ind_vec.iter().map(|&ind| segments[ind].id()).collect())
-                })
+                .map(|segments| MergeCandidate(segments.iter().map(|&seg| seg.id()).collect()))
                 .collect()
         } else {
             return vec![];
```
PSeitz: this is dead code btw because of the early exit above

src/indexer/log_merge_policy.rs
```diff
@@ -1,19 +1,20 @@
...
 use crate::core::SegmentMeta;
+use itertools::Itertools;
 use std::cmp;
 use std::f64;
 
 const DEFAULT_LEVEL_LOG_SIZE: f64 = 0.75;
 const DEFAULT_MIN_LAYER_SIZE: u32 = 10_000;
-const DEFAULT_MIN_MERGE_SIZE: usize = 8;
-const DEFAULT_MAX_MERGE_SIZE: usize = 10_000_000;
+const DEFAULT_MIN_NUM_SEGMENTS_IN_MERGE: usize = 8;
```
fulmicoton, line 9: Thank you!

PR #1044, opened 2021-05-17
add iterator over documents in docstore

PSeitz:
When profiling, I saw that around 8% of the time in a merge was spent in look-ups into the skip index. Since the documents in the merge case are read continuously, we can replace the random access with an iterator over the documents.

Merge Time on Sorted Index Before/After:
24s / 19s

Merge Time on Unsorted Index Before/After:
15s / 13,5s

So we can expect 10-20% faster merges.
This iterator is also important if we add sorting based on a field in the documents.

### DIFF
--- a/src/indexer/merger.rs
+++ b/src/indexer/merger.rs
@@ -920,19 +920,26 @@ impl IndexMerger {
             .iter()
             .map(|reader| reader.get_store_reader())
             .collect::<Result<_, _>>()?;
+        let mut document_iterators: Vec<_> = store_readers
+            .iter()
+            .enumerate()
+            .map(|(i, store)| store.iter_raw(self.readers[i].delete_bitset()))
+            .collect();
         if let Some(doc_id_mapping) = doc_id_mapping {
             for (old_doc_id, reader_with_ordinal) in doc_id_mapping {
-                let store_reader = &store_readers[reader_with_ordinal.ordinal as usize];
-                let raw_doc = store_reader.get_raw(*old_doc_id)?;
+                let store_reader = &mut document_iterators[reader_with_ordinal.ordinal as usize];
+                let raw_doc = store_reader.next().expect(&format!(
+                    "unexpected missing document in docstore on merge, doc id {:?}",
+                    old_doc_id
+                ))?;
                 store_writer.store_bytes(raw_doc.get_bytes())?;
             }
         } else {
             for reader in &self.readers {
                 let store_reader = reader.get_store_reader()?;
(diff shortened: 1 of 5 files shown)


src/indexer/merger.rs
```diff
@@ -920,19 +920,26 @@ impl IndexMerger {
...
+            .iter()
+            .enumerate()
+            .map(|(i, store)| store.iter_raw(self.readers[i].delete_bitset()))
+            .collect();
         if let Some(doc_id_mapping) = doc_id_mapping {
             for (old_doc_id, reader_with_ordinal) in doc_id_mapping {
-                let store_reader = &store_readers[reader_with_ordinal.ordinal as usize];
-                let raw_doc = store_reader.get_raw(*old_doc_id)?;
+                let store_reader = &mut document_iterators[reader_with_ordinal.ordinal as usize];
+                let raw_doc = store_reader.next().expect(&format!(
```
fulmicoton, line 931: can we return an error here? (the error message is great.)

fulmicoton:
a) this is great.
b) @ppodolsky  will like that.

src/store/reader.rs
```diff
@@ -128,6 +128,102 @@ impl StoreReader {
...
+            let raw_doc = raw_doc?;
+            let mut cursor = raw_doc.get_bytes();
+            Ok(Document::deserialize(&mut cursor)?)
+        })
+    }
+
+    /// Iterator over all RawDocuments in their order as they are stored in the doc store.
+    /// Use this, if you want to extract all Documents from the doc store.
+    /// The delete_bitset has to be forwarded from the `SegmentReader` or the results maybe wrong.
+    pub fn iter_raw<'a: 'b, 'b>(
```
fulmicoton: this should probably be `pub(crate)`?

src/store/reader.rs
```diff
@@ -128,6 +128,102 @@ impl StoreReader {
...
+        let mut checkpoint_block_iter = self.block_checkpoints();
+        let mut curr_checkpoint = checkpoint_block_iter.next();
+        let mut curr_block = curr_checkpoint
+            .as_ref()
+            .map(|checkpoint| self.read_block(&checkpoint));
+        let mut block_start_pos = 0;
+        let mut num_skipped = 0;
+        (0..last_docid)
+            .filter_map(move |doc_id| {
+                // filter_map is only used to resolve lifetime issues between the two closures on
```
fulmicoton, line 166: I am incapable of reading this :)

But I think this is ok considering it has no adherence outside of this function

PR #1046, opened 2021-05-18
Replaced RawDocument by OwnedBytes

fulmicoton:
This PR removes RawDocument and uses OwnedBytes instead.

OwnedBytes is a self referring struct that offers a slice of data (&[u8]) and its ownership.
It is similar in many ways to RawDocument.

It implements Read.

### DIFF
--- a/src/indexer/merger.rs
+++ b/src/indexer/merger.rs
@@ -1,4 +1,5 @@
 use super::doc_id_mapping::DocIdMapping;
+use crate::error::DataCorruption;
 use crate::fastfield::DeleteBitSet;
 use crate::fastfield::FastFieldReader;
 use crate::fastfield::FastFieldSerializer;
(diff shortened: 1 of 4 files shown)


src/indexer/merger.rs
```diff
@@ -927,19 +928,25 @@ impl IndexMerger {
...
+                let doc_bytes_it = &mut document_iterators[reader_with_ordinal.ordinal as usize];
+                if let Some(doc_bytes_res) = doc_bytes_it.next() {
+                    let doc_bytes = doc_bytes_res?;
+                    store_writer.store_bytes(&doc_bytes)?;
+                } else {
+                    return Err(DataCorruption::comment_only(&format!(
+                        "unexpected missing document in docstore on merge, doc id {:?}",
+                        old_doc_id
+                    ))
+                    .into());
```
PSeitz, line 940: ok_or_else would be nicer here

https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else

As a general note, I think it would be good to add backtraces to the Errors. If something fails this can be really helpful. It seems like `thiserror` supports this

PSeitz (approved):
This is much nicer

src/indexer/merger.rs
fulmicoton: I used ok_or_else at first and judged that the double ?? was confusing

fulmicoton: backtrace would be useful indeed.

PSeitz: since this reads like a book, more `?` are better for dramatic purposes

`let it_was_not_an_error = is_it_an_error???;`

fulmicoton: :)

PR #1054, opened 2021-05-19
Detect if segments are stackackable with sorting, fixes #1038

### LINKED ISSUE #1038: Important and easy merge speed optimization when sorting by a field
There is one common special case that we should optimize for...

We should try and detect whether the segments are actually trivially stackable.
For instance, when sorting by timestamp, it is very common to merge segments that are covering
disjoint interval of time.

In that case, the logic stacking the docids is correct.

PSeitz:
Detect if segments are stackable when their data ranges on the sort property are disjunct.
Presort segments by their min value on merge, to enable easier stacking.

### DIFF
--- a/src/indexer/merger.rs
+++ b/src/indexer/merger.rs
@@ -184,25 +184,7 @@ impl IndexMerger {
             }
         }
         if let Some(sort_by_field) = index_settings.sort_by_field.as_ref() {
-            // presort the readers by their min_values, so that when they are disjunct, we can use
-            // the regular merge logic (implicitly sorted)
-            let mut readers_with_min_sort_values = readers
-                .into_iter()
-                .map(|reader| {
-                    let accessor = Self::get_sort_field_accessor(&reader, &sort_by_field)?;
-                    Ok((reader, accessor.min_value()))
-                })
-                .collect::<crate::Result<Vec<_>>>()?;
-            if sort_by_field.order.is_asc() {
-                readers_with_min_sort_values.sort_by_key(|(_, min_val)| *min_val);
-            } else {
-                readers_with_min_sort_values
-                    .sort_by_key(|(_, min_val)| std::cmp::Reverse(*min_val));
-            }
-            readers = readers_with_min_sort_values
-                .into_iter()
-                .map(|(reader, _)| reader)
-                .collect();
+            readers = Self::sort_readers_by_min_sort_field(readers, sort_by_field)?;
         }
         // sort segments by their natural sort setting
         if max_doc >= MAX_DOC_LIMIT {
@@ -219,6 +201,30 @@ impl IndexMerger {
             readers,
             max_doc,
         })
+    }
+
+    fn sort_readers_by_min_sort_field(
+        readers: Vec<SegmentReader>,
+        sort_by_field: &IndexSortByField,
+    ) -> crate::Result<Vec<SegmentReader>> {
+        // presort the readers by their min_values, so that when they are disjunct, we can use
+        // the regular merge logic (implicitly sorted)
+        let mut readers_with_min_sort_values = readers
+            .into_iter()
+            .map(|reader| {
+                let accessor = Self::get_sort_field_accessor(&reader, &sort_by_field)?;
+                Ok((reader, accessor.min_value()))
+            })
+            .collect::<crate::Result<Vec<_>>>()?;
+        if sort_by_field.order.is_asc() {
+            readers_with_min_sort_values.sort_by_key(|(_, min_val)| *min_val);
+        } else {
+            readers_with_min_sort_values.sort_by_key(|(_, min_val)| std::cmp::Reverse(*min_val));
+        }
+        Ok(readers_with_min_sort_values
+            .into_iter()
+            .map(|(reader, _)| reader)
+            .collect())
     }
 
     fn write_fieldnorms(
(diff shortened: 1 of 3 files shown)


src/indexer/merger.rs
```diff
@@ -355,6 +377,55 @@ impl IndexMerger {
...
+    pub(crate) fn get_sort_field_accessor<'a, 'b>(
+        reader: &SegmentReader,
+        sort_by_field: &'b IndexSortByField,
+    ) -> crate::Result<FastFieldReader<u64>> {
+        let field_id = expect_field_id_for_sort_field(&reader.schema(), &sort_by_field)?; // for now expect fastfield, but not strictly required
+        let value_accessor = reader.fast_fields().u64_lenient(field_id)?;
+        Ok(value_accessor)
+    }
+    /// Collecting value_accessors into a vec to bind the lifetime.
+    pub(crate) fn get_reader_with_sort_field_accessor<'a, 'b>(
```
fulmicoton, line 417: I believe this could be private, and be a function rather than a static method.

src/indexer/merger.rs
```diff
@@ -183,6 +183,28 @@ impl IndexMerger {
                 readers.push(reader);
             }
         }
+        if let Some(sort_by_field) = index_settings.sort_by_field.as_ref() {
```
fulmicoton, line 186: No need to fix: 

Just a remark.
`sort_by_cached_key` would be have been cleaner, but does not allow for error handling.
I almost miss Exceptions.

https://doc.rust-lang.org/std/primitive.slice.html#method.sort_by_cached_key

src/indexer/merger.rs
```diff
@@ -183,6 +183,28 @@ impl IndexMerger {
...
+            if sort_by_field.order.is_asc() {
+                readers_with_min_sort_values.sort_by_key(|(_, min_val)| *min_val);
+            } else {
+                readers_with_min_sort_values
+                    .sort_by_key(|(_, min_val)| std::cmp::Reverse(*min_val));
+            }
+            readers = readers_with_min_sort_values
+                .into_iter()
+                .map(|(reader, _)| reader)
+                .collect();
```
fulmicoton: Can we isolate this block in a function 

```
fn sort_readers_by_min_sort_field(readers: Vec<SegmentReader>, sort_by_field: &IndexSortByField) -> crate::Result<Vec<SegmentReader>> {
   //...  
}
```

src/indexer/merger.rs
```diff
@@ -355,6 +377,55 @@ impl IndexMerger {
...
+    /// Checks if the readers are disjunct for their sort property and in the correct order to be
+    /// able to just stack them.
+    pub(crate) fn is_disjunct_and_sorted_on_sort_property(
+        &self,
+        sort_by_field: &IndexSortByField,
+    ) -> crate::Result<bool> {
+        let reader_and_field_accessors =
+            Self::get_reader_with_sort_field_accessor(&self.readers, sort_by_field)?;
+
+        let everything_is_in_order = reader_and_field_accessors
```
fulmicoton, line 394: Kudos on that. I love these variable used only to make the code more readable and provide the intent.

src/indexer/merger.rs
```diff
@@ -355,6 +377,55 @@ impl IndexMerger {
...
+            .into_iter()
+            .map(|reader| reader.1)
+            .tuple_windows()
```
fulmicoton, line 397: I didn't know about tuple_windows

src/indexer/merger.rs
```diff
@@ -355,6 +377,55 @@ impl IndexMerger {
...
+            .all(|(field_accessor1, field_accessor2)| {
```
fulmicoton, line 398: Nitpick:

No need to change anything.

I like to use `_left`, and `_right` instead of `_0`, `_1`.
The most popular alternative might be _lhs and _rhs maybe?

fulmicoton (changes requested):
Great job.

One minor comment to isolate a block of the code in a function.

src/indexer/merger.rs
PSeitz: It's an extension of itertools, imo an awesome lib

fulmicoton: AAAAAh I see. It is useful indeed.

src/indexer/merger.rs
PSeitz: I feel like this is a missing building block in Rust, to convert any function into one which returns a Result. But it would probably add too much complexity at this stage into the language.

fulmicoton: I think this is a contrarian opinion but I actually like that Java semantics of Exception.

src/indexer/merger.rs
PSeitz: in this case we want to check the order in the array, and imo `1`, `2` better expresses the elements order than left right or is this just me?

src/indexer/merger.rs
PSeitz: yep, done

src/indexer/merger.rs
PSeitz: done, thanks for the copy pasta

src/indexer/merger.rs
fulmicoton: I see what you mean. I agree!

PR #1056, opened 2021-05-19
Provide a means to deal with malformed facet text representation for the query parser

moriyoshi:
**This patch requires API change.**

As `Facet::from_text` is not designed to return a `Result`, the query parser cannot handle invalid facet representations.   This patch enables it to do so by modifying `Facet::from_text()` to return a `Result<Facet, TantivyError>` instead of a bare Facet.  Also, I added a new error kind to `QueryParserError`.

BTW, I will agree to not insist any copyright on this patch.

### DIFF
--- a/src/query/query_parser/query_parser.rs
+++ b/src/query/query_parser/query_parser.rs
@@ -13,6 +13,7 @@ use crate::schema::{Field, Schema};
 use crate::schema::{FieldType, Term};
 use crate::tokenizer::TokenizerManager;
 use crate::Score;
+use crate::TantivyError;
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::num::{ParseFloatError, ParseIntError};
@@ -68,6 +69,9 @@ pub enum QueryParserError {
     /// The format for the date field is not RFC 3339 compliant.
     #[error("The date field has an invalid format")]
     DateFormatError(chrono::ParseError),
+    /// The format for the facet field invalid.
+    #[error("The facet field is malformed")]
+    FacetFormatError(String),
 }
 
 impl From<ParseIntError> for QueryParserError {
@@ -358,10 +362,15 @@ impl QueryParser {
                     ))
                 }
             }
-            FieldType::HierarchicalFacet(_) => {
-                let facet = Facet::from_text(phrase);
-                Ok(vec![(0, Term::from_field_text(field, facet.encoded_str()))])
-            }
+            FieldType::HierarchicalFacet(_) => match Facet::from_text(phrase) {
+                Ok(facet) => Ok(vec![(0, Term::from_field_text(field, facet.encoded_str()))]),
+                Err(e) => match e {
+                    TantivyError::InvalidArgument(path) => {
+                        Err(QueryParserError::FacetFormatError(path))
+                    }
+                    _ => Err(QueryParserError::SyntaxError),
+                },
+            },
             FieldType::Bytes(_) => {
                 let bytes = base64::decode(phrase).map_err(QueryParserError::ExpectedBase64)?;
                 let term = Term::from_field_bytes(field, &bytes);
--- a/src/query/query_parser/query_parser.rs
+++ b/src/query/query_parser/query_parser.rs
@@ -8,12 +8,11 @@ use crate::query::Query;
 use crate::query::RangeQuery;
 use crate::query::TermQuery;
 use crate::query::{AllQuery, BoostQuery};
-use crate::schema::{Facet, IndexRecordOption};
+use crate::schema::{Facet, FacetParseError, IndexRecordOption};
 use crate::schema::{Field, Schema};
 use crate::schema::{FieldType, Term};
 use crate::tokenizer::TokenizerManager;
 use crate::Score;
-use crate::TantivyError;
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::num::{ParseFloatError, ParseIntError};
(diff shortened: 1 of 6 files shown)


src/schema/facet.rs
```diff
@@ -75,11 +76,47 @@ impl Facet {
...
     where
         T: ?Sized + AsRef<str>,
     {
-        From::from(path)
+        #[derive(Copy, Clone)]
+        enum State {
+            Escaped,
+            Idle,
+        }
+        let path_ = path.as_ref();
```
PSeitz: `path_ref` or `path` [although @fulmicoton doesn't like shadowing :)] would be a better name

PSeitz (commented):
Thanks for the PR! lgtm besides the two comments

src/schema/facet.rs
fulmicoton: path_ref please.

src/query/query_parser/query_parser.rs
```diff
@@ -68,6 +69,9 @@ pub enum QueryParserError {
     /// The format for the date field is not RFC 3339 compliant.
     #[error("The date field has an invalid format")]
     DateFormatError(chrono::ParseError),
+    /// The format for the facet field invalid.
+    #[error("The facet field is malformed")]
+    FacetFormatError(String),
```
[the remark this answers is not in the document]
moriyoshi: While an additional entry in `TantivyError` for this will possibly be handy later, I prefer this to be a query parser error like others (e.g. date time format error)  because one can think Facet string representations are coincidentally the same between Facet and QueryParser.

src/schema/facet.rs
moriyoshi: I don't have any preference here.

src/query/query_parser/query_parser.rs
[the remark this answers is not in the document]
moriyoshi: From the view point of `Facet::from_text()`, it will just suffice, for now, to return with a single kind of error (meaning `TantivyError::InvalidArgument`) because the parser isn't made possible to report a detailed error.

moriyoshi:
As a sidenote, I intended to use Tantivy in a production and found that user-supplied queries including wrong facet strings can effectively cause DoS (the server itself is written in Python).  That's why I made this fix.  Anyway, thanks for offering such a great product for free with a generous license!

src/query/query_parser/query_parser.rs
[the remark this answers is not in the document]
PSeitz: I had another look and think `from_text` should return a `FacetParseError`. This error would then be included in the tantivty and parse error sets. Currently there is a conversion to `_ => Err(QueryParserError::SyntaxError),`, but this is dead code as I understand it, which would be removed.

```
/// Error that may occur when parsing a facet
#[derive(Debug, Error)]
pub enum FacetParseError {
    /// The format for the facet field invalid.
    #[error("The facet field is malformed")]
    FacetParseError(String),
}
```


```
pub enum TantivyError {
    /// Failed to parse facet.
    #[error("{0}")]
    FacetParseError(#[from] FacetParseError),
```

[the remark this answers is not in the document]
moriyoshi: Having a specific error enum for the facet parser totally makes sense to me.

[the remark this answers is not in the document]
moriyoshi: I just added an enum to facet as I wouldn't think extending `TantivyError` was necessary.

src/query/query_parser/query_parser.rs
```diff
@@ -68,6 +68,9 @@ pub enum QueryParserError {
     /// The format for the date field is not RFC 3339 compliant.
     #[error("The date field has an invalid format")]
     DateFormatError(chrono::ParseError),
+    /// The format for the facet field is invalid.
+    #[error("The facet field is malformed")]
```
PSeitz: I'm not sure, but I think this error should forward the original error message or the user would not see which part of the query failed

PSeitz: like this `#[error("{0}")]`

PSeitz (approved):
Thanks! Looks good!

src/schema/facet.rs
```diff
@@ -137,39 +181,7 @@ impl Borrow<str> for Facet {
 
 impl<'a, T: ?Sized + AsRef<str>> From<&'a T> for Facet {
     fn from(path_asref: &'a T) -> Facet {
-        #[derive(Copy, Clone)]
```
fulmicoton, line 140: We should probably remove the `From` implementation and possibly add a `TryFrom`.

moriyoshi: Sounds good. Is that necessary to do in this PR? or a separate PR is better?

fulmicoton: Can you do this in this PR?

moriyoshi: It just turned out that a generic TryFrom cannot be implemented in the way From is implemented here.  See the following thread for detail.  Actually I've succeeded with a specific implementation for `String` but I don't think it does ever have any use.

https://github.com/rust-lang/rust/issues/50133 (the reason is detailed in https://github.com/rust-lang/rust/issues/50133#issuecomment-488512355 and https://github.com/rust-lang/rust/issues/50133#issuecomment-722546899)

I would suggest keeping this as is now.

PSeitz: We should consider replacing TryFrom with a custom trait. Would be ok from my side to merge it like that.

fulmicoton: @PSeitz @moriyoshi ...

Merging. We cannot decenlty ask you to fix all our mistakes :). Thank you!

PR #1060, opened 2021-05-24
Change Footer version handling, Make compression dynamic 

### LINKED ISSUE #904: Dynamic compression format
Right now, tantivy relies on compile-time feature flags to decide which store compression format (lz4, snappy, brotli) should be used.
The store compression algorithm is serialized within VersionFooter in every file and the compatibility check ensures that a meaningful error message is emitted if a user tries to open an index using a compression scheme that is incompatible with the feature flag they used at compilation time.

Ideally, a user should be able to compile tantivy with several compression algorithm  to enable to define in the schema which compression format is used.

Some user want to keep tantivy as light as possible, or prefer to work with a "pure rust" project with no C bindings. (for instance when working with WASM)
For this reason, the compression algorithm would still be enabled or disabled at compile time.

The compression algorithm would be specified in the meta.json AND this compression should also be written in the store file.
We can keep the compression algorithm in the VersionFooter for a bit, but only for backward compatibility version.

On read, the compression algorithm will be decided by read a specific store file.
On write, the compression algorithm used will be decided by what is defined in the meta.json file.

Changing the compression algorithm of an existing index will be handled. In that case, old segment and new segment may use a different compression algorithm. On merge, the new compression algorithm should be used systematically.

Note for the implementer: An optimization exists when merging store file that do not contain any deleted docs. In that case, the store blocks just get stacked. Please make sure that this optimization is disabled if the compression scheme are different.

### LINKED ISSUE #1058: Change Footer versioning handling
We append `Footer` to every File in tantivy with some metadata, so when opening a file we can validate if the format can still be read.

### Current State

Currently `Footer` includes `Version` and `VersionedFooter` and a meta String, which is a String version of `Version`. Footer is de/serialized as bytes.
`VersionedFooter` includes crc, and the doc store compression format, which will be obsolete with https://github.com/tantivy-search/tantivy/issues/904. This would leave only crc in VersionedFooter. `VersionedFooter` is enabling versioned information in the `Footer`.

### Proposal

With JSON as metadata, we could simplify the Versioning by dropping `VersionedFooter` in `Footer` and moving crc to `Footer`.
The format could look as follows |magic_byte_u32 | footer_len_u32 | json serialized Version|

Footer {Version, Crc}

We can add  and remove Fields in Footer. The magic byte would be used to detect old versions of tantivy or corruption and returning an error. Also meta would be dropped, since it is duplicate information.

### Limitations

Tantivy versions <0.15 trying to read a index created by version >=0.15 would fail without a nice error.
Changing the type of a field is not supported and would cause JSON parsing to fail.

E.g. `{metaData: String}` to `{metaData: {moreData:String}}` is not supported. This can be mitigated either by using an enum with both types or using a different variable name. For that reason, we should also track the names of deleted properties to not reuse them.

We can remove only fields of type `Option` from the footer, or older versions would fail parsing. Non optional fields should be marked as obsolete 

### Open Question
Should we add a footer version for larger changes to the footer metadata, e.g. json to messagepack serialization

PSeitz:
Change Footer version handling
Simplify version handling by switching to JSON instead of binary serialization.
fixes #1058

Make compression dynamic
Instead of choosing the compression during compile time via a feature flag, you can now have multiple compression algorithms enabled and decide during runtime which one to choose via IndexSettings. Changing the compression algorithm on an index is also supported. The information which algorithm was used in the doc store is stored in the DocStoreFooter. The default is the lz4 block format.
fixes #904


src/common/serialize.rs
```diff
@@ -14,7 +14,20 @@ pub trait BinarySerializable: fmt::Debug + Sized {
     fn deserialize<R: Read>(reader: &mut R) -> io::Result<Self>;
 }
 
-/// `FixedSize` marks a `BinarySerializable` as
```
fulmicoton, line 17: Did you remove this comment line on purpose?

src/core/index_meta.rs
```diff
@@ -233,7 +233,10 @@ impl InnerSegmentMeta {
 pub struct IndexSettings {
     /// Sorts the documents by information
     /// provided in `IndexSortByField`
+    #[serde(skip_serializing_if = "Option::is_none")]
```
fulmicoton, line 236: yes

src/core/index_meta.rs
```diff
@@ -233,7 +233,10 @@ impl InnerSegmentMeta {
...
     pub sort_by_field: Option<IndexSortByField>,
+    /// The `Compressor` used to compress the doc store.
+    pub compressor: Compressor,
```
fulmicoton: considering we serialize this thing, 

`s/compressor/docstore_compression/g` ?

src/store/compressors.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+/// Compressor can be used on `IndexSettings` to choose
+/// the compressor used to compress the doc store.
+///
+/// The default is Lz4Block, but also depends on the enabled feature flags.
+#[derive(Clone, Debug, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum Compressor {
+    /// Use the lz4 block format compressor
+    Lz4Block,
```
fulmicoton: Can we serialize to "lz4-block", "lz4-frame", "brotli", "snappy" ?

src/store/compressors.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Use the lz4 frame format compressor
+    Lz4Frame,
```
fulmicoton: Is there a rationale behind having lz4frame AND lz4block?

If not could we just call it lz4 and pick the best one?

src/store/compressors.rs
```diff
@@ -0,0 +1,155 @@
...
+impl Default for Compressor {
+    fn default() -> Self {
+        if cfg!(feature = "lz4-block-compression") {
+            Compressor::Lz4Block
+        } else if cfg!(feature = "lz4-compression") {
+            Compressor::Lz4Frame
+        } else if cfg!(feature = "brotli-compression") {
+            Compressor::Brotli
+        } else if cfg!(feature = "snappy-compression") {
+            Compressor::Snap
```
fulmicoton: s/Snap/Snappy/g

(Snap is the name of the crate)

src/store/footer.rs
```diff
@@ -0,0 +1,62 @@
...
+                    file.len()
+                ),
+            ));
+        }
+        let (body, footer_slice) = file.split_from_end(DocStoreFooter::SIZE_IN_BYTES);
+        let mut footer_bytes = footer_slice.read_bytes()?;
+        let footer = DocStoreFooter::deserialize(&mut footer_bytes)?;
+        Ok((footer, body))
+    }
+}
```
fulmicoton, line 62: could we add unit tests?

To check the fixed size for instance...
The goal is not to find bugs, but to ensure we spot on CI if someone updates DocStoreFooter without updating FixedSize  and serialization.

src/store/mod.rs
```diff
@@ -33,66 +33,26 @@ and should rely on either
...
+mod compressors;
+mod footer;
 mod index;
 mod reader;
 mod writer;
+pub use self::compressors::Compressor;
 pub use self::reader::StoreReader;
 pub use self::writer::StoreWriter;
 
-// compile_error doesn't scale very well, enum like feature flags would be great to have in Rust
```
fulmicoton, line 42: that feels good

fulmicoton (changes requested):
I added some minor comments.

Mostly renaming.

src/common/serialize.rs
PSeitz: that was by accident

src/core/index_meta.rs
PSeitz: yes that's better

src/store/compressors.rs
PSeitz: I added the block format for backwards compatibility instead of replacing it. But since this is a breaking change anyway, I think we can just delete the lz4 frame stuff, since lz4 block is better imo in this case (less overhead and faster).

I think the removal with some ff renaming can be done in a separate PR

src/store/footer.rs
PSeitz: I added a small test, but for exact tests we would need some custom derives

PSeitz:
Should be fine now, except removing the lz-frame stuff

Edit also removed the lz4 frame stuff

src/common/serialize.rs
```diff
@@ -14,6 +14,20 @@ pub trait BinarySerializable: fmt::Debug + Sized {
     fn deserialize<R: Read>(reader: &mut R) -> io::Result<Self>;
 }
 
+pub trait DeserializeFrom<T: BinarySerializable> {
+    fn deserialize(&mut self) -> io::Result<T>;
+}
+
+/// Implement deserialize from &[u8] for all types which implement BinarySerializable.
+///
+/// TryFrom would actually be preferrable, but not possible because of the orphan
```
fulmicoton, line 23: I am not sure this trait makes the code any simpler.

Also if try_from what the target, why not move self instead of passing (&mut self) ?

src/directory/footer.rs
```diff
@@ -1,69 +1,45 @@
...
+use serde::{Deserialize, Serialize};
 use std::io;
 use std::io::Write;
 
-const FOOTER_MAX_LEN: usize = 10_000;
+const FOOTER_MAX_LEN: u32 = 50_000;
+
+/// The magic byte of the footer to identify corruption
+/// or an old version of the footer.
+const FOOTER_MAGIC_BYTE: u32 = 1337;
```
fulmicoton: that's not single byte. FOOTER_MAGIC_NUMBER?

src/store/mod.rs
```diff
@@ -216,6 +194,67 @@ pub mod tests {
         }
         Ok(())
     }
+
+    #[cfg(feature = "snap")]
+    #[cfg(feature = "lz4_flex")]
```
fulmicoton: I'm confused by the feature name in this file.
Aren't they supposed to be called respectively `{snappy, lz4, brotli}_compression` .

fulmicoton (changes requested):
Sorry I added a bunch of extra comments

src/directory/footer.rs
PSeitz: yes that's better

src/common/serialize.rs
PSeitz: It's used below, imo quite nice to use
```     
   let (footer_len, footer_magic_byte): (u32, u32) = file
            .slice_from_end(footer_metadata_len)
            .read_bytes()?
            .as_ref()
            .deserialize()?;
```

It's `&mut self` because the trait wants `&mut  Read`

with try_from it would be converting from `&[u8]` to `T` I think.

PR #1062, opened 2021-05-26
use bit packing in fieldnorm

PSeitz:


src/fieldnorm/serializer.rs
```diff
@@ -19,9 +24,23 @@ impl FieldNormsSerializer {
...
+    pub fn serialize_field(
+        &mut self,
+        field: Field,
+        fieldnorms_data: &[u8],
+        max_value: u8,
+    ) -> io::Result<()> {
         let write = self.composite_write.for_field(field);
-        write.write_all(fieldnorms_data)?;
+        max_value.serialize(write)?;
+        dbg!(max_value);
```
fulmicoton: we should probably remove the dbg

fulmicoton (commented):
The benefit is probably not fantastic: fieldnorms only take one byte per doc, and fieldnorm_id is called once per matched doc & term.

Do you observe no performance hit ? You can use the search benchmark for instance.
(Top10 + Count, type of query: "term").

The concern might feel ridiculous, be we currently spend around 7ns per doc today.
I would not be entirely surprise if tantivy is 10% slower on that query after this PR.

src/fieldnorm/serializer.rs
PSeitz: sounds reasonable

PSeitz:
There are a lot of fields with fieldsnorms where the max number of terms is quite small, e.g. a title field, where we can compress to ~25%. Here is the saving for wikipedia:

```
1432    wikipedia-index/4e8f182175ec4459912b19587268aa33.fieldnorm
1844    wikipedia-index/55b1b7aad942490f99cc54480c3d6ba0.fieldnorm
116     wikipedia-index/390ec451d46a49ef9d82f64cb0bfe53c.fieldnorm
11360   wikipedia-index/d31c548ca4434f0090c136cb744924e6.fieldnorm
14752   total


8524    wikipedia-index/5be1a883812c42e6b2ffeeff64c2950a.fieldnorm
892     wikipedia-index/00217d87e07a4774a6310ce4a04cb19d.fieldnorm
1228    wikipedia-index/358fd760bbc14347b76ae30eb292035b.fieldnorm
76      wikipedia-index/f118b1f8e2d04e2ca46260531b883f18.fieldnorm
10720   total
```
The concern is not ridiculous at all, there is some performance hit in TOP10_COUNT:


Query | tantivy-0.13 | tantivy-0.14 | tantivy-0.15 | tantivy-0.15-normcomp
-- | -- | -- | -- | --
AVERAGE | 4,684 μs | 3,964 μs | 3,947 μs | 4,178 μs

There was a inline missing in the bitpacker, these are the numbers **without inline**:
Query | tantivy-0.13 | tantivy-0.14 | tantivy-0.15 | tantivy-0.15-normcomp
-- | -- | -- | -- | --
AVERAGE | 4,633 μs | 3,978 μs | 3,958 μs | 4,414 μs

There is further potential for improvement in the bitpacker. Currently it works on u64, but we have only u8 values here.

Generally it would be best to able to choose what you want, compressed/uncompressed. On S3 compression is preferable, on a local ssd maybe not.

PSeitz:
Closing for now due to performance impact

PR #1072, opened 2021-06-04
Enable support for multiple fastfield codecs, add linear interpolation

### LINKED ISSUE #1042: Optimize fast field representation
Right now fast fields are only represented using bitpacking.

It would be great to improve their representation.

An important target would be sorted u64 fields, like:
- a u64 field if the index is sorted by it
- an index in a multivalued field.

There are different ways allowing for random access:
- Frame of reference
- Interpolation
- what else?

PSeitz:
This PR enables support for multiple fast field codec and adds linear interpolation as fast field compression.
Fast field codecs are moved to their own crate. This simplifies codec implementation by enforcing a minimal API and enables fast test feedback cycles.
Fast field serialization gets now access to all the data to be serialized at once, instead of single values.

A codec is identified by NAME and ID. Every codec has to implement an estimation to guess (or calculate) the compression ratio. Estimation should be fast, in the linear interpolation case some samples are taken for the estimation.

The fast field data now has a u8 prepended, which is the ID of the codec used. So total number of codecs is 256. (should be fine?). All data afterwards is handled by the fast field. (although a common statistics would make sense)

fixes #1042


fastfield_codecs/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "fastfield_codecs"
+version = "0.1.0"
```
fulmicoton, line 3: can you add the mit license, and a description

common/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "common"
+version = "0.1.0"
+authors = ["Pascal Seitz <pascal.seitz@gmail.com>"]
+edition = "2018"
```
fulmicoton, line 6: license and description.

common/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "common"
+version = "0.1.0"
+authors = ["Pascal Seitz <pascal.seitz@gmail.com>"]
```
fulmicoton: quickwit email please. It can be yours.

fastfield_codecs/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+authors = ["Pascal Seitz <pascal.seitz@gmail.com>"]
```
fulmicoton: quickwit.io email please.

src/fastfield/reader.rs
```diff
@@ -61,73 +65,100 @@ pub trait FastFieldReader<Item: FastValue>: Clone {
...
+    LinearInterpol(FastFieldReaderCodecWrapper<Item, LinearinterpolFastFieldReader>),
 }
 
 impl<Item: FastValue> DynamicFastFieldReader<Item> {
     /// Returns correct the reader wrapped in the `DynamicFastFieldReader` enum for the data.
     pub fn open(file: FileSlice) -> crate::Result<DynamicFastFieldReader<Item>> {
-        Ok(DynamicFastFieldReader::Bitpacked(
-            BitpackedFastFieldReader::open(file)?,
-        ))
+        let bytes = file.read_bytes()?;
```
fulmicoton: there is a read_u8 method in OwnedBytes.
```
let mut bytes = file.read_bytes();
let id = bytes.read_u8();
```

src/fastfield/reader.rs
```diff
@@ -61,73 +65,100 @@ pub trait FastFieldReader<Item: FastValue>: Clone {
...
-    bit_unpacker: BitUnpacker,
-    min_value_u64: u64,
-    max_value_u64: u64,
     _phantom: PhantomData<Item>,
 }
 
-impl<Item: FastValue> BitpackedFastFieldReader<Item> {
+impl<Item: FastValue, C: CodecReader> FastFieldReaderCodecWrapper<Item, C> {
     /// Opens a fast field given a file.
     pub fn open(file: FileSlice) -> crate::Result<Self> {
```
fulmicoton, line 159: When is this function useful? Is it for when we are sure of the encoding and we want static dispatch or something?

src/fastfield/reader.rs
```diff
@@ -61,73 +65,100 @@ pub trait FastFieldReader<Item: FastValue>: Clone {
...
         let mut bytes = file.read_bytes()?;
-        let min_value = u64::deserialize(&mut bytes)?;
-        let amplitude = u64::deserialize(&mut bytes)?;
-        let max_value = min_value + amplitude;
-        let num_bits = compute_num_bits(amplitude);
-        let bit_unpacker = BitUnpacker::new(num_bits);
-        Ok(BitpackedFastFieldReader {
+        let _id = u8::deserialize(&mut bytes)?;
```
fulmicoton: Should we at least defensively test the id?

src/fastfield/serializer/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+
+impl CompositeFastFieldSerializer {
+    /// Constructor
+    pub fn from_write(write: WritePtr) -> io::Result<CompositeFastFieldSerializer> {
+        // just making room for the pointer to header.
+        let composite_write = CompositeWrite::wrap(write);
+        Ok(CompositeFastFieldSerializer { composite_write })
+    }
+
+    /// Serialize data into a new u64 fast field. The compression will be detected automatically.
```
fulmicoton: s/the compression will be detected automaticlly/The best compression codec will be chosen automatically.

src/fastfield/serializer/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+        }
+        {
+            let (ratio, name, id) = (
+                LinearInterpolFastFieldSerializer::estimate(&fastfield_accessor, stats.clone()),
+                LinearInterpolFastFieldSerializer::NAME,
+                LinearInterpolFastFieldSerializer::ID,
+            );
+            estimations.push((ratio, name, id));
+        }
+        estimations.sort_by(|a, b| a.0.partial_cmp(&b.0).unwrap());
```
fulmicoton, line 104: Can we remove nan from the estimations vec before?

src/fastfield/serializer/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+        let (_ratio, name, id) = estimations[0];
```
fulmicoton, line 105: max_by also exists.

src/fastfield/serializer/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+        //estimations.sort_by_key(|el| el.0);
+        id.serialize(field_write)?;
+        match name {
```
fulmicoton, line 111: I'm curious, why did you choose to match on name rather than id?

fastfield_codecs/src/linearinterpol.rs
```diff
@@ -0,0 +1,279 @@
...
+impl FastFieldSerializerEstimate for LinearInterpolFastFieldSerializer {
+    /// estimation for linear interpolation is hard because, you don't know
+    /// where the local maxima are for the deviation of the calculated value and
+    /// the offset is also unknown.
+    fn estimate(fastfield_accessor: &impl FastFieldDataAccess, stats: FastFieldStats) -> f32 {
+        let first_val = fastfield_accessor.get(0);
+        let last_val = fastfield_accessor.get(stats.num_vals as u32 - 1);
+        let slope = get_slope(first_val, last_val, stats.num_vals);
+
+        // let's sample at 10%, 20%, 25%, 50%, 75%, 90%
```
fulmicoton: why these values? A uniform law would make more sense here I believe.

fastfield_codecs/src/linearinterpol.rs
```diff
@@ -0,0 +1,279 @@
...
+                let calculated_value = get_calculated_value(first_val, *pos as u64, slope);
+                let actual_value = fastfield_accessor.get(*pos as u32);
+                distance(calculated_value, actual_value)
+            })
+            .max()
+            .unwrap();
+
+        // the theory would be that we don't have the actual max_distance, but we are close within 50%
+        // threshold.
+        // It is multiplied by 2 because in a log case scenario the line would be as much above as
```
fulmicoton, line 216: I agree, but log isn't the bad scenario is it?

fulmicoton (changes requested):

src/fastfield/serializer/mod.rs
PSeitz: we need sort_by eventually anyway when the estimate needs to be big enough in difference to the other ones to justify a slower algorithm

fastfield_codecs/src/linearinterpol.rs
PSeitz: They are pretty much random. you mean e.g. 20 samples, every 5%?

src/fastfield/serializer/mod.rs
PSeitz: to log the name of the chosen algorithm eventually

fastfield_codecs/src/linearinterpol.rs
fulmicoton: yes or every 10 %

src/fastfield/serializer/mod.rs
fulmicoton: won't it be a max_by a  some function using  cost and benefit?

fastfield_codecs/src/linearinterpol.rs
```diff
@@ -0,0 +1,279 @@
...
+        footer.serialize(write)?;
+        Ok(())
+    }
+}
+fn get_slope(first_val: u64, last_val: u64, num_vals: u64) -> f64 {
+    (last_val as f64 - first_val as f64) / (num_vals as u64 - 1) as f64
+}
+
+fn get_calculated_value(first_val: u64, pos: u64, slope: f64) -> u64 {
+    first_val + (pos as f64 * slope) as u64
```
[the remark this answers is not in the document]
PSeitz: even with f64 it's surprisingly fast

```
running 4 tests
test tests::bench_fastfield_bitpack_create        ... bench:      30,234 ns/iter (+/- 1,360)
test tests::bench_fastfield_bitpack_get           ... bench:      28,106 ns/iter (+/- 767)
test tests::bench_fastfield_linearinterpol_create ... bench:     139,701 ns/iter (+/- 3,147)
test tests::bench_fastfield_linearinterpol_get    ... bench:      48,430 ns/iter (+/- 13,202)
```
f32 should be fine also, but with the interpolate_via_int_arithmetic I think we could run into out of bounds issues for very large values?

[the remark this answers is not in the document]
PSeitz: f32 precision:

```
running 4 tests
test tests::bench_fastfield_bitpack_create        ... bench:      31,102 ns/iter (+/- 1,242)
test tests::bench_fastfield_bitpack_get           ... bench:      28,628 ns/iter (+/- 874)
test tests::bench_fastfield_linearinterpol_create ... bench:     134,130 ns/iter (+/- 6,145)
test tests::bench_fastfield_linearinterpol_get    ... bench:      43,267 ns/iter (+/- 697)
```

fastfield_codecs/src/linearinterpol.rs
PSeitz: It's also not the best case. I think the estimation is reasonable for a lot of cases, but it could make sense to add a worst case estimation?

src/fastfield/reader.rs
PSeitz: there are some cases, where we always use bitpacked for now and there we can use static dispatch

src/fastfield/reader.rs
PSeitz: yep that's a good idea

src/fastfield/serializer/mod.rs
PSeitz: hmm yes probably

fastfield_codecs/README.md
```diff
@@ -0,0 +1,62 @@
+
+
+# Fast Field Codecs
+
+This crate contains various fast field codecs, used to compress/decompress fast field data in tantivy.
+
+## Contributing
+
+Contributing is pretty straightforward. Since the bitpacking is the simplest compressor, you can check it for reference.
```
fulmicoton, line 9: Awesome job.

fastfield_codecs/src/linearinterpol.rs
```diff
@@ -0,0 +1,292 @@
...
+        };
+        footer.serialize(write)?;
+        Ok(())
+    }
+    /// estimation for linear interpolation is hard because, you don't know
+    /// where the local maxima for the deviation of the calculated value are and
+    /// the offset to shift all values to >=0 is also unknown.
+    fn estimate(fastfield_accessor: &impl FastFieldDataAccess, stats: FastFieldStats) -> f32 {
+        if stats.num_vals < 3 {
+            return f32::MAX; //disable compressor for this case
```
fulmicoton: That's a scary way to disable a compressor. Could we think of another way? For instance introduce an extra method 
`.is_applicable()` or something like that.

fastfield_codecs/src/linearinterpol.rs
```diff
@@ -0,0 +1,292 @@
...
+#[inline]
+fn get_calculated_value(first_val: u64, pos: u64, slope: f32) -> u64 {
+    first_val + (pos as f32 * slope) as u64
+}
+
+impl FastFieldCodecSerializer for LinearInterpolFastFieldSerializer {
+    const NAME: &'static str = "LinearInterpol";
+    const ID: u8 = 2;
+    /// Creates a new fast field serializer.
+    fn create(
```
fulmicoton: This is not a very good name for a method.
`create`this is more about serializing than creating anything here.

fastfield_codecs/src/linearinterpol.rs
```diff
@@ -0,0 +1,292 @@
...
+
+        let max_distance = sample_positions
+            .iter()
+            .map(|pos| {
+                let calculated_value = get_calculated_value(first_val, *pos as u64, slope);
+                let actual_value = fastfield_accessor.get(*pos as u32);
+                distance(calculated_value, actual_value)
+            })
+            .max()
+            .unwrap();
```
fulmicoton: nitpick: 

You can unwrap here, but I would still rather have unwrap_or(0f32).
I think this is not a widely shared opinion.

Here is my logic:
I like to be able to grep against unwrap or expect. If we keep the list short, (ideally inexistent) auditing the code is simpler.

On the other hand, people like unwrap because on a perfect code base it is a way to tell the reader this case never happens.

fulmicoton (approved):
LGTM can you fix the `fn create` function name though?
Some other nitpicks to your discretion.

Great job!

fastfield_codecs/src/linearinterpol.rs
PSeitz: I thought about `.is_applicable()`, and discarded it initially to share potentially expensive computations between applicable and estimation, but on second thought it probably would be best to pass some cache around

fastfield_codecs/src/linearinterpol.rs
PSeitz: agreed, it's called `serialize` now

fastfield_codecs/src/linearinterpol.rs
PSeitz: is_applicable added

fastfield_codecs/src/linearinterpol.rs
PSeitz: unwrap_or is better here, but sometimes I prefer except with an explanation why it can't happen when the program would be in an invalid state afterwards

fastfield_codecs/README.md
```diff
@@ -0,0 +1,62 @@
+
+
+# Fast Field Codecs
+
+This crate contains various fast field codecs, used to compress/decompress fast field data in tantivy.
```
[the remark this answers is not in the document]
PSeitz: Thanks!

fastfield_codecs/benches/bench.rs
```diff
@@ -0,0 +1,108 @@
...
+        data.insert(1000, 2000);
+        data.insert(2000, 100);
+        data.insert(3000, 4100);
+        data.insert(4000, 100);
+        data.insert(5000, 800);
+        data
+    }
+
+    fn value_iter() -> impl Iterator<Item = u64> {
+        0..20_000
```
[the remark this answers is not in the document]
PSeitz: I'm not sure how it would mess up the bench, but I moved it into a vec

fastfield_codecs/benches/bench.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+    #[bench]
+    fn bench_fastfield_multilinearinterpol_get(b: &mut Bencher) {
+        let data: Vec<_> = get_data();
+        bench_get::<MultiLinearInterpolFastFieldSerializer, MultiLinearInterpolFastFieldReader>(
+            b, &data,
+        );
+    }
+    pub fn stats_from_vec(data: &[u64]) -> FastFieldStats {
+        let min_value = data.iter().cloned().min().unwrap_or(0);
```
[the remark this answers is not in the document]
PSeitz: that's cool, but the enum seems to be unwieldy for the test case

fastfield_codecs/src/bitpacked.rs
```diff
@@ -0,0 +1,170 @@
...
+            create_and_validate(&data, name);
+        }
+    }
+
+    #[test]
+    fn bitpacked_fast_field_rand() {
+        for _ in 0..500 {
+            let mut data = (0..1 + rand::random::<u8>() as usize)
+                .map(|_| rand::random::<i64>() as u64 / 2 as u64)
+                .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
PSeitz: Yes I agree proptest or similar would be better. I think this can be done in a separate PR or if an issue arises

PR #1091, opened 2021-06-17
Prepare TermInfoStore for lazy access

PSeitz:


src/termdict/fst_termdict/term_info_store.rs
```diff
@@ -120,31 +118,26 @@ fn extract_bits(data: &[u8], addr_bits: usize, num_bits: u8) -> u64 {
...
 impl TermInfoStore {
-    pub fn open(term_info_store_file: FileSlice) -> crate::Result<TermInfoStore> {
-        let (len_slice, main_slice) = term_info_store_file.split(16);
-        let mut bytes = len_slice.read_bytes()?;
-        let len = u64::deserialize(&mut bytes)? as usize;
-        let num_terms = u64::deserialize(&mut bytes)? as usize;
-        let (block_meta_file, term_info_file) = main_slice.split(len);
-        let term_info_bytes = term_info_file.read_bytes()?;
+    pub fn open(term_info_store_bytes: &[u8]) -> crate::Result<TermInfoStore> {
+        let mut footer_bytes = &term_info_store_bytes[term_info_store_bytes.len() - 16..];
```
fulmicoton, line 122: Why wouldn't we own a copy of the `OwnedBytes` in the TermInfoStore?
I think that would remove the argument in `get`, and make the code an the unit test a bit cleaner.

fulmicoton (commented):
Can we discuss offline the reason for this change? I am sure it will be all make sense with the next change, but I'd like to understand before approving.

PR #1094, opened 2021-06-22
Exposing min/max value interface on MultiValuedFastField Reader

evanxg852000:
Issue: as pointed out by @fmassot We want to avoid issues related to schema evolution in Quickwit. so currently fast fields are implicitly declared as multivalued.
We need the max/min value interface to be available for `MultiValuedFastFieldReader` as it is for  `FastFieldReader`
Currently, we need it to allow us to perform the timestamp filtering optimization. That is when a split time range is already within the user query bound, we should not attach filtering.

### DIFF
--- a/src/fastfield/multivalued/reader.rs
+++ b/src/fastfield/multivalued/reader.rs
@@ -46,6 +46,24 @@ impl<Item: FastValue> MultiValuedFastFieldReader<Item> {
         self.vals_reader.get_range_u64(range.start, &mut vals[..]);
     }
 
+    /// Returns the minimum value for this fast field.
+    ///
+    /// The max value does not take in account of possible
+    /// deleted document, and should be considered as an upper bound
+    /// of the actual maximum value.
+    pub fn min_value(&self) -> Item {
+        self.vals_reader.max_value()
+    }
+
+    /// Returns the maximum value for this fast field.
+    ///
+    /// The max value does not take in account of possible
+    /// deleted document, and should be considered as an upper bound
+    /// of the actual maximum value.
+    pub fn max_value(&self) -> Item {
+        self.vals_reader.min_value()
+    }
+
     /// Returns the number of values associated with the document `DocId`.
     pub fn num_vals(&self, doc: DocId) -> usize {
         let range = self.range(doc);


src/fastfield/multivalued/reader.rs
```diff
@@ -46,6 +46,24 @@ impl<Item: FastValue> MultiValuedFastFieldReader<Item> {
...
+    /// The max value does not take in account of possible
+    /// deleted document, and should be considered as an upper bound
+    /// of the actual maximum value.
+    pub fn min_value(&self) -> Item {
+        self.vals_reader.max_value()
+    }
+
+    /// Returns the maximum value for this fast field.
+    ///
+    /// The max value does not take in account of possible
```
shikhar, line 60: this comment seems like a copy-pasta

src/fastfield/multivalued/reader.rs
```diff
@@ -46,6 +46,24 @@ impl<Item: FastValue> MultiValuedFastFieldReader<Item> {
         self.vals_reader.get_range_u64(range.start, &mut vals[..]);
     }
 
+    /// Returns the minimum value for this fast field.
+    ///
+    /// The max value does not take in account of possible
```
shikhar, line 51: this comment seems like a copy-pasta

src/fastfield/multivalued/reader.rs
```diff
@@ -46,6 +46,24 @@ impl<Item: FastValue> MultiValuedFastFieldReader<Item> {
...
+    /// deleted document, and should be considered as an upper bound
+    /// of the actual maximum value.
+    pub fn min_value(&self) -> Item {
+        self.vals_reader.max_value()
```
shikhar, line 55: out-of-context, is it deliberate to have `min_value()` return `vals_reader.max_value()` and vice versa?

fulmicoton: Thanks for catching that. @evanxg852000 can you fix?

fmassot: good catch indeed, By the way, I was worried by the fact they were no test on that.

PSeitz:
> Issue: as pointed out by @fmassot We want to avoid issues related to schema evolution in Quickwit. so currently fast fields are implicitly declared as multivalued.

In that case we should check in tantivy if a field is actually multivalued on serialization and use single value if applicable to reduce index size and read performance.

src/fastfield/multivalued/reader.rs
evanxg852000: Yes the comment looked like a warning that need to be clear to users from the internal reader. That's why I felt it's necessary to also have it on this methods.

evanxg852000:
Let's move discussion to this [new pr](https://github.com/tantivy-search/tantivy/pull/1096)

PR #1097, opened 2021-06-23
Use dynamic fastfield codes for multivalues fixes #1093

### LINKED ISSUE #1093: Unlock the use of fast field codecs for multivalued fast field
Currently, we can use [3 different codecs](https://github.com/tantivy-search/tantivy/blob/cd049e28bc1d2b3b25e4eda35124e12d9f7428e9/src/fastfield/reader.rs#L68-L75) for single value fast field: `Bitpacked`, `LinearInterpol` and `MultiLinearInterpol`.

But this is not the case for `MultiValuedFastFieldReader` where the only available codec for the data is `BitpackedFastFieldReader` , see [code](https://github.com/tantivy-search/tantivy/blob/cd049e28bc1d2b3b25e4eda35124e12d9f7428e9/src/fastfield/multivalued/reader.rs#L17-L20).

The request is thus to replace the harcoded `BitpackedFastFieldReader` by the enum `DynamicFastFieldReader` for the `MultiValuedFastFieldReader`.

PSeitz:
Use dynamic fastfield codes for multivalues fixes (only sorting case covered)
Rename get to get_val due to conflict with Vec
use u64 precision instead of u32 for get_range, to allow use of existing fast field interface interface (actually not sure if it would be better have a different interface)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ rayon = "1.5"
 lru = "0.6.5"
 fastdivide = "0.3"
 itertools = "0.10.0"
-measure_time = "0.6.0"
+measure_time = "0.7.0"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.3.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ rayon = "1.5"
 lru = "0.6.5"
 fastdivide = "0.3"
 itertools = "0.10.0"
+measure_time = "0.6.0"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.3.9"
--- a/fastfield_codecs/src/lib.rs
+++ b/fastfield_codecs/src/lib.rs
@@ -51,14 +51,14 @@ pub trait FastFieldCodecSerializer {
 
 /// FastFieldDataAccess is the trait to access fast field data during serialization and estimation.
 pub trait FastFieldDataAccess {
-    /// Return the value associated to the given document.
+    /// Return the value associated to the given position.
     ///
     /// Whenever possible use the Iterator passed to the fastfield creation instead, for performance reasons.
     ///
     /// # Panics
     ///
-    /// May panic if `doc` is greater than the segment
-    fn get(&self, doc: u32) -> u64;
+    /// May panic if `position` is greater than the index.
+    fn get_val(&self, position: u64) -> u64;
 }
 
 #[derive(Debug, Clone)]
@@ -69,20 +69,20 @@ pub struct FastFieldStats {
 }
 
 impl<'a> FastFieldDataAccess for &'a [u64] {
-    fn get(&self, doc: u32) -> u64 {
-        self[doc as usize]
+    fn get_val(&self, position: u64) -> u64 {
+        self[position as usize]
     }
 }
 
 impl<'a> FastFieldDataAccess for &'a Vec<u64> {
-    fn get(&self, doc: u32) -> u64 {
-        self[doc as usize]
+    fn get_val(&self, position: u64) -> u64 {
+        self[position as usize]
     }
 }
 
 impl FastFieldDataAccess for Vec<u64> {
-    fn get(&self, doc: u32) -> u64 {
-        self[doc as usize]
+    fn get_val(&self, position: u64) -> u64 {
+        self[position as usize]
     }
 }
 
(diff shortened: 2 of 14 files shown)


src/core/segment_reader.rs
```diff
@@ -63,11 +62,9 @@ impl SegmentReader {
         self.max_doc
     }
 
-    /// Returns the number of documents.
+    /// Returns the number of alive documents.
```
fulmicoton, line 65: thanks

src/fastfield/multivalued/reader.rs
```diff
@@ -16,13 +14,13 @@ use crate::DocId;
 #[derive(Clone)]
 pub struct MultiValuedFastFieldReader<Item: FastValue> {
     idx_reader: DynamicFastFieldReader<u64>,
```
fulmicoton, line 16: ah ok so the most important one (the index) was already commited.

src/indexer/merger.rs
```diff
@@ -739,17 +797,68 @@ impl IndexMerger {
...
+            }
+            impl<'a> FastFieldDataAccess for SortedDocidMultiValueAccessProvider<'a> {
+                fn get_val(&self, pos: u64) -> u64 {
+                    // use the offsets index to find the doc_id which will contain the position.
+                    // the offsets are stricly increasing so we can do a simple search on it.
+                    let new_docid = self
+                        .offsets
+                        .iter()
+                        .position(|&offset| offset > pos)
+                        .unwrap()
```
fulmicoton: `expect` with some kind of message

"Pos is out of bound"... or something like that.

fulmicoton (approved):
LGTM except one comment

src/indexer/merger.rs
```diff
@@ -519,7 +574,7 @@ impl IndexMerger {
         fast_field_serializer: &mut CompositeFastFieldSerializer,
         doc_id_mapping: &Option<Vec<(DocId, SegmentReaderWithOrdinal)>>,
         reader_and_field_accessors: &[(&SegmentReader, T)],
-    ) -> crate::Result<()> {
+    ) -> crate::Result<Vec<u64>> {
```
fulmicoton, line 492: Can we document in the rustdoc what the return value is? (If I understood correctly it is the index.)

fastfield_codecs/src/lib.rs
```diff
@@ -69,20 +69,20 @@ pub struct FastFieldStats {
...
 
 impl<'a> FastFieldDataAccess for &'a [u64] {
-    fn get(&self, doc: u32) -> u64 {
-        self[doc as usize]
+    fn get_val(&self, position: u64) -> u64 {
+        self[position as usize]
     }
 }
 
 impl<'a> FastFieldDataAccess for &'a Vec<u64> {
```
fulmicoton: I suspect this one is useless...

src/indexer/merger.rs
```diff
@@ -379,31 +379,86 @@ impl IndexMerger {
...
+                    let mut num_docs_so_far = 0;
+                    let reader_ordinal = self
+                        .segment_and_field_readers
+                        .iter()
+                        .position(|(segment_reader, _)| {
+                            num_docs_so_far += segment_reader.num_docs() as u64;
+
+                            num_docs_so_far > doc
+                        })
+                        .unwrap();
```
fulmicoton: expect with a decent error message pls

fastfield_codecs/src/lib.rs
PSeitz: because &[u64] exists? type deduction doesn't work in some case properly

fulmicoton: let's remove it...

Cargo.toml
```diff
@@ -53,6 +53,7 @@ rayon = "1.5"
 lru = "0.6.5"
 fastdivide = "0.3"
 itertools = "0.10.0"
+measure_time = "0.6.0"
```
fulmicoton: it is unused.

src/lib.rs
```diff
@@ -100,7 +100,8 @@
 
 #[cfg_attr(test, macro_use)]
 extern crate serde_json;
-
+#[macro_use]
+extern crate measure_time;
```
fulmicoton: it is unused.

fulmicoton (approved):

PR #1098, opened 2021-06-27
Fix delete documents with sort by field

appaquet:
Deleting documents in the same commit as they had been created in an index using the new `sort_by_field` was broken. Depending on the order of deletes, they may not have been applied.

The problem was that `doc_opstamps` returned by the segment writer's finalizer was in the original documents order as opposed to the `sort_by_field` order. Because of that, `compute_deleted_bitset` was discarding deletions because their opstamps were not correct.

### DIFF
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -1324,12 +1324,18 @@ mod tests {
         index_writer.commit().unwrap();
         index_reader.reload().unwrap();
 
-        // only non-deleted should be returned
         let searcher = index_reader.searcher();
-        let count = searcher
-            .search(&crate::query::AllQuery, &crate::collector::Count)
+        let results = searcher
+            .search(&crate::query::AllQuery, &TopDocs::with_limit(15))
             .unwrap();
-        assert_eq!(count, docs.len());
+
+        // only non-deleted should be returned
+        assert_eq!(results.len(), docs.len());
+        for (_, addr) in results {
+            let doc = searcher.doc(addr).unwrap();
+            let id = doc.get_first(id_field).unwrap().u64_value().unwrap();
+            assert!(docs.contains(&id));
+        }
     }
 
     #[test]
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -795,6 +795,7 @@ mod tests {
     use crate::Index;
     use crate::ReloadPolicy;
     use crate::Term;
+    use crate::{IndexSettings, IndexSortByField, Order};
 
     #[test]
     fn test_operations_group() {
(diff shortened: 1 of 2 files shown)


fulmicoton:
Again, thank you so much @appaquet !

src/indexer/index_writer.rs
```diff
@@ -1282,6 +1283,55 @@ mod tests {
...
+        }
+        for id in 5..10 {
+            index_writer.add_document(doc!(id_field => id));
+            docs.insert(id);
+        }
+
+        index_writer.commit().unwrap();
+        index_reader.reload().unwrap();
+
+        // only non-deleted should be returned
```
PSeitz: could you apply the comment as an actual assertion?

PSeitz (commented):
thanks @appaquet, really good catch. I just have added one comment about the test

src/indexer/index_writer.rs
appaquet: Updated. Let me know if it's what you had in mind !

src/indexer/segment_writer.rs
```diff
@@ -317,12 +318,13 @@ impl SegmentWriter {
 
 // This method is used as a trick to workaround the borrow checker
 fn write(
+    doc_opstamps: &Vec<Opstamp>,
```
fulmicoton, line 321: `&Vec<_>` is almost never a good idea. `&[Opstamp]` would have been better, but here I think we can pass `Vec<Opstamp>` and hence avoid the copy.

src/indexer/segment_writer.rs
```diff
@@ -356,13 +359,22 @@ fn write(
...
         for old_doc_id in doc_id_map.iter_old_doc_ids() {
+            remapped_opstamps.push(doc_opstamps[*old_doc_id as usize]);
             let doc_bytes = store_read.get_document_bytes(*old_doc_id)?;
             serializer.get_store_writer().store_bytes(&doc_bytes)?;
         }
-    }
+
+        remapped_opstamps
+    } else {
+        doc_opstamps.clone()
```
fulmicoton, line 372: (I was talking about that copy above.)

src/indexer/segment_writer.rs
```diff
@@ -356,13 +359,22 @@ fn write(
                 .segment()
                 .open_read(SegmentComponent::TempStore)?,
         )?;
+
+        let mut remapped_opstamps = Vec::with_capacity(doc_opstamps.len());
         for old_doc_id in doc_id_map.iter_old_doc_ids() {
+            remapped_opstamps.push(doc_opstamps[*old_doc_id as usize]);
```
fulmicoton, line 365: It would improve readability to separate the computation of the remapped_opstamps from writing the `store_reader`.

fulmicoton (commented):
I added some comments in the CR. No need to do any change.
I will address them myself and merge. 

Once again, thank you so much @appaquet

src/indexer/segment_writer.rs
appaquet: Yeah, my first iteration was to move the Vec, but I saw that write was also being used in a &self method.

src/indexer/segment_writer.rs
appaquet: But that would mean 2 iterations though ?

fulmicoton:
There is one bug remaining in apply_deletes. The max opstamp is assumed to be the last opstamp in the vec...

fulmicoton:
There is another bug in compute_deleted_bitset.. The simple skipping logic is not usable anymore.

src/indexer/segment_writer.rs
fulmicoton: Yes. I'd do that trade every single day.

src/indexer/segment_writer.rs
fulmicoton: Yes. I ended up doing a bit of refactoring.

src/indexer/index_writer.rs
PSeitz: yes exactly, thanks!

PSeitz:
Is this superseeded by https://github.com/tantivy-search/tantivy/pull/1099?

Issue #1089, opened 2021-06-16
Proptest/Fuzzytests for deletion


PSeitz:
# Proptesting Deletion

Following up to https://github.com/tantivy-search/tantivy/pull/1088 and https://github.com/tantivy-search/tantivy/pull/1076 we need to add proptests for deleting random elements on an index.

Deletion is often a special case with its own code path with an increased complexity. To cover against regressions we should create an index and delete random elements and check if the returned contents are respecting deletion. Ideally this would cover different features, like doc store, facets, fast fields and especially merging since they all have special logic for deletion. 

There are also some thresholds to consider e.g. the flush threshold for blocks in the doc store.

Issue #1100, opened 2021-06-30
More proptesting for interaction between deletes and index sorting.


fulmicoton:
Some proptesting was added to check interaction between deleted and index_sorting, 
but the coverage only tests which docs are deleted, on one sorting setting, and on a single commit.

We should cover all data structures here eventually. Facets, Multivaluefastfields, doc store etc, and different sort settings.

PR #1102, opened 2021-06-30
extend proptests for sorting and merge

PSeitz:
extend proptests for sorting and merge
test multivalue fastfields
#1100

### DIFF
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -781,6 +781,7 @@ impl Drop for IndexWriter {
 
 #[cfg(test)]
 mod tests {
+    use futures::executor::block_on;
     use proptest::prelude::*;
     use proptest::prop_oneof;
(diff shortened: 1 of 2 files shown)


src/indexer/index_writer.rs
```diff
@@ -1401,15 +1434,41 @@ mod tests {
                     .map(move |doc| ff_reader.get(doc))
             })
             .collect();
+        ids.sort();
```
fulmicoton: we should probably disable the sorted is not enabled? (that requires disabling it in the expected_ids function too)

PSeitz: the sort is required because the order is not kept with the merges. I guess a set would be better suited?

fulmicoton: ah I see. I guess that's ok.

PSeitz:
fixes #1089

PR #1108, opened 2021-07-02
move ownedbytes to own crate

### LINKED ISSUE #1106: Moved OwnedBytes to an external crate
(OwnedBytes would be useful in quickwit as well.)

PSeitz:
fixes #1106


ownedbytes/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+            self.as_slice()
+        };
+        write!(f, "OwnedBytes({:?}, len={})", bytes_truncated, self.len())
+    }
+}
+
+impl Deref for OwnedBytes {
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
```
fulmicoton, line 131: should we inline this guy too?

ownedbytes/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+                io::ErrorKind::UnexpectedEof,
+                "failed to fill whole buffer",
+            ));
+        }
+        Ok(())
+    }
+}
+
+impl AsRef<[u8]> for OwnedBytes {
+    fn as_ref(&self) -> &[u8] {
```
fulmicoton, line 179: inline?

fulmicoton (changes requested):
Do we need to add the inline I suggested? Youare the expert

fulmicoton (approved):
Do we need to add the inline I suggested? Youare the expert

ownedbytes/src/lib.rs
PSeitz: yes, good catch

ownedbytes/src/lib.rs
PSeitz: added

PR #1109, opened 2021-07-03
Handle field names with any characters with a known set of special 

fmassot:
…characters and an escape one for the query parser.

Special characters : ```+ ^ ` :  {  } " [ ] ( ) ~ ! \  * \s```.
Almost like Lucene but without `&& || -`.
The escape character is `\`.
I kept the restriction on the `-` which cannot be a first character of a field name. 
`-` can be considered as a special character as it is used by the language query, I did not put it in the special characters for retro compatibility.

- [x] update the query parser to parse such field names
- [x] update `is_valid_field_name` function to check only if field name has at least one character and does not start with `-`

### DIFF
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -1,21 +1,49 @@
 use super::user_input_ast::{UserInputAst, UserInputBound, UserInputLeaf, UserInputLiteral};
 use crate::Occur;
-use combine::parser::char::{char, digit, letter, space, spaces, string};
+use combine::parser::char::{char, digit, space, spaces, string};
+use combine::parser::range::{take_while, take_while1};
+use combine::parser::repeat::escaped;
 use combine::parser::Parser;
 use combine::{
     attempt, choice, eof, many, many1, one_of, optional, parser, satisfy, skip_many1, value,
 };
 use combine::{error::StringStreamError, parser::combinator::recognize};
+use once_cell::sync::Lazy;
+use regex::Regex;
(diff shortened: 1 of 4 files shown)


PSeitz:
Ah nice the coverage report I added yesterday. Seems to be helpful, but not completely accurate.

fmassot:
> Ah nice the coverage report I added yesterday. Seems to be helpful, but not completely accurate.

very nice indeed, would be nice to add it to quickwit.

query-grammar/src/query_grammar.rs
```diff
@@ -1,21 +1,49 @@
...
+
+// Note: '-' char is only forbidden at the beginning of a field name, would be clearer to add it to special characters.
+const SPECIAL_CHARS: &[char] = &[
+    '+', '^', '`', ':', '{', '}', '"', '[', ']', '(', ')', '~', '!', '\\', '*', ' ',
+];
+const ESCAPED_SPECIAL_CHARS_PATTERN: &str = r#"\\(\+|\^|`|:|\{|\}|"|\[|\]|\(|\)|\~|!|\\|\*| )"#;
+
+/// Parses a field_name
+/// A field name must have at least one character and be followed by a colon.
+/// All characters are allowed except special characters `SPECIAL_CHARS`
```
fulmicoton: including special characters, but these needs to be escaped with escape character `\`.

fulmicoton (approved):
(I edited a comment and simplified the code a tiny bit)

PR #1110, opened 2021-07-05
Slop support for phrase queries

halvorboe:
Implements slop for phrase queries (https://github.com/tantivy-search/tantivy/issues/1068).

Currently the phrase query only supports exact matching. The algorithm used does a set intersection between the position of the last term in the phrase and the positions of all the prior terms in the phrase offset by their position in the phrase.

Looked at the Lucene implementation and the description in "Information Retrieval: Implementing and evaluating search engines". 

I modified the current algorithm slightly to keep track of:
- The start of each candidate phrase. Stored in a separate array.
- The first possible end of the phrase. (searching for `a b .` in `a c b d e` it stores the 0 as the start and 1 as the end). There is never any benefit from picking a later instance of the term. This is stored in what was previously called left.

I'll test the efficiency and try to optimize, but the is no change in complexity.

### DIFF
--- a/src/query/phrase_query/phrase_query.rs
+++ b/src/query/phrase_query/phrase_query.rs
@@ -26,6 +26,7 @@ use crate::schema::{Field, Term};
 pub struct PhraseQuery {
     field: Field,
     phrase_terms: Vec<(usize, Term)>,
+    slop: u32,
 }
 
 impl PhraseQuery {
@@ -56,7 +57,13 @@ impl PhraseQuery {
         PhraseQuery {
             field,
             phrase_terms: terms,
+            slop: 0,
         }
+    }
+
+    /// Slop allowed for the phrase.
+    pub fn slop(&mut self, value: u32) {
+        self.slop = value;
     }
 
     /// The `Field` this `PhraseQuery` is targeting.
@@ -97,11 +104,11 @@ impl PhraseQuery {
         }
         let terms = self.phrase_terms();
         let bm25_weight = Bm25Weight::for_terms(searcher, &terms)?;
-        Ok(PhraseWeight::new(
-            self.phrase_terms.clone(),
-            bm25_weight,
-            scoring_enabled,
-        ))
+        let mut weight = PhraseWeight::new(self.phrase_terms.clone(), bm25_weight, scoring_enabled);
+        if self.slop > 0 {
+            weight.slop(self.slop);
+        }
+        Ok(weight)
     }
 }
 
--- a/src/query/phrase_query/phrase_scorer.rs
+++ b/src/query/phrase_query/phrase_scorer.rs
@@ -170,6 +170,7 @@ impl<TPostings: Postings> PhraseScorer<TPostings> {
     }
 
     fn phrase_match(&mut self) -> bool {
+        // Need to add support for slop in phrase_count and phrase_exists
         if self.scoring_enabled {
             let count = self.compute_phrase_count();
             self.phrase_count = count;
--- a/src/query/phrase_query/phrase_scorer.rs
+++ b/src/query/phrase_query/phrase_scorer.rs
@@ -109,6 +109,7 @@ fn intersection(left: &mut [u32], right: &[u32]) -> usize {
     let mut count = 0;
     let left_len = left.len();
     let right_len = right.len();
+    // TODO: Modify this logic to add fuzziness to the logic.
     while left_i < left_len && right_i < right_len {
         let left_val = left[left_i];
         let right_val = right[right_i];
@@ -170,7 +171,7 @@ impl<TPostings: Postings> PhraseScorer<TPostings> {
     }
 
     fn phrase_match(&mut self) -> bool {
-        // Need to add support for slop in phrase_count and phrase_exists
+        // Need to add support for slop in phrase_count and phrase_exists.
         if self.scoring_enabled {
             let count = self.compute_phrase_count();
             self.phrase_count = count;
--- a/src/query/phrase_query/phrase_scorer.rs
+++ b/src/query/phrase_query/phrase_scorer.rs
@@ -51,10 +51,11 @@ pub struct PhraseScorer<TPostings: Postings> {
     fieldnorm_reader: FieldNormReader,
     similarity_weight: Bm25Weight,
     scoring_enabled: bool,
+    slop: u32,
 }
 
 /// Returns true iff the two sorted array contain a common element
-fn intersection_exists(left: &[u32], right: &[u32]) -> bool {
+fn intersection_exists(left: &[u32], right: &[u32], slop: u32) -> bool {
     let mut left_i = 0;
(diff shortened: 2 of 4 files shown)


fulmicoton:
I did not understand the description. Is this really about slop queries?

Slop queries are about relaxing the constraints of phrase query matching.
For instance, with a slop of 1, the query "my bag" matches on a document that contains "my favorite bag".

halvorboe:
@fulmicoton I'm still working on the description. 

It's the same logic as here right?
https://lucene.apache.org/core/7_4_0/core/org/apache/lucene/search/spans/SpanNearQuery.html
https://github.com/apache/lucene/blob/main/lucene/queries/src/java/org/apache/lucene/queries/spans/SpanNearQuery.java
https://github.com/apache/lucene/blob/main/lucene/queries/src/java/org/apache/lucene/queries/spans/NearSpansOrdered.java

The exact definition I'm working off is:
Given a slop value, a phrase matches the query if the sequence from the phrase appears within `length of the phrase + slop` token positions.

If ~ marks slop:
`a b b c d a c d`
`"c d"` -> 2 matches
`"a d"` -> 0 matches
`"a d"~1` -> 1 matches (`a c d`) 
`"a d"~3` -> 2 matches (`a b b c d`, `a c d`)
`"a d"~100` -> 3 matches (`a b b c d`, `a c d`, `a b b c d a c d`)

The challenges I'm currently thinking about:
1. ~~Keeping track of both where _all_ the phrases ending at the same token started. Exact matching of a phrase does not allow for multiple matches to end on the same token position, while slop matching does. Also if we want to know how much of the slop was _used_ for the match we can use the start and compare it against the end.~~
2. How is the score for a match calculated? In the case above `a c d` should have a higher score than `a b b c d` when both are a match.
3.  ~~What are the edge cases? For example a document with 1 million a b a b and then searching for "a b"~1000000.~~

src/query/phrase_query/phrase_weight.rs
```diff
@@ -77,8 +80,13 @@ impl PhraseWeight {
             similarity_weight,
             fieldnorm_reader,
             self.scoring_enabled,
+            self.slop,
         )))
     }
+
+    pub fn slop(&mut self, slop: u32) {
```
fulmicoton, line 87: Rename set_slop

src/query/phrase_query/phrase_query.rs
```diff
@@ -56,9 +57,15 @@ impl PhraseQuery {
         PhraseQuery {
             field,
             phrase_terms: terms,
+            slop: 0,
         }
     }
 
+    /// Slop allowed for the phrase.
+    pub fn slop(&mut self, value: u32) {
```
fulmicoton, line 65: set_slop

src/query/phrase_query/phrase_query.rs
```diff
@@ -56,9 +57,15 @@ impl PhraseQuery {
         PhraseQuery {
             field,
             phrase_terms: terms,
+            slop: 0,
         }
     }
 
+    /// Slop allowed for the phrase.
```
fulmicoton, line 64: Please document the behavior, with examples and an example where this is useful.

For instance "java developer"~1 matching "java software developer"

fulmicoton: You need to document the "slop keeps the order", and also the slop is "not an overall budget" specification.

In other words
"senior java developer"~1 does match "senior level distributed java software developer". 
The slop is not a budget that is spread amongst the different tokens. It is a maximum quantity that can be applied on every positions

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -130,12 +131,82 @@ fn intersection(left: &mut [u32], right: &[u32]) -> usize {
     count
 }
 
+/// Intersect twos sorted arrays `left` and `right` and outputs the
+/// resulting array in left.
+///
+/// Condition for match is that the value stored in left is less than the value in right,
+/// and that the difference between the value in left_begin is within
```
fulmicoton, line 138: I think your comment was truncated here.

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -130,12 +131,82 @@ fn intersection(left: &mut [u32], right: &[u32]) -> usize {
...
+///
+/// Returns the length of the intersection
+fn intersection_with_distance(
+    left: &mut [u32],
+    left_begin: &mut [u32],
+    right: &mut [u32],
+    max_distance_to_begin: u32,
+) -> usize {
+    // TODO: Improve variable names?
```
fulmicoton, line 147: I think the names are ok.

fulmicoton (commented):
Sorry for the late review.

I skimmed through it. Overall it looks good but I did not go into the detail.
Let me ask @PSeitz  to review that in depth.

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -130,12 +131,82 @@ fn intersection(left: &mut [u32], right: &[u32]) -> usize {
...
+    let mut left_i = 0;
+    let mut right_i = 0;
```
PSeitz, line 149: left_index, right_index would be better imo

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -170,6 +242,7 @@ impl<TPostings: Postings> PhraseScorer<TPostings> {
     }
 
     fn phrase_match(&mut self) -> bool {
+        // Need to add support for slop in phrase_count and phrase_exists.
```
PSeitz, line 245: outdated comment?

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -130,12 +131,82 @@ fn intersection(left: &mut [u32], right: &[u32]) -> usize {
...
+/// resulting array in left.
+///
+/// Condition for match is that the value stored in left is less than the value in right,
+/// and that the difference between the value in left_begin is within
+///
+/// Returns the length of the intersection
+fn intersection_with_distance(
+    left: &mut [u32],
+    left_begin: &mut [u32],
+    right: &mut [u32],
```
PSeitz, line 144: would be good to add a comment about the content of left and right

edko99:
Is there any way to help completing this feature and having it merged?

halvorboe:
I'm looking into it this weekend.

fulmicoton:
@halvorboe Ping!

halvorboe:
Sorry, I've not had any capacity. Going back to work on the PR now.

fulmicoton:
Do you need help to land this PR? @halvorboe 

Also if you are interested we have created a discord channel. I think you are one of the oldest user of tantivy. :) 
https://discord.gg/MT27AG5EVE

halvorboe:
Hey @fulmicoton. Sorry again. I'm going to spend a few hours on this

PR #1113, opened 2021-07-16
stale comments in segment_reader.rs

shikhar:


fmassot:
nice catch, thanks again @shikhar !

PR #1114, opened 2021-07-16
FilterCollector doc fix

shikhar:
Other types supported since https://github.com/tantivy-search/tantivy/pull/953/files


fmassot:
thanks @shikhar!

PR #1129, opened 2021-08-02
add long running test in ci

PSeitz:


PSeitz:
just waiting for tests to finish, to see it exits before 6h to not count as a failure

fulmicoton:
How long does it run?

PSeitz:
I don't know how fast the machine is, I think maybe around 3hours. But could be very different for increasing index sizes and dev.

PR #1130, opened 2021-08-02
Checksum validation using active files

### LINKED ISSUE #1127: validate_checksum should target live files, not managed files
Right now validate_checksum checks all managed files.

It should only work on the files that are actually part of the index.

evanxg852000:
added small changes to make sure checksum validation only checks active files instead of all managed files

@fulmicoton on https://github.com/tantivy-search/tantivy/blob/main/src/core/index_meta.rs#L139
Maybe we could change the api to only return `Option<PathBuf>`  that is if `delete_opstamp().is_some()  `xxx.del` files ?
To avoid listing a file that may not exist.

Closes https://github.com/tantivy-search/tantivy/issues/1127

### DIFF
--- a/src/core/index_meta.rs
+++ b/src/core/index_meta.rs
@@ -112,10 +112,16 @@ impl SegmentMeta {
             .load(std::sync::atomic::Ordering::Relaxed)
         {
             SegmentComponent::iterator()
+                .filter(|comp| {
+                    !(*comp == &SegmentComponent::Delete && self.delete_opstamp().is_none())
+                })
                 .map(|component| self.relative_path(*component))
                 .collect::<HashSet<PathBuf>>()
         } else {
             SegmentComponent::iterator()
+                .filter(|comp| {
+                    !(*comp == &SegmentComponent::Delete && self.delete_opstamp().is_none())
+                })
                 .filter(|comp| *comp != &SegmentComponent::TempStore)
                 .map(|component| self.relative_path(*component))
                 .collect::<HashSet<PathBuf>>()
--- a/src/core/index_meta.rs
+++ b/src/core/index_meta.rs
@@ -101,6 +101,7 @@ impl SegmentMeta {
 
     /// Returns the list of files that
     /// are required for the segment meta.
+    /// Note: Some of the returned files may not exist depending on the state of the segment.
     ///
     /// This is useful as the way tantivy removes files
     /// is by removing all files that have been created by tantivy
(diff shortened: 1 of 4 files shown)


src/core/index_meta.rs
```diff
@@ -112,10 +112,16 @@ impl SegmentMeta {
             .load(std::sync::atomic::Ordering::Relaxed)
```
PSeitz: instead of `list_files`, it could make sense to have `list_active_files` and `list_all_files`

src/core/index_meta.rs
```diff
@@ -112,10 +112,16 @@ impl SegmentMeta {
...
         {
             SegmentComponent::iterator()
+                .filter(|comp| {
```
fulmicoton: Yeah as you spotted, the comment is not accurate.

It contains a superset of the files that are useful for the index.
Some of the file in the returned list may not exist.

You spotted the delete file, but it is not the only one.
One thing you can do is to use the list of managed files and do
`managed files - list_files()`

src/directory/managed_directory.rs
```diff
@@ -267,6 +261,17 @@ impl ManagedDirectory {
         }
         Ok(damaged_files)
     }
+
+    /// List all managed files
+    pub fn list_files(&self) -> HashSet<PathBuf> {
```
fulmicoton: rename list_managed_files

src/directory/managed_directory.rs
```diff
@@ -248,15 +248,9 @@ impl ManagedDirectory {
         Ok(footer.crc() == crc)
     }
 
-    /// List files for which checksum does not match content
+    /// List managed files for which checksum does not match content
     pub fn list_damaged(&self) -> result::Result<HashSet<PathBuf>, OpenReadError> {
```
fulmicoton: sorry this is not part of your PR, but can you  rename it list_damaged_files ..

src/core/index_meta.rs
```diff
@@ -101,6 +101,7 @@ impl SegmentMeta {
 
     /// Returns the list of files that
     /// are required for the segment meta.
+    /// Note: Some of the returned files may not exist depending on the state of the segment.
```
fulmicoton, line 104: awesome

fulmicoton (approved):
Could you rename two functions?
Then we should be good to go!

src/directory/managed_directory.rs
```diff
@@ -248,15 +248,9 @@ impl ManagedDirectory {
...
-    pub fn list_damaged(&self) -> result::Result<HashSet<PathBuf>, OpenReadError> {
-        let mut managed_paths = self
-            .meta_informations
-            .read()
-            .expect("Managed directory rlock poisoned in list damaged.")
-            .managed_paths
-            .clone();
-
+    /// List managed files for which checksum does not match content
+    pub fn list_damaged_files(&self) -> result::Result<HashSet<PathBuf>, OpenReadError> {
```
fulmicoton: I think we can remove this method now.

PR #1135, opened 2021-08-13
add Cargo.lock

DavHau:
For reproducible builds, the repo should contain a Cargo.lock


fulmicoton:
This is a bad idea for a library.

PR #1136, opened 2021-08-13
more docs detail

PSeitz:
remove code duplicate


PSeitz:
I think `INDEXED` is an too ambiguous name and should be deprecated in favor of `SEARCHABLE`

PR #1140, opened 2021-08-19
rename common to tantivy-common

PSeitz:


PSeitz:
https://github.com/tantivy-search/tantivy/issues/1139

PR #1144, opened 2021-08-20
Resolve some clippys, format

sigaloid:
Lots of f32->f64 comparisons need to be resolved as well, an error margin should be determined for comparison.

Passes tests


fulmicoton (approved):
Thank you!

PR #1148, opened 2021-09-02
Fix incorrect comment for Index::create_in_dir

k-yomo:
The original comment says, it'll destroy existing meta file but the implementation is returning error when meta file exists.
https://github.com/tantivy-search/tantivy/blob/dd81e38e53343b08bea3315c8b0ea626a2cd15fc/src/core/index.rs#L125-L129

### DIFF
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -120,7 +120,7 @@ impl IndexBuilder {
     /// Creates a new index in a given filepath.
     /// The index will use the `MMapDirectory`.
     ///
-    /// If a previous index was in this directory, it returns error.
+    /// If a previous index was in this directory, it returns an `IndexAlreadyExists` error.
     #[cfg(feature = "mmap")]
     pub fn create_in_dir<P: AsRef<Path>>(self, directory_path: P) -> crate::Result<Index> {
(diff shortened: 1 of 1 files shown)


src/core/index.rs
```diff
@@ -120,7 +120,7 @@ impl IndexBuilder {
     /// Creates a new index in a given filepath.
     /// The index will use the `MMapDirectory`.
     ///
-    /// If a previous index was in this directory, then its meta file will be destroyed.
+    /// If a previous index was in this directory, it returns error.
```
```suggestion
    /// If a previous index was in this directory, it returns an `IndexAlreadyExists` error.
```

src/core/index.rs
```diff
@@ -238,7 +238,7 @@ impl Index {
     /// Creates a new index in a given filepath.
     /// The index will use the `MMapDirectory`.
     ///
-    /// If a previous index was in this directory, then its meta file will be destroyed.
+    /// If a previous index was in this directory, then it returns error.
```
```suggestion
    /// If a previous index was in this directory, then it returns  an `IndexAlreadyExists` error.
```

fulmicoton:
Thank you!

PR #1154, opened 2021-09-15
add DeleteBitSet iterator

PSeitz:


fulmicoton:
Is this CR making something faster?

src/indexer/merger.rs
```diff
@@ -98,27 +98,25 @@ pub struct IndexMerger {
 
 fn compute_min_max_val(
     u64_reader: &impl FastFieldReader<u64>,
+    segment_reader: &SegmentReader,
     max_doc: DocId,
     delete_bitset_opt: Option<&DeleteBitSet>,
```
fulmicoton: why do we keep delete_bitset_opt then?

src/core/segment_reader.rs
```diff
@@ -287,8 +287,19 @@ impl SegmentReader {
...
 
     /// Returns an iterator that will iterate over the alive document ids
-    pub fn doc_ids_alive(&self) -> impl Iterator<Item = DocId> + '_ {
-        (0u32..self.max_doc).filter(move |doc| !self.is_deleted(*doc))
+    pub fn doc_ids_alive(&self) -> Box<dyn Iterator<Item = DocId> + '_> {
+        if let Some(delete_bitset) = &self.delete_bitset_opt {
+            Box::new(
+                delete_bitset
+                    .iter()
+                    .take(self.max_doc() as usize)
```
fulmicoton: why is the take required?

src/core/segment_reader.rs
```diff
@@ -287,8 +287,19 @@ impl SegmentReader {
     }
 
     /// Returns an iterator that will iterate over the alive document ids
-    pub fn doc_ids_alive(&self) -> impl Iterator<Item = DocId> + '_ {
-        (0u32..self.max_doc).filter(move |doc| !self.is_deleted(*doc))
+    pub fn doc_ids_alive(&self) -> Box<dyn Iterator<Item = DocId> + '_> {
+        if let Some(delete_bitset) = &self.delete_bitset_opt {
+            Box::new(
+                delete_bitset
+                    .iter()
```
fulmicoton: If it is a set, can't we directly have semantics where we iterate the doc it contains?
I'm ok with renaming it "AliveDocSet" and inverting its semantics.

Iterating over `true, false` feels inefficient.

common/src/bitset.rs
```diff
@@ -28,6 +31,16 @@ impl IntoIterator for TinySet {
 }
 
 impl TinySet {
+    pub fn serialize(&self, writer: &mut dyn Write) -> io::Result<()> {
+        writer.write_all(self.0.to_le_bytes().as_ref())
```
fulmicoton, line 36: `WritePtr` has the merit of removing the dynamic dispatch.

common/src/bitset.rs
```diff
@@ -132,6 +149,66 @@ fn num_buckets(max_val: u32) -> u32 {
...
+            tinyset.serialize(writer)?;
+        }
+        writer.flush()?;
+        Ok(())
+    }
+
+    /// Deserialize a `BitSet`.
+    ///
+    pub fn deserialize(mut data: &[u8]) -> io::Result<Self> {
+        let max_value: u32 = u32::from_le_bytes(data[..4].try_into().unwrap());
```
fulmicoton, line 197: Do we actually need deserialize?

common/src/bitset.rs
```diff
@@ -132,6 +149,66 @@ fn num_buckets(max_val: u32) -> u32 {
...
+            len,
+            max_value,
+        })
+    }
+
+    /// Iterate the tinyset on the fly from serialized data.
+    ///
+    #[inline]
+    pub fn iter_from_bytes<'a>(data: &'a [u8]) -> impl Iterator<Item = TinySet> + 'a {
+        data[4..].chunks_exact(8).map(move |chunk| {
```
fulmicoton: Maybe we should assert that `data.len()` is a multiple of 8?

common/src/bitset.rs
```diff
@@ -132,6 +149,66 @@ fn num_buckets(max_val: u32) -> u32 {
...
+        let max_val: u32 = u32::from_le_bytes(data[..4].try_into().unwrap());
+        Self::iter_from_bytes(data)
+            .map(|tinyset| tinyset.complement())
+            .enumerate()
+            .flat_map(move |(chunk_num, tinyset)| {
+                let chunk_base_val = chunk_num as u32 * 64;
+                tinyset
+                    .into_iter()
+                    .map(move |val| val + chunk_base_val)
+                    .take_while(move |doc| *doc < max_val)
```
fulmicoton: With padding we will break backward compatibility anyway....
I suggest we inverse the semantic of bit = 1 to is_alive and remove the `complement` and `take_while`

common/src/bitset.rs
```diff
@@ -132,6 +149,66 @@ fn num_buckets(max_val: u32) -> u32 {
...
+            })
+    }
+
     /// Create a new `BitSet` that may contain elements
     /// within `[0, max_val[`.
```
fulmicoton: We use this notation in France too, but I think globally people use `[0, max_val)`.

common/src/bitset.rs
```diff
@@ -168,6 +245,15 @@ impl BitSet {
         };
     }
 
+    /// Returns true iff the elements is in the `BitSet`.
+    #[inline]
+    pub fn contains_from_bytes(el: u32, data: &[u8]) -> bool {
+        let byte_offset = 4 + el / 8u32;
+        let b: u8 = data[byte_offset as usize];
+        let shift = (el & 7u32) as u8;
```
fulmicoton: nitpick: `% 8` might be less cryptic than `& 7`

common/src/bitset.rs
```diff
@@ -132,6 +149,66 @@ fn num_buckets(max_val: u32) -> u32 {
...
+            tinysets: tinysets.into_boxed_slice(),
+            len,
+            max_value,
+        })
+    }
+
+    /// Iterate the tinyset on the fly from serialized data.
+    ///
+    #[inline]
+    pub fn iter_from_bytes<'a>(data: &'a [u8]) -> impl Iterator<Item = TinySet> + 'a {
```
fulmicoton: The function name should contain "tinyset". As a reader, I expect this function to iterate over docs.

src/fastfield/delete.rs
```diff
@@ -67,29 +45,34 @@ impl DeleteBitSet {
     /// Opens a delete bitset given its file.
     pub fn open(file: FileSlice) -> crate::Result<DeleteBitSet> {
         let bytes = file.read_bytes()?;
-        let num_deleted: usize = bytes
-            .as_slice()
-            .iter()
-            .map(|b| b.count_ones() as usize)
+        let num_deleted = BitSet::iter_from_bytes(bytes.as_slice())
```
fulmicoton: Can we make a `iter_from_bytes` private and introduce a public `len_from_bytes` method?

The prototype of `iter_from_bytes` is a bit too complex for a public API.

src/indexer/merger.rs
fulmicoton: Can you address this comment?

fulmicoton: (and same question for max_doc)

src/indexer/merger.rs
```diff
@@ -326,7 +323,7 @@ impl IndexMerger {
                 .fast_fields()
                 .typed_fast_field_reader(field)
                 .expect("Failed to find a reader for single fast field. This is a tantivy bug and it should never happen.");
-                compute_min_max_val(&u64_reader, reader.max_doc(), reader.delete_bitset())
+                compute_min_max_val(&u64_reader, reader, reader.max_doc(), reader.delete_bitset())
```
fulmicoton: remove reader.max_doc() and reader.delete_bitset()

fulmicoton (changes requested):

src/indexer/merger.rs
PSeitz: fixed

src/core/segment_reader.rs
PSeitz: It's now part of the iterator and not the api

Each block contains 64 elements, but we only want to iterate until max_doc, since the last elements will be emitted as alive

src/core/segment_reader.rs
PSeitz: I think this is addressed in the new version with `unset_iter`

common/src/bitset.rs
PSeitz: good idea

common/src/bitset.rs
PSeitz: It's not used currently. I can remove it, but it's tested and may be useful later. Also kinda asserts the format of serialize via the test.

fulmicoton: `#[cfg(test)]` then

src/indexer/merger.rs
PSeitz: done

common/src/bitset.rs
PSeitz: I can also complement before serialization. 

The alternative use 1 in the bitset internally could have some performance implications, since all methods like insert, pop_lowest etc. would also need to be flipped and the operations may be more expensive.

fulmicoton: I don't think this is true but maybe I am missing something.

My point is:  let's not have a deletebitset but an alive bitset.
Initialization is different. Marking a doc as deleted has the same cost.
pop_lowest does not change...

common/src/bitset.rs
PSeitz: I just copied it, in Germany both notations are used, but [0, max_val) is indeed more common

fulmicoton (changes requested):
See comments

common/src/bitset.rs
PSeitz: actually we can use static dispatch here

src/fastfield/alive_bitset.rs
```diff
@@ -0,0 +1,200 @@
...
+        write_alive_bitset(&bitset, &mut wrt).unwrap();
+        wrt.terminate().unwrap();
+        let file = directory.open_read(path).unwrap();
+        Self::open(file).unwrap()
+    }
+
+    /// Opens a delete bitset given its file.
+    pub fn open(file: FileSlice) -> crate::Result<AliveBitSet> {
+        let bytes = file.read_bytes()?;
+        let num_deleted = BitSet::count_unset_from_bytes(bytes.as_slice());
```
fulmicoton: This is incorrect. You are including the padding here.

src/fastfield/alive_bitset.rs
```diff
@@ -0,0 +1,200 @@
...
+    #[inline]
+    pub fn is_alive(&self, doc: DocId) -> bool {
+        !self.is_deleted(doc)
+    }
+
+    /// Returns true iff the document has been marked as deleted.
+    #[inline]
+    pub fn is_deleted(&self, doc: DocId) -> bool {
+        let data = self.data.as_slice();
+        !BitSet::contains_from_bytes(doc, data)
```
fulmicoton: the logic between is_alive and is_deleted is silly.

src/fastfield/alive_bitset.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /// Iterate over the positions of the set elements
+    #[inline]
+    pub fn iter_unset(&self) -> impl Iterator<Item = u32> + '_ {
```
fulmicoton: s/u32/DocId

fulmicoton: iter_deleted

src/fastfield/alive_bitset.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /// Returns true iff the document has been marked as deleted.
+    #[inline]
+    pub fn is_deleted(&self, doc: DocId) -> bool {
+        let data = self.data.as_slice();
+        !BitSet::contains_from_bytes(doc, data)
+    }
+
+    /// Iterate over the positions of the set elements
```
fulmicoton: comment is wrong.

src/fastfield/alive_bitset.rs
```diff
@@ -0,0 +1,200 @@
...
+/// Set of alive `DocId`s.
+#[derive(Clone)]
+pub struct AliveBitSet {
+    data: OwnedBytes,
+    num_deleted: usize,
+}
+
+impl AliveBitSet {
+    #[cfg(test)]
+    pub(crate) fn for_test(not_alive_docs: &[DocId], max_doc: u32) -> AliveBitSet {
```
fulmicoton: deleted_docs

src/fastfield/alive_bitset.rs
```diff
@@ -0,0 +1,200 @@
...
+        let mut wrt = directory.open_write(path).unwrap();
+        write_alive_bitset(&bitset, &mut wrt).unwrap();
+        wrt.terminate().unwrap();
+        let file = directory.open_read(path).unwrap();
+        Self::open(file).unwrap()
+    }
+
+    /// Opens a delete bitset given its file.
+    pub fn open(file: FileSlice) -> crate::Result<AliveBitSet> {
+        let bytes = file.read_bytes()?;
```
fulmicoton, line 49: What happened to the 4 byte with max_doc in the beginning?

OwnedBytes is sliceable for free.

common/src/bitset.rs
```diff
@@ -132,8 +180,78 @@ fn num_buckets(max_val: u32) -> u32 {
...
+    #[inline]
+    fn iter_tinysets_from_bytes<'a>(data: &'a [u8]) -> impl Iterator<Item = TinySet> + 'a {
+        assert!((data.len() - 4) % 8 == 0);
+        data[4..].chunks_exact(8).map(move |chunk| {
+            let tinyset: TinySet = TinySet::deserialize(chunk.try_into().unwrap()).unwrap();
+            tinyset
+        })
+    }
+
+    /// Iterate over the positions of the unset elements.
```
fulmicoton: Don't we want `iter_set`? What is iter_unset for?

src/fastfield/alive_bitset.rs
fulmicoton: and actually we don't care about iter_deleted,(and iter_unset is erroneous)...

What we want is iter_alive

src/core/segment_reader.rs
```diff
@@ -72,14 +72,14 @@ impl SegmentReader {
     /// Return the number of documents that have been
     /// deleted in the segment.
     pub fn num_deleted_docs(&self) -> DocId {
-        self.delete_bitset()
+        self.alive_bitset()
             .map(|delete_set| delete_set.num_deleted() as DocId)
```
fulmicoton: rename alive_bitset

src/fastfield/alive_bitset.rs
PSeitz: which padding do you mean?

src/fastfield/alive_bitset.rs
PSeitz: true

src/indexer/index_writer.rs
```diff
@@ -167,20 +167,21 @@ pub(crate) fn advance_deletes(
...
-                delete_bitset.insert(doc);
+            if seg_alive_bitset.is_deleted(doc) {
+                alive_bitset.remove(doc);
             }
         }
     }
 
-    let num_deleted_docs: u32 = delete_bitset.len() as u32;
+    let num_alive_docs: u32 = alive_bitset.num_set_bits() as u32;
+    let num_deleted_docs = max_doc - num_alive_docs;
```
fulmicoton: Actually there is `alive_bitset.num_deleted()`

src/indexer/merger.rs
```diff
@@ -503,13 +497,11 @@ impl IndexMerger {
...
         for (reader, u64s_reader) in reader_and_field_accessors.iter() {
-            if let Some(delete_bitset) = reader.delete_bitset() {
-                num_docs += reader.max_doc() as u64 - delete_bitset.len() as u64;
-                for doc in 0u32..reader.max_doc() {
-                    if delete_bitset.is_alive(doc) {
-                        let num_vals = u64s_reader.get_len(doc) as u64;
-                        total_num_vals += num_vals;
-                    }
+            if let Some(alive_bitset) = reader.alive_bitset() {
+                num_docs += reader.max_doc() as u64 - alive_bitset.num_deleted() as u64;
```
fulmicoton, line 500: I think there is a method for num_alive()

common/src/bitset.rs
```diff
@@ -28,56 +31,101 @@ impl IntoIterator for TinySet {
...
+    /// Returns true iff the `TinySet` bit is set at position `pos`.
+    pub fn contains(self, pos: u32) -> bool {
+        !self.intersect(TinySet::singleton(pos)).is_empty()
     }
 
-    /// Returns the number of elements in the TinySet.
-    pub fn len(self) -> u32 {
+    #[inline]
+    /// Returns the number of set bits in the TinySet.
+    pub fn num_set(self) -> u32 {
```
fulmicoton: len was ok actually. It is a set. `len` is pretty well defined and unambiguous for a set.

common/src/bitset.rs
```diff
@@ -132,8 +180,78 @@ fn num_buckets(max_val: u32) -> u32 {
...
+            tinysets: tinysets.into_boxed_slice(),
+            len,
+            max_value,
+        })
+    }
+
+    /// Count the number of unset bits from serialized data.
+    ///
+    #[inline]
+    pub fn count_unset_from_bytes<'a>(data: &'a [u8]) -> usize {
```
fulmicoton: I'd rather remove this guy if possible, and rely on TinySet::len and max_doc to infer num_deletes

common/src/bitset.rs
```diff
@@ -28,56 +31,101 @@ impl IntoIterator for TinySet {
...
-    pub fn len(self) -> u32 {
+    #[inline]
+    /// Returns the number of set bits in the TinySet.
+    pub fn num_set(self) -> u32 {
         self.0.count_ones()
     }
 
+    #[inline]
+    /// Returns the number of unset bits in the TinySet.
+    pub fn num_unset(self) -> u32 {
```
fulmicoton: I'd rather not have this function.

common/src/bitset.rs
```diff
@@ -28,56 +31,101 @@ impl IntoIterator for TinySet {
...
+    }
+
+    #[inline]
     /// Returns the intersection of `self` and `other`
     pub fn intersect(self, other: TinySet) -> TinySet {
         TinySet(self.0 & other.0)
     }
 
-    /// Creates a new `TinySet` containing only one element
+    /// Creates a new `TinySet` with only one bit set at `pos`.
```
fulmicoton: Again... element was perfectly fine. It's a set. It has element.

I understand it is possibly  to sell it as a bitfield, but I  don't think it is better, and certainly not useful to rename the world partially for it.

common/src/bitset.rs
```diff
@@ -28,56 +31,101 @@ impl IntoIterator for TinySet {
...
     /// within `[0; 64[`
     #[inline]
-    pub fn singleton(el: u32) -> TinySet {
-        TinySet(1u64 << u64::from(el))
+    pub fn singleton(pos: u32) -> TinySet {
+        TinySet(1u64 << u64::from(pos))
     }
 
-    /// Insert a new element within [0..64[
+    /// Insert a new element within [0..64)
```
fulmicoton, line 95: thx

src/fastfield/alive_bitset.rs
PSeitz: they are still there, but now private data of the BitSet, since it's part of the serialization format of BitSet

common/src/bitset.rs
```diff
@@ -28,56 +31,101 @@ impl IntoIterator for TinySet {
...
         TinySet(self.0 & other.0)
     }
 
-    /// Creates a new `TinySet` containing only one element
+    /// Creates a new `TinySet` with only one bit set at `pos`.
     /// within `[0; 64[`
     #[inline]
-    pub fn singleton(el: u32) -> TinySet {
-        TinySet(1u64 << u64::from(el))
+    pub fn singleton(pos: u32) -> TinySet {
```
fulmicoton: Again this rename just adds entropy.

common/src/bitset.rs
PSeitz: yes, I forgot to flip the name to `iter_set`

common/src/bitset.rs
```diff
@@ -168,6 +298,27 @@ impl BitSet {
         };
     }
 
+    /// Inserts an element in the `BitSet`
```
fulmicoton, line 268: Comment is wrong.

common/src/bitset.rs
```diff
@@ -152,8 +282,8 @@ impl BitSet {
     }
 
     /// Returns the number of elements in the `BitSet`.
-    pub fn len(&self) -> usize {
-        self.len
+    pub fn num_set_bits(&self) -> usize {
```
fulmicoton: len is perfectly fined and well defined.

src/docset.rs
```diff
@@ -85,11 +85,11 @@ pub trait DocSet: Send {
 
     /// Returns the number documents matching.
     /// Calling this method consumes the `DocSet`.
-    fn count(&mut self, delete_bitset: &DeleteBitSet) -> u32 {
+    fn count(&mut self, alive_bitset: &AliveBitSet) -> u32 {
         let mut count = 0u32;
         let mut doc = self.doc();
         while doc != TERMINATED {
-            if !delete_bitset.is_deleted(doc) {
+            if !alive_bitset.is_deleted(doc) {
```
fulmicoton: There is an is_alive function now.

src/indexer/merger.rs
```diff
@@ -98,27 +96,23 @@ pub struct IndexMerger {
 
 fn compute_min_max_val(
     u64_reader: &impl FastFieldReader<u64>,
-    max_doc: DocId,
-    delete_bitset_opt: Option<&DeleteBitSet>,
+    segment_reader: &SegmentReader,
 ) -> Option<(u64, u64)> {
-    if max_doc == 0 {
+    if segment_reader.max_doc() == 0 {
         None
```
fulmicoton: Let's use the return early pattern on that kind of stuff.

```rust
 if segment_reader.max_doc() == 0 {
        None
}

if segment_reader.alive_bitset().is_none() {
    // no deleted documents,
    // we can use the previous min_val, max_val.
    return Some((u64_reader.min_value(), u64_reader.max_value()));
}

// some deleted documents,
// we need to recompute the max / min
minmax(
    segment_reader
        .doc_ids_alive()
        .map(|doc_id| u64_reader.get(doc_id))
```

(I understand it was not that way before)

common/src/bitset.rs
```diff
@@ -144,6 +262,18 @@ impl BitSet {
         }
     }
 
+    /// Create a new `BitSet` that may contain elements
```
fulmicoton: The docstring does not explain the bitset is full.

common/src/bitset.rs
```diff
@@ -144,6 +262,18 @@ impl BitSet {
...
+    /// within `[0, max_val)`.
+    pub fn with_max_value_and_filled(max_value: u32) -> BitSet {
```
fulmicoton: I am not sure filled is easy to understand. Maybe `with_all_elements(max_value: u32)`?
(I can't think of a great name either)

fulmicoton: "full" would have the benefit of sticking with the TinySet naming.

fulmicoton: I would change the remaining bits (above max_doc) of the  last TinyBitSet to 0.

PSeitz: full seems like a good name to me

> I would change the remaining bits (above max_doc) of the last TinyBitSet to 0.

I considered that, but they aren't supposed to be used anyway?

src/indexer/merger.rs
PSeitz: done, I also prefer it that way (easier to read due to less indentation)

src/indexer/index_writer.rs
PSeitz: the alive_bitset is not the AliveBitSet, but just the im BitSet

common/src/bitset.rs
```diff
@@ -198,16 +301,160 @@ impl BitSet {
...
 
+/// Lazy Read a serialized BitSet.
+#[derive(Clone)]
+pub struct ReadSerializedBitSet {
+    data: OwnedBytes,
+    max_value: u32,
+}
+
+impl ReadSerializedBitSet {
+    pub fn new(data: OwnedBytes) -> Self {
```
fulmicoton: Rename `open`, or `read`, `from_bytes` maybe?

common/src/bitset.rs
```diff
@@ -198,16 +301,160 @@ impl BitSet {
...
+
+                let tinyset: TinySet = TinySet::deserialize(chunk.try_into().unwrap()).unwrap();
+                (tinyset, is_last)
+            })
+    }
+
+    /// Iterate over the positions of the unset elements.
+    ///
+    #[inline]
+    pub fn iter_unset<'a>(&'a self) -> impl Iterator<Item = u32> + 'a {
```
fulmicoton: Rename `iter()`.

common/src/bitset.rs
```diff
@@ -28,30 +32,51 @@ impl IntoIterator for TinySet {
...
+    }
+
     pub fn clear(&mut self) {
         self.0 = 0u64;
     }
 
+    #[inline]
     /// Returns the complement of the set in `[0, 64[`.
-    fn complement(self) -> TinySet {
+    pub fn complement(self) -> TinySet {
```
fulmicoton: nitpick: If it is not used I'd prefer to make it private.

common/src/bitset.rs
```diff
@@ -198,16 +301,160 @@ impl BitSet {
...
+        self.max_value as usize - num_set
+    }
+
+    /// Iterate the tinyset on the fly from serialized data.
+    ///
+    /// Iterator returns (TinySet, is_last) element, so the consumer can ignore up to max_doc in the
+    /// last block.
+    ///
+    #[inline]
+    fn iter_tinysets<'a>(&'a self) -> impl Iterator<Item = (TinySet, bool)> + 'a {
```
fulmicoton: If it reaches this level of complexity, I think we'd better access the remaining bits but make sure they are set to their right value.

We can do that by either reading the actually bits or masking the like TinySet.
I would rather not care, and read the actual bits and fix the initialization `with_max_value_and_full`

common/src/bitset.rs
```diff
@@ -198,16 +301,160 @@ impl BitSet {
...
+            .enumerate()
+            .map(move |(chunk_num, chunk)| {
+                let is_last = (chunk_num + 1) * 8 == self.data.len();
+
+                let tinyset: TinySet = TinySet::deserialize(chunk.try_into().unwrap()).unwrap();
+                (tinyset, is_last)
+            })
+    }
+
+    /// Iterate over the positions of the unset elements.
```
fulmicoton: fix comment. This is not iterating unset elements.

common/src/bitset.rs
PSeitz: yep

fulmicoton (approved):
Please rename iter_unset and ideally address the other comments as well.

PR #1156, opened 2021-09-20
Add unit tests for tokenizers and filters

mocobeta:
Some tokenizers/filters seem to have no unit test (e.g. SimpleTokenizer and StopWordFilter).
I think it would be nice to add basic tests for them for future development. To start with, I added a test module for SimpleTokenizer; does it make sense?

### DIFF
--- a/src/tokenizer/lower_caser.rs
+++ b/src/tokenizer/lower_caser.rs
@@ -56,31 +56,30 @@ impl<'a> TokenStream for LowerCaserTokenStream<'a> {
 
 #[cfg(test)]
 mod tests {
-    use crate::tokenizer::{LowerCaser, SimpleTokenizer, TextAnalyzer};
+    use crate::tokenizer::tests::assert_token;
+    use crate::tokenizer::{LowerCaser, SimpleTokenizer, TextAnalyzer, Token};
 
     #[test]
     fn test_to_lower_case() {
-        assert_eq!(
-            lowercase_helper("Русский текст"),
-            vec!["русский".to_string(), "текст".to_string()]
-        );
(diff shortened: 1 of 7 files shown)


PSeitz:
@mocobeta That's a really good idea to cover these with tests

mocobeta:
@PSeitz thanks for your reply. I'll try to add tests for other components.

fulmicoton:
@mocobeta awesome. I'm happy to merge this, but the PR is in draft state.

mocobeta:
Thanks, @fulmicoton. I'd like to include a few more tests for other tokenizers, and then I will open this soon.

src/tokenizer/lower_caser.rs
```diff
@@ -56,31 +56,30 @@ impl<'a> TokenStream for LowerCaserTokenStream<'a> {
...
+    use crate::tokenizer::tests::assert_token;
+    use crate::tokenizer::{LowerCaser, SimpleTokenizer, TextAnalyzer, Token};
 
     #[test]
     fn test_to_lower_case() {
-        assert_eq!(
-            lowercase_helper("Русский текст"),
-            vec!["русский".to_string(), "текст".to_string()]
-        );
+        let tokens = token_stream_helper("Tree");
```
mocobeta, line 64: I refactored LowerCaser test so that positions and offsets are also verified.

mocobeta:
I think this covers all existing tokenizers/filters. Tests added here are very basic ones though, I hope this will be of some help.

fulmicoton:
@mocobeta It definitely helps! Thank you!

PR #1159, opened 2021-09-22
add index accessor for index writer

Mesteery:
I think it would be interesting to be able to access the Index from the IndexWriter. In my (specific) case, it is in particular to access the fields, from an index writer. Otherwise I don't have many ways to access it.

### DIFF
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -348,6 +348,13 @@ impl IndexWriter {
         self.operation_sender = sender;
     }
 
+    /// Accessor to the index
+    ///
+    /// The index is actually cloned.
+    pub fn index(&self) -> Index {
+        self.index.clone()
+    }
+
     /// If there are some merging threads, blocks until they all finish their work and
     /// then drop the `IndexWriter`.
     pub fn wait_merging_threads(mut self) -> crate::Result<()> {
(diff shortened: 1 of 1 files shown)


src/indexer/index_writer.rs
```diff
@@ -348,6 +348,13 @@ impl IndexWriter {
         self.operation_sender = sender;
     }
 
+    /// Accessor to the index
+    ///
+    /// The index is actually cloned.
+    pub fn index(&self) -> Index {
+        self.index.clone()
```
fulmicoton: Cloning Index is not super cheap.
Why not returning a reference? People could then clone it if required?

```suggestion
    pub fn index(&self) -> &Index {
        &self.index
```

fulmicoton (changes requested):

src/indexer/index_writer.rs
Mesteery: All right! Is there any reason not to do the same in https://github.com/tantivy-search/tantivy/blob/367f5da7827f5213b04efb8409f5ecc8374a9786/src/core/index.rs#L474?

fulmicoton: yes :)
One big difference is that Schema is super cheap to copy.

PR #1163, opened 2021-09-28
reduce mem usage

PSeitz:
- use only segment ordinal in docidmapping
- reserve vec
- prealloc vec in postinglist

### DIFF
--- a/src/indexer/merger.rs
+++ b/src/indexer/merger.rs
@@ -67,26 +67,6 @@ fn compute_total_num_tokens(readers: &[SegmentReader], field: Field) -> crate::R
             .sum::<u64>())
 }
 
-/// `ReaderWithOrdinal` is used to be able to easier associate
-/// data with a `SegmentReader`. The ordinal is supposed to be
-/// used as an index access.
-///
-/// The ordinal identifies the position within `Merger` readers.
-#[derive(Clone, Copy)]
-pub(crate) struct SegmentReaderWithOrdinal<'a> {
-    pub reader: &'a SegmentReader,
-    pub ordinal: SegmentOrdinal,
-}
-
-impl<'a> From<(usize, &'a SegmentReader)> for SegmentReaderWithOrdinal<'a> {
-    fn from(data: (usize, &'a SegmentReader)) -> Self {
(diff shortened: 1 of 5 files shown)


src/indexer/merger.rs
```diff
@@ -482,20 +467,20 @@ impl IndexMerger {
...
     // Important: reader_and_field_accessor needs
     // to have the same order as self.readers since ReaderWithOrdinal
-    // is used to index the reader_and_field_accessors vec.
+    // is used to index the reader_ordinal_and_field_accessors vec.
     fn write_1_n_fast_field_idx_generic<T: MultiValueLength>(
         field: Field,
         fast_field_serializer: &mut CompositeFastFieldSerializer,
         doc_id_mapping: &SegmentDocidMapping,
-        reader_and_field_accessors: &[(&SegmentReader, T)],
+        reader_ordinal_and_field_accessors: &[(&SegmentReader, T)],
```
fulmicoton: !?

fulmicoton: It is still the reader here?

src/indexer/merger.rs
```diff
@@ -482,20 +467,20 @@ impl IndexMerger {
     //
     // Important: reader_and_field_accessor needs
     // to have the same order as self.readers since ReaderWithOrdinal
-    // is used to index the reader_and_field_accessors vec.
+    // is used to index the reader_ordinal_and_field_accessors vec.
```
fulmicoton: ?

src/indexer/merger.rs
```diff
@@ -766,7 +753,7 @@ impl IndexMerger {
         fast_field_serializer: &mut CompositeFastFieldSerializer,
         doc_id_mapping: &SegmentDocidMapping,
     ) -> crate::Result<()> {
-        let reader_and_field_accessors = self
+        let reader_ordinal_and_field_accessors = self
```
fulmicoton: it's not the ordinal here either

fulmicoton (changes requested):
See comments. You mistakenly renamed a bunch of segment -> segment_ordinal

fulmicoton (approved):

PR #1164, opened 2021-10-05
Leaning more on the alive (vs delete) semantics.

fulmicoton:


common/src/bitset.rs
```diff
@@ -369,7 +366,11 @@ impl ReadSerializedBitSet {
         b & (1u8 << shift) != 0
     }
 
-    /// Returns the max_value.
+    /// Maximum value the bitset may contain.
+    /// (Note this is not the maximum value contained in the set.)
+    ///
+    /// A bitset has an intrinsic capacity.
+    /// It only stores elements within [0..max_value).
```
PSeitz, line 373: this comment suggests to maybe rename it to capacity

src/core/segment_reader.rs
```diff
@@ -69,17 +69,9 @@ impl SegmentReader {
         &self.schema
     }
 
-    /// Return the number of documents that have been
-    /// deleted in the segment.
-    pub fn num_deleted_docs(&self) -> DocId {
```
PSeitz, line 74: Isn't this a public API someone may use currently?

PSeitz (approved):
LGTM, except maybe avoid API change

src/core/segment_reader.rs
fulmicoton: That's true. yes let's readd this.

common/src/bitset.rs
fulmicoton: Yeah it does feel weird right? Capacity is not great either though.

src/core/segment_reader.rs
PSeitz: Ah we still need this method in demux

common/src/bitset.rs
PSeitz: yes I agree

shikhar: maybe limit?

fulmicoton: `num_bits`, `num_capacity_bits` ?

shikhar: `width`!

fulmicoton: `bit_width` is quite good actually.

PR #1169, opened 2021-10-06
Code simplification.

fulmicoton:


common/src/bitset.rs
```diff
@@ -36,10 +36,14 @@ impl TinySet {
         writer.write_all(self.0.to_le_bytes().as_ref())
     }
 
+    pub fn into_bytes(self) -> [u8; 8] {
+        self.0.to_le_bytes()
+    }
```
PSeitz, line 41: `From` would be the better approach imo

common/src/bitset.rs
```diff
@@ -326,16 +309,35 @@ impl BitSet {
...
 
-impl ReadSerializedBitSet {
+pub fn intersect_bitsets(left: &ReadOnlyBitSet, other: &ReadOnlyBitSet) -> ReadOnlyBitSet {
+    assert_eq!(left.max_value(), other.max_value());
+    assert_eq!(left.data.len(), other.data.len());
+    let union_tinyset_it = left
+        .iter_tinysets()
+        .zip(other.iter_tinysets())
+        .map(|(left_tinyset, right_tinyset)| left_tinyset.intersect(right_tinyset))
+        .map(|union_tinyset| union_tinyset);
```
PSeitz: map does not map anything

common/src/bitset.rs
```diff
@@ -326,16 +309,35 @@ impl BitSet {
 
 /// Serialized BitSet.
 #[derive(Clone)]
-pub struct ReadSerializedBitSet {
+pub struct ReadOnlyBitSet {
```
PSeitz, line 312: I think ReadOnlyBitSet is misleading, it's not read only

common/src/bitset.rs
```diff
@@ -326,16 +309,35 @@ impl BitSet {
...
+    ReadOnlyBitSet {
+        data: OwnedBytes::new(output_dataset),
+        max_value: left.max_value(),
+    }
+}
+
+impl ReadOnlyBitSet {
     pub fn open(data: OwnedBytes) -> Self {
         let (max_value_data, data) = data.split(4);
+        assert_eq!(data.len() % 8, 0);
```
PSeitz, line 337:
```suggestion
        assert_eq!(data.len() % 8, 0, "bitset size not multiple of 8byte {}", data.len());
```

common/src/bitset.rs
fulmicoton: Is it not?

common/src/bitset.rs
fulmicoton: Nice catch!

common/src/bitset.rs
fulmicoton: It has pros and cons.

With `From<[u8; 8]> for TinySet` the compiler will not be able to infer the types from time to time
without hinting. For instance, when we actually want a `&[u8]`.

PR #1180, opened 2021-10-23
Remove travis config

palfrey:
On the basis that https://app.travis-ci.com/github/quickwit-inc/tantivy isn't activated and there's a Github Actions setup, I'm assuming the travis config isn't needed anymore?


fulmicoton:
Very good question! @pseitz are we done with Travis?

PSeitz:
Yep, we can delete it

fulmicoton:
Thank you @palfrey

PR #1181, opened 2021-10-24
Update codecov-action to v2 and make it possible to keep it up-to-date with dependabot

k-yomo:
Updated codecov-action v2 because v1 is depracated and will not function soon.
https://github.com/codecov/codecov-action#%EF%B8%8F--deprecration-of-v1

Also, added github-actions to dependabot to keep actions up-to-date.


fulmicoton:
Thank you @k-yomo

PR #1182, opened 2021-10-24
use Box<dyn Directory> as parameter to open/create an Index

### LINKED ISSUE #1171: Remove Directory generic
Index::create etc. are generic over `Directory`. But the generic is wrapped in a `Box<dyn Directory>` internally

PSeitz:
Replace <D: Directory> generic and support `Box<dyn Directory>` instead. Box<T> does not implement the trait for T, so this it the more powerful api.
Remove boxing in ManagedDirectory.

This is an breaking API change, all `Directory` parameters in Index::create etc. need to be wrapped in Box::new().

Closes #1171

### DIFF
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -123,8 +123,8 @@ impl IndexBuilder {
     /// If a previous index was in this directory, it returns an `IndexAlreadyExists` error.
     #[cfg(feature = "mmap")]
     pub fn create_in_dir<P: AsRef<Path>>(self, directory_path: P) -> crate::Result<Index> {
-        let mmap_directory = MmapDirectory::open(directory_path)?;
-        if Index::exists(&mmap_directory)? {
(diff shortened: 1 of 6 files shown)


fulmicoton (approved):

fulmicoton (changes requested):
wer

src/core/index.rs
```diff
@@ -149,8 +149,8 @@ impl IndexBuilder {
             .ok_or(TantivyError::IndexBuilderMissingArgument("schema"))
     }
     /// Opens or creates a new index in the provided directory
-    pub fn open_or_create<Dir: Directory>(self, dir: Dir) -> crate::Result<Index> {
-        if !Index::exists(&dir)? {
+    pub fn open_or_create(self, dir: Box<dyn Directory>) -> crate::Result<Index> {
```
fulmicoton: ?

fulmicoton (approved):
wer

PR #1183, opened 2021-10-24
Fix outdated comment for IndexWriter::new

k-yomo:
Fix outdated comments since `IndexWriter::new` doesn't cause panic by too small heap size per thread anymore and returns an error instead
https://github.com/quickwit-inc/tantivy/blob/c412a46105e87f92142eba159aa3413808a04252/src/indexer/index_writer.rs#L285-L296


fulmicoton (approved):

fulmicoton:
Thank you @k-yomo

PR #1188, opened 2021-10-29
fix incorrect padding in bitset for multiple of 64

PSeitz:


common/src/bitset.rs
```diff
@@ -213,8 +213,9 @@ impl BitSet {
 
         // Fix padding
         let lower = max_value % 64u32;
-        tinybisets[tinybisets.len() - 1] = TinySet::range_lower(lower);
-
+        if lower != 0 {
+            tinybisets[tinybisets.len() - 1] = TinySet::range_lower(lower);
```
fulmicoton: rename tinysets

fulmicoton (approved):

fulmicoton (approved):

PR #1190, opened 2021-10-30
Optimize block wand for one and several TermScorer.

fmassot:
I made some changes to your initial commit @fulmicoton 

- first, based on the [paper](http://engineering.nyu.edu/~suel/papers/bmw.pdf) of Ding and Suel I made a small change on the block wand algorithm. The function `block_max_was_too_low_advance_one_scorer` now takes the min of all last doc id scorers until the pivot scorer. Before, we were taking the `doc()` of the pivot scorer.
- secondly, I think there was a tiny bug in `block_max_was_too_low_advance_one_scorer`: we were doing `doc_to_seek_after + 1` even on scorer after the pivot. There is a very small probability to miss a document here, I think we should not add +1 in this case.
- thirdly, I updated the way the scorer to advance is chosen. Now we take the scorer which has the best score among scorers[..pivot_len].
- at last, I added the implementation for the case where we have one `TermScorer` because the code is simple and readable and the performance is far better (around x3 faster). With that, we are on par with Lucene.

I also created a branch with the benchmark results here: https://github.com/quickwit-inc/search-benchmark-game/tree/blockwand-for-termquery

I noticed that the performance is better on average and for union we have:
- Top 10 - UNION: 1,081 μs | 1,419 μs

### DIFF
--- a/src/query/boolean_query/block_wand.rs
+++ b/src/query/boolean_query/block_wand.rs
@@ -195,6 +195,46 @@ pub fn block_wand(
     }
 }
 
+pub fn block_wand_single_scorer(
+    mut scorer: TermScorer,
+    mut threshold: Score,
+    callback: &mut dyn FnMut(u32, Score) -> Score,
+) {
+    let mut doc = scorer.doc();
+    loop {
+        // We position the scorer on a block that can reach
+        // the threshold.
+        while scorer.block_max_score() < threshold {
+            let last_doc_in_block = scorer.last_doc_in_block();
+            if doc == TERMINATED {
+                return;
+            }
+            doc = last_doc_in_block + 1;
+            scorer.shallow_seek(doc);
+        }
+        // Seek will effectively load that block.
+        doc = scorer.seek(doc);
+        if doc == TERMINATED {
+            break;
+        }
+        loop {
+            let score = scorer.score();
+            if score > threshold {
+                threshold = callback(doc, score);
+            }
+            if doc >= scorer.last_doc_in_block() {
+                break;
+            }
+            doc = scorer.advance();
+            if doc == TERMINATED {
+                return;
+            }
+        }
+        doc = doc + 1;
+        scorer.shallow_seek(doc);
+    }
+}
+
 struct TermScorerWithMaxScore<'a> {
     scorer: &'a mut TermScorer,
     max_score: Score,
@@ -245,6 +285,8 @@ mod tests {
     use std::collections::BinaryHeap;
     use std::iter;
 
+    use super::block_wand_single_scorer;
+
     struct Float(Score);
 
     impl Eq for Float {}
@@ -278,7 +320,8 @@ mod tests {
         let mut heap: BinaryHeap<Float> = BinaryHeap::with_capacity(n);
         let mut checkpoints: Vec<(DocId, Score)> = Vec::new();
         let mut limit: Score = 0.0;
-        super::block_wand(term_scorers, Score::MIN, &mut |doc, score| {
+
+        let callback = &mut |doc, score| {
             heap.push(Float(score));
             if heap.len() > n {
                 heap.pop().unwrap();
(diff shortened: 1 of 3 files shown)


fmassot:
Ok there is one more subtleties to take into account: we need to advance the scorer that have the greatest idf. I will update the algorithm and the benchmark.

src/query/boolean_query/block_wand.rs
```diff
@@ -42,27 +42,39 @@ fn find_pivot_doc(
...
+/// Note: before and after calling this method, scorers need to be sorted by their `.doc()`.
 fn block_max_was_too_low_advance_one_scorer(
     scorers: &mut Vec<TermScorerWithMaxScore>,
     pivot_len: usize,
 ) {
     debug_assert!(is_sorted(scorers.iter().map(|scorer| scorer.doc())));
     let mut scorer_to_seek = pivot_len - 1;
-    let mut doc_to_seek_after = scorers[scorer_to_seek].doc();
+    let mut global_max_score = scorers[scorer_to_seek].max_score;
+    let mut doc_to_seek_after = scorers[scorer_to_seek].last_doc_in_block();
```
fulmicoton, line 56: I don't think this is correct, is it?

src/query/boolean_query/block_wand.rs
```diff
@@ -42,27 +42,39 @@ fn find_pivot_doc(
...
     for scorer_ord in (0..pivot_len - 1).rev() {
         let scorer = &scorers[scorer_ord];
         if scorer.last_doc_in_block() <= doc_to_seek_after {
             doc_to_seek_after = scorer.last_doc_in_block();
+        }
+        if scorers[scorer_ord].max_score > global_max_score {
+            global_max_score = scorers[scorer_ord].max_score;
```
fulmicoton, line 63: ah yes this is way better! Advancing any scorer is "correct", but we want to advance the termscorer with the lowest docfreq.

src/query/boolean_query/block_wand.rs
fulmicoton, line 56: Are we sure the next doc cannot be in the same block?

src/query/boolean_query/block_wand.rs
```diff
@@ -42,27 +42,39 @@ fn find_pivot_doc(
...
             scorer_to_seek = scorer_ord;
         }
     }
+    // Add +1 to go to the next block unless we are already at the end.
+    if doc_to_seek_after != TERMINATED {
+        doc_to_seek_after += 1;
```
fulmicoton, line 69: you are right this was a bug.

src/query/boolean_query/block_wand.rs
```diff
@@ -195,6 +211,55 @@ pub fn block_wand(
...
+    mut threshold: Score,
+    callback: &mut dyn FnMut(u32, Score) -> Score,
+) {
+    let mut doc = scorer.doc();
+    loop {
+        // We position the scorer on a block that can reach
+        // the threshold.
+        while scorer.block_max_score() < threshold {
+            let last_doc_in_block = scorer.last_doc_in_block();
+            if doc == TERMINATED {
```
fulmicoton: did you mean last_doc_in_block here maybe?

src/query/boolean_query/block_wand.rs
```diff
@@ -195,6 +211,55 @@ pub fn block_wand(
...
+        doc = scorer.seek(doc);
+        if doc == TERMINATED {
+            break;
+        }
+        loop {
+            let score = scorer.score();
+            if score > threshold {
+                threshold = callback(doc, score);
+            }
+            if doc >= scorer.last_doc_in_block() {
```
fulmicoton: This `>=` only triggers as `==` ? 
If this is correct, should we add a debug_assert on == maybe?

fulmicoton (approved):

fulmicoton (approved):

src/query/boolean_query/block_wand.rs
fmassot: Indeed, we don't need the assert, this is pretty straightforward.

fulmicoton (approved):

PR #1199, opened 2021-11-09
Fix the deserialization error of FieldEntry when the 'options' field …

azerowall:
Currently, FieldEntry deserialization is done manually by using a custom deserializer. During the parsing process, if  the "options" field appears before the "type" field, an error will occur.
Example of json on which the error occurs:
```
{
  "name": "title",
  "options": {
    "indexing": {
      "record": "position",
      "tokenizer": "default"
    },
    "stored": false
  },
  "type": "text"
}
```
This pull request changes the custom serializer to serde_derive macro, which fixes the described error.


fulmicoton:
Thank you!

fulmicoton (approved):

Issue #1204, opened 2021-11-14
Add shingle token filter or token n-grams


fmassot:
I thought this was present in tantivy but for now, there is only a `NgramTokenizer` that tokenizes words into the n-grams.

Lucene offers a [ShinglerFilter](https://lucene.apache.org/core/8_9_0/analyzers-common/org/apache/lucene/analysis/shingle/ShingleFilter.html) which creates shingles or token n-grams, it creates combinations of tokens and not letters.

For example, this [dataset](https://archive.org/details/GeneralIndex) published token n-grams and that would be interesting to index it with tantivy instead of having some SQL dump.

mocobeta:
Hi,
I was also trying to implement a shingle filter. I left a PR, but it's incomplete - I tried to explain where I've stuck in the description.

fulmicoton:
@fmassot I am not sure I understand how the shingle filter could help with the ngram dataset.

fmassot:
@fulmicoton ah yes, that was not clear, my idea was to be able to process directly articles contents not the ngram dataset which is there because of legal constraints.

PR #1205, opened 2021-11-14
Add FixedShingleFilter

mocobeta:
Relates to #1204.

Shingle token filter is useful for some advanced applications (e.g., fast phrase search, or implementing word n-gram language models).
Recently I was trying to implement a shingle token filter for my own project. This is more like Lucene's [FixedShingleFilter](https://lucene.apache.org/core/8_10_0/analyzers-common/org/apache/lucene/analysis/shingle/FixedShingleFilter.html) that emits only fixed-sized shingles rather than ShingleFilter; it's easier to implement, and I think fixed-sized shingles are often sufficient in many practical use-cases.

This is my prototype and works well for my application - making shingles is possible by maintaining a buffer that remembers the past tokens (keeps only tokens that are needed to generate the next shingles). However, to me, there is one problem - emitting the correct token position. The current implementation is not compatible with StopWordFilter (or filters that remove some tokens from the stream). `token()` and `token_mut()` methods return a reference rather than a new Token object, so the original stream's position information is not available after overwriting it. I tried to align that problem but have failed so far. But I could be missing something... your help or feedback is welcome.

### DIFF
--- a/src/tokenizer/shingle_filter.rs
+++ b/src/tokenizer/shingle_filter.rs
@@ -1,0 +1,207 @@
+use super::{BoxTokenStream, Token, TokenFilter, TokenStream};
+
+#[derive(Clone)]
+pub struct FixedShingleFilter {
+    size: usize,
+    separator: char,
+}
+
+impl FixedShingleFilter {
+    pub fn new(size: usize) -> Self {
+        FixedShingleFilter::new_with_separator(size, '_')
+    }
+
+    pub fn new_with_separator(size: usize, separator: char) -> Self {
+        FixedShingleFilter { size, separator }
+    }
+}
+
+impl TokenFilter for FixedShingleFilter {
+    fn transform<'a>(&self, token_stream: BoxTokenStream<'a>) -> BoxTokenStream<'a> {
+        BoxTokenStream::from(ShingleFilterStream {
+            size: self.size,
+            separator: self.separator,
+            buffer: String::with_capacity(100),
+            pendings: Vec::with_capacity(self.size - 1),
+            tail: token_stream,
+            original_token_position: 0,
+        })
+    }
+}
+
+pub struct ShingleFilterStream<'a> {
+    size: usize,
+    separator: char,
+    buffer: String,
+    pendings: Vec<Token>,
+    tail: BoxTokenStream<'a>,
+    original_token_position: usize,
+}
+
+impl<'a> ShingleFilterStream<'a> {
+    fn is_window_filled(&self) -> bool {
+        self.pendings.len() == self.size - 1
+    }
+
+    fn push_token(&mut self, mut token: Token) {
+        if self.is_window_filled() {
+            self.pendings.remove(0);
+        }
+        token.position = self.original_token_position;
+        self.pendings.push(token);
+    }
+}
+
+fn make_shingle(token: &mut Token, pendings: &Vec<Token>, buffer: &mut String, separator: char) {
+    buffer.clear();
+    buffer.push_str(token.text.as_str());
+
+    // set text
+    token.text.clear();
+    for pend_token in pendings {
+        token.text.push_str(pend_token.text.as_str());
+        token.text.push(separator);
+    }
+    token.text.push_str(buffer.as_str());
+
+    // set offset and position
+    if let Some(first_token) = pendings.get(0) {
+        token.offset_from = first_token.offset_from;
+        token.position = first_token.position;
+    }
+}
+
+impl<'a> TokenStream for ShingleFilterStream<'a> {
+    fn advance(&mut self) -> bool {
+        while !self.is_window_filled() {
+            if self.tail.advance() {
+                self.push_token(self.tail.token().clone());
+                self.original_token_position += 1;  // This is not compatible with StopWordFilter
+            } else {
+                return false;
+            }
+        }
+        if self.tail.advance() {
+            let next_token = self.tail.token().clone();
+            make_shingle(
+                &mut self.tail.token_mut(),
+                &self.pendings,
+                &mut self.buffer,
+                self.separator,
+            );
+            self.push_token(next_token);
+            self.original_token_position += 1;
+            return true;
+        }
+        false
+    }
+
+    fn token(&self) -> &Token {
+        self.tail.token()
+    }
+
+    fn token_mut(&mut self) -> &mut Token {
+        self.tail.token_mut()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::tokenizer::{FixedShingleFilter, SimpleTokenizer, TextAnalyzer, Token};
+
+    #[test]
+    fn test_shingle_bigram() {
+        let tokens = shingle_helper("i am a cat", 2);
+        let expected = vec![
+            Token {
+                text: "i_am".to_string(),
+                offset_from: 0,
+                offset_to: 4,
+                position: 0,
+                position_length: 1,
+            },
+            Token {
+                text: "am_a".to_string(),
+                offset_from: 2,
+                offset_to: 6,
+                position: 1,
+                position_length: 1,
+            },
+            Token {
+                text: "a_cat".to_string(),
+                offset_from: 5,
+                offset_to: 10,
+                position: 2,
+                position_length: 1,
+            },
+        ];
+        assert_eq!(&tokens, &expected);
+    }
+
+    #[test]
+    fn test_shingle_trigram() {
+        let tokens = shingle_helper("as yet i have no name", 3);
+        let expected = vec![
+            Token {
+                text: "as_yet_i".to_string(),
+                offset_from: 0,
+                offset_to: 8,
+                position: 0,
+                position_length: 1,
+            },
+            Token {
+                text: "yet_i_have".to_string(),
+                offset_from: 3,
+                offset_to: 13,
+                position: 1,
+                position_length: 1,
+            },
+            Token {
+                text: "i_have_no".to_string(),
+                offset_from: 7,
+                offset_to: 16,
+                position: 2,
+                position_length: 1,
+            },
+            Token {
+                text: "have_no_name".to_string(),
+                offset_from: 9,
+                offset_to: 21,
+                position: 3,
+                position_length: 1,
+            },
+        ];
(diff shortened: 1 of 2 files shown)


fulmicoton:
Here is what I'd love to have for shingles to improve phrase query speed incredibly at a very low cost on indexing.
A shinglefilter that takes a trigger dictionary...

At indexing:
Emit all tokens all of the times
If a given token is encounterred, then also emit the 2-gram starting at the trigger token.

If "and" is a trigger then:

"Of mice and men" would emit at indexing
- "of"{pos=0, length=1}
-  "mice"{pos=1, length=1}
-  "and"{pos=2, length=1}
-  "and+men"{pos=2, length=2} # where + would actually be one these exotic unicode separator codepoint.
- "men"{pos=3, length=1}  

And at query time:
- "of"{pos=0, length=1}
-  "mice"{pos=1, length=1}
-  "and+men"{pos=2, length=2} 


The point here is that the shingle trick is mostly potent for some specific token with a very high doc frequency.
For other words it is not as beneficial to emit shingles, and the cost at indexing is quite high.

fulmicoton:
@mocobeta For this to be useful, we need to have some asymetry between the indexing tokenizer and the search tokenizer isn't it? We need to land #1073 or some variant of it?

mocobeta:
> At indexing:
> Emit all tokens all of the times
> If a given token is encounterred, then also emit the 2-gram starting at the trigger token.

I didn't think of that, but this is an interesting idea to me.
Besides indexing performance, the obvious downside of shingles is their vocabulary size... Generating shingles only when some criteria (dictionary- or rule-based?) are met significantly reduces the index size.

Here are some stats of my tantivy index with bi-gram and tri-gram tokens made by this patch. (data: [47 million web pages](https://lemurproject.org/clueweb09/))
```
Total indexed docs: 47187784
Total term freq in unigram field: 9171726767 (unique terms: 29185)
Total term freq in bigram field: 9124538983 (unique terms: 96691414)
Total term freq in trigram field: 9077354364 (unique terms: 913295070)
```

The whole index size is 100GB (unigram + bi-gram + tri-gram) though, indexing was done within an hour with 36 threads - impressively fast :)

mocobeta:
I think the shingle filter alone works well without #1073, but the combination could be useful in some situations?

src/tokenizer/shingle_filter.rs
```diff
@@ -0,0 +1,207 @@
...
+    fn push_token(&mut self, mut token: Token) {
+        if self.is_window_filled() {
+            self.pendings.remove(0);
+        }
+        token.position = self.original_token_position;
+        self.pendings.push(token);
+    }
+}
+
+fn make_shingle(token: &mut Token, pendings: &Vec<Token>, buffer: &mut String, separator: char) {
```
fulmicoton: `&[T]` is almost always a better argument type than `&Vec<T>` .

(that comment does not apply to `&mut Vec<T>`)

```suggestion
fn make_shingle(token: &mut Token, pendings: &[Token], buffer: &mut String, separator: char) {
```

src/tokenizer/shingle_filter.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+impl<'a> ShingleFilterStream<'a> {
+    fn is_window_filled(&self) -> bool {
+        self.pendings.len() == self.size - 1
+    }
+
+    fn push_token(&mut self, mut token: Token) {
+        if self.is_window_filled() {
+            self.pendings.remove(0);
```
fulmicoton, line 48: nitpick: I wonder if a rolling buffer would be more appropriate here.

src/tokenizer/shingle_filter.rs
```diff
@@ -0,0 +1,207 @@
...
+
+impl<'a> TokenStream for ShingleFilterStream<'a> {
+    fn advance(&mut self) -> bool {
+        while !self.is_window_filled() {
+            if self.tail.advance() {
+                self.push_token(self.tail.token().clone());
+                self.original_token_position += 1;  // This is not compatible with StopWordFilter
+            } else {
+                return false;
+            }
```
fulmicoton: nitpick:
The whole codebase is not consistent but we try to use early returns these days..

```suggestion
            if !self.tail.advance() {
                return false;
            }
            self.push_token(self.tail.token().clone());
            self.original_token_position += 1;  // This is not compatible with StopWordFilter
```

src/tokenizer/shingle_filter.rs
```diff
@@ -0,0 +1,207 @@
...
+            make_shingle(
+                &mut self.tail.token_mut(),
+                &self.pendings,
+                &mut self.buffer,
+                self.separator,
+            );
+            self.push_token(next_token);
+            self.original_token_position += 1;
+            return true;
+        }
```
fulmicoton, line 95:
```suggestion
        if !self.tail.advance() {
            return false;
        }
        let next_token = self.tail.token().clone();
        make_shingle(
            &mut self.tail.token_mut(),
            &self.pendings,
            &mut self.buffer,
            self.separator,
        );
        self.push_token(next_token);
        self.original_token_position += 1;
        true
   }
```

src/tokenizer/shingle_filter.rs
```diff
@@ -0,0 +1,207 @@
...
+        false
```
fulmicoton:
```suggestion
        if !self.tail.advance() {
            return false;
        }
        let next_token = self.tail.token().clone();
        make_shingle(
            &mut self.tail.token_mut(),
            &self.pendings,
            &mut self.buffer,
            self.separator,
        );
        self.push_token(next_token);
        self.original_token_position += 1;
        true
```

src/tokenizer/shingle_filter.rs
```diff
@@ -0,0 +1,207 @@
...
+            self.pendings.remove(0);
+        }
+        token.position = self.original_token_position;
+        self.pendings.push(token);
+    }
+}
+
+fn make_shingle(token: &mut Token, pendings: &Vec<Token>, buffer: &mut String, separator: char) {
+    buffer.clear();
+    buffer.push_str(token.text.as_str());
```
fulmicoton, line 57: nitpick
That just removes one line and one token copy so it does not matter much...
But mem::swap could work too here.

src/tokenizer/shingle_filter.rs
```diff
@@ -0,0 +1,207 @@
...
+        self.tail.token_mut()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::tokenizer::{FixedShingleFilter, SimpleTokenizer, TextAnalyzer, Token};
+
+    #[test]
+    fn test_shingle_bigram() {
```
fulmicoton, line 112: The code handles this case just fine but can you add a test where we have less than "window size" tokens to begin with?
(The expected result is that we should emit not token).

fulmicoton: And one test with the empty string too.

fulmicoton (changes requested):
I put minor comments inline

fulmicoton:
@mocobeta ah yes, you are right. We don't need any asymetry between search and indexing here.
It is only helpful with the extension I suggested.

LGTM.
I just added a few minor comments. Please ignore all of them but the unit tests.

src/tokenizer/shingle_filter.rs
fulmicoton: (please do not modify the code... It is fine as it is)

src/tokenizer/shingle_filter.rs
fulmicoton: (please do not address this comment. the code is fine as is.)

src/tokenizer/shingle_filter.rs
```diff
@@ -0,0 +1,207 @@
...
+        token.position = first_token.position;
+    }
+}
+
+impl<'a> TokenStream for ShingleFilterStream<'a> {
+    fn advance(&mut self) -> bool {
+        while !self.is_window_filled() {
+            if self.tail.advance() {
+                self.push_token(self.tail.token().clone());
+                self.original_token_position += 1;  // This is not compatible with StopWordFilter
```
fulmicoton: Can you comment in the code the reason why you chose not to use `self.tail.token().position` for this?
Ideally with an example with a phrase query?

fulmicoton (changes requested):
Added request for comment.

fulmicoton:
```
Total indexed docs: 47187784
Total term freq in unigram field: 9171726767 (unique terms: 29185)
Total term freq in bigram field: 9124538983 (unique terms: 96691414)
Total term freq in trigram field: 9077354364 (unique terms: 913295070)
The whole index size is 100GB (unigram + bi-gram + tri-gram) though, indexing was done within an hour with 36 threads - impressively fast :)
```

Awesome!
So my theory (never tested it) is that the "shingle on trigger" trick gives most of the benefit of shingles, but cuts down the 
number of bigrams drastically.

One mostly wants shingles for super high docfreq terms. In english with 100 static stop words, that would lead to at most 2.9M bigrams (and actually way less than that).

src/tokenizer/shingle_filter.rs
mocobeta: Thanks for your suggestion. Actually, I still don't have a clear idea about position handling in tantivy; this silly hack should be replaced with more sensible code. I'll try to find the right way for this.

fulmicoton: I would not be surprise if we are inconsistent on positions. I would advise you to be very critical on tantivy approach, and I am ok to adopt the lucene way in all tokenizer.

My understanding is that for something like a stop word filter, depending on how we deal with position we will have a different outcome for the following query / doc pair.
- query: the cheapest on internet 
- doc: the cheapest on the internet

In other words, should queries act as 
`the (STOP WORDS)* cheapest (STOP WORDS)* internet`
or 
`the cheapest (STOP WORDS){2} internet`

There is a decent case for both. The best is probably to adopt the lucene way and stay consistent with all TokenFilter

fulmicoton: @mocobeta Do you want help to dig into what should be the right choice for position here?

mocobeta: @fulmicoton I can't find the time right now, but it could be good homework for the coming winter vacation.

PR #1208, opened 2021-11-16
Add `none` compression

shikhar:
If none of the `{lz4,brotli,snappy}-compression` feature flags are present, `Compressor::None` is used.

It just copies the input slice to destination vector. The copy may be avoidable with some `Compressor` API changes, but left to a future optimization if it's even worthwhile - someone exercising the docstore will probably enable compression.

### DIFF
--- a/src/store/compressors.rs
+++ b/src/store/compressors.rs
@@ -53,7 +53,7 @@ impl Compressor {
     }
     pub(crate) fn get_id(&self) -> u8 {
         match self {
-            Compressor::NoOp => 0,
+            Self::NoOp => 0,
             Self::Lz4 => 1,
             Self::Brotli => 2,
             Self::Snappy => 3,
--- a/src/store/compressors.rs
+++ b/src/store/compressors.rs
@@ -63,7 +63,7 @@ impl Compressor {
     pub(crate) fn compress(&self, uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
         match self {
             Self::None => {
-                debug_assert!(compressed.is_empty());
+                compressed.clear();
                 compressed.extend_from_slice(uncompressed);
                 Ok(())
             }
@@ -108,7 +108,7 @@ impl Compressor {
     ) -> io::Result<()> {
         match self {
             Self::None => {
-                debug_assert!(decompressed.is_empty());
+                decompressed.clear();
                 decompressed.extend_from_slice(compressed);
                 Ok(())
             }
--- a/src/store/compressors.rs
+++ b/src/store/compressors.rs
@@ -13,6 +13,9 @@ pub trait StoreCompressor {
 /// The default is Lz4Block, but also depends on the enabled feature flags.
 #[derive(Clone, Debug, Copy, PartialEq, Eq, Serialize, Deserialize)]
 pub enum Compressor {
+    #[serde(rename = "noop")]
+    /// No-op compressor
+    NoOp,
     #[serde(rename = "lz4")]
     /// Use the lz4 compressor (block format)
     Lz4,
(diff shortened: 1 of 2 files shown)


src/store/compressors.rs
```diff
@@ -59,6 +62,11 @@ impl Compressor {
     #[inline]
     pub(crate) fn compress(&self, uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
         match self {
+            Self::NoOp => {
+                debug_assert!(compressed.is_empty());
```
shikhar: All the compressor implementations first clear the vec first, although caller already pre-clears. I can replace those with `debug_assert`, or switch this to an explicit `clear()`, whatever you prefer for consistency!

src/store/compressors.rs
```diff
@@ -33,16 +36,15 @@ impl Default for Compressor {
...
-            );
+            Compressor::NoOp
         }
     }
 }
 
 impl Compressor {
     pub(crate) fn from_id(id: u8) -> Compressor {
         match id {
+            0 => Compressor::NoOp,
```
shikhar: Are you ok with using 0 for this?

src/store/compressors.rs
```diff
@@ -13,6 +13,9 @@ pub trait StoreCompressor {
 /// The default is Lz4Block, but also depends on the enabled feature flags.
 #[derive(Clone, Debug, Copy, PartialEq, Eq, Serialize, Deserialize)]
 pub enum Compressor {
+    #[serde(rename = "noop")]
+    /// No-op compressor
+    NoOp,
```
fulmicoton: Can we rename this `NoCompression`

src/store/compressors.rs
```diff
@@ -13,6 +13,9 @@ pub trait StoreCompressor {
 /// The default is Lz4Block, but also depends on the enabled feature flags.
 #[derive(Clone, Debug, Copy, PartialEq, Eq, Serialize, Deserialize)]
 pub enum Compressor {
+    #[serde(rename = "noop")]
```
fulmicoton: See below, rename "no-compression"

src/store/compressors.rs
fulmicoton: ok

src/store/compressors.rs
fulmicoton: I think we should remove hidden contract, call `compressed.clear()` here and remove the assert.

fulmicoton (changes requested):
See the minor comments

src/store/compressors.rs
shikhar: How about just `None`, serialized as `"none"`? The other variants don't have a `Compression` suffix and are serialized as `"lz4"`, `"snappy"`, `"brotli"`.

src/store/compressors.rs
shikhar: I added explicit `clear()`, like the other implementations have it. Let me know if by removing hidden contract, you meant to also remove the redundant clear from `compress()` / `decompress()` callers! Otherwise I guess we are good.

src/store/compressors.rs
fulmicoton: ok

src/store/compressors.rs
fulmicoton: I think we can eventually remove the redundant clear but this is not necessary here. 

I have a PR about Quickwit's code style.
There is a section about hidden contracts you can read.
https://github.com/quickwit-inc/quickwit/pull/779

shikhar: Thanks, that helped - "If it is not too much work, just redoing" .. certainly a very tiny amount of work in question here

fulmicoton: yes exactly :)

PSeitz:
When would someone want not any compressor?

fulmicoton:
Removing dependencies. @shikhar only uses fast field, probably like a foreign key. I encounterred the same need in the past when working on a small index and web assembly.

PR #1213, opened 2021-11-23
Add random dataset on fastfield bench.

fmassot:


PSeitz (approved):
Only 256 different values on purpose?

fmassot:
I tried `u32` and `u64`.
If I remember correctly `u32` gives the same kind of results, `u64` is not chosen by the multilinear interpolation because it's not applicable.

fastfield_codecs/src/main.rs
```diff
@@ -91,6 +91,10 @@ pub fn get_codec_test_data_sets() -> Vec<(Vec<u64>, &'static str)> {
         .collect::<Vec<_>>();
     data_and_names.push((data, "Almost monotonically increasing"));
 
+    let data = (1000..=200_000_u64)
+        .map(|_| rand::random::<u8>() as u64)
```
fulmicoton, line 95: Can we seed that?

fulmicoton (changes requested):
I would rather seed that one to avoid extra noise in the bench.

fastfield_codecs/src/main.rs
fmassot: I will. But the the results is already very stable on 200 000 values

fmassot:
Closing this PR as it will be done in #1217

PR #1216, opened 2021-11-25
Switch to non-strict schema

### LINKED ISSUE #1211: Remove tantivy schema strictness
Currently all fields in the data need to be defined in the tantivy schema, even if these fields should be ignored.
This can be an issue if you have a dataset with many fields and are interested in only few fields.

On the other side fields can be missing in the data while indexing, so there is some asymmetry.

For that I would propose to remove the strict schema default in tantivy and simply ignore fields which are not defined. It will also provide a lower entry barrier for new users (e.g. just indexing one field as a poke test).

PSeitz:
Fixes #1211

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+Tantivy 0.17
+================================
+- Change to non-strict schema. Ignore fields in data which are not defined in schema. Previously this returned an error. #1211
+
 Tantivy 0.16.2
 ================================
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+Tantivy 0.17
+================================
+- Change to non-strict schema. Ignore fields in data which are not defined in schema. Previously this returned an error. #1211
```
fulmicoton, line 3: Thank you for updating the changelog!

fulmicoton (approved):

Issue #1212, opened 2021-11-23
Possible to save 16 bytes on linear block metadata in the multilinear interpolation


fmassot:
Currently, we serialize these values for each linear block:
- `data_start_offset` (u64) 
- `value_start_pos` (u64)
- `positive_val_offset` (u64)
- `slope` (f32)
- `num_bits` u8

It represents 29 bytes. But it's possible to remove 16 bytes.

Let's try to understand what we really need first and without bitpacking. We want to model some y values [y1, y2,...yn] as a linear function
`y = a.x + b`

But we want all values to be positive, so we find `b'` to satisfy this requirement and in the end, we have:
`y = a.x + b'`

We know that our y values will not match the linear function but we will write the difference between the expected values and the real ones on the disk:
`ywritten = yreal - (ax + b')` (1)

To rebuild `yreal`, we only need to know `ywritten`, `b'`, `a`. The `x` value is given by the caller so we don't need to write it. 

But we also want to bitpack `yw` values. For that we need to store additional information, the `num_bits` used for bitpacking. And that's all: 
- `data_start_offset` can be deduced by summing `num_bits * block_size` for each block before the block where we want to read.  we can do it when we open the fast field file.
- `value_start_pos` is not needed as we can see in (1).

## Is it worth saving 16 bytes?

Saving 16 bytes is not changing the world nor is a priority. But removing variables that we don't need is a good thing because it makes the code more readable. 

## What can we do?

For backward compatibility, I think we can add a new codec in the future release that will be more optimized than the current one.

PSeitz:
> data_start_offset can be deduced by summing num_bits * block_size for each block before the block where we want to read. we can do it when we open the fast field file.

>Saving 16 bytes is not changing the world nor is a priority. But removing variables that we don't need is a good thing because it >makes the code more readable.

We would still have the struct with data_start_offset, since we shouldn't calculate it on every access. So the removal of data_start_offset would be only in the serialized data, not in the code.

fmassot:
> We would still have the struct with data_start_offset, since we shouldn't calculate it on every access. So the removal of data_start_offset would be only in the serialized data, not in the code.

True. But I think we can also save some lines of code ^^.

fulmicoton:
Is the slope actually useful in the blockwise linear thing or frame of reference sufficient for blocks of 512 elements?
An alternative could be for instance to have a simpler code that takes 9 bytes per block, have smaller blocks (say 128 bytes), and no linear interpolation.

[...]

PR #1217, opened 2021-11-25
Add piecewise linear codec, deprecate linear and mulitinear,  complete benchmark with real world datasets.

### LINKED ISSUE #1215: Slope value unexpected in multilinear interpolation
```rust
  #[test]
  fn test_compression_on_simple_data() {
      let data = (0..=CHUNK_SIZE*11 as u64).collect::<Vec<_>>();
      let (estimate, actual_compression) =
          create_and_validate(&data, "simple monotonically");
      println!("actual compression {} estimate {} data len {}", actual_compression, estimate, data.len());
      assert!(actual_compression < 0.2);
      assert!(estimate < 0.20);
      assert!(estimate > 0.15);
      assert!(actual_compression > 0.1);
  }
```

I printed the slopes and the length of the interpolation

```
slope 1.0019569
slope 1.0019569
slope 1.0019569
slope 1.0019569
slope 1.0019569
slope 1.0019569
slope 1.0019569
slope 1.0019569
slope 1.0019569
slope 1.0019569
slope 1.0019569
slope NaN
interpolations len 12
```

I would have expected a slope of 1 and only 11 interpolation functions.

fmassot:
Ok PR is ready for review.

Here is a summary of the PR:
- Piecewise linear codec added: basically it's a refactor and it fixes #1215
- #1212 partially fixed that by saving 8 bytes.
- Linear interpolation and multilinear deprecated. It's now only there for the reading part so that users can still read these fast field codecs
- FOR codec added with feature `unstable`, it's not used in tantivy for the moment.
- Bench also compression time/reading time and on real-world datasets.


fastfield_codecs/src/multilinearinterpol_2.rs
```diff
@@ -0,0 +1,353 @@
...
+use tantivy_bitpacker::BitPacker;
+
+use common::BinarySerializable;
+use common::DeserializeFrom;
+use tantivy_bitpacker::BitUnpacker;
+
+const BLOCK_SIZE: u64 = 512;
+
+#[derive(Clone)]
+pub struct MultiLinearInterpol2FastFieldReader {
```
PSeitz:
```suggestion
pub struct MultiLinearInterpolV2FastFieldReader {
```

fastfield_codecs/README.md
```diff
@@ -29,40 +40,108 @@ cargo run --features bin
...
++----------------------------------+-------------------+------------------------------+--------------------------+----------------------+
+| Monotonically increasing concave |                   |                              |                          |                      |
++----------------------------------+-------------------+------------------------------+--------------------------+----------------------+
+| LinearInterpol                   | 0.25003937        | 0.26562938                   | 775                      | 113                  |
++----------------------------------+-------------------+------------------------------+--------------------------+----------------------+
+| MultiLinearInterpol              | 0.05403           | 0.1883836                    | 914                      | 223                  |
++----------------------------------+-------------------+------------------------------+--------------------------+----------------------+
+| MultiLinearInterpol2             | 0.005955          | 0.18813984                   | 899                      | 245                  |
```
PSeitz:
```suggestion
| MultiLinearInterpolV2             | 0.005955          | 0.18813984                   | 899                      | 245                  |
```

fastfield_codecs/src/multilinearinterpol_2.rs
```diff
@@ -0,0 +1,353 @@
+/*!
+
+MultiLinearInterpol compressor uses linear interpolation to guess a values and stores the offset, but in blocks of 512.
+
+With a CHUNK_SIZE of 512 and 29 byte metadata per block, we get a overhead for metadata of 232 / 512 = 0,45 bits per element.
```
PSeitz: Outdated comment. I think outlining the V2 changes would make sense

fastfield_codecs/src/multilinearinterpol_2.rs
```diff
@@ -0,0 +1,353 @@
...
+    min_value: u64,
+    max_value: u64,
+    block_metadatas: Vec<BlockMetadata>,
+    block_start_offsets: Vec<u64>,
+    block_bit_unpackers: Vec<BitUnpacker>,
+}
+
+#[derive(Clone, Debug, Default)]
+struct BlockMetadata {
+    constant: u64,
```
PSeitz: I don't think constant is a good name

fastfield_codecs/src/multilinearinterpol_2.rs
```diff
@@ -0,0 +1,353 @@
...
+    num_bits: u8,
+}
+
+
+impl BinarySerializable for BlockMetadata {
+    fn serialize<W: Write>(&self, write: &mut W) -> io::Result<()> {
+        self.constant.serialize(write)?;
+        self.constant_positive_offset.serialize(write)?;
+        self.slope.serialize(write)?;
+        self.num_bits.serialize(write)?;
```
PSeitz: actually we could get the max vals and apply bitpacking to save some more bits (except from the f32)

(store the num bits in the codec header)

fmassot: I won't do that for now :)

fastfield_codecs/src/multilinearinterpol_2.rs
fmassot: Yes indeed, I replaced it with `first_value`.

fulmicoton (approved):

fastfield_codecs/src/frame_of_reference.rs
```diff
@@ -0,0 +1,303 @@
...
+use tantivy_bitpacker::BitPacker;
+
+use common::BinarySerializable;
+use common::DeserializeFrom;
+use tantivy_bitpacker::BitUnpacker;
+
+const BLOCK_SIZE: u64 = 128;
+
+#[derive(Clone)]
+pub struct FrameOfReferenceFastFieldReader {
```
fulmicoton: I think we can use the abbreviation `FOR` for once :)
It's one of those acronyms that is more famous than the underlying expression.

fastfield_codecs/src/frame_of_reference.rs
```diff
@@ -0,0 +1,303 @@
...
+
+impl FastFieldCodecReader for FrameOfReferenceFastFieldReader {
+    /// Opens a fast field given a file.
+    fn open_from_bytes(bytes: &[u8]) -> io::Result<Self> {
+        let footer_len: u32 = (&bytes[bytes.len() - 4..]).deserialize()?;
+        let (_, mut footer) = bytes.split_at(bytes.len() - (4 + footer_len) as usize);
+        let footer = FrameOfReferenceFooter::deserialize(&mut footer)?;
+        let mut block_readers = Vec::with_capacity(footer.block_metadatas.len());
+        let mut current_data_offset = 0;
+        for block_metadata in footer.block_metadatas.into_iter() {
```
fulmicoton:
```suggestion
        for block_metadata in footer.block_metadatas {
```

fastfield_codecs/src/frame_of_reference.rs
```diff
@@ -0,0 +1,303 @@
...
+        Ok(Self {
+            num_vals: footer.num_vals,
+            min_value: footer.min_value,
+            max_value: footer.max_value,
+            block_readers,
+        })
+    }
+
+    #[inline]
+    fn get_u64(&self, doc: u64, data: &[u8]) -> u64 {
```
fulmicoton: doc -> DocId (which is a u32 actually)

fastfield_codecs/src/multilinearinterpol_v2.rs
```diff
@@ -0,0 +1,376 @@
...
+use tantivy_bitpacker::BitPacker;
+
+use common::BinarySerializable;
+use common::DeserializeFrom;
+use tantivy_bitpacker::BitUnpacker;
+
+const BLOCK_SIZE: u64 = 512;
+
+#[derive(Clone)]
+pub struct MultiLinearInterpolV2FastFieldReader {
```
fulmicoton: I suggest we stop calling this MultiLinear. MultiLinear usually means linear against with regard to several variables.
PiecewiseLinear is the correct terminology I believe.

fastfield_codecs/src/frame_of_reference.rs
fmassot: This is defined at the trait level, I prefer to change that in another PR.

fulmicoton: Actually scratch that!

This is not necessarily a doc here, which is why this is u64.
The multivalued fast field rely on 2 fast field as you know.

In this case the second one is *not* a DocId. Let's call doc "idx" or something in this function.
We should fix other fastfields as well as the trait.

fmassot: Ok I totally forgot about multivalued ff, I updated the trait.

fmassot:
@PSeitz @fulmicoton : here is my take on what to do for next tantivy release and what to do after:

These remarks are valid for integers only (for floats, our codecs are useless):
- piecewise linear interpolation should have the best compression ratio for monotonically increasing series
- bitpacking should have the best compression ratio on random stuff
- frame of reference (FOR) should have the best compression for series in between

What can we do for tantivy next release?
- unplug linear interpolation and replace multilinear interpolation by new piecewise linear interpolation (the bug fix increases a little bit the compression ratio but nothing noteworthy)
- update the benchmark with new datasets
- I don't want to plug immediately the FOR stuff because I think we can do slightly better and I would also want to activate piecewise interpolation only on monotonically increasing series, I would introduce that later.

Future work:
- finalize the FOR codecs, activate piecewise linear for only monotonically increasing series
- sparse codecs
- float codecs

fulmicoton:
@fmassot agreeing with  the above.

fulmicoton:
closing as obsolete. it was going in the right direction IMHO.

PR #1220, opened 2021-11-29
WIP: Add support for fields of type Vector

catunlock:
The objective of this pull request is to have support for fields of type Vector in the schema so a document can have stored it's embedded representation (or even multiple representations in different fields) and the user should be able to make querys using as a parameter a vector  to search documents close in distance to this one.

The following changes has to be made:
- [ ] Support for fields of type Vector in the schema.
- [ ] Store these fields in the schema using a HNSW index (a modification of qdrant-segment for the moment).
- [ ] Add VectorQuery.
- [ ] Add VectorCollector (to retrieve the documents address with its Vector representation).


fulmicoton:
Can you extract the qdrant-vector stuff from the project? @catunlock

src/collector/vector_collector.rs
```diff
@@ -0,0 +1,100 @@
...
+        trace!("Segment local id: {}", segment_local_id);
+
+        let a = reader.vector_readers.read().unwrap();
+        let vector_reader = a.get(&self.field).unwrap();
+
+        Ok(VectorSegmentCollector::new(Arc::clone(vector_reader), segment_local_id))
+    }
+
+    fn requires_scoring(&self) -> bool {
+        true
```
fulmicoton, line 35:
```suggestion
        false
```

src/collector/vector_collector.rs
```diff
@@ -0,0 +1,100 @@
...
+
+
+
+use crate::{DocAddress, SegmentOrdinal, schema::Field, vector::VectorReader};
+
+use super::{Collector, SegmentCollector};
+
+/// Collector for vectors
+///
+/// The collector collects all vectors though all segments.
```
fulmicoton: What is the point of this collector? Is this for test only?

src/collector/vector_collector.rs
fulmicoton: do we really require scoring here?

fulmicoton:
@catunlock I looks like this is going somewhere! Pinging @mocobeta who has more expertise on vector search and I suspect it might spike her interest.
Have you discussed with the qdrant team to see if it possible to progress without an actual fork?

Also, this PR is a draft so no worries... to eventually get it merge, there will be some work required to clean up some of the code. You can have a look at this document to get an idea of what the preferred code style should look like.
https://github.com/quickwit-inc/quickwit/blob/main/CODE_STYLE.md

Cargo.toml
```diff
@@ -56,6 +56,8 @@ fastdivide = "0.3"
 itertools = "0.10.0"
 measure_time = "0.8.0"
 
+nuclia_vectors = { git = "https://github.com/nuclia/nuclia_vectors.git" }
```
shikhar, line 59: Will this be behind a feature?

src/schema/term.rs
```diff
@@ -143,6 +150,13 @@ impl Term {
         self.0.extend(bytes);
     }
 
+    /// Sets the value of a `Vector` field.
+    pub fn set_vector(&mut self, vector: &Vec<f32>) {
```
shikhar, line 154: you can use a slice here (and probably some other places you are passing in a `&Vec<f32>`)

Cargo.toml
catunlock: I'm not sure. I can see that if the vector indexation is not well integrated into the Segments readers and writers, and it's like it's calling an external library, it could be behind a feature.

Although, if the code of this PR is well integrated (and there is still work to do :-) ), there is no need to make a feature to support Vector type fields. This will be only one more dependency.

src/collector/vector_collector.rs
catunlock: I did this collector to understand better how tantivy works. A TopDocs collector would indeed suffice. I will remove it from the PR.

src/schema/term.rs
catunlock: Ah yes! Clippy usually tells me to change it! I was just not using it yet.

catunlock:
> Have you discussed with the qdrant team to see if it possible to progress without an actual fork?

For the moment, I'm stripping all the features that are redundant from the Qdrant vector segments that are already on Tantivy (redundant or not needed for this use case, as the payload store or the filters in the search. I'm more interested in the HNSW indexing.).

src/collector/vector_collector.rs
catunlock: No

catunlock:
I'm closing this pull request since we are not going to use qdrant to add this feature. I will open another one in the future cleaner with the desired features implemented.

PR #1226, opened 2021-12-02
We are missing 4 bytes in the LZ4 compression buffer.

fulmicoton:
Closes #831

### DIFF
--- a/src/store/compression_lz4_block.rs
+++ b/src/store/compression_lz4_block.rs
@@ -1,28 +1,30 @@
 use std::io::{self};
 
 use core::convert::TryInto;
+use std::mem;
 use lz4_flex::{compress_into, decompress_into};
 
 #[inline]
+#[allow(clippy::uninit_vec)]
 pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
     compressed.clear();
-    let maximum_ouput_size = lz4_flex::block::get_maximum_output_size(uncompressed.len());
+    let maximum_ouput_size = mem::size_of::<u32>() + lz4_flex::block::get_maximum_output_size(uncompressed.len());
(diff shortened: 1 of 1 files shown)


PSeitz (approved):
The 4 bytes are already included in get_maximum_output_size, but let's not rely on that

src/store/compression_lz4_block.rs
```diff
@@ -1,23 +1,23 @@
...
 
 use core::convert::TryInto;
+use std::mem;
 use lz4_flex::{compress_into, decompress_into};
 
 #[inline]
 pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
     compressed.clear();
-    let maximum_ouput_size = lz4_flex::block::get_maximum_output_size(uncompressed.len());
+    let maximum_ouput_size = mem::size_of::<u32>() + lz4_flex::block::get_maximum_output_size(uncompressed.len());
```
PSeitz, line 11: reserve 4 bytes more

fulmicoton: I don't understand this comment?

fulmicoton:
> The 4 bytes are already included in get_maximum_output_size, but let's not rely on that

The previous code was creating a slice with unallocated memory. (in addition to being uninitialized)

PSeitz:
> 
> The previous code was creating a slice with unallocated memory. (in addition to being uninitialized)

With the `reserve` call, it should be allocated?
Uninitialized data is being handled in lz4_flex via `MaybeUninit`.

fulmicoton:
@PSeitz  Now it is allocated. Before this PR, the last 4 bytes were not necessarily allocated.
If you know how to deal with the MaybeUninitialized thing that would be awesome to remove the allow clippy thing.

PR #1228, opened 2021-12-02
Reducing the number of call to fsync on the directory.

fulmicoton:


src/directory/managed_directory.rs
```diff
@@ -222,9 +223,21 @@ impl ManagedDirectory {
...
-        if has_changed {
-            save_managed_paths(self.directory.as_ref(), &meta_wlock)?;
+        if !has_changed {
+            return Ok(());
+        }
+        save_managed_paths(self.directory.as_ref(), &meta_wlock)?;
+        if meta_wlock.managed_paths.len() > 1 {
+            // This is not the first file we add.
+            // `.managed.json` has been already properly created and we do not need to
+            // sync its parent directory.
```
PSeitz: I don't understand why, the existence of .managed.json is important, but not its content?

src/directory/mmap_directory.rs
```diff
@@ -367,22 +340,17 @@ impl Directory for MmapDirectory {
     /// removed before the file is deleted.
     fn delete(&self, path: &Path) -> result::Result<(), DeleteError> {
         let full_path = self.resolve_path(path);
-        match fs::remove_file(&full_path) {
-            Ok(_) => self.sync_directory().map_err(|e| DeleteError::IoError {
-                io_error: e,
-                filepath: path.to_path_buf(),
-            }),
```
PSeitz, line 374: removed, because we rely on the os now?

src/directory/mmap_directory.rs
```diff
@@ -470,6 +441,30 @@ impl Directory for MmapDirectory {
     fn watch(&self, watch_callback: WatchCallback) -> crate::Result<WatchHandle> {
         Ok(self.inner.watch(watch_callback))
     }
+
+    fn sync_directory(&self) -> Result<(), io::Error> {
+        let mut open_opts = OpenOptions::new();
+
+        // Linux needs read to be set, otherwise returns EINVAL
+        // write must not be set, or it fails with EISDIR
+        open_opts.read(true);
```
PSeitz, line 450: this code is also active in windows

src/store/compression_lz4_block.rs
```diff
@@ -1,14 +1,15 @@
...
 use lz4_flex::{compress_into, decompress_into};
+use std::mem;
 
 #[inline]
 #[allow(clippy::uninit_vec)]
 pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
     compressed.clear();
-    let maximum_ouput_size = mem::size_of::<u32>() + lz4_flex::block::get_maximum_output_size(uncompressed.len());
+    let maximum_ouput_size =
+        mem::size_of::<u32>() + lz4_flex::block::get_maximum_output_size(uncompressed.len());
```
PSeitz: change from different PR

src/indexer/segment_updater.rs
```diff
@@ -82,6 +84,7 @@ fn save_metas(metas: &IndexMeta, directory: &dyn Directory) -> crate::Result<()>
         io::ErrorKind::Other,
         msg.unwrap_or_else(|| "Undefined".to_string())
     ))));
+    directory.sync_directory()?;
```
PSeitz, line 87: sync before the atomic write?

src/directory/managed_directory.rs
fulmicoton: I am not sure what you read here. We do fsync its content today upon each write.
However, we do not flush the parent directory every time we update it.

src/directory/mmap_directory.rs
fulmicoton: No... Removed because we call sync_directory a single time after deleting all of the file, instead of doing it after deleting every single file.

fulmicoton: see the change in managed_directory l.195

src/directory/mmap_directory.rs
fulmicoton: Yeah I genuinely have no clue if it has any effect what it does on windows... but it has the merit of having been tested by some user a long time ago :D

src/indexer/segment_updater.rs
fulmicoton: Yes. Again this is a directory sync. Its point is to make sure all of the component files we created  are durably created.

src/directory/managed_directory.rs
```diff
@@ -222,9 +223,21 @@ impl ManagedDirectory {
             .write()
             .expect("Managed file lock poisoned");
         let has_changed = meta_wlock.managed_paths.insert(filepath.to_owned());
-        if has_changed {
-            save_managed_paths(self.directory.as_ref(), &meta_wlock)?;
+        if !has_changed {
+            return Ok(());
+        }
+        save_managed_paths(self.directory.as_ref(), &meta_wlock)?;
+        if meta_wlock.managed_paths.len() > 1 {
```
PSeitz: being explicit here would increase readability. It would also remove the dependency to `meta_wlock.managed_paths`. Could have unexpected side effects when someone alters `meta_wlock.managed_paths` elsewhere.

if meta_json_already_synced_to_directory {

src/directory/mmap_directory.rs
PSeitz: I think it's additive, but if it has been tested should be fine :)

src/indexer/segment_updater.rs
PSeitz: I thought that the atomic write may cause a dangling file if it only calls sync_data, but it doesn't because in the mmap_dir we call sync_directory afterwards, so we can remove that call before?

```
    fn atomic_write(&self, path: &Path, content: &[u8]) -> io::Result<()> {
        debug!("Atomic Write {:?}", path);
        let full_path = self.resolve_path(path);
        atomic_write(&full_path, content)?;
        self.sync_directory()
    }
```

fulmicoton: Yes! So the idea is we just need to be sure the files are durably persisted right before the commit, so we flush the directory right before the commit atomic_write

src/directory/managed_directory.rs
fulmicoton: Good point

fulmicoton: I'll keep the lock though. It makes it easy to reason with. Of course it might poison it, but here if we fail a write, I think halting and catch fire is a decent policy.

src/indexer/segment_updater.rs
PSeitz: I think we could save some calls of sync_directory after atomic_write. We need to sync the file only once.

This may require some bookkeeping. `atomic_write` seems to be used only by managed and meta.json, so we could think about making a specialized api

PSeitz (approved):
LGTM, see https://github.com/quickwit-inc/tantivy/pull/1228/files#r761617748 for potential more gains

src/indexer/segment_updater.rs
fulmicoton: I am not sure what you are suggesting exactly. There are many ways to reduce the number of fsync. 

Probably the simplest, least leaking solutions are:
- add a `open_writes(...)` method that takes a list of path instead of a single one.
- remove the managed.json file and rely on listing the directory.
 
A more involved solution but drastically more efficient could be to transparently bundle files into one if they share a prefix and they are small.

With that we would have only
- 3 file flush
- 1 directory flush

If we remove the managed.json too, we could get down to 2 file flush and 1 directory flush.
The cost would be of < 3ms per commit on an SSD.

That might be enough for NRT use cases (usually 50ms or so) without doing all of the NRT plumbing.

PSeitz: I'm not sure how multiple open_writes would reduce the amount of fsync calls.

`open_write` returns a SafeFileWriter which calls sync_all on terminate.
In both cases we have one sync_data call per file on commit?

PR #1231, opened 2021-12-06
canonicalize path when opening `MmapDirectory`

### LINKED ISSUE #1229: Resolve Path once and for all.
When using a relative path, we end up calling the `getcwd` syscalls 8 times per commit with the mmapdirectory.
Overall we end up losing `~0.1ms` on this.

It is not too terrible but still... It would be nicer to canonicalize our root Path once and for all when we open the directory.

saroh:
fixes #1229

- [x] test on linux box
- [ ] factorise `wrap_io_error`s in a trait ?

### DIFF
--- a/src/directory/mmap_directory.rs
+++ b/src/directory/mmap_directory.rs
@@ -196,7 +196,7 @@ impl MmapDirectory {
             |io_err| OpenDirectoryError::wrap_io_error(io_err, PathBuf::from(directory_path))
         )?;
         if !canonical_path.exists() {
-            Err(OpenDirectoryError::DoesNotExist(PathBuf::from(
+            return Err(OpenDirectoryError::DoesNotExist(PathBuf::from(
                 directory_path,
             )))
         } else if !canonical_path.is_dir() {
--- a/src/directory/mmap_directory.rs
+++ b/src/directory/mmap_directory.rs
@@ -192,17 +192,19 @@ impl MmapDirectory {
     /// exist or if it is not a directory.
     pub fn open<P: AsRef<Path>>(directory_path: P) -> Result<MmapDirectory, OpenDirectoryError> {
         let directory_path: &Path = directory_path.as_ref();
-        if !directory_path.exists() {
+        let canonical_path: PathBuf = directory_path.canonicalize().map_err(
+            |io_err| OpenDirectoryError::wrap_io_error(io_err, PathBuf::from(directory_path))
+        )?;
+        if !canonical_path.exists() {
             Err(OpenDirectoryError::DoesNotExist(PathBuf::from(
                 directory_path,
             )))
-        } else if !directory_path.is_dir() {
+        } else if !canonical_path.is_dir() {
             Err(OpenDirectoryError::NotADirectory(PathBuf::from(
                 directory_path,
             )))
-        } else {
-            Ok(MmapDirectory::new(PathBuf::from(directory_path), None))
         }
+        Ok(MmapDirectory::new(canonical_path, None))
     }
 
     /// Joins a relative_path to the directory `root_path`
--- a/src/directory/mmap_directory.rs
+++ b/src/directory/mmap_directory.rs
@@ -199,10 +199,11 @@ impl MmapDirectory {
             return Err(OpenDirectoryError::DoesNotExist(PathBuf::from(
                 directory_path,
             )))
-        } else if !canonical_path.is_dir() {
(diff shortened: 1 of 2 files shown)


src/directory/mmap_directory.rs
```diff
@@ -192,17 +192,19 @@ impl MmapDirectory {
     /// exist or if it is not a directory.
     pub fn open<P: AsRef<Path>>(directory_path: P) -> Result<MmapDirectory, OpenDirectoryError> {
         let directory_path: &Path = directory_path.as_ref();
-        if !directory_path.exists() {
+        let canonical_path: PathBuf = directory_path.canonicalize().map_err(
+            |io_err| OpenDirectoryError::wrap_io_error(io_err, PathBuf::from(directory_path))
+        )?;
+        if !canonical_path.exists() {
             Err(OpenDirectoryError::DoesNotExist(PathBuf::from(
```
fulmicoton: You need the  `return` keyword here. 

It was ok before without the `return` because we were returning the `if {} else {}` expression

src/directory/mmap_directory.rs
```diff
@@ -192,17 +192,19 @@ impl MmapDirectory {
...
                 directory_path,
             )))
-        } else if !directory_path.is_dir() {
+        } else if !canonical_path.is_dir() {
             Err(OpenDirectoryError::NotADirectory(PathBuf::from(
```
fulmicoton: same as above. 

We can remove the `else` too.

src/directory/mmap_directory.rs
```diff
@@ -192,17 +192,19 @@ impl MmapDirectory {
     /// exist or if it is not a directory.
     pub fn open<P: AsRef<Path>>(directory_path: P) -> Result<MmapDirectory, OpenDirectoryError> {
         let directory_path: &Path = directory_path.as_ref();
-        if !directory_path.exists() {
+        let canonical_path: PathBuf = directory_path.canonicalize().map_err(
+            |io_err| OpenDirectoryError::wrap_io_error(io_err, PathBuf::from(directory_path))
+        )?;
+        if !canonical_path.exists() {
             Err(OpenDirectoryError::DoesNotExist(PathBuf::from(
                 directory_path,
```
fulmicoton:
```suggestion
            return Err(OpenDirectoryError::DoesNotExist(PathBuf::from(
                directory_path,
```

src/directory/mmap_directory.rs
```diff
@@ -192,17 +192,19 @@ impl MmapDirectory {
...
             )))
-        } else if !directory_path.is_dir() {
+        } else if !canonical_path.is_dir() {
```
fulmicoton, line 202:
```suggestion
        if !canonical_path.is_dir() {
```

src/directory/mmap_directory.rs
```diff
@@ -192,17 +192,19 @@ impl MmapDirectory {
...
             Err(OpenDirectoryError::NotADirectory(PathBuf::from(
                 directory_path,
             )))
```
fulmicoton:
```suggestion
        }
        if !canonical_path.is_dir() {
            return Err(OpenDirectoryError::NotADirectory(PathBuf::from(
                directory_path,
            )));
```

fulmicoton (changes requested):
We need to fix the code to return errors.

fulmicoton (approved):

src/directory/mmap_directory.rs
saroh: > `we were returning the if {} else {} expression`

Ooooh... Ok I get it, the if block is evaluated so its value is returned. Thx, my "rust by example" is rusty.

saroh:
FYI / nitpick: 
+ there are probably OS dependent optims doable eg it seems under linux canonicalize fails if it doesn't point to an existing file, making the `exists` test redundant. (cf code cov)

+ I've copy/pasted the `wrap_io_error` implem, total rust newbie here so I'd say there might be a way to create a macro or a common base trait to derive for errors in this file ? It's short code though.

fulmicoton:
@saroh We really don't care about the optimisation aspect of it, but in order to get the right error, could you make the existence check over the relative path and before the canonicalization? (Otherwise we won't have the right error.)

fulmicoton (changes requested):
@saroh We really don't care about the optimisation aspect of it, but in order to get the right error, could you make the existence check over the relative path and before the canonicalization? (Otherwise we won't have the right error.)

fulmicoton (approved):

PR #1234, opened 2021-12-10
fix typ in doc

kkanas:
Signed-off-by: Krzysztof Kanas <kkanas@fastmail.com>

### DIFF
--- a/src/schema/field_type.rs
+++ b/src/schema/field_type.rs
@@ -84,7 +84,7 @@ impl FieldType {
         }
     }
 
-    /// returns true iff the field is indexed.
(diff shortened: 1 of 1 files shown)


fulmicoton:
It is not a typo https://en.wikipedia.org/wiki/If_and_only_if

kkanas:
Oh, didn't came across those type of notation very often in doc. Usually if works, ok. Thanks for clarification.

fulmicoton:
iff means "if and only if".
We can think of expanding the abbrevation everywhere it if it is not common enough.

fulmicoton:
Created issue #1236

Issue #115, opened 2017-04-08
Take in account the number of deletes in the LogMergePolicy


fulmicoton:
The current LogMergePolicy does not take in account deletes today. As a result, some bad behavior may appear
under the following natural workload.

- Documents are inserted continually.
- Documents are deleted continually. For instance as they get too old.

As merge are occuring one very large segment may appear.
It might take a very long time for segment of an equivalent size to appear, so that this segment virtually never get merged.
Yet, all or almost all of its docs might be deleted.

It would be beneficial to refine the heuristics of the log merge policy to take in account the ratio of delete documents as well.

fulmicoton:
Assigned informally to @blitzerr

fulmicoton:
@guilload Can you mentor @blitzerr on this ticket?

PR #1238, opened 2021-12-16
LogMergePolicy knob `del_docs_percentage_before_merge`

shikhar:
Addresses #115

If this % of deleted documents is exceeded for a segment, it will always be proposed for a merge, even if it is the only one at that level.

TODOs
* [x] ~determine what the default threshold should be~ https://github.com/quickwit-inc/tantivy/issues/1240
* [x] add test cases

### DIFF
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -974,7 +974,7 @@ mod tests {
         assert_eq!(
             format!("{:?}", index_writer.get_merge_policy()),
             "LogMergePolicy { min_num_segments: 8, max_docs_before_merge: 10000000, min_layer_size: 10000, \
-             level_log_size: 0.75, del_docs_percentage_before_merge: 100 }"
+             level_log_size: 0.75, del_docs_ratio_before_merge: 1.0 }"
         );
         let merge_policy = Box::new(NoMergePolicy::default());
         index_writer.set_merge_policy(merge_policy);
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -974,7 +974,7 @@ mod tests {
         assert_eq!(
             format!("{:?}", index_writer.get_merge_policy()),
             "LogMergePolicy { min_num_segments: 8, max_docs_before_merge: 10000000, min_layer_size: 10000, \
-             level_log_size: 0.75 }"
+             level_log_size: 0.75, max_del_docs_pct: 100 }"
         );
         let merge_policy = Box::new(NoMergePolicy::default());
         index_writer.set_merge_policy(merge_policy);
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -974,7 +974,7 @@ mod tests {
         assert_eq!(
             format!("{:?}", index_writer.get_merge_policy()),
             "LogMergePolicy { min_num_segments: 8, max_docs_before_merge: 10000000, min_layer_size: 10000, \
-             level_log_size: 0.75, max_del_docs_pct: 100 }"
+             level_log_size: 0.75, del_docs_percentage_before_merge: 100 }"
         );
         let merge_policy = Box::new(NoMergePolicy::default());
         index_writer.set_merge_policy(merge_policy);
--- a/src/indexer/log_merge_policy.rs
+++ b/src/indexer/log_merge_policy.rs
@@ -7,7 +7,7 @@ const DEFAULT_LEVEL_LOG_SIZE: f64 = 0.75;
 const DEFAULT_MIN_LAYER_SIZE: u32 = 10_000;
 const DEFAULT_MIN_NUM_SEGMENTS_IN_MERGE: usize = 8;
 const DEFAULT_MAX_DOCS_BEFORE_MERGE: usize = 10_000_000;
-const DEFAULT_MAX_DEL_DOCS_PCT: u8 = 100;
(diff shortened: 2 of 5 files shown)


shikhar:
Need to add additional test cases.

I tried it out at 10% and in initial usage it is seeming effective for us:
image


vs previously:
image

src/indexer/log_merge_policy.rs
```diff
@@ -8,6 +8,7 @@ const DEFAULT_LEVEL_LOG_SIZE: f64 = 0.75;
 const DEFAULT_MIN_LAYER_SIZE: u32 = 10_000;
 const DEFAULT_MIN_NUM_SEGMENTS_IN_MERGE: usize = 8;
 const DEFAULT_MAX_DOCS_BEFORE_MERGE: usize = 10_000_000;
+const DEFAULT_MAX_DEL_DOCS_PCT: u8 = 100;
```
shikhar: Lucene default for a similar knob is 33% https://github.com/apache/lucene/blob/c64e5fe84c4990968844193e3a62f4ebbba638ea/lucene/core/src/java/org/apache/lucene/index/TieredMergePolicy.java#L91

100% is effectively a no-op over the current policy. Lowering it to 33% causes some tests to fail, probably worth working through it though, if the approach makes sense!

src/indexer/index_writer.rs
```diff
@@ -974,7 +974,7 @@ mod tests {
         assert_eq!(
             format!("{:?}", index_writer.get_merge_policy()),
             "LogMergePolicy { min_num_segments: 8, max_docs_before_merge: 10000000, min_layer_size: 10000, \
-             level_log_size: 0.75 }"
+             level_log_size: 0.75, max_del_docs_pct: 100 }"
         );
```
PSeitz, line 978: I would prefer
max_del_docs_percentage

shikhar: Alternatively `max_del_docs_ratio: f32`?

PSeitz: I think max_del_docs_percentage is a little bit easier to understand. To be more precise maybe
`del_docs_percentage_before_merge`
 
I think we can drop the max part?

src/indexer/log_merge_policy.rs
```diff
@@ -82,7 +99,12 @@ impl MergePolicy for LogMergePolicy {
 
         levels
             .iter()
-            .filter(|level| level.len() >= self.min_num_segments)
+            .filter(|level| {
+                level.len() >= self.min_num_segments
+                    || level
+                        .iter()
+                        .any(|segment| deletes_pct(segment) > self.max_del_docs_pct)
```
PSeitz: Is it intended that we merge the whole level, if one segment is above the delete threshold?

I think it's better to assign them to variables first e.g.
```
let contains_segment_above_delete_threshold = level.iter().any(|segment| deletes_pct(segment) > self.max_del_docs_pct);
```

src/indexer/log_merge_policy.rs
PSeitz: I think we should lower it, not sure if 33% is not too early though.

src/indexer/log_merge_policy.rs
shikhar: I factored out a function `has_segment_above_deletes_threshold`

src/indexer/index_writer.rs
shikhar: `del_docs_percentage_before_merge` is more consistent with the current naming, renamed!

fulmicoton:
> Need to add additional test cases.
> 
> I tried it out at 10% and in initial usage it is seeming effective for us: image
> 
> vs previously: image

Seeing that kind of graph makes me happy :)

src/indexer/log_merge_policy.rs
```diff
@@ -2,12 +2,12 @@ use super::merge_policy::{MergeCandidate, MergePolicy};
 use crate::core::SegmentMeta;
 use itertools::Itertools;
 use std::cmp;
-use std::f64;
 
 const DEFAULT_LEVEL_LOG_SIZE: f64 = 0.75;
 const DEFAULT_MIN_LAYER_SIZE: u32 = 10_000;
 const DEFAULT_MIN_NUM_SEGMENTS_IN_MERGE: usize = 8;
 const DEFAULT_MAX_DOCS_BEFORE_MERGE: usize = 10_000_000;
+const DEFAULT_DEL_DOCS_PERCENTAGE_BEFORE_MERGE: u8 = 100;
```
fulmicoton:
```suggestion
// The default value of 100% means that deletes are not taken in account when
// identifying merge candidates. This is not a very sensible default: it was
// set like that for backward compatibility and might change in the near future.
const DEFAULT_DEL_DOCS_PERCENTAGE_BEFORE_MERGE: u8 = 100;
```

fulmicoton:
@shikhar LGTM! Thank you for taking care of that very old issue!
I'll open an issue to eventually change the issue.

fulmicoton (approved):

fulmicoton:
@shikhar @PSeitz

I added a bunch of unit test, and switch back to from (percentage, u8) into (ratio, f32) for different "soft reasons".

I had to make a change anyway, because in the previous implem, the the actual first value where we observed the switch was pretty unexpected. 
For 10,000 docs and a pct of 25%, the threshold was expressed as > 26,000 deleted docs and not > 25,000 deleted documents.  It is possible to fix it using only integer operations using the cross product, but then the code would not be exactly as simple.

PR #1241, opened 2021-12-26
Slop support for phrase queries

halvorboe:
Duplicate of https://github.com/quickwit-inc/tantivy/pull/1110 (because of permissions)

Implements slop for phrase queries (https://github.com/tantivy-search/tantivy/issues/1068).

Currently the phrase query only supports exact matching. The algorithm used does a set intersection between the position of the last term in the phrase and the positions of all the prior terms in the phrase offset by their position in the phrase.

Looked at the Lucene implementation and the description in "Information Retrieval: Implementing and evaluating search engines". 

I modified the current algorithm slightly to keep track of:
- The start of each candidate phrase. Stored in a separate array.
- The first possible end of the phrase. (searching for `a b .` in `a c b d e` it stores the 0 as the start and 1 as the end). There is never any benefit from picking a later instance of the term. This is stored in what was previously called left.

I'll test the efficiency and try to optimize, but the is no change in complexity.


src/query/phrase_query/phrase_scorer.rs
```diff
@@ -104,38 +105,104 @@ fn intersection_count(left: &[u32], right: &[u32]) -> usize {
...
             }
         }
     }
     count
 }
 
+/// Intersect twos sorted arrays `left` and `right` and outputs the
+/// resulting array in left.
+///
+/// Condition for match is that the value stored in left is less than
```
fulmicoton:
```suggestion
/// Condition for match is that the value stored in left is less than or equal to
```

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -104,38 +105,104 @@ fn intersection_count(left: &[u32], right: &[u32]) -> usize {
...
-                right_i += 1;
+                right_index += 1;
             }
         }
     }
     count
 }
 
+/// Intersect twos sorted arrays `left` and `right` and outputs the
+/// resulting array in left.
```
fulmicoton, line 135:
```suggestion
/// resulting array inline in both `left` and `right`.
```

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -104,38 +105,104 @@ fn intersection_count(left: &[u32], right: &[u32]) -> usize {
...
+            Ordering::Greater => {
+                right_index += 1;
+                // Given the constraint that left cannot be greater than right we know that the value in left is
+                // final.
+                is_temporary = false;
+            }
+        }
+    }
+    for i in 0..count {
+        left[i] = right[i];
```
fulmicoton: I'm confused... Why would we fill `left` and `right` , if it is to trash `left` in the end?

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -104,38 +105,104 @@ fn intersection_count(left: &[u32], right: &[u32]) -> usize {
...
+    let mut right_index = 0;
+    let mut count = 0;
+    let left_len = left.len();
+    let right_len = right.len();
+    // Is the current last value guaranteed to be the final value.
+    let mut is_temporary = false;
+    while left_index < left_len && right_index < right_len {
+        let left_val = left[left_index];
+        let right_val = right[right_index];
+        match left_val.cmp(&right_val) {
```
fulmicoton: The actual three cases are

- `left_val < right - slop`
- `right_slop <= left_val <= right`
- `left_val > right`

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -104,38 +105,104 @@ fn intersection_count(left: &[u32], right: &[u32]) -> usize {
...
+            Ordering::Less => {
+                if right_val - left[left_index] <= max_distance_to_begin {
+                    if is_temporary {
```
fulmicoton: Once the condition is switch to the ternary choice above, the is_temporary contraption can be avoided with an inner loop.

src/query/phrase_query/phrase_scorer.rs
fulmicoton: I think we are only interested in right.
Can't we just write right into left, and keep right non-mutable?

fulmicoton (changes requested):
See comments inline

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -138,61 +138,54 @@ fn intersection(left: &mut [u32], right: &[u32]) -> usize {
...
+        };
+
+        if left_val < right_slop {
+            left_index += 1;
+        } else if right_slop <= left_val && left_val <= right_val {
+            loop {
+                if left_index + 1 >= left_len {
+                    // current one is the best as there are no more values.
+                    break;
                 }
```
fulmicoton:
```suggestion
            while left_index + 1 < left_len {
```

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -138,61 +138,54 @@ fn intersection(left: &mut [u32], right: &[u32]) -> usize {
...
-                    is_temporary = true;
-                } else {
-                    left_index += 1;
+
+        /*
+         * The three conditions are:
+         * left_val < right_slop -> left index increment.
+         * right_slop <= left_val <= right -> find the best match.
+         * left_val > right -> right index increment.
+         */
```
fulmicoton:
```suggestion
        // The three conditions are:
        // left_val < right_slop -> left index increment.
        // right_slop <= left_val <= right -> find the best match.
        // left_val > right -> right index increment.
```

fulmicoton: Great comment

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -192,39 +270,49 @@ impl<TPostings: Postings> PhraseScorer<TPostings> {
...
-                return false;
+                return 0;
             }
         }
 
         self.intersection_docset
             .docset_mut_specialized(self.num_terms - 1)
             .positions(&mut self.right);
-        intersection_exists(&self.left[..intersection_len], &self.right[..])
+        intersection_len
```
fulmicoton, line 280: We are forgetting the last term here.

fulmicoton, line 281: I understand you are trying to be DRY here, but the contract of the `compute_match*` function is horrible.
This makes the code very difficult to understand.

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -104,38 +105,93 @@ fn intersection_count(left: &[u32], right: &[u32]) -> usize {
...
 }
 
+/// Intersect twos sorted arrays `left` and `right` and outputs the
+/// resulting array inline in both `left` and `right`.
+///
+/// Condition for match is that the value stored in left is less than or equal to
+/// the value in right and that the distance to the previous token is lte to the slop.
+///
+/// Returns the length of the intersection
+fn intersection_with_distance(left: &mut [u32], right: &[u32], slop: u32) -> usize {
```
fulmicoton: `intersection_with_slop` maybe?

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -104,38 +105,93 @@ fn intersection_count(left: &[u32], right: &[u32]) -> usize {
...
-                right_i += 1;
+                right_index += 1;
             }
         }
     }
     count
 }
 
+/// Intersect twos sorted arrays `left` and `right` and outputs the
+/// resulting array inline in both `left` and `right`.
```
fulmicoton: The result of the `right` array is stored in `left`

halvorboe:
@fulmicoton Do you have time to take a look?

fulmicoton (approved):

fulmicoton:
Thank you @halvorboe

PR #1244, opened 2022-01-01
Levenshtein distance score with custom score function (POC for #998)

Restioson:
Hi, I've tried to implement the ability to use a custom score function (`Fn(u8) -> f32`) for fuzzy queries, which take the DFA distance as the input and return the score for the term. This is a **proof of concept** for #998, and I am opening this as a draft so that people can more easily see the diff and comment on it. If the solution is good, it could perhaps be merged into @lerouxrgd's branch first, and then #998 could be merged, or whatever people see fit.

### DIFF
--- a/src/query/automaton_weight.rs
+++ b/src/query/automaton_weight.rs
@@ -1,5 +1,5 @@
 use crate::core::SegmentReader;
-use crate::query::fuzzy_query::DFAWrapper;
+use crate::query::fuzzy_query::DfaWrapper;
 use crate::query::score_combiner::SumCombiner;
 use crate::query::Explanation;
 use crate::query::{ConstScorer, Union};
@@ -8,7 +8,6 @@ use crate::schema::{Field, IndexRecordOption};
 use crate::termdict::{TermDictionary, TermWithStateStreamer};
 use crate::TantivyError;
 use crate::{DocId, Score};
-use common::BitSet;
 use std::any::{Any, TypeId};
 use std::io;
 use std::sync::Arc;
@@ -88,8 +87,8 @@ where
     A::State: Clone,
     F: Fn(u8) -> f32,
 {
-    if TypeId::of::<DFAWrapper>() == automaton.type_id() && TypeId::of::<u32>() == state.type_id() {
-        let dfa = automaton as *const A as *const DFAWrapper;
+    if TypeId::of::<DfaWrapper>() == automaton.type_id() && TypeId::of::<u32>() == state.type_id() {
+        let dfa = automaton as *const A as *const DfaWrapper;
         let dfa = unsafe { &*dfa };
 
         let id = &state as *const A::State as *const u32;
@@ -162,7 +161,7 @@ mod tests {
     fn test_automaton_weight() -> crate::Result<()> {
         let index = create_index()?;
         let field = index.schema().get_field("title").unwrap();
-        let automaton_weight = AutomatonWeight::new(field, PrefixedByA);
+        let automaton_weight = AutomatonWeight::new(field, PrefixedByA, |_| 1.0);
         let reader = index.reader()?;
         let searcher = reader.searcher();
         let mut scorer = automaton_weight.scorer(searcher.segment_reader(0u32), 1.0)?;
@@ -179,7 +178,7 @@ mod tests {
     fn test_automaton_weight_boost() -> crate::Result<()> {
         let index = create_index()?;
         let field = index.schema().get_field("title").unwrap();
-        let automaton_weight = AutomatonWeight::new(field, PrefixedByA);
+        let automaton_weight = AutomatonWeight::new(field, PrefixedByA, |_| 1.0);
         let reader = index.reader()?;
         let searcher = reader.searcher();
         let mut scorer = automaton_weight.scorer(searcher.segment_reader(0u32), 1.32)?;
--- a/src/query/automaton_weight.rs
+++ b/src/query/automaton_weight.rs
@@ -14,21 +14,24 @@ use std::sync::Arc;
 use tantivy_fst::Automaton;
 
 /// A weight struct for Fuzzy Term and Regex Queries
-pub struct AutomatonWeight<A> {
+pub struct AutomatonWeight<A, F> {
     field: Field,
     automaton: Arc<A>,
+    score_fn: F,
 }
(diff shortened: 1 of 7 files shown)


Restioson:
I'll run rustfmt on this in a moment

src/query/automaton_weight.rs
```diff
@@ -1,80 +1,78 @@
...
 use crate::{DocId, Score};
-use common::BitSet;
+use std::any::{Any, TypeId};
 use std::io;
 use std::sync::Arc;
 use tantivy_fst::Automaton;
 
 /// A weight struct for Fuzzy Term and Regex Queries
-pub struct AutomatonWeight<A> {
+pub struct AutomatonWeight<A, F> {
```
fulmicoton, line 17: This approach is not good...

The automaton weight is for regex and fuzzy queries, as mentionned above.
F has no meaning for regex queries.

Let's try something nicer.
We can have F match F: `Fn(A::State) -> Score` .

For better usability, we can also offer a default to the Const function that returns one.

The tricky part then, is that we want the user to pass an array over the distance. Not a function of the state.
Well, that's a job for the FuzzyQuery itself.

The FuzzyQuery has access to the automaton. It can therefore build a function that picks a state, computes the distance, and return a score.

fulmicoton (changes requested):
Add comment inline.

src/query/automaton_weight.rs
Restioson: >  F has no meaning for regex queries.

In this case, it can default to some no-op function like `|_| 1.0`

>  We can have F match F: Fn(A::State) -> Score .

I don't really understand this because there would still be an F for a regex query then, wouldn't there?

> For better usability, we can also offer a default to the Const function that returns one.

If `F` is a `Fn` then we cannot provide `struct AutomatonWeight<A, F = <some default>>` because we can't name `|_| 1.0`, nor implement it for a `struct DefaultScorer;` because that's unstable. Otherwise, a scorer trait could be added, which can then have a named default implementation, and be blanket implemented for Fn(A::State) -> Score.

Restioson: I also don't quite understand what you mean by "we want the user to pass an array over the distance, not a function of the state", if you are also suggesting `Fn(A::State) -> Score`. Apologies if I am misinterpreting this

Restioson:
Another question: is there any way to factor in the length of the suffix of the term (when using the query as the prefix) to the score? If possible it seems like something that should be exposed.

src/query/automaton_weight.rs
fulmicoton:
```
I don't really understand this because there would still be an F for a regex query then, wouldn't there?
```
Ah yes... :)
But there is one big notable difference.

The role of F can then be explained without necessary talking about the fuzzy term query.

```
 because we can't name |_| 1.0
```

> nor implement it for a struct DefaultScorer; because that's unstable

Which feature is unstable exactly?

> I also don't quite understand what you mean by "we want the user to pass an array over the distance, not a function of the state", if you are also suggesting Fn(A::State) -> Score. Apologies if I am misinterpreting this

We can keep the function.
The user would give the FuzzyTermQuery an array over the distance rather than a function. It's tiny anyway.

user gives `[f32; max_distance + 1]`
`FuzzyTermQuery` takes that an transforms into a `Fn(u32) -> f32`,
The implementation take a state, uses the automaton to get a distance, and do a lookup in the user array.

src/query/automaton_weight.rs
```diff
@@ -83,6 +81,26 @@ where
     }
 }
 
+fn automaton_score<A, F>(automaton: &A, state: A::State, score_fn: F) -> f32
+where
+    A: Automaton + Send + Sync + 'static,
+    A::State: Clone,
+    F: Fn(u8) -> f32,
+{
+    if TypeId::of::<DfaWrapper>() == automaton.type_id() && TypeId::of::<u32>() == state.type_id() {
```
fulmicoton, line 90: this it the code we want to get rid of.

fulmicoton:
> Another question: is there any way to factor in the length of the suffix of the term (when using the query as the prefix) to the score? If possible it seems like something that should be exposed.

Great question...
Factoring in the length of the overall term should be relatively easy.

Factoring in the length of the suffix is harder, but we can probably cook something if we dig deep in the levenshtein automata crate. I suggest we use the `length - querylength` as a first approximation. That should probably do the job.

One case where that will give us different result is 

-A: target: elephant, query: eleh
vs
-B: target: elephant, query: elepj 

In this case, B will get a better score. But A could be the deletion of a character bringing us to the actual same suffix size, same levenshtein distance.

src/query/automaton_weight.rs
Restioson: > Which feature is unstable exactly?}

Implementing a Fn trait is unstable: https://github.com/rust-lang/rust/issues/29625. This can be mitigated by making `trait AutomatonScorer { fn score(A::State) -> f32  }` or similar.

>  The user would give the FuzzyTermQuery an array over the distance rather than a function. It's tiny anyway.

So, in `FuzzyTermQuery` the user provides a function, and then at the AutomatonWeight boundary it is converted to a suitable `F`?

Restioson:
> I suggest we use the length - querylength as a first approximation. That should probably do the job.

This sounds good to me. How can we expose this to the user in such a way that they can factor it in?

fulmicoton:
@Restioson a funciton is good I think.

`Fn(lev_distance: u8, approximate_suffix_length: u32) -> f32`

We can cache the result for 
lev_distance = 0,1,2,3,>=3
and
approximate_suffix_length = 0..10 or something like that.

Restioson:
As far as caching goes, would it maybe be cleaner to let the user do their own caching? They could always wrap it with a memoisation function if they wanted to, perhaps? I'm just wondering if caching might be extraneous in a situation of just reading off a predefined const value table.

ChillFish8:
> As far as caching goes, would it maybe be cleaner to let the user do their own caching? They could always wrap it with a memoisation function if they wanted to, perhaps? I'm just wondering if caching might be extraneous in a situation of just reading off a predefined const value table.

I agree that caching should be left to the user, I imagine in most cases anyway the calculations are going to be pretty simple and easily optimised by the compiler.

PR #1246, opened 2022-01-03
Searcher warming API

shikhar:
`Warmer`s can be registered with the `IndexReaderBuilder` and used to maintain any segment-level caches effectively. By integrating with the reload mechanism, we can ensure that any new segments are warmed before they can be exercised via a new generation of `Searcher`s.

### DIFF
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -168,11 +168,23 @@ impl TryInto<IndexReader> for IndexReaderBuilder {
     }
 }
 
+#[derive(Debug, Clone)]
+struct SearcherGeneration {
+    liveness_token: Weak<()>,
+    segment_ids: Vec<SegmentId>,
+}
+
+impl SearcherGeneration {
+    fn is_live(&self) -> bool {
+        self.liveness_token.strong_count() > 0
+    }
+}
+
 /// Warming-related state with interior mutability.
 struct WarmingState {
     warmers: RwLock<Vec<Weak<dyn Warmer>>>,
     executor: Executor,
-    searcher_generations: RwLock<Vec<(Weak<()>, Vec<SegmentId>)>>,
+    searcher_generations: RwLock<Vec<SearcherGeneration>>,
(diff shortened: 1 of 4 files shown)


shikhar:
As context, also baking some abstractions inspired by Lucene and Solr:
* `ValueSource<V>`, abstracting over `Field` for `V=FastValue` Additionally implemented by an `AuxiliaryField` which maintains in-memory u64:V map with a specified fast field used for associating docID to u64 key and thus value.
* `FunctionQuery` that can operate on `ValueSource` with a Rust function that tests/scores a value.
* `CachedQuery` implementing the `Warmer` API that can maintain segment-level cache of a wrapped query's results (either const score as a bitmap, or with score in sparse|dense array), and warm stale/missing segment data.

fulmicoton (commented):
That sounds like a nice addition, but can you add unit tests and nice rustdoc so we can merge?

src/reader/mod.rs
```diff
@@ -143,7 +191,19 @@ impl InnerIndexReader {
         })
         .take(self.num_searchers)
         .collect::<io::Result<_>>()?;
+        let warmers = self.list_warmers();
+        self.warming_executor
+            .map(|warmer| warmer.warm(&searchers[0]), warmers.iter().cloned())?;
         self.searcher_pool.publish_new_generation(searchers);
```
shikhar, line 299: I would ideally like to ensure that old `Searcher`s have been dropped, so that we are not calling `retain()` early while there are still active searchable segments with previously leased searchers. Not sure how to architect that.

this test tweak fails: https://gist.github.com/314e5029010c42c4e39f9d91da4c7a1f

shikhar: Addressed in https://github.com/quickwit-inc/tantivy/pull/1246/commits/c9346cbf8aaf4160c1276bc81b5f5353bbf55ee0

shikhar:
plan to refactor & reopen

Issue #1232, opened 2021-12-07
Check and fix doc.rs pages


evanxg852000:
Right now the documentation page for `https://docs.rs/tantivy/latest/tantivy/termdict/type.TermDictionary.html` renders almost empty without listing methods and other item info. This could be the case for other types as well.

This task is to check the documentation rendered at  https://docs.rs/tantivy/latest/tantivy and fix pages that are not rendering documentation content.

saroh:
by public modules in the tantivy crate: 
- [x] tantivy https://github.com/quickwit-oss/tantivy/pull/1391
- [x] aggregation #1299 
- [x] collector #1247 
- [x] directory #1247 
- [x] error #1299 
- [x] fastfield #1301 
- [x] fieldnorm
- [x] merge_policy
- [x] positions https://github.com/quickwit-oss/tantivy/pull/1391
- [x] postings https://github.com/quickwit-oss/tantivy/pull/1391
- [x] query https://github.com/quickwit-oss/tantivy/pull/1391
- [x] schema https://github.com/quickwit-oss/tantivy/pull/1391
- [x] space_usage https://github.com/quickwit-oss/tantivy/pull/1391
- [x] store https://github.com/quickwit-oss/tantivy/pull/1391
- [x] termdict #1242 
- [x] tokenizer https://github.com/quickwit-oss/tantivy/pull/1391

PR #1247, opened 2022-01-03
Collector + directory doc fixes

saroh:
Just fixing a few links :) 

#1232

### DIFF
--- a/src/directory/directory.rs
+++ b/src/directory/directory.rs
@@ -43,7 +43,6 @@ impl RetryPolicy {
 }
 
 /// The `DirectoryLock` is an object that represents a file lock.
-/// See  [`LockType`](struct.LockType.html)
 ///
 /// It is transparently associated to a lock file, that gets deleted
 /// on `Drop.` The lock is released automatically on `Drop`.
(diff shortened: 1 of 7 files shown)


src/directory/directory.rs
```diff
@@ -43,7 +43,6 @@ impl RetryPolicy {
 }
 
 /// The `DirectoryLock` is an object that represents a file lock.
-/// See  [`LockType`](struct.LockType.html)
```
saroh, line 46: `DirectoryLockGuard` ? It's private though.

fulmicoton: No clue what it was

fulmicoton (approved):
Thank you @saroh 
Do you know if there is a command to check for broken links in the rustdoc?
thtat would be cool to add that to the CI

saroh:
> Do you know if there is a command to check for broken links in the rustdoc?

@fulmicoton seems supported https://doc.rust-lang.org/rustdoc/lints.html I'll give it a go for the future packages :)

saroh:
https://github.com/saroh/tantivy/commit/6ef8e20c122a7c49ff79cdb3bab3acd503d584cc should make `cargo doc` fail providing it's in your CI 
the doc passes as is.

PR #1248, opened 2022-01-03
Use test-log instead of test-env-log

d-e-s-o:
The test-env-log crate has been renamed to test-log to better reflect
its intent of not only catering to env_logger specific initialization
but also tracing (and potentially others in the future).
This change updates the crate to use test-log instead of the now
deprecated test-env-log.


fulmicoton (approved):
I confirmed the repository is from the same author as the previous crate and that the redirection exists.

PR #1254, opened 2022-01-06
Searcher warming API, take 2

shikhar:
`Warmer` implementations can be registered with the `IndexReaderBuilder` and used to maintain any segment-level caches effectively.

By integrating `Warmer::warm()` with the reload mechanism, we can ensure that any new segments are warmed before they can be exercised via a new generation of `Searcher`s.

`Warmer::garbage_collect()` is called when some searcher generation has been fully dropped, with the set of segment IDs that are still active. This allows implementations to prune redundant state.

### DIFF
--- a/src/core/searcher.rs
+++ b/src/core/searcher.rs
@@ -1,5 +1,6 @@
 use crate::collector::Collector;
 use crate::core::Executor;
+use crate::SegmentId;
 
 use crate::core::SegmentReader;
 use crate::query::Query;
@@ -11,8 +12,51 @@ use crate::store::StoreReader;
 use crate::DocAddress;
 use crate::Index;
 
+use std::collections::BTreeMap;
+use std::ops::Deref;
 use std::sync::Arc;
+use std::sync::Weak;
 use std::{fmt, io};
+
+/// Identifies the data accessed by a generation of [Searcher].
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct SearcherGeneration(BTreeMap<SegmentId, u32>);
+
+impl SearcherGeneration {
+    pub(crate) fn from_segment_readers(readers: &[SegmentReader]) -> Self {
+        let mut segment_id_to_del_docs = BTreeMap::new();
+        for reader in readers {
+            segment_id_to_del_docs.insert(reader.segment_id(), reader.num_deleted_docs());
+        }
+        Self(segment_id_to_del_docs)
+    }
+
+    /// Segment IDs represented in this generation.
+    pub fn segment_ids(&self) -> impl Iterator<Item = SegmentId> + '_ {
+        self.0.keys().copied()
+    }
+
+    /// Number of deleted documents for the specified segment ID if it is present in this generation.
+    pub fn num_deleted_docs(&self, segment_id: &SegmentId) -> Option<u32> {
+        self.0.get(segment_id).copied()
+    }
+}
+
+/// A token that is coupled with the lifetime of a [SearcherGeneration].
+#[derive(Debug, Clone)]
+pub struct SearcherGenerationToken(Weak<SearcherGeneration>);
+
+impl SearcherGenerationToken {
+    /// Whether the [SearcherGeneration] is still alive.
+    pub fn is_live(&self) -> bool {
+        self.0.strong_count() > 0
+    }
+
+    /// Access the [SearcherGeneration] if it is still alive.
+    pub fn access(&self) -> Option<SearcherGeneration> {
+        Weak::upgrade(&self.0).map(|gen| gen.deref().clone())
+    }
+}
 
 /// Holds a list of `SegmentReader`s ready for search.
 ///
@@ -24,7 +68,7 @@ pub struct Searcher {
     index: Index,
     segment_readers: Vec<SegmentReader>,
     store_readers: Vec<StoreReader>,
-    _liveness_token: Arc<()>,
+    generation: Arc<SearcherGeneration>,
 }
 
 impl Searcher {
@@ -33,7 +77,7 @@ impl Searcher {
         schema: Schema,
         index: Index,
         segment_readers: Vec<SegmentReader>,
-        liveness_token: Arc<()>,
+        generation: Arc<SearcherGeneration>,
     ) -> io::Result<Searcher> {
         let store_readers: Vec<StoreReader> = segment_readers
             .iter()
@@ -44,13 +88,23 @@ impl Searcher {
             index,
             segment_readers,
             store_readers,
-            _liveness_token: liveness_token,
+            generation,
         })
     }
 
     /// Returns the `Index` associated to the `Searcher`
     pub fn index(&self) -> &Index {
         &self.index
+    }
+
+    /// [SearcherGeneration] which uniquely identifies that data accessed by this generation of `Searcher`.
+    pub fn generation(&self) -> &SearcherGeneration {
+        &self.generation
+    }
+
+    /// [SearcherGenerationToken] which is coupled with the lifetime of this generation of `Searcher`.
+    pub fn generation_token(&self) -> SearcherGenerationToken {
+        SearcherGenerationToken(Arc::downgrade(&self.generation))
     }
 
     /// Fetches a document from tantivy's store given a `DocAddress`.
--- a/src/core/searcher.rs
+++ b/src/core/searcher.rs
@@ -1,6 +1,8 @@
 use crate::collector::Collector;
 use crate::core::Executor;
+use crate::Opstamp;
 use crate::SegmentId;
+use crate::SegmentMeta;
 
 use crate::core::SegmentReader;
 use crate::query::Query;
@@ -20,15 +22,17 @@ use std::{fmt, io};
 
 /// Identifies the index generation accessed by a [Searcher].
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct SearcherIndexGeneration(BTreeMap<SegmentId, u32>);
+pub struct SearcherIndexGeneration(BTreeMap<SegmentId, Option<Opstamp>>);
 
 impl SearcherIndexGeneration {
-    pub(crate) fn from_segment_readers(readers: &[SegmentReader]) -> Self {
-        let mut segment_id_to_del_docs = BTreeMap::new();
-        for reader in readers {
-            segment_id_to_del_docs.insert(reader.segment_id(), reader.num_deleted_docs());
+    pub(crate) fn from_segment_metas<'a>(
+        segment_metas: impl Iterator<Item = &'a SegmentMeta>,
+    ) -> Self {
+        let mut segment_id_to_del_opstamp = BTreeMap::new();
+        for meta in segment_metas {
+            segment_id_to_del_opstamp.insert(meta.id(), meta.delete_opstamp());
         }
-        Self(segment_id_to_del_docs)
+        Self(segment_id_to_del_opstamp)
     }
 
     /// Segment IDs represented in this generation.
@@ -36,9 +40,9 @@ impl SearcherIndexGeneration {
         self.0.keys().copied()
     }
 
-    /// Number of deleted documents for the specified segment ID if it is present in this generation.
-    pub fn num_deleted_docs(&self, segment_id: &SegmentId) -> Option<u32> {
-        self.0.get(segment_id).copied()
+    /// `Opstamp` of the last delete operation accounted for in the specified segment, if present.
+    pub fn delete_opstamp(&self, segment_id: &SegmentId) -> Option<Opstamp> {
+        self.0.get(segment_id).copied().flatten()
     }
 }
 
--- a/src/core/searcher.rs
+++ b/src/core/searcher.rs
@@ -11,6 +11,7 @@ use crate::store::StoreReader;
 use crate::DocAddress;
 use crate::Index;
 
+use std::sync::Arc;
 use std::{fmt, io};
 
 /// Holds a list of `SegmentReader`s ready for search.
@@ -23,6 +24,7 @@ pub struct Searcher {
     index: Index,
     segment_readers: Vec<SegmentReader>,
     store_readers: Vec<StoreReader>,
+    _liveness_token: Arc<()>,
 }
 
 impl Searcher {
@@ -31,6 +33,7 @@ impl Searcher {
         schema: Schema,
         index: Index,
         segment_readers: Vec<SegmentReader>,
+        liveness_token: Arc<()>,
     ) -> io::Result<Searcher> {
         let store_readers: Vec<StoreReader> = segment_readers
             .iter()
@@ -41,6 +44,7 @@ impl Searcher {
             index,
             segment_readers,
             store_readers,
+            _liveness_token: liveness_token,
         })
     }
 
--- a/src/core/searcher.rs
+++ b/src/core/searcher.rs
@@ -18,11 +18,11 @@ use std::sync::Arc;
 use std::sync::Weak;
 use std::{fmt, io};
 
-/// Identifies the data accessed by a generation of [Searcher].
+/// Identifies the index generation accessed by a [Searcher].
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct SearcherGeneration(BTreeMap<SegmentId, u32>);
+pub struct SearcherIndexGeneration(BTreeMap<SegmentId, u32>);
 
-impl SearcherGeneration {
+impl SearcherIndexGeneration {
     pub(crate) fn from_segment_readers(readers: &[SegmentReader]) -> Self {
         let mut segment_id_to_del_docs = BTreeMap::new();
         for reader in readers {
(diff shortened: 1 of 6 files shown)


shikhar:
hello, this should be ready for review now!

src/reader/mod.rs
```diff
@@ -107,11 +112,27 @@ impl IndexReaderBuilder {
         self
     }
 
-    /// Sets the number of `Searcher` in the searcher pool.
+    /// Sets the number of [Searcher] to pool.
+    ///
+    /// When more than `num_searchers` are requested, the caller will block.
```
fulmicoton:
```suggestion
    /// See [Self::searcher()].
```
This might seem like a weird suggestion. The documentation seemed to suggest that the called to `num_searchers()` would block.

src/reader/mod.rs
```diff
@@ -107,11 +112,27 @@ impl IndexReaderBuilder {
...
     pub fn num_searchers(mut self, num_searchers: usize) -> IndexReaderBuilder {
         self.num_searchers = num_searchers;
         self
     }
+
+    /// Set the [Warmer]s that are invoked when reloading searchable segments.
+    pub fn warmers(mut self, warmers: Vec<Weak<dyn Warmer>>) -> IndexReaderBuilder {
```
fulmicoton, line 124:
```suggestion
    pub fn set_warmers(mut self, warmers: Vec<Weak<dyn Warmer>>) -> IndexReaderBuilder {
```

src/reader/warming.rs
```diff
@@ -0,0 +1,384 @@
...
+            let weak_inner = Arc::downgrade(this);
+            let handle = std::thread::Builder::new()
+                .name("tantivy-warm-gc".to_owned())
+                .spawn(|| Self::gc_loop(weak_inner))
+                .map_err(|_| {
+                    TantivyError::SystemError("Failed to spawn warmer GC thread".to_owned())
+                })?;
+            self.gc_thread = Some(handle);
+            Ok(true)
+        }
```
fulmicoton:
```suggestion
        if self.gc_thread.is_some() {
            return Ok(false);
        }
        let weak_inner = Arc::downgrade(this);
        let handle = std::thread::Builder::new()
            .name("tantivy-warm-gc".to_owned())
            .spawn(|| Self::gc_loop(weak_inner))
            .map_err(|_| {
               TantivyError::SystemError("Failed to spawn warmer GC thread".to_owned())
             })?;
        self.gc_thread = Some(handle);
        Ok(true)
```

fulmicoton: Super minor. The code style in tantivy is quite a mess, but we are trying to move to early returns.

fulmicoton:
@shikhar Great job. But before merging this, I'd like to discuss an alternate design to drop the need for a gc thread within tantivy, and possibly ease the user-side usage.

My understanding is that warmers compute some kind of byproduct associated to a searcher or a segment readers.
For instance, some statistics, or a cache, or whatever.

We cannot just assume that any searcher that we use is in sync with the last state of the warmer due. 
For instance, the following sequence of action is possible
- acquire searcher  gen N
- new generation of searcher. last warmer state is gen N+1
- actual use of searcher, using the warmer state

Your GC approach made it possible to handle the lifecycles of the warmer byproduct (which is great). 
However it does not help the user in a possession of a searcher to access the right warmer byproduct.

If the byproduct is actually a per-segment object it is possible to key these byproducts by segment id+delete opstamp.
If it is searcher level, the set of (segment_id + delete opstamp) and their is the right identifier.

This is possible. But this might be tricky.

As an alternative, should we could associate that byproduct directly with the Searcher with an Arc shared by all searcher?
`Drop` would take care of the cleaning up if needed (sometimes it might mean file deletion).
(Error handling would not be possible, but well it is not done either today).

That would solve:
- cleanup for things living in RAM
- the problem of accessing the right byproduct from the  searcher.

Of course the type system will not be nice with us...
But we could do some dynamic magic,  and use `Any` or even use a `typedmap`.

```
let my_cache: MyWarmer::ByProduct = searcher.warmer_byproduct(&warmer)
```

shikhar:
Thanks for the detailed feedback @fulmicoton 

I see your point about deletes being the mutable layer on top of immutable segments, making it possibly incorrect to treat warmer byproducts as immutable for a given segment...

I believe this won't affect the `CachedQuery` that I was developing because it does not bother with the alive bitset in accumulating doc sets for caching by `SegmentId` (not exercising `Searcher` or `Collector`, iterating over underlying query's scorers directly [1]). AFAIU, [deletes would still be respected eventually using the live `Searcher` during collection](https://github.com/quickwit-inc/tantivy/blob/732f6847c0632bc086ea37f39d5f9d623aaa473d/src/collector/mod.rs#L181-L187).

The incremental indexing interval is sub-minute so I was hoping to maximize cache effectiveness by keeping it segment-level (keyed by `SegmentId`) rather than index-level (keyed by some `Searcher` generation token). There is a decent rate of clobbering old documents so delete opstamp in the key would probably cause similar churn.

Wondering if there is an API possibility that allows for safe segment-level caching. Would it suffice to warn users in the docs   about ignoring deletes for their byproducts?


[1]  ED: just checked and do need to pass `Searcher` to the underlying `query.weight()`..

shikhar:
Maybe changing signature to

`warm(readers: &SegmentReader)`

where readers have been opened/cloned with empty `AliveSet`s

src/reader/warming.rs
```diff
@@ -0,0 +1,383 @@
...
+    fn is_live(&self) -> bool {
+        self.liveness_token.strong_count() > 0
+    }
+}
+
+#[derive(Default)]
+struct WarmingStateInner {
+    num_warming_threads: usize,
+    warmers: Vec<Weak<dyn Warmer>>,
+    searcher_generations: Vec<SearcherGeneration>,
```
shikhar: Just came across the https://docs.rs/census/latest/census/ crate, which tantivy already uses...

fulmicoton:
> I see your point about deletes being the mutable layer on top of immutable segments, making it possibly incorrect to treat warmer byproducts as immutable for a given segment...

And if the byproduct is a function not of a segment, but of the searcher entirely, we make it quite difficult to the user.

> The incremental indexing interval is sub-minute so I was hoping to maximize cache effectiveness by keeping it segment-level (keyed by SegmentId)

There are a lot of use cases for which segment-level byproducts is the most reasonable approach.
In your case, you can rely on the DeleteBitset to save the day and that's great. But someone who wants to pre-compute aggregates (average of price per product type for instance) would not be able to just key by segment.

I think we must leave it to the user to decide what they want to key their cache on.
This is not something I am disputing here.

Let's come back to your case...
You could keep warming up at the segment level.

In your case, the warmed-up by products would probably live in a cache (global or attached to your Warmer object) that goes :
```
HashMap<SegmentId, Weak<SegmentLevelCache>>
```

The warmer would first check if the byproduct is already cached.
If it is isn't, it would produce it and add it to the cache.
The byproduct itself would be an `HashMap<SegmentId, Arc<SegmentLevelCache>>` for all of the segments.

GC would be replaced by the drop of the last `Arc<SegmentLevelCache>` for a given `SegmentId`.

On consumption, you could get access to your data by requesting it from your `Searcher`, or through your cache (if you decided to make it global for instance).

shikhar:
I made updates to identify the data accessed by searcher generations more accurately, using a map from `SegmentId -> NumDeletedDocs`. This is different from delete `Opstamp` but I think gives us the same effect while exposing a more meaningful value, since  number of deleted docs can only grow with new generations. Happy to change this to `Opstamp` if that seems preferable!

A use-case which is aggregating can retrieve `Searcher::generation()` and use that as a cache key at the index-level, or `(SegmentId, NumDeletedDocs)` at the segment-level.

And `SegmentId` is a reasonable cache key if storing document-level info - I was referring to this as segment-level before, but I think the distinguishing feature is it is at the `DocId` level and hence delete filtering during eventual collection can be effective.

---

I am still trying to understand your alternate API suggestion where a GC thread is not needed. Sounds like the idea is to move ownership of the byproducts to the `Searcher`. Would you mind explaining it again? 😅 

> The byproduct itself would be an HashMap<SegmentId, Arc<SegmentLevelCache>> for all of the segments.

`for all the searchers`, right?

p.s. learned about `typedmap` from your mention, pretty cool

shikhar:
BTW to make the usage a bit more concrete, here are some of the types in my `CachedQuery` (still in flux)

```rust
type DataGeneration = u64;

// Implemented for `FunctionQuery` over a `ValueSource`, that can filter and optionally score using a Rust function over some per-doc value.
// `ValueSource` is implemented for a type providing mutable in-memory key:value map (also for `Field` assuming it's a fastfield...)
// the `ValueSource` generation is incremented when the data changes (bulk replacement to start with via files in object storage, but can also stream updates to a concurrent map)
// this allows joining against data which changes very frequently, ala Solr EFFs (https://solr.apache.org/guide/8_11/working-with-external-files-and-processes.html)
pub trait CacheableQuery: Query + Clone {
    fn generation(&self) -> DataGeneration;
}

enum Matches {
    Empty,
    Sparse(Arc<[(DocId, Score)]>),
    Dense(Arc<[Score]>),
    Const(Score, Arc<RoaringBitmap>),
}

type Cache = DashMap<SegmentId, (DataGeneration, Matches)>;

// scorer() returns a Matches-variant-specific `Scorer`
pub struct CachedQuery<Q>
where
    Q: CacheableQuery,
{
    query: Q,
    scoring_enabled: bool,
    cache: Arc<Cache>,
}

impl<Q> Warmer for CachedQuery<Q>
where
    Q: CacheableQuery
{
// Warming updates cache if a SegmentId is missing OR data generation for cached data is out-of-date.
// GC removes non-live SegmentId from cache
}

```

src/core/searcher.rs
```diff
@@ -11,8 +12,52 @@ use crate::store::StoreReader;
...
 
+/// Identifies the index generation accessed by a [Searcher].
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct SearcherIndexGeneration(BTreeMap<SegmentId, u32>);
+
+impl SearcherIndexGeneration {
+    pub(crate) fn from_segment_readers(readers: &[SegmentReader]) -> Self {
+        let mut segment_id_to_del_docs = BTreeMap::new();
+        for reader in readers {
+            segment_id_to_del_docs.insert(reader.segment_id(), reader.num_deleted_docs());
```
fulmicoton: Ah interesting. Technically this is correct because we cannot undelete a message, but I'd rather we used the `delete_opstamp` because this is what we use to identify a segment version in the code.
I'm ok to consider refactoring tantivy to `num_deleted_docs` though if we see some cool benefits.

That value is accessible from the `SegmentMeta`.

src/core/searcher.rs
```diff
@@ -11,8 +12,52 @@ use crate::store::StoreReader;
...
+        }
+        Self(segment_id_to_del_docs)
+    }
+
+    /// Segment IDs represented in this generation.
+    pub fn segment_ids(&self) -> impl Iterator<Item = SegmentId> + '_ {
+        self.0.keys().copied()
```
fulmicoton, line 40: Oh I knew cloned, but I didn't know about `.copied()`. Well now I am not sure of the benefits of using `.copied()` instead of `.cloned()`. It tells the reader it is "just a copy"?

src/core/searcher.rs
fulmicoton: I think the change is simple, rename the function `from_segment_metas` and pass it SegmentMetas.

The call site already has them.

src/reader/mod.rs
```diff
@@ -138,12 +161,21 @@ impl InnerIndexReader {
                 .map(SegmentReader::open)
                 .collect::<crate::Result<_>>()?
         };
+        let searcher_index_generation = Arc::new(SearcherIndexGeneration::from_segment_readers(
```
fulmicoton: searchable_segments above is your segment meta.

The reason for the weird lock is to make sure merge+GC does not remove files as we are in the middle of opening segments.

src/reader/warming.rs
```diff
@@ -0,0 +1,353 @@
...
+    #[cfg(test)]
+    fn gc_maybe(&self) -> bool {
+        self.0.lock().unwrap().gc_maybe()
+    }
+}
+
+struct WarmingStateInner {
+    num_warming_threads: usize,
+    warmers: Vec<Weak<dyn Warmer>>,
+    searcher_generations: HashMap<SearcherIndexGeneration, Vec<SearcherGenerationToken>>,
```
fulmicoton, line 56: I am confused. Why do we need `Vec<SearcherGenerationToken>` here?

Isn't `HashMap<SearcherIndexGeneration, SearcherGenerationToken>` suficient ?
I think all `SearcherGenerationToken` for a given SearcherIndexGeneration are identical `Weak` reference coupled to the same set of `Arc` clones.

fulmicoton: If the above is correct, the HashMap is useless too...
It would become 

`Vec<SearcherGenerationToken>`

src/reader/warming.rs
```diff
@@ -0,0 +1,353 @@
...
+            warmers,
+            searcher_generations: HashMap::new(),
+            gc_thread: None,
+        }))))
+    }
+
+    /// Start tracking a new generation of [Searcher], and [Warmer::warm] it if there are active warmers.
+    ///
+    /// A background GC thread for [Warmer::garbage_collect] calls is uniquely created if there are active warmers.
+    pub fn new_searcher_generation(&self, searcher: &Searcher) -> crate::Result<()> {
```
fulmicoton: Rename `warm_new_searcher_generation`

src/reader/warming.rs
```diff
@@ -0,0 +1,353 @@
...
+    warmers: Vec<Weak<dyn Warmer>>,
+    searcher_generations: HashMap<SearcherIndexGeneration, Vec<SearcherGenerationToken>>,
+    gc_thread: Option<JoinHandle<()>>,
+}
+
+impl WarmingStateInner {
+    /// Start tracking provided searcher as an exemplar of a new generation.
+    /// If there are active warmers, warm them with the provided searcher, and kick background GC thread if it has not yet been kicked.
+    /// Otherwise, prune state for dropped searcher generations inline.
+    fn new_searcher_generation(
```
fulmicoton: Rename `warm_new_search_generation`

src/reader/warming.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    /// Prune dropped searcher generations from our state.
+    /// Return count of removed generations.
+    fn prune_searcher_generations(&mut self) -> usize {
+        for tokens in self.searcher_generations.values_mut() {
+            // This can be implemented neatly using Vec::drain_filter() when that is stable.
+            let mut i = 0;
+            while i < tokens.len() {
+                if !tokens[i].is_live() {
+                    tokens.remove(i);
```
fulmicoton: Performance does not matter but well.... `swap_remove` is more indicated here..

fulmicoton (changes requested):
See comments inline

fulmicoton:
@shikhar  Let's forget about my suggestion and focus on landing your PR as is. I made a couple of comments however.

src/reader/warming.rs
shikhar: > I think all `SearcherGenerationToken` for a given `SearcherIndexGeneration` are identical `Weak` reference coupled to the same set of Arc clones.

I believe `reload()` on an unchanged index can result in distinct sets of `Searcher` with the same underlying `SearcherIndexGeneration`?

src/reader/warming.rs
shikhar: ooh, makes sense!

src/core/searcher.rs
shikhar: Yes, I guess that's the only benefit since `Clone` will (unless manually implemented against expectations...) also just copy for a `Copy` type.

src/core/searcher.rs
shikhar: > I'm ok to consider refactoring tantivy to num_deleted_docs though if we see some cool benefits.

Hmm, I think the only benefit is that it is a bit more meaningful in interpreting, opstamp feels opaque. Also `Option<u64>` vs `u32` - there is no need for optionality.

I changed it to the delete opstamp for consistency.

src/reader/mod.rs
shikhar: Factored it out into new method `open()` that returns the generation & readers

src/core/searcher.rs
fulmicoton: Yes that's interesting. I opened #1257.
I am not entirely sure there are no negative side effects, but on the paper that looks like a good idea.

src/reader/warming.rs
fulmicoton: That's true. They will be identical though...

In your use case, your warmed data is keyed by `SegmentId` only and it might not really matter much... But let's dig a little bit.

One could imagine a Warmer that grabs updated values from an external source (say price from a SQL db), grabs the docid->someprimaryid from a fast field and creates a column `docid -> newprice using the `Warmer`. 
The Searcher is supposed to a consistent view of the index throughout its lifetime.

Calling reload manually, even without any change in the index is meaningful in that world: it updates the price.

So we want the API to allow:
- Warming stuff, and keeping things per `(Segment, DeleteOpstamp)`
- Warming stuff, and keeping things per `Searcher generation`.

fulmicoton:
@shikhar I added one commit here:
https://github.com/quickwit-inc/tantivy/pull/1258/

    
- Using Inventory for the SearcherIndexGeneration
- Added an example.
- Added a searcher generation to warmer implementers to rely on that granularity
- Exporting census objects.
- Updated Changelog.

Can you have a look?

src/reader/warming.rs
shikhar: > Calling reload manually, even without any change in the index is meaningful in that world

`warm()` will indeed get called -- it's a redundant `garbage_collect()` call that can be avoided when one of those `SearcherGenerationToken`s that point to different object but equal `SearcherIndexGeneration` are dropped. But I think not that big a deal to have a redundant call, as we always provide the set of live `SearcherIndexGeneration` to test state against.

shikhar:
Superseded by #1258

PR #1258, opened 2022-01-19
Warming pr

fulmicoton:
Warmer implementations can be registered with the IndexReaderBuilder and used to maintain any segment-level caches effectively.

By integrating Warmer::warm() with the reload mechanism, we can ensure that any new segments are warmed before they can be exercised via a new generation of Searchers.

Warmer::garbage_collect() is called when some searcher generation has been fully dropped, with the set of segment IDs that are still active. This allows implementations to prune redundant state.

### DIFF
--- a/examples/warmer.rs
+++ b/examples/warmer.rs
@@ -1,0 +1,224 @@
+use std::cmp::Reverse;
+use std::collections::{HashMap, HashSet};
+use std::sync::{Arc, RwLock, Weak};
+
+use tantivy::collector::TopDocs;
+use tantivy::fastfield::FastFieldReader;
+use tantivy::query::QueryParser;
+use tantivy::schema::{Field, Schema, FAST, TEXT};
+use tantivy::{doc, DocAddress, DocId, Index, IndexReader, SegmentReader, TrackedObject};
+use tantivy::{Opstamp, Searcher, SearcherIndexGeneration, SegmentId, Warmer};
+
+// This example shows how warmers can be used to
+// load a values from an external sources using the Warmer API.
+//
+// In this example, we assume an e-commerce search engine.
+
+type ProductId = u64;
+
+/// Price
+type Price = u32;
+
+pub trait PriceFetcher: Send + Sync + 'static {
+    fn fetch_prices(&self, product_ids: &[ProductId]) -> Vec<Price>;
+}
+
+struct DynamicPriceColumn {
+    field: Field,
+    price_cache: RwLock<HashMap<(SegmentId, Option<Opstamp>), Arc<Vec<Price>>>>,
+    price_fetcher: Box<dyn PriceFetcher>,
+}
+
+impl DynamicPriceColumn {
+    pub fn with_product_id_field<T: PriceFetcher>(field: Field, price_fetcher: T) -> Self {
+        DynamicPriceColumn {
+            field,
+            price_cache: Default::default(),
+            price_fetcher: Box::new(price_fetcher),
+        }
+    }
+
+    pub fn price_for_segment(&self, segment_reader: &SegmentReader) -> Option<Arc<Vec<Price>>> {
+        let segment_key = (segment_reader.segment_id(), segment_reader.delete_opstamp());
+        self.price_cache.read().unwrap().get(&segment_key).cloned()
+    }
+}
+impl Warmer for DynamicPriceColumn {
+    fn warm(&self, searcher: &Searcher) -> tantivy::Result<()> {
+        for segment in searcher.segment_readers() {
+            let key = (segment.segment_id(), segment.delete_opstamp());
+            let product_id_reader = segment.fast_fields().u64(self.field)?;
+            let product_ids: Vec<ProductId> = segment
+                .doc_ids_alive()
+                .map(|doc| product_id_reader.get(doc))
+                .collect();
+            let mut prices_it = self.price_fetcher.fetch_prices(&product_ids).into_iter();
+            let mut price_vals: Vec<Price> = Vec::new();
+            for doc in 0..segment.max_doc() {
+                if segment.is_deleted(doc) {
+                    price_vals.push(0);
+                } else {
+                    price_vals.push(prices_it.next().unwrap())
+                }
+            }
+            self.price_cache
+                .write()
+                .unwrap()
+                .insert(key, Arc::new(price_vals));
+        }
+        Ok(())
+    }
+
+    fn garbage_collect(&self, live_generations: &[TrackedObject<SearcherIndexGeneration>]) {
+        let live_segment_id_and_delete_ops: HashSet<(SegmentId, Option<Opstamp>)> =
+            live_generations
+                .iter()
+                .flat_map(|gen| gen.segments())
+                .map(|(&segment_id, &opstamp)| (segment_id, opstamp))
+                .collect();
+        let mut price_cache_wrt = self.price_cache.write().unwrap();
+        // let price_cache = std::mem::take(&mut *price_cache_wrt);
+        // Drain would be nicer here.
+        *price_cache_wrt = std::mem::take(&mut *price_cache_wrt)
+            .into_iter()
+            .filter(|(seg_id_and_op, _)| !live_segment_id_and_delete_ops.contains(seg_id_and_op))
+            .collect();
+    }
+}
+
+/// For the sake of this example, the table is just an editable HashMap behind a RwLock.
+/// This map represents a map (ProductId -> Price)
+///
+/// In practise, it could be fetching things from an external service, like a SQL table.
+///
+#[derive(Default, Clone)]
+pub struct ExternalPriceTable {
+    prices: Arc<RwLock<HashMap<ProductId, Price>>>,
+}
+
+impl ExternalPriceTable {
+    pub fn update_price(&self, product_id: ProductId, price: Price) {
+        let mut prices_wrt = self.prices.write().unwrap();
+        prices_wrt.insert(product_id, price);
+    }
+}
+
+impl PriceFetcher for ExternalPriceTable {
+    fn fetch_prices(&self, product_ids: &[ProductId]) -> Vec<Price> {
+        let prices_read = self.prices.read().unwrap();
+        product_ids
+            .iter()
+            .map(|product_id| prices_read.get(product_id).cloned().unwrap_or(0))
+            .collect()
+    }
+}
+
+fn main() -> tantivy::Result<()> {
+    // Declaring our schema.
+    let mut schema_builder = Schema::builder();
+    // The product id is assumed to be a primary id for our external price source.
+    let product_id = schema_builder.add_u64_field("product_id", FAST);
+    let text = schema_builder.add_text_field("text", TEXT);
+    let schema: Schema = schema_builder.build();
+
+    let price_table = ExternalPriceTable::default();
+    let price_dynamic_column = Arc::new(DynamicPriceColumn::with_product_id_field(
+        product_id,
+        price_table.clone(),
+    ));
+    price_table.update_price(OLIVE_OIL, 12);
+    price_table.update_price(GLOVES, 13);
+    price_table.update_price(SNEAKERS, 80);
+
+    const OLIVE_OIL: ProductId = 323423;
+    const GLOVES: ProductId = 3966623;
+    const SNEAKERS: ProductId = 23222;
+
+    let index = Index::create_in_ram(schema);
+    let mut writer = index.writer_with_num_threads(1, 10_000_000)?;
+    writer.add_document(doc!(product_id=>OLIVE_OIL, text=>"cooking olive oil from greece"))?;
+    writer.add_document(doc!(product_id=>GLOVES, text=>"kitchen gloves, perfect for cooking"))?;
+    writer.add_document(doc!(product_id=>SNEAKERS, text=>"uber sweet sneakers"))?;
+    writer.commit()?;
+
+    let warmers: Vec<Weak<dyn Warmer>> = vec![Arc::downgrade(
+        &(price_dynamic_column.clone() as Arc<dyn Warmer>),
+    )];
+    let reader: IndexReader = index
+        .reader_builder()
+        .warmers(warmers)
+        .num_searchers(1)
+        .try_into()?;
+    reader.reload()?;
+
+    let query_parser = QueryParser::for_index(&index, vec![text]);
+    let query = query_parser.parse_query("cooking")?;
+
+    let searcher = reader.searcher();
+    let score_by_price = move |segment_reader: &SegmentReader| {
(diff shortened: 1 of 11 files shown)


src/reader/mod.rs
```diff
@@ -124,33 +156,53 @@ impl TryInto<IndexReader> for IndexReaderBuilder {
...
+        // Prevents segment files from getting deleted while we are in the process of opening them
+        let _meta_lock = self.index.directory().acquire_lock(&META_LOCK)?;
+        let searchable_segments = self.index.searchable_segments()?;
+        let generation = self
+            .searcher_generation_counter
+            .fetch_add(1, atomic::Ordering::Relaxed);
+        let searcher_generation = SearcherIndexGeneration::from_segment_metas(
+            searchable_segments.iter().map(|segment| segment.meta()),
+            generation,
+        );
```
shikhar: Minor, not a big deal either way: now that `SegmentReader` provides `delete_opstamp()`, we could  `from_segment_readers(&[SegmentReader])` using the `segment_readers` vec.

src/reader/warming.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    /// [Warmer::garbage_collect] active warmers if some searcher generation is observed to have been dropped.
+    fn gc_maybe(&mut self) -> bool {
+        // TODO avoid GC if no gen was deleted
+        let live_generations = self.searcher_generation_inventory.list();
+        let live_generation_ids: HashSet<u64> = live_generations
+            .iter()
+            .map(|searcher_generation| searcher_generation.generation_id())
+            .collect();
+        let warmup_not_required = self
```
shikhar: `gc_not_required` :-)

src/core/searcher.rs
```diff
@@ -10,9 +9,62 @@ use crate::space_usage::SearcherSpaceUsage;
...
+use crate::SegmentId;
+use crate::SegmentMeta;
+use crate::TrackedObject;
 
+use std::collections::BTreeMap;
 use std::{fmt, io};
 
+/// Identifies the index generation accessed by a [Searcher].
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct SearcherIndexGeneration {
```
shikhar: Shall we name this back to `SearcherGeneration`? When it was wrapping `segments` it was only concerned with index state, but with `generation_id` it also contains an identifier for a generation of `Searcher`.

src/reader/warming.rs
```diff
@@ -0,0 +1,341 @@
...
+            .flat_map(|weak_warmer| weak_warmer.upgrade())
+            .collect::<Vec<_>>();
+        self.warmers = strong_warmers.iter().map(Arc::downgrade).collect();
+        strong_warmers
+    }
+
+    /// [Warmer::garbage_collect] active warmers if some searcher generation is observed to have been dropped.
+    fn gc_maybe(&mut self) -> bool {
+        // TODO avoid GC if no gen was deleted
+        let live_generations = self.searcher_generation_inventory.list();
```
shikhar, line 106: much neater thanks to `census`, makes a lot of sense to reuse it

examples/warmer.rs
```diff
@@ -0,0 +1,224 @@
...
+use tantivy::fastfield::FastFieldReader;
+use tantivy::query::QueryParser;
+use tantivy::schema::{Field, Schema, FAST, TEXT};
+use tantivy::{doc, DocAddress, DocId, Index, IndexReader, SegmentReader, TrackedObject};
+use tantivy::{Opstamp, Searcher, SearcherIndexGeneration, SegmentId, Warmer};
+
+// This example shows how warmers can be used to
+// load a values from an external sources using the Warmer API.
+//
+// In this example, we assume an e-commerce search engine.
```
shikhar, line 15: Awesome, thanks for adding this!

src/reader/warming.rs
```diff
@@ -0,0 +1,341 @@
...
+            .iter()
+            .flat_map(|weak_warmer| weak_warmer.upgrade())
+            .collect::<Vec<_>>();
+        self.warmers = strong_warmers.iter().map(Arc::downgrade).collect();
+        strong_warmers
+    }
+
+    /// [Warmer::garbage_collect] active warmers if some searcher generation is observed to have been dropped.
+    fn gc_maybe(&mut self) -> bool {
+        // TODO avoid GC if no gen was deleted
```
shikhar: I agree with not doing this right now, but to be more explicit "avoid GC if the only generations dropped are semantically equivalent to a live generation"

src/reader/mod.rs
```diff
@@ -124,33 +156,53 @@ impl TryInto<IndexReader> for IndexReaderBuilder {
...
     }
 
-    /// Returns the list of segments that are searchable
-    fn searchable_segments(&self) -> crate::Result<Vec<Segment>> {
-        self.index.searchable_segments()
+    fn open(&self) -> crate::Result<(TrackedObject<SearcherIndexGeneration>, Vec<SegmentReader>)> {
+        // Prevents segment files from getting deleted while we are in the process of opening them
+        let _meta_lock = self.index.directory().acquire_lock(&META_LOCK)?;
+        let searchable_segments = self.index.searchable_segments()?;
+        let generation = self
```
shikhar: `generation_id`

src/core/searcher.rs
```diff
@@ -10,9 +9,62 @@ use crate::space_usage::SearcherSpaceUsage;
...
+        self.segments.get(segment_id).copied().flatten()
+    }
+
+    /// Return an iterator over the `(SegmentId, Option<Opstamp>)`.
+    pub fn segments(&self) -> &BTreeMap<SegmentId, Option<Opstamp>> {
+        &self.segments
+    }
+
+    /// Returns an ID identifying a searcher generation uniquely.
+    pub fn searcher_gen(&self) -> u64 {
```
shikhar: just `id()`?

src/core/searcher.rs
```diff
@@ -10,9 +9,62 @@ use crate::space_usage::SearcherSpaceUsage;
...
+    pub fn generation_id(&self) -> u64 {
+        self.generation_id
+    }
+
+    /// `Opstamp` of the last delete operation accounted for in the specified segment, if present.
+    pub fn delete_opstamp(&self, segment_id: &SegmentId) -> Option<Opstamp> {
+        self.segments.get(segment_id).copied().flatten()
+    }
+
+    /// Return an iterator over the `(SegmentId, Option<Opstamp>)`.
```
shikhar: `s/iterator over/mapping from/`?

src/core/searcher.rs
shikhar: Ah actually this fn looks like an accidental a dupe of `generation_id()` above

src/core/searcher.rs
```diff
@@ -10,9 +9,62 @@ use crate::space_usage::SearcherSpaceUsage;
...
+
+    /// Returns the searcher generation id.
+    pub fn generation_id(&self) -> u64 {
+        self.generation_id
+    }
+
+    /// `Opstamp` of the last delete operation accounted for in the specified segment, if present.
+    pub fn delete_opstamp(&self, segment_id: &SegmentId) -> Option<Opstamp> {
+        self.segments.get(segment_id).copied().flatten()
+    }
```
shikhar: I'm good with omitting `segment_ids()` and `delete_opstamp()`, users can get it directly from `segments()`.

shikhar (approved):
Looks great, just minor comments!

src/reader/warming.rs
```diff
@@ -0,0 +1,341 @@
...
+    fn gc_maybe(&self) -> bool {
+        self.0.lock().unwrap().gc_maybe()
+    }
+}
+
+struct WarmingStateInner {
+    num_warming_threads: usize,
+    warmers: Vec<Weak<dyn Warmer>>,
+    gc_thread: Option<JoinHandle<()>>,
+    searcher_generation_ids: HashSet<u64>,
```
shikhar: Maybe `warmed_generation_ids` is more descriptive? (and update `warm_new_searcher_generation()` to only insert into it after warmup...)

src/reader/mod.rs
fulmicoton: Yes... By the way I had to add it, because it is necessary on the consumer side.

fulmicoton: Addressed.

src/reader/warming.rs
fulmicoton: ah thanks for catching that.

src/core/searcher.rs
fulmicoton: Yes I think that would be better.

src/reader/mod.rs
fulmicoton: Renamed

src/reader/warming.rs
fulmicoton: I need to drop this TODO. It is obsolete.

src/core/searcher.rs
fulmicoton: Ok! I removed the methods.

src/reader/warming.rs
fulmicoton: Very good point.

fulmicoton: I moved it a little bit down...
But not after the warm up :).
Here is my train of thoughts.

This list is here as an optimization to avoid triggers Warmer GC when we identify this is not useful.
In other words:
If we have too many elements in it, this is not a bug.
If we have not enough elements in it, it can induce a bug.

The invariant we want to keep is "we warmed up a generation => that generation id is in that list" (where => means implies).

The invariant require us to add stuff in the list before we warm up.

Being strict on that kind of invariant is useful when doing multithreading or if there is a state that is persisted.
Here I think we don't have this problem, so this is just a defensive practise.

fulmicoton: I moved it a little bit down...
But not after the warm up :).
Here is my train of thoughts.

This list is here as an optimization to avoid triggers Warmer GC when we identify this is not useful.
In other words:
If we have too many elements in it, this is not a bug.
If we have not enough elements in it, it can induce a bug.

The invariant we want to keep is "we warmed up a generation => that generation id is in that list" (where => means implies).

The invariant require us to add stuff in the list before we warm up.

Being strict on that kind of invariant is useful when doing multithreading or if there is a state that is persisted.
Here I think we don't have this problem, so this is just a defensive practise.

fulmicoton:
Thanks @shikhar for the great comments!
I ended up pushing my PR a bit late last night, so it was very sloppy. Sorry for that.

src/core/searcher.rs
```diff
@@ -49,6 +103,11 @@ impl Searcher {
         &self.index
     }
 
+    /// [SearcherIndexGeneration] which identifies the index data accessed by this generation of `Searcher`.
+    pub fn index_generation(&self) -> &SearcherGeneration {
```
shikhar: I think now you can call this fn just `generation()` + the doc reference needs a fix

src/core/searcher.rs
```diff
@@ -10,9 +9,61 @@ use crate::space_usage::SearcherSpaceUsage;
...
+use crate::Opstamp;
+use crate::SegmentId;
+use crate::TrackedObject;
 
+use std::collections::BTreeMap;
 use std::{fmt, io};
 
+/// Identifies the searcher generation accessed by a [Searcher].
+///
+/// While this may seem, it contains both a `generation_id` AND a list of
```
shikhar: may seem.... :)

src/core/searcher.rs
```diff
@@ -10,9 +9,61 @@ use crate::space_usage::SearcherSpaceUsage;
...
+/// `(SegmentId, DeleteOpstamp)`.
+///
+/// The reason for this, is that this object is used by the `Warmer` API.
+/// It makes it possible to identify which artifact should be refreshed
+/// or garbage collected.
+///
+/// Some `Warmer` might product artifact keyed by:
```
shikhar: might key product artifacts by

src/core/searcher.rs
```diff
@@ -23,6 +74,7 @@ pub struct Searcher {
     index: Index,
     segment_readers: Vec<SegmentReader>,
     store_readers: Vec<StoreReader>,
+    index_generation: TrackedObject<SearcherGeneration>,
```
shikhar: can name this just `generation` now perhaps

shikhar: might be worth grepping for usage of `index_generation` and considering a rename

src/reader/warming.rs
```diff
@@ -0,0 +1,343 @@
...
+pub const GC_INTERVAL: Duration = Duration::from_secs(1);
+
+/// `Warmer` can be used to maintain segment-level state e.g. caches.
+///
+/// They must be registered with the [IndexReaderBuilder].
+pub trait Warmer: Sync + Send {
+    /// Perform any warming work using the provided [Searcher].
+    fn warm(&self, searcher: &Searcher) -> crate::Result<()>;
+
+    /// Discard internal state for any [SearcherIndexGeneration] not provided.
```
shikhar: Also worth grepping for `IndexGeneration` to fixup those doc refs

examples/warmer.rs
```diff
@@ -0,0 +1,224 @@
...
+        ]
+    );
+
+    // Olive oil just got more expensive!
+    price_table.update_price(OLIVE_OIL, 15);
+
+    // The price update are directly reflected on `reload`.
+    //
+    // Be careful here though!...
+    // You may have spotted that we are still using the same `Searcher`.
```
shikhar, line 194: using.. in the sense `searcher` above not dropped yet? I see you acquiring a new one after the reload.

src/core/searcher.rs
shikhar: very useful to have this kind of guidance btw!

examples/warmer.rs
fulmicoton: ah oops yes the reacquire is an error.

src/core/searcher.rs
```diff
@@ -10,9 +9,62 @@ use crate::space_usage::SearcherSpaceUsage;
...
+///
+/// While this might seem redundant, a [SearcherGeneration] contains
+/// both a `generation_id` AND a list of `(SegmentId, DeleteOpstamp)`.
+///
+/// This is on purpose. This object is used by the `Warmer` API.
+/// Having both information makes it possible to identify which
+/// artifact should be refreshed or garbage collected.
+///
+/// Depending on the use case, `Warmer`'s implementers can decide to
+/// product artifacts per:
```
shikhar: i think you meant produce/key instead of product

shikhar (approved):
final pass LG!

src/core/searcher.rs
fulmicoton: Grep for dexGen and dex_gen and removed occurences.

PR #1262, opened 2022-01-21
Aggregation

PSeitz:


src/aggregation/agg_result.rs
```diff
@@ -0,0 +1,37 @@
...
+    BucketResult(BucketAggregationResult),
+    MetricResult(MetricResult),
+}
+
+pub enum MetricResult {
+    Average(i64),
+}
+
+pub struct BucketAggregationResult {
+    buckets: Vec<Box<BucketDataEntry>>,
```
fulmicoton: Why Box?

src/aggregation/agg_result.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+pub enum Key {
+    String(String),
```
fulmicoton: `Str` maybe?

src/aggregation/agg_result.rs
```diff
@@ -0,0 +1,37 @@
...
+    ValueF64(f64),
```
fulmicoton: `F64` Maybe?

src/aggregation/agg_result.rs
```diff
@@ -0,0 +1,37 @@
...
+    Valuei64(i64),
```
fulmicoton: `I64` Maybe?

src/aggregation/agg_result.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+pub enum BucketDataEntry {
+    KeyCount {
```
fulmicoton: What will be the alternative here?
Why is this an enum?

src/aggregation/bucket/histogram.rs
```diff
@@ -0,0 +1,15 @@
+use crate::{aggregation::agg_result::BucketAggregationResult, collector::SegmentCollector};
+
+struct Histogram {}
+
+impl SegmentCollector for Histogram {
+    type Fruit = BucketAggregationResult;
```
fulmicoton: Can we keep the Fruit specific to each type at this level, and have type conversion into BucketAggregationResult?

src/aggregation/segment_agg_result.rs
```diff
@@ -0,0 +1,35 @@
+use std::collections::HashMap;
+
+use crate::DocId;
+
+use super::metric::AverageData;
+
+pub type AggregationResultTree = HashMap<String, AggregationResult>;
+
+pub enum AggregationResult {
```
fulmicoton: you already have a file with aggregation result... Is there something specific at the aggregation level?

src/aggregation/agg_result.rs
PSeitz: I moved the Box now into BucketDataEntry to break up the circular structure

src/aggregation/agg_result.rs
PSeitz: yep

src/aggregation/agg_result.rs
PSeitz: We have different bucket types for different bucket aggregations, but most common is something with a key and doc count, so I just put that there for now.
Alternative would be a trait, but I think this would cause issues when merging buckets

src/aggregation/bucket/histogram.rs
PSeitz: Good point, this is what I do in the metric aggregation already, but for the bucket it is a tree, e.g. something like
- bucketL1 Color Key Green docs: 50
   - bucketL2 Year Key 1900 docs: 20
   - bucketL2 Year Key 2000 docs: 30
- bucketL1 Color Key Red docs: 20
   - bucketL2 Year Key 1900 docs: 5
   - bucketL2 Year Key 2000 docs: 15

I think the alternative would be that the aggregation is called for every subbucket, in that case twice (Green and Red) and the tree is not know in the aggregation itself. The handling would then be on the execution side, which is missing currently, so I'm not sure how good that approach is.

src/aggregation/segment_agg_result.rs
PSeitz: Yes, we have an intermediate result tree, and a final result tree. The intermediate result tree would be returned from the segment.

src/aggregation/agg_result.rs
fulmicoton: I thought no Box was required. I mean it is already on the stack, and therefore should be sized.

PSeitz: `BucketDataEntry` can have a sub aggregation tree

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -0,0 +1,126 @@
...
+/// Aggregation tree with fast field accessors.
+///
+#[derive(Clone)]
+pub enum AggregationWithAccessor {
+    Bucket(BucketAggregationWithAccessor),
+    Metric(MetricAggregationWithAccessor),
+}
+
+impl AggregationWithAccessor {
+    pub fn as_bucket(&self) -> &BucketAggregationWithAccessor {
```
fulmicoton: nitpick: Can we return `Option<&BucketAggregationWithAccessor>`?

This is mostly to align with what is done in `serde_json::Value::as_*` and in `tantivy::schema::Value::as_*` prototype.

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -0,0 +1,126 @@
...
+            AggregationWithAccessor::Bucket(_) => panic!("wrong aggregation type"),
+            AggregationWithAccessor::Metric(metric) => metric,
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct BucketAggregationWithAccessor {
+    /// In general there can be buckets without fast field access, e.g. buckets that are created
+    /// based on search terms. So eventually this needs to be Option or moved.
```
fulmicoton, line 29: or reorganize the code introducing a BucketAggregationTypeWithAccessor.

src/aggregation/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+use itertools::Itertools;
+
+use self::agg_result::AggregationResults;
+use self::intermediate_agg_result::IntermediateAggregationResults;
+
+/// VecWithNames will be used for th
+#[derive(Clone, PartialEq)]
+pub struct VecWithNames<T: Clone> {
+    data: Vec<T>,
+    data_names: Vec<String>,
```
fulmicoton: Interesting choice. I like it.

I would have just used a `Vec` and required to pass the `Aggregations` object upon conversion but your solution is good too.

src/aggregation/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+            data: Default::default(),
+            data_names: Default::default(),
+        }
+    }
+}
+
+impl<T: Clone + std::fmt::Debug> std::fmt::Debug for VecWithNames<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("VecWithNames")
+            .field("data", &self.data)
```
fulmicoton: nitpick: The debug impl is a bit sad.

It could have printed something closer to what a map does.

src/aggregation/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+    fn from(map: HashMap<String, T>) -> Self {
+        VecWithNames::from_iter(map.into_iter())
+    }
+}
+
+impl<T: Clone> VecWithNames<T> {
+    fn from_iter(iter: impl Iterator<Item = (String, T)>) -> Self {
+        let mut entries = iter.collect_vec();
+        // Sort to ensure order of elements match across multiple instances
+        entries.sort_by_cached_key(|entry| entry.0.to_string());
```
fulmicoton: haha. Yeah that's an annoying pitfalls of the borrow checker.

nitpick: A common lighter  workaround is to rely on `sort_by`. 

```suggestion
        entries.sort_by(|left, right| left.0.cmp(&right.0));
```

src/aggregation/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+impl<T: Clone> From<HashMap<String, T>> for VecWithNames<T> {
+    fn from(map: HashMap<String, T>) -> Self {
+        VecWithNames::from_iter(map.into_iter())
+    }
+}
+
+impl<T: Clone> VecWithNames<T> {
+    fn from_iter(iter: impl Iterator<Item = (String, T)>) -> Self {
```
fulmicoton: nitpick: I'd take a Vec straight away.

I added a ticket to add a point in the code style and assigned it to you if that's ok.
https://github.com/quickwit-oss/quickwit/issues/1129

We do not observe it everywhere in tantivy or quickwit yet.

src/aggregation/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+
+pub use agg_req::Aggregation;
+pub use agg_req::BucketAggregationType;
+pub use agg_req::MetricAggregation;
+use itertools::Itertools;
+
+use self::agg_result::AggregationResults;
+use self::intermediate_agg_result::IntermediateAggregationResults;
+
+/// VecWithNames will be used for th
```
```suggestion
/// Represents an associative array `(key => values)` in a very efficient manner.
///
/// This array can be iterated in a specific order defined at construction. 
```

src/aggregation/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+#[derive(Clone, PartialEq)]
+pub struct VecWithNames<T: Clone> {
+    data: Vec<T>,
```
fulmicoton: nitpick:
I am not fond of the name `(data, data_names)`

I suggest `(keys, values)`.

src/aggregation/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+        self.data_names.iter().zip(self.data.iter_mut())
+    }
+    fn into_iter(self) -> impl Iterator<Item = (String, T)> {
+        self.data_names.into_iter().zip(self.data.into_iter())
+    }
+    fn iter(&self) -> impl Iterator<Item = (&String, &T)> + '_ {
+        self.data_names.iter().zip(self.data.iter())
+    }
+    fn keys(&self) -> impl Iterator<Item = &String> + '_ {
+        self.data_names.iter()
```
fulmicoton: super nitpick: I think I would still use `Item=&str` here but I agree this is one of the case where it can be argued otherwise.

src/aggregation/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+    fn values(&self) -> impl Iterator<Item = &T> + '_ {
+        self.data.iter()
+    }
+    fn values_mut(&mut self) -> impl Iterator<Item = &mut T> + '_ {
+        self.data.iter_mut()
+    }
+    fn entries(&self) -> impl Iterator<Item = (&String, &T)> + '_ {
```
fulmicoton: super nitpick: Same as above. I would prefer `&str`, but I would understand why one would disagree with this.

src/aggregation/segment_agg_result.rs
```diff
@@ -0,0 +1,157 @@
...
+        agg_with_accessor: &AggregationsWithAccessor,
+    ) {
+        for (agg_res, agg_with_accessor) in self.0.values_mut().zip(agg_with_accessor.values()) {
+            agg_res.collect(doc, agg_with_accessor);
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum SegmentAggregationResultCollector {
```
fulmicoton:
```suggestion
/// TODO Once we have a bench, test if it is helpful to remove the enum here by having two typed vecs in `SegmentAggregationResultCollectors`.
pub enum SegmentAggregationResultCollector {
```

fulmicoton:
@PSeitz : Good job! I think you solved all of the technical difficulties.

We can now focus on :
- Making the code easier to enter: I think this is already quite readable.
Doing a definitive pass on the object names, and then writing a short README for the agg directory explaning the relationship of the different objects however would be helpful.
- Add unit tests ?
- Some cleanup (remove commented code, etc.?)
- Eventually add more metric aggregation... but that's not necessary right away I guess. We can merge, and integrate with quickwit, and then add functionality if you prefer.

src/aggregation/agg_req_with_accessor.rs
PSeitz: I don't understand, the struct is already effectively BucketAggregationTypeWithAccessor

src/aggregation/mod.rs
PSeitz: yep the switch is super easy with the iter on VecWithNames

src/aggregation/mod.rs
PSeitz: yep Vec is better

src/aggregation/mod.rs
PSeitz: I think we can use &str here

src/aggregation/agg_req_with_accessor.rs
fulmicoton: You missed `Type` in the name. I meant that we could introduce a `BucketAggregationTypeWithAccessor` enum that would contain type specific accessor.

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -0,0 +1,152 @@
...
+            .into(),
+    ))
+}
+
+fn get_aggregation_with_accessor(
+    agg: &Aggregation,
+    reader: &SegmentReader,
+) -> crate::Result<AggregationWithAccessor> {
+    match agg {
+        Aggregation::Bucket(b) => {
```
fulmicoton: rename `b`

src/aggregation/README.md
```diff
@@ -0,0 +1,27 @@
+# Code Organization
+
+The code is organized in submodules:
```
fulmicoton, line 6: Can we add a link to elasticsearch doc somewhere and explain that this aggregation collector has been designed to mimick its behavior?

src/aggregation/bucket/range.rs
```diff
@@ -0,0 +1,206 @@
...
+    fn collect_val(
+        &mut self,
+        val: u64,
+        doc: DocId,
+        bucket_with_accessor: &BucketAggregationWithAccessor,
+    ) {
+        let bucket = self
+            .buckets
+            .iter_mut()
+            .find(|bucket| bucket.range.contains(&val))
```
fulmicoton: :-|

Ok so this is really terrible code here.
For one, I don't think the partition covers all of the space...

Then elasticsearch weirdly decided to allow for overlapping ranges.
Finally, if we do not accept overlapping buckets, then a binary search would have been way more efficient.

I suggest:
- We refuse overlapping buckets.
- We artificially extend the partition to cover the entire space. That means we add two buckets starting from u64::min and ending at u64::max collector to cover everything.
- We represent the bucket scale using a Vec<u64>
- We remap the values to the correct buckets.
- We use a binary search.

If we have a need for overlapping buckets in the future, we can add a specific/slower implementation for them.

src/aggregation/bucket/range.rs
```diff
@@ -0,0 +1,206 @@
...
+/// the same monotonic mapping function, so that the provided ranges contain the u64 values in the
+/// fast field.
+/// The alternative would be that every value read would be converted to the f64 range, but that is
+/// more computational expensive when many documents are hit.
+fn to_u64_range(range: &Range<f64>, field_type: &Type) -> Range<u64> {
+    f64_to_fastfield_u64(range.start, field_type)..f64_to_fastfield_u64(range.end, &field_type)
+}
+
+/// Extends the provided buckets to contain the whole value range, by inserting buckets at the
+/// beginning and end.
```
fulmicoton, line 263: Ah! Hence the unwrap above.
I see. Yes let's do that.

See comment above.

src/aggregation/README.md
PSeitz: yep good idea

src/aggregation/bucket/range.rs
PSeitz: binary search is not necessarily much more efficient because of the low cache efficiency. It depends on the number of buckets and if the incoming data is sorted (improved branch prediction).

11 buckets, data sorted
```
test aggregation::bucket::range::bench::bench_small_range_contains_binary_search                                         ... bench:     126,452 ns/iter (+/- 9,607)
test aggregation::bucket::range::bench::bench_small_range_contains_linear_search                                         ... bench:     163,070 ns/iter (+/- 24,590)
```

5 buckets, data sorted
```
test aggregation::bucket::range::bench::bench_small_range_contains_binary_search                                         ... bench:      83,141 ns/iter (+/- 3,984)
test aggregation::bucket::range::bench::bench_small_range_contains_linear_search                                         ... bench:      81,279 ns/iter (+/- 8,091)
```

> If we have a need for overlapping buckets in the future, we can add a specific/slower implementation for them.

Yes, that should be easy to add later. Performance wise we can expect a little more than 2x

src/aggregation/agg_req.rs
```diff
@@ -0,0 +1,86 @@
...
+/// An aggregation is either a bucket or a metric.
+#[derive(Clone, Debug, PartialEq)]
+pub enum Aggregation {
+    /// Bucket aggregation, see [BucketAggregation] for details.
+    Bucket(BucketAggregation),
+    /// Metric aggregation, see [MetricAggregation] for details.
+    Metric(MetricAggregation),
+}
+
+#[derive(Clone, Debug, PartialEq)]
```
fulmicoton: can we put the decorator below the comment.

src/aggregation/agg_req.rs
```diff
@@ -0,0 +1,86 @@
...
+/// BucketAggregations create buckets of documents. Each bucket is associated with a rule which
+/// determines whether or not a document in the falls into it. In other words, the buckets
+/// effectively define document sets. Buckets are not necessarily disjunct, therefore a document can
+/// fall into multiple buckets. In addition to the buckets themselves, the bucket aggregations also
```
fulmicoton, line 74: Great two lines...

"///In other words, the buckets
/// effectively define document sets. Buckets are not necessarily disjunct,"

src/aggregation/bucket/range.rs
```diff
@@ -0,0 +1,437 @@
...
+/// to value for each range.
+///
+/// Result type is
+/// [BucketDataEntryKeyCount](crate::aggregation::agg_result::BucketDataEntryKeyCount)
+pub struct RangeAggregation {
+    /// The field to aggregate on.
+    pub field_name: String,
+    /// Note that this aggregation includes the from value and excludes the to value for each
+    /// range. Extra buckets will be created until the first to, and last from.
+    pub buckets: Vec<Range<f64>>,
```
fulmicoton: Could that be `Vec<f64>`? The first value would be u64::MIN_VALUE 

The benefit is that the invariant "buckets do not overlap" and partition the space would be expressed by the type.

src/aggregation/bucket/range.rs
```diff
@@ -0,0 +1,437 @@
...
+    bucket: SegmentBucketDataEntry,
+}
+
+/// The collector puts values from the fast field into the correct buckets and does a conversion to
+/// the correct datatype.
+#[derive(Clone, Debug, PartialEq)]
+pub struct SegmentRangeCollector {
+    /// The buckets containing the aggregation data.
+    buckets: Vec<SegmentRangeBucketEntry>,
+    field_type: Type,
```
fulmicoton, line 93: I suspect it might be more efficient and proofreadable to have a `Vec<u64>` directly in the `SegmentRangeCollector`.

RIght now the property that the SegmentReangeBucketEntry do not overlap and has no gap is not expressed by the type.

src/aggregation/bucket/range.rs
```diff
@@ -0,0 +1,437 @@
...
+    #[inline]
+    fn collect_val(
+        &mut self,
+        val: u64,
+        doc: DocId,
+        bucket_with_accessor: &BucketAggregationWithAccessor,
+    ) {
+        let bucket_pos = self
+            .buckets
+            .binary_search_by(|probe| match probe.range.contains(&val) {
```
fulmicoton, line 231: We are reaching "Early optimization is the root of all evil" territory I guess...
But binary search in an array of float should be much faster than binary search in an array of objects.

It comes with an arguable hit on intent readability and an improvement on proofreadability.

src/aggregation/bucket/range.rs
PSeitz: Yes could be, but I think this is more future-proof, when we add support for overlapping buckets, and the api feels more natural to me with ranges

src/aggregation/bucket/range.rs
PSeitz: That's what I thought, but my benchmark says otherwise, where it is consistently slower. I think mainly because now we need to manage two arrays. linear search does look quite quite good though, probably since it is more sensitive to cache optimizations.

```
running 5 tests
test aggregation::bucket::range::bench::bench_small_range_binary_search_only_float                                       ... bench:     132,952 ns/iter (+/- 26,978)
test aggregation::bucket::range::bench::bench_small_range_contains_binary_search                                         ... bench:     120,086 ns/iter (+/- 6,664)
test aggregation::bucket::range::bench::bench_small_range_contains_linear_search                                         ... bench:     151,267 ns/iter (+/- 4,616)
test aggregation::bucket::range::bench::bench_small_range_contains_linear_search_only_float                              ... bench:      94,946 ns/iter (+/- 33,240)
test aggregation::bucket::range::bench::bench_small_range_contains_linear_search_overlapping_buckets                     ... bench:     315,108 ns/iter (+/- 42,815)
```

`cargo bench --features unstable small_range`

src/aggregation/bucket/range.rs
fulmicoton: 5 and 11 buckets is very small. Think about a time histogram.

PR #1264, opened 2022-01-24
Avoid exposing `census::TrackedObject` from `Warmer` API

shikhar:
neater and more misuse resistant (`TrackedObject` is `Clone`)


fulmicoton (approved):
Nice catch!

PR #1275, opened 2022-02-17
added a bench to measure the perf of indexing logs

fulmicoton:


benches/index-bench.rs
```diff
@@ -0,0 +1,39 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use tantivy::schema::{INDEXED, STRING, TEXT};
+use tantivy::Index;
+
+const HDFS_LOGS: &str = include_str!("hdfs.json");
```
PSeitz, line 5: this includes the file in the binary

PSeitz (commented):
hdfs is not a good dataset, because values in the timestamp field are weird

benches/index-bench.rs
fulmicoton: yes I know :) but that's the bench binary.

PR #1276, opened 2022-02-17
Aggregation

PSeitz:


src/aggregation/segment_agg_result.rs
```diff
@@ -0,0 +1,171 @@
...
+};
+use super::bucket::SegmentRangeCollector;
+use super::metric::{
+    AverageAggregation, SegmentAverageCollector, SegmentStatsCollector, StatsAggregation,
+};
+use super::{Key, VecWithNames};
+use crate::aggregation::agg_req::BucketAggregationType;
+use crate::DocId;
+
+pub(crate) type DocBlock = [DocId; 256];
```
fulmicoton: Can you make 256 a const?

fulmicoton:
As discussed multiple times, I would have prefered you did not go into batching yet.
Now we have no point of reference to know if it helps performance or not.

src/aggregation/bucket/range.rs
```diff
@@ -0,0 +1,546 @@
...
+mod bench {
+
+    use std::cmp::Ordering;
+
+    use rand::seq::SliceRandom;
+    use rand::thread_rng;
+
+    use super::*;
+
+    #[derive(Clone)]
```
fulmicoton: These bench have been super helpful, thanks.

They are however disconnected from the code.
Could you remove them and add one bench that actually use the path code?

Ideally that bench should be an util function that take the number of buckets as a parameter so we can have a few bucket and a many bucket bench.

src/aggregation/bucket/range.rs
```diff
@@ -0,0 +1,546 @@
...
+    #[inline]
+    pub(crate) fn collect_block(
+        &mut self,
+        doc: &[DocId],
+        bucket_with_accessor: &BucketAggregationWithAccessor,
+        force_flush: bool,
+    ) {
+        let mut iter = doc.chunks_exact(4);
+        for docs in iter.by_ref() {
+            let val1 = bucket_with_accessor.accessor.get(docs[0]);
```
fulmicoton, line 155: why all the manual unrolling?

src/aggregation/bucket/range.rs
```diff
@@ -0,0 +1,546 @@
...
+            .unwrap()
+            .iter()
+            .map(|bucket| range_to_string(bucket, &Type::F64))
+            .collect_vec();
+        assert_eq!(buckets[0], "*-0");
+        assert_eq!(buckets[buckets.len() - 1], "10-*");
+    }
+
+    #[test]
+    fn range_binary_search_test() {
```
fulmicoton: Remove this test. (it tests nothing)

src/aggregation/bucket/range.rs
```diff
@@ -0,0 +1,546 @@
...
+            field_type,
+        })
+    }
+
+    #[inline]
+    pub(crate) fn collect_block(
+        &mut self,
+        doc: &[DocId],
+        bucket_with_accessor: &BucketAggregationWithAccessor,
+        force_flush: bool,
```
fulmicoton, line 151: Why is the `force_flush` argument necessary?

src/aggregation/bucket/range.rs
fulmicoton: Replace it with a test actually testing the target SegmentRangeCollector object .

src/aggregation/collector.rs
```diff
@@ -0,0 +1,135 @@
...
+
+/// Collector for distributed aggregations.
+///
+/// The collector collects all aggregations by the underlying aggregation request.
+///
+/// # Purpose
+/// AggregationCollector returns `IntermediateAggregationResults` and not the final
+/// `AggregationResults`, so that results from differenct indices can be merged and then converted
+/// into the final `AggregationResults` via the `into()` method.
+pub struct DistributedAggregationCollector {
```
fulmicoton, line 32: Alternatively we could do this 
```
     fn merge_fruits(
         &self,
         segment_fruits: Vec<<Self::Child as SegmentCollector>::Fruit>,
-    ) -> crate::Result<Self::Fruit>;
+    ) -> crate::Result<<Self::Child as SegmentCollector>::Fruit>;
+
+    fn transform(&self, child_fruit: <Self::Child as SegmentCollector>::Fruit) -> Self::Fruit;
```

fulmicoton: I'll open a ticket.

fulmicoton: https://github.com/quickwit-oss/tantivy/issues/1277

fulmicoton (changes requested):
Can you fix the unit tests/ bench?

src/aggregation/bucket/range.rs
PSeitz: It's faster

```
test aggregation::tests::bench::bench_aggregation_average_f64                                                            ... bench:   7,063,803 ns/iter (+/- 1,769,966)
test aggregation::tests::bench::bench_aggregation_average_u64                                                            ... bench:   5,622,493 ns/iter (+/- 100,153)
test aggregation::tests::bench::bench_aggregation_average_u64_and_f64                                                    ... bench:   8,695,562 ns/iter (+/- 109,397)
test aggregation::tests::bench::bench_aggregation_stats_f64                                                              ... bench:   7,493,152 ns/iter (+/- 214,351)
test aggregation::tests::bench::bench_aggregation_sub_tree                                                               ... bench:  13,159,764 ns/iter (+/- 589,236)
```

removed unrolling in average metric
```
test aggregation::tests::bench::bench_aggregation_average_f64                                                            ... bench:   7,631,327 ns/iter (+/- 174,053)
test aggregation::tests::bench::bench_aggregation_average_u64                                                            ... bench:   7,451,904 ns/iter (+/- 410,011)
test aggregation::tests::bench::bench_aggregation_average_u64_and_f64                                                    ... bench:  10,471,694 ns/iter (+/- 140,192)
test aggregation::tests::bench::bench_aggregation_stats_f64                                                              ... bench:   7,465,887 ns/iter (+/- 344,009)
test aggregation::tests::bench::bench_aggregation_sub_tree                                                               ... bench:  16,609,919 ns/iter (+/- 399,840)
```

src/aggregation/bucket/range.rs
PSeitz: uses now actual code and is parameterized

src/aggregation/bucket/range.rs
PSeitz: There's a cache of 256 in each (sub)tree, that handles the docs when it is filled in one block.
After all docs have been collected the caches need to be flushed.
The parameter is to pass down the flush operation to the sub_trees. Each bucket can have a subtree (recursion)

src/aggregation/bucket/range.rs
fulmicoton: Awesome ! Thanks

fulmicoton (approved):

PR #1279, opened 2022-02-18
Fix opening bytes index with dynamic codec

### LINKED ISSUE #1278: Bytes fastfield codec mismatch
Bytes fast field consists of two indices, the raw bytes data and a fast field index to find the bytes for a doc.

When writing a bytes fast field for the first time, the index is always written with the bitpack codec https://github.com/quickwit-oss/tantivy/blob/505e6a440ca4a7e0749b1138aa16a1c4f60ad346/src/fastfield/bytes/writer.rs#L112-L113

When merging indices there is generic code used, which has an auto-detection for the best codec https://github.com/quickwit-oss/tantivy/blob/505e6a440ca4a7e0749b1138aa16a1c4f60ad346/src/indexer/merger.rs#L506-L510

The index part of the bytes field is always read as Bitpacked serialized index, which can lead to following error after a merge:

```
thread 'merge_thread0' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `2`: Tried to open fast field as bitpacked encoded (id=1), but got serializer with different id', C:\Users\ChillFish8\.cargo\git\checkouts\tantivy-f70b7ea03dadae9a\13a4473\src\fastfield\reader.rs:160:9
```

PSeitz:
Fix #1278

### DIFF
--- a/src/fastfield/bytes/reader.rs
+++ b/src/fastfield/bytes/reader.rs
@@ -1,5 +1,5 @@
 use crate::directory::{FileSlice, OwnedBytes};
-use crate::fastfield::{BitpackedFastFieldReader, FastFieldReader, MultiValueLength};
+use crate::fastfield::{DynamicFastFieldReader, FastFieldReader, MultiValueLength};
 use crate::DocId;
 
 /// Reader for byte array fast fields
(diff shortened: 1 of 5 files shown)


fulmicoton:
Please add a unit test that failed before and succeeds after @PSeitz

fulmicoton (changes requested):
unit test plz

PSeitz:
I extended the proptest to cover that case and confirmed they fail before

fulmicoton (approved):

PR #1292, opened 2022-02-23
improve validation in aggregation, extend invalid field test

### LINKED ISSUE #1291: Aggregation over invalid should return an error and be unit tested.
Apologies if it is already covered. The code does not make it clear.

I someone runs an aggregation (any kind) over a non existing field or a field that does not support the aggregation type, we need to return an error.

This corner case needs to be unit tested.

Also, we should to remove the 2 `unimplemented!()` occurrences (one is easier than the other.)

PSeitz:
improve validation in aggregation
extend invalid field test
Fixes #1291

improve validation in aggregation, extend invalid field test
collect fast field names on request structure

### DIFF
--- a/src/aggregation/metric/stats.rs
+++ b/src/aggregation/metric/stats.rs
@@ -95,7 +95,7 @@ impl IntermediateStats {
         self.max = self.max.max(other.max);
     }
 
-    /// compute final result
+    /// compute final resultimprove_docs
     pub fn finalize(&self) -> Stats {
         Stats {
             count: self.count,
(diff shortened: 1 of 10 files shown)


src/aggregation/metric/stats.rs
```diff
@@ -95,7 +95,7 @@ impl IntermediateStats {
         self.max = self.max.max(other.max);
     }
 
-    /// compute final result
+    /// compute final resultimprove_docs
```
fulmicoton, line 98: typo

fulmicoton (approved):

Issue #1236, opened 2021-12-13
Remove the use of iff in the code


fulmicoton:
I love this abbrevation because its expresses precisely something very common, and in a very concise manner.
However, at least two people mentioned they thought it was a typo, so we should probably replace its usage to the verbose "if and only if".

PR #1298, opened 2022-03-01
iff->if or if and only if 

saroh:
rationale : has_XXX, is_XXX functions are nearly always defined as a iff in code. Other bolean returns you may need to specify when to return what value.

#1236

### DIFF
--- a/src/collector/top_collector.rs
+++ b/src/collector/top_collector.rs
@@ -173,8 +173,7 @@ impl<T: PartialOrd + Clone> TopSegmentCollector<T> {
             .collect()
     }
 
-    /// Return true iff at least K documents have gone through
-    /// the collector.
+    /// Return true if more documents have been collected than the limit.
     #[inline]
     pub(crate) fn at_capacity(&self) -> bool {
         self.heap.len() >= self.limit
--- a/src/termdict/fst_termdict/termdict.rs
+++ b/src/termdict/fst_termdict/termdict.rs
@@ -148,7 +148,7 @@ impl TermDictionary {
     /// Term ordinals are defined as the position of the term in
     /// the sorted list of terms.
     ///
-    /// Returns true iff the term has been found.
+    /// Returns true if and only if the term has been found.
     ///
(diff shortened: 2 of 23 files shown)


fulmicoton:
iff means if and only if. It has confused people in the past so I opened an issue to replace it with if and only if
https://github.com/quickwit-oss/tantivy/issues/1236

I guess if alone is ok too where it is not amibguous.
https://en.wikipedia.org/wiki/If_and_only_if#:~:text=In%20logic%20and%20related%20fields,true%20or%20both%20are%20false.

saroh:
Oh I had the feelingh it may have ben something else, I'm re-writing the PR :)

saroh:
So I feel I could put put `if` on functions/methods that implement an obvious single responsability, so most of your code. There are a few interfaces where the `if and only if` should be stressed imo, as a guideline for the implementer: methods which answer a boolean but are not obvious booleans (eg. advance vs is_something). Or exotic truth values such as bloomfilters https://en.wikipedia.org/wiki/Bloom_filter

src/collector/top_collector.rs
```diff
@@ -173,8 +173,7 @@ impl<T: PartialOrd + Clone> TopSegmentCollector<T> {
             .collect()
     }
 
-    /// Return true iff at least K documents have gone through
-    /// the collector.
+    /// Return true if more documents have been collected than the limit.
```
saroh, line 176: wording might be iffy here :)

src/termdict/fst_termdict/termdict.rs
```diff
@@ -143,7 +143,8 @@ impl TermDictionary {
         Ok(self.fst_index.get(key))
     }
 
-    /// Returns the term associated to a given term ordinal.
+    /// Stores the term associated to a given term ordinal in
+    /// a `bytes` buffer.
     ///
```
saroh, line 148: tocheck, out of PR scope ;)

fulmicoton (approved):

fulmicoton:
Thank you @saroh !

PR #1301, opened 2022-03-02
update fastfield doc

saroh:
#1232

### DIFF
--- a/fastfield_codecs/src/lib.rs
+++ b/fastfield_codecs/src/lib.rs
@@ -63,6 +63,7 @@ pub trait FastFieldDataAccess {
 }
 
 #[derive(Debug, Clone)]
+/// Contains statistics used for compression in field writing.
 pub struct FastFieldStats {
     pub min_value: u64,
     pub max_value: u64,
--- a/fastfield_codecs/src/lib.rs
+++ b/fastfield_codecs/src/lib.rs
@@ -63,7 +63,7 @@ pub trait FastFieldDataAccess {
 }
 
 #[derive(Debug, Clone)]
-/// Contains statistics used for compression in field writing.
+/// Statistics are used in codec detection and stored in the fast field footer.
 pub struct FastFieldStats {
     pub min_value: u64,
     pub max_value: u64,
--- a/src/fastfield/alive_bitset.rs
+++ b/src/fastfield/alive_bitset.rs
@@ -7,7 +7,7 @@ use ownedbytes::OwnedBytes;
 use crate::space_usage::ByteCount;
 use crate::DocId;
 
-/// Write a alive `BitSet`
+/// Write an alive `BitSet`
 ///
 /// where `alive_bitset` is the set of alive `DocId`.
 /// Warning: this function does not call terminate. The caller is in charge of
@@ -55,7 +55,7 @@ impl AliveBitSet {
         AliveBitSet::from(readonly_bitset)
     }
 
-    /// Opens a delete bitset given its file.
+    /// Opens an alive bitset given its file.
     pub fn open(bytes: OwnedBytes) -> AliveBitSet {
         let bitset = ReadOnlyBitSet::open(bytes);
         AliveBitSet::from(bitset)
@@ -79,13 +79,13 @@ impl AliveBitSet {
         self.bitset.iter()
     }
 
-    /// Get underlying bitset
+    /// Get underlying bitset.
     #[inline]
     pub fn bitset(&self) -> &ReadOnlyBitSet {
         &self.bitset
     }
 
-    /// The number of deleted docs
+    /// The number of alive documents.
     pub fn num_alive_docs(&self) -> usize {
         self.num_alive_docs
     }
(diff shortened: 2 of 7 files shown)


fastfield_codecs/src/lib.rs
```diff
@@ -63,6 +63,7 @@ pub trait FastFieldDataAccess {
 }
 
 #[derive(Debug, Clone)]
+/// Contains statistics used for compression in field writing.
```
saroh: I assumed so. Might be a little incorrect.

src/fastfield/alive_bitset.rs
```diff
@@ -55,7 +55,7 @@ impl AliveBitSet {
         AliveBitSet::from(readonly_bitset)
     }
 
-    /// Opens a delete bitset given its file.
+    /// Opens an alive bitset given its file.
```
saroh, line 58: The doc in this file is written as if we were tombstoning, but the code seems to say we're "inverse tombstoning".

src/fastfield/mod.rs
```diff
@@ -12,8 +12,8 @@
 //!
 //!
 //! Fields have to be declared as `FAST` in the  schema.
-//! Currently only 64-bits integers (signed or unsigned) are
-//! supported.
+//! Currently supported fields are: u64, i64 and f64.
+//! u8 are supported for advanced usage.
```
saroh: unsure.

src/fastfield/alive_bitset.rs
PSeitz: yes thanks, that was inverted recently

fulmicoton:
@PSeitz can you review

fastfield_codecs/src/lib.rs
PSeitz:
```suggestion
/// Statistics are used in codec detection and stored in the fast field footer.
```

src/fastfield/mod.rs
PSeitz: where do you get the u8 part from?

src/fastfield/writer.rs
```diff
@@ -14,7 +14,7 @@ use crate::postings::UnorderedTermId;
 use crate::schema::{Cardinality, Document, Field, FieldEntry, FieldType, Schema};
 use crate::termdict::TermOrdinal;
 
-/// The fastfieldswriter regroup all of the fast field writers.
+/// The `FastFieldsWriter` regroups all of the fast field writers.
```
PSeitz:
```suggestion
/// The `FastFieldsWriter` groups all of the fast field writers.
```

PSeitz (commented):
Thanks, I appreciate these doc PRs.
I made some suggestions and unclear about the u8 fast field

src/fastfield/mod.rs
saroh: https://github.com/quickwit-oss/tantivy/pull/1301/files#diff-a5377d7ddb22106434828e863850fa6d6ec58e5cb0cdefa8cc4a58ecce8c3543L8

+ changelog for v0.6 
https://github.com/quickwit-oss/tantivy/blob/main/CHANGELOG.md#tantivy-06

made me infer the u8

saroh:
You're welcome, the technical doc is good. I work as a user with other search libs/engines and I think there wouldn't be a lot needed to turn it in a good user manual/documentation by front loading a few concepts (some I'm yet to read so I currently accept as a black box). Namely segments (parts of indices you maybe compact later on because it's a WORM fs ? ), collecting fruits, term ordinals (internal id for values ? segment specific ?).

src/fastfield/mod.rs
PSeitz: This is a &[u8] bytes fast field, not u8. You can use it like this:

```
let bytes_field = schema_builder.add_bytes_field("bytes", FAST | INDEXED | STORED);
```
Instead of a number you can associate arbitrary bytes with a doc and retrieve them fast via the fast field.

PSeitz:
Yes segments are immutable, but they can be merged into a new segment.

term ordinals are monotonically increasing segment specific ids to reference a term in a field. They are more efficient than handling the terms themselves.

src/fastfield/mod.rs
```diff
@@ -12,11 +12,10 @@
...
 //!
 //! Fields have to be declared as `FAST` in the  schema.
-//! Currently only 64-bits integers (signed or unsigned) are
-//! supported.
+//! Currently supported fields are: u64, i64, f64 and bytes.
 //!
-//! They are stored in a bit-packed fashion so that their
-//! memory usage is directly linear with the amplitude of the
+//! u64, i64 and f64 fields are stored in a bit-packed fashion so that
+//! their memory usage is directly linear with the amplitude of the
```
saroh, line 18: maybe like this then ?

src/fastfield/mod.rs
saroh: Thx for the explanations. So something like this : https://github.com/quickwit-oss/tantivy/pull/1301/commits/d36e0a954986c35565e545886dd500ae4d4d19cd ?

PSeitz (approved):

PR #1306, opened 2022-03-09
add Histogram aggregation

PSeitz:


src/aggregation/agg_result.rs
```diff
@@ -106,6 +113,99 @@ impl From<IntermediateBucketResult> for BucketResult {
                 });
                 BucketResult::Range { buckets }
             }
+            IntermediateBucketResult::Histogram { buckets, req } => {
+                let buckets = if req.min_doc_count() == 0 {
+                    // We need to fill up the buckets for the total ranges, so that there are no
+                    // gaps
+                    let max = buckets
+                        .iter()
+                        .map(|bucket| bucket.key)
```
fulmicoton: There is a min_max in itertools and a separate min/max in Iterator. No need to use fold.

src/aggregation/agg_result.rs
```diff
@@ -81,12 +82,18 @@ impl From<IntermediateMetricResult> for MetricResult {
...
     /// sub_aggregations.
     Range {
         /// The range buckets sorted by range.
         buckets: Vec<RangeBucketEntry>,
     },
+    /// This is the histogram entry for a bucket, which contains a key, count, and optionally
+    /// sub_aggregations.
+    Histogram {
+        /// The buckets.
+        buckets: Vec<BucketEntry>,
```
fulmicoton, line 141: I don't understand. Can key be anything?
Can there be holes?

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -0,0 +1,822 @@
...
+use crate::{DocId, TantivyError};
+
+/// Histogram is a bucket aggregation, where buckets are created dynamically for given `interval`.
+/// Each document value is rounded down to its bucket.
+///
+/// E.g. if we have a price 18 and an interval of 5, the document will fall into the bucket with
+/// the key 15. The formula used for this is:
+/// `((val - offset) / interval).floor() * interval + offset`
+///
+/// For this calculation all fastfield values are converted to f64.
```
fulmicoton, line 30: I wonder if this is a good idea. u64 -> f64 is not super cheap and will yield wrong results sometimes.
On the other hand, the main use for this is timestamp which are i64/u64

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -0,0 +1,822 @@
...
+pub struct HistogramBounds {
+    /// The lower bounds.
+    pub min: f64,
+    /// The upper bounds.
+    pub max: f64,
+}
+
+impl HistogramBounds {
+    fn contains(&self, val: f64) -> bool {
+        val >= self.min && val <= self.max
```
fulmicoton, line 157: Really? I was expecting a semi-open interval here.

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -0,0 +1,822 @@
...
+            .count();
+        let skip_end = self
+            .buckets
+            .iter()
+            .rev()
+            .take_while(|bucket| bucket.doc_count == 0)
+            .count();
+        let num_buckets = self.buckets.len();
+
+        let buckets = self
```
fulmicoton: I understand the saturating_sub tries to address the edge case, but that's a cryptic way to write code.

Buckets is a Vec... There is no need for magic tricks.
For instance, you could compute `skip_end` first,  working on the slice `[skip_start..]`.

You could also, compute `skip_end` first, and thank skip heading element and collect on the same iterator.

src/aggregation/agg_result.rs
PSeitz: BucketEntry is a general purpose bucket result, because many bucket aggregations have `key`, `doc_count` and `sub_aggregations`.

Key is in this case always Key::F64. Do you think a specialized version would be better?

If there are holes depends on the request, if min_doc_count is 0, then there are no holes

src/aggregation/agg_result.rs
PSeitz: yeah that's much better

src/aggregation/segment_agg_result.rs
```diff
@@ -177,10 +186,20 @@ impl SegmentBucketResultCollector {
...
+            SegmentBucketResultCollector::Histogram(histogram) => {
+                histogram.collect_block(doc, bucket_with_accessor, force_flush)
+            }
         }
     }
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct SegmentHistogramBucketEntry {
+    pub key: f64,
```
fulmicoton, line 209: Why do we need the key in here?

src/aggregation/mod.rs
```diff
@@ -332,18 +355,17 @@ mod tests {
         let index = Index::create_in_ram(schema_builder.build());
         {
             let mut index_writer = index.writer_for_tests()?;
-            for i in 0..num_docs {
+            for i in values {
```
fulmicoton: You can use `for &i in values` I think.

src/aggregation/mod.rs
```diff
@@ -239,6 +248,10 @@ pub enum Key {
     F64(f64),
 }
 
+trait MergeFruits {
+    fn merge_fruits(&mut self, other: &Self);
```
fulmicoton: Can we drop pass other by value? That would remove some clones.

src/aggregation/intermediate_agg_result.rs
```diff
@@ -142,22 +157,96 @@ impl IntermediateBucketResult {
...
+                panic!("try merge on different types")
+            }
+            (IntermediateBucketResult::Histogram { .. }, _) => {
+                panic!("try merge on different types")
             }
         }
     }
 }
 
+// fn merge_sorted_vecs<V: MergeFruits + Clone>(entries_left: &mut Vec<V>, entries_right: &Vec<V>) {
```
fulmicoton: please remove

src/aggregation/mod.rs
```diff
@@ -239,6 +248,10 @@ pub enum Key {
     F64(f64),
 }
 
+trait MergeFruits {
```
fulmicoton: Is this trait useful? I'm not sure I understand its point.

fulmicoton: Ah it is used to share code in the merge_maps thingy?

Why is it not declared in `intermediate_agg_result` then, close to that function then?

src/aggregation/bucket/histogram/histogram.rs
PSeitz: oh yeah I think you are right, I have to check elastic on this

fulmicoton (changes requested):
see inline comments.

src/aggregation/agg_result.rs
fulmicoton: Why is it different from `RangeBucketEntry`? I thought histogram was like range but with a shorter API and the opportunity for optimization. Is it a "output much match elastic's" story?

Can we put some comment in there about the presence of holes ?

src/aggregation/segment_agg_result.rs
```diff
@@ -177,10 +186,20 @@ impl SegmentBucketResultCollector {
...
+    pub doc_count: u64,
+    pub sub_aggregation: Option<SegmentAggregationResultsCollector>,
```
fulmicoton: I wonder if we should also keep sub_aggregation as an external `Vec<Option<SegmentAggregationResultsCollector>>`.

src/aggregation/bucket/histogram/histogram.rs
PSeitz: since you solved the riddle I can remove it ;)
I just realized we can simply filter all empty buckets.

Edit: I started with the slice, but here we want to get the elements by value to avoid clones

src/aggregation/segment_agg_result.rs
PSeitz: yes, that's something I wanted to test as well, but I would probably go for `Option<Vec<SegmentAggregationResultsCollector>>`

fulmicoton: Oh right.

src/aggregation/agg_result.rs
PSeitz: `RangeBucketEntry` contains the from and to values, histogram has just the f64 key (the min value of the bucket)

> Can we put some comment in there about the presence of holes ?
Yep I'll add it

src/aggregation/segment_agg_result.rs
PSeitz: we could compute it at the end before filtering the collector for 0 buckets, because we implicitly know which bucket index has which key, but I think I'd not do this optimization now, since it will limit how we can change the algorithm

src/aggregation/mod.rs
PSeitz: nice

src/aggregation/mod.rs
PSeitz: yep, done

src/aggregation/mod.rs
PSeitz: yes, so multiple buckets can share the merge code on collections

I moved it to intermediate_agg_result

src/aggregation/segment_agg_result.rs
PSeitz: Yep, it's much faster like that, 2x in some cases

src/aggregation/agg_result.rs
fulmicoton: Yes but why? Is it for serialization purpose?

src/aggregation/agg_result.rs
```diff
@@ -106,6 +112,94 @@ impl From<IntermediateBucketResult> for BucketResult {
                 });
                 BucketResult::Range { buckets }
             }
+            IntermediateBucketResult::Histogram { buckets, req } => {
+                let buckets = if req.min_doc_count() == 0 {
+                    // We need to fill up the buckets for the total ranges, so that there are no
+                    // gaps
+                    let minmax = buckets.iter().minmax_by_key(|bucket| bucket.key);
+                    let all_buckets = match minmax {
+                        itertools::MinMaxResult::MinMax(min, max) => {
```
fulmicoton: there is a bug if htere is a single bucket, I suggest you to use into_option()


```rust
let min_max_opt: Option<(f64, f64)> = 
     buckets.iter().minmax_by_key(|bucket| bucket.key).into_option();
let all_buckets: Vec<WhatEverTypeIsCorrecthere> = min_max_opt
     .map(|(min, max)| {
         generate_buckets(&req, min, max) 
     })
     .unwrap_or_default();
```

Please write the type when it is not trivial. Here I have no clue what `all_buckets` is

src/aggregation/bucket/histogram/mod.rs
```diff
@@ -0,0 +1,2 @@
+mod histogram;
+pub use histogram::*;
```
fulmicoton, line 2: I suggest we put everything in `../histogram.rs` or in `histogram/mod.rs`

src/aggregation/mod.rs
```diff
@@ -189,6 +199,15 @@ impl<T: Clone> From<HashMap<String, T>> for VecWithNames<T> {
 }
 
 impl<T: Clone> VecWithNames<T> {
+    fn from_other<K: Clone + Into<T>>(entries: VecWithNames<K>) -> Self {
+        let mut values = Vec::with_capacity(entries.len());
+        values.extend(entries.values.into_iter().map(Into::into));
```
fulmicoton: !??

`let values = entries.into_iter().map(Into::into).collect();`

src/aggregation/mod.rs
```diff
@@ -310,6 +335,16 @@ mod tests {
...
         merge_segments: bool,
         num_docs: usize,
+    ) -> crate::Result<Index> {
+        get_test_index_from_values(
+            merge_segments,
+            &(0..num_docs).map(|el| el as f64).collect::<Vec<f64>>(),
+        )
+    }
+
+    pub fn get_test_index_from_values(
```
fulmicoton, line 341: why pub

fulmicoton (changes requested):
See comment inline.

src/aggregation/agg_result.rs
PSeitz: all_buckets is used to fill gaps. In case of a single bucket there is nothing to do, so empty vec is fine

src/aggregation/mod.rs
PSeitz: that was a small performance optimization, but I don't think we need it

src/aggregation/mod.rs
PSeitz: we reuse it in other modules

src/aggregation/agg_result.rs
```diff
@@ -106,6 +146,112 @@ impl From<IntermediateBucketResult> for BucketResult {
                 });
                 BucketResult::Range { buckets }
             }
+            IntermediateBucketResult::Histogram { buckets } => {
+                let histogram_req = req.as_histogram();
+                let buckets = if histogram_req.min_doc_count() == 0 {
```
fulmicoton: This is still needlessly complicated.

fulmicoton (changes requested):
same file same comment. the code is needlessly complicated.

src/aggregation/agg_result.rs
```diff
@@ -10,25 +10,80 @@ use std::collections::HashMap;
...
-impl From<IntermediateAggregationResults> for AggregationResults {
-    fn from(tree: IntermediateAggregationResults) -> Self {
-        Self(
-            tree.0
-                .into_iter()
-                .map(|(key, agg)| (key, agg.into()))
-                .collect(),
-        )
+impl AggregationResults {
+    /// Convert and intermediate result and its aggregation request to the final result
```
fulmicoton, line 27: very helpful

src/aggregation/agg_result.rs
```diff
@@ -10,25 +10,80 @@ use std::collections::HashMap;
...
+    /// Internal function, CollectorAggregations is used instead Aggregations, which is optimized
+    /// for internal processing
+    fn from_intermediate_and_req_internal(
+        results: IntermediateAggregationResults,
+        req: &AggregationsInternal,
+    ) -> Self {
+        let mut result = HashMap::default();
+
+        // Important assumption:
+        // When the tree contains buckets/metric, we expect it to have all buckets/metrics from the
```
fulmicoton, line 45: What happens if this is not the case?
Does it panic, returns a wrong result or returns an error?

We want to avoid panic here, because IntermediateAggregationResults is coming from another node.

fulmicoton, line 45: The code contradicts the comment a little.
You do handle the case where the buckets are empty.

I assume this is to handle the special case where there are no splits. Can you put that in a comment?

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -0,0 +1,1352 @@
...
+/// See [BucketEntry](crate::aggregation::agg_result::BucketEntry)
+
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
+pub struct HistogramAggregation {
+    /// The field to aggregate on.
+    pub field: String,
+    /// The interval to chunk your data range. The buckets span ranges of [0..interval).
+    /// Must be a positive value.
+    pub interval: f64,
+    /// Intervals intersect at 0 by default, offset can move the interval.
```
```suggestion
    /// Intervals implicitely defines an absolute grid of buckets `[interval * k, interval * (k + 1))`.
    ///
    /// Offset makes it possible to shift this grid into `[offset + interval * k, offset + interval * (k + 1))
```

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -0,0 +1,1352 @@
...
+    /// Offset has to be in the range [0, interval).
+    ///
+    /// As an example. If there are two documents with value 8 and 12 and interval 10.0, they would
+    /// fall into the buckets with the key 0 and 10.
+    /// With offset 5 and interval 10, they would both fall into the bucket with they key 5 and the
+    /// range [5..15)
+    pub offset: Option<f64>,
+    /// The minimum number of documents in a bucket to be returned. Defaults to 0.
+    pub min_doc_count: Option<u64>,
+    /// Limit the data range.
```
```suggestion
    /// Limits the data range to `[min, max]` closed interval.
```

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -0,0 +1,1352 @@
...
+    pub fn into_intermediate_bucket_result(self) -> IntermediateBucketResult {
+        let mut buckets = Vec::with_capacity(
+            self.buckets
+                .iter()
+                .filter(|bucket| bucket.doc_count != 0)
+                .count(),
+        );
+
+        // Below we remove empty buckets for two reasons
+        // 1. To reduce the size of the intermediate result, which may be passed on the wire.
```
fulmicoton, line 185: very helpful comment.

Im a bit lost though. I thought IntermediateBucketResult always contain all of the bucket?

fulmicoton (approved):

PR #1307, opened 2022-03-09
Replace `chrono` with `time`

### LINKED ISSUE #1304: Replace `chrono` with `time`
`chrono` seems to be unmaintained and depends on an outdated version of `time`. For all features required by the engine and the API `time` should suffice.

As `chrono` types and modules are re-exported by the public API this will be a breaking change. I also suggest to use this opportunity for removing those re-exports from the API which IMHO do not provide any value and are more confusing than helpful. Explicit, external dependencies are fine and more transparent.

uklotzde:
Fixes #1304.

**BREAKING API CHANGE**: The type `DateTime` is no longer (re-)exported. Instead you need to use `time::OffsetDateTime` explicitly.

Remarks:
- Please check the error handling and mappings. `time` returns results instead of panicking like `chrono` does. For the `FastValue` impl I had to use `.expect("valid UNIX timestamp").
- I had to adjust some debug strings in tests that are used for matching test results. Copy&paste from the failed test outputs. Not very handy, please verify.

TODO:
- [x] Add changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+Unreleased
+================================
+- For date values `chrono` has been replaced with `time`:
+  - The `time` crate is re-exported as `tantivy::time` instead of `tantivy::chrono`.
+  - The type alias `tantivy::DateTime` has been removed.
+  - `Value::Date` wraps `time::PrimitiveDateTime` without time zone information.
+  - Internally date/time values are stored as seconds since UNIX epoch in UTC.
+  - Converting a `time::OffsetDateTime` to `Value::Date` implicitly converts the value into UTC.
+    If this is not desired do the time zone conversion yourself and use `time::PrimitiveDateTime`
+    directly instead.
+
 Tantivy 0.17
 ================================
 - LogMergePolicy now triggers merges if the ratio of deleted documents reaches a threshold (@shikhar @fulmicoton) [#115](https://github.com/quickwit-oss/tantivy/issues/115)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 Unreleased
 ================================
-- For date values `chrono` has been replaced with `time`:
+- For date values `chrono` has been replaced with `time` (@uklotzde) #1304 :
   - The `time` crate is re-exported as `tantivy::time` instead of `tantivy::chrono`.
   - The type alias `tantivy::DateTime` has been removed.
   - `Value::Date` wraps `time::PrimitiveDateTime` without time zone information.
--- a/src/error.rs
+++ b/src/error.rs
@@ -149,8 +149,20 @@ impl<Guard> From<PoisonError<Guard>> for TantivyError {
     }
 }
 
-impl From<chrono::ParseError> for TantivyError {
-    fn from(err: chrono::ParseError) -> TantivyError {
+impl From<time::error::Format> for TantivyError {
+    fn from(err: time::error::Format) -> TantivyError {
+        TantivyError::InvalidArgument(err.to_string())
+    }
+}
+
+impl From<time::error::Parse> for TantivyError {
+    fn from(err: time::error::Parse) -> TantivyError {
+        TantivyError::InvalidArgument(err.to_string())
+    }
+}
+
+impl From<time::error::ComponentRange> for TantivyError {
+    fn from(err: time::error::ComponentRange) -> TantivyError {
         TantivyError::InvalidArgument(err.to_string())
     }
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -151,19 +151,19 @@ impl<Guard> From<PoisonError<Guard>> for TantivyError {
 
 impl From<time::error::Format> for TantivyError {
     fn from(err: time::error::Format) -> TantivyError {
-        TantivyError::InvalidArgument(err.to_string())
+        TantivyError::InvalidArgument(format!("Date formatting error: {err}"))
     }
 }
 
 impl From<time::error::Parse> for TantivyError {
     fn from(err: time::error::Parse) -> TantivyError {
-        TantivyError::InvalidArgument(err.to_string())
+        TantivyError::InvalidArgument(format!("Date parsing error: {err}"))
     }
 }
 
 impl From<time::error::ComponentRange> for TantivyError {
     fn from(err: time::error::ComponentRange) -> TantivyError {
-        TantivyError::InvalidArgument(err.to_string())
+        TantivyError::InvalidArgument(format!("Date range error: {err}"))
     }
 }
 
--- a/src/fastfield/readers.rs
+++ b/src/fastfield/readers.rs
@@ -1,5 +1,3 @@
-use time::OffsetDateTime;
-
 use super::reader::DynamicFastFieldReader;
 use crate::directory::{CompositeFile, FileSlice};
 use crate::fastfield::{
@@ -7,6 +5,7 @@ use crate::fastfield::{
 };
 use crate::schema::{Cardinality, Field, FieldType, Schema};
 use crate::space_usage::PerFieldSpaceUsage;
+use crate::time::OffsetDateTime;
 use crate::TantivyError;
 
 /// Provides access to all of the BitpackedFastFieldReader.
--- a/src/fastfield/readers.rs
+++ b/src/fastfield/readers.rs
@@ -1,3 +1,5 @@
+use time::OffsetDateTime;
+
 use super::reader::DynamicFastFieldReader;
 use crate::directory::{CompositeFile, FileSlice};
 use crate::fastfield::{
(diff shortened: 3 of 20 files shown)


src/error.rs
```diff
@@ -149,8 +149,20 @@ impl<Guard> From<PoisonError<Guard>> for TantivyError {
     }
 }
 
-impl From<chrono::ParseError> for TantivyError {
-    fn from(err: chrono::ParseError) -> TantivyError {
+impl From<time::error::Format> for TantivyError {
+    fn from(err: time::error::Format) -> TantivyError {
+        TantivyError::InvalidArgument(err.to_string())
```
fulmicoton:
```suggestion
        TantivyError::InvalidArgument(format!("Date parsing error: {err}"))
```

src/error.rs
```diff
@@ -149,8 +149,20 @@ impl<Guard> From<PoisonError<Guard>> for TantivyError {
...
+    }
+}
+
+impl From<time::error::Parse> for TantivyError {
+    fn from(err: time::error::Parse) -> TantivyError {
+        TantivyError::InvalidArgument(err.to_string())
```
fulmicoton:
```suggestion
        TantivyError::InvalidArgument(format!("Date parsing error: {err}"))
```

src/error.rs
```diff
@@ -149,8 +149,20 @@ impl<Guard> From<PoisonError<Guard>> for TantivyError {
...
+    }
+}
+
+impl From<time::error::ComponentRange> for TantivyError {
+    fn from(err: time::error::ComponentRange) -> TantivyError {
         TantivyError::InvalidArgument(err.to_string())
```
fulmicoton:
```suggestion
        TantivyError::InvalidArgument(format!("Date parsing error: {err}"))
```

src/fastfield/readers.rs
```diff
@@ -147,10 +149,10 @@ impl FastFieldReaders {
         self.typed_fast_field_reader(field)
     }
 
-    /// Returns the `i64` fast field reader reader associated to `field`.
+    /// Returns the `date` fast field reader reader associated to `field`.
```
fulmicoton, line 150: thx

src/lib.rs
```diff
@@ -124,8 +124,6 @@ mod functional_test;
 #[macro_use]
 mod macros;
 
-pub use chrono;
```
fulmicoton, line 128: I want to keep the reexport. Let me explain in the ticket.

fulmicoton (changes requested):
Awesome! Thank you!

Can you re-add the re-export or make the case against it on the ticket? I commented there?
Also can you add your contribution on CHANGELOG.md. Don't forget to put your github username.
Make sure to highlight the breaking changes:
`chrono -> time`
`timestamp -> unix_timestamp` <---- I'm not sure that one is useful tbh.

src/lib.rs
uklotzde: https://github.com/quickwit-oss/tantivy/pull/1307/commits/431a901d0889cc373851082a09dacce8fcb4475e

src/error.rs
uklotzde: https://github.com/quickwit-oss/tantivy/pull/1307/commits/2d787e861de6a28e7cd90e057e934d81762264e0

src/error.rs
uklotzde: https://github.com/quickwit-oss/tantivy/pull/1307/commits/2d787e861de6a28e7cd90e057e934d81762264e0

src/error.rs
uklotzde: https://github.com/quickwit-oss/tantivy/pull/1307/commits/2d787e861de6a28e7cd90e057e934d81762264e0

uklotzde:
I decided against introducing a type alias `DateTime` for `time::OffsetDateTime`. The code is more concise and less confusing when explicitly using the actual types from `crate::time`. Re-exporting the crate as a module should be sufficient.

uklotzde:
@fulmicoton I am still unsure if it would be better to use `PrimitiveDateTime` instead of `OffsetDateTime` in the public API? Since the time zone information gets lost when storing an `OffsetDateTime` I consider `PrimitiveDateTime` more appropriate. The time zone UTC is implicit and only matters for the internal integer conversions.

uklotzde:
I am convinced that storing `PrimitiveDateTime` instead of `OffsetDateTime` is the right choice and prevents inconsistent behavior.

For convenience `impl From<OffsetDateTime> for Value` is provided that performs the conversion to UTC. In the opposite direction you have to invoke `PrimitiveDateTime::assume_utc()` explicitly to get back an `OffsetDateTime`.

Please verify: https://github.com/quickwit-oss/tantivy/pull/1307/commits/304013f9c413730f19b6ee2ff3ea77efbb23a7f9

fulmicoton:
> I am convinced that storing PrimitiveDateTime instead of OffsetDateTime is the right choice and prevents inconsistent behavior.

Haha, I noticed that yesterday before going to sleep 
I'm glad you already made the change.

fulmicoton (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+Unreleased
+================================
+- For date values `chrono` has been replaced with `time`:
```
fulmicoton:
```suggestion
- For date values `chrono` has been replaced with `time` (@uklotzde) #1304 :
```

fulmicoton (approved):

uklotzde:
Converted to draft because there are still some API inconsistencies that need to be resolved. Moreover all time zone conversions should be explicit, not implicit.

fulmicoton:
@uklotzde Understood! Please let me know when I should resume the review.

uklotzde:
@fulmicoton Done. Please see the commit messages for a summary of the changes.

I would also like to drop the convenient but lossy `impl From<OffsetDateTime> for Value` and replace it with an explicit conversion function `Value::date_utc()` that resembles the newly added `Document::add_date_utc()`. But unfortunately that would break the `doc!` macro for `OffsetDateTime` (e.g. used by many tests) and I don't know how to fix that easily.

uklotzde:
A simple extension of the `time` API would serve us well: https://github.com/time-rs/time/pull/454

uklotzde:
More and more I come to the conclusion that using a dedicated newtype with conversion functions would be the correct approach. Using `PrimitiveDateTime`/`OffsetDateTime` is unsound because storing those values is lossy by implicitly reducing the precision to seconds.

uklotzde:
Phew, that escalated quickly.

Fortunately, introducing the newtype revealed how scattered the conversion logic was. All conversions are now implemented by dedicated functions of `DateTime`.

uklotzde:
Ready again. Hopefully I didn't mess anything up. Especially the `FastValue` conversions `to_u64()` and `as_64()` are easy to confuse. A less generic and more distinctive naming would help.

src/lib.rs
```diff
@@ -124,7 +124,81 @@ mod functional_test;
...
+///
+/// This timestamp does not carry any time zone information.
+/// Users are responsible for applying the provided conversion
+/// functions consistently.
+///
+/// All constructors and conversions are provided as explicit
+/// functions and not by implementing any `From`/`Into` traits
+/// to prevent unintended usage.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct DateTime {
```
fulmicoton, line 147: Can we have a nicer implem of Debug showing Rfc3339?

fulmicoton: I wonder if we should rename it `UtcDateTime`?

fulmicoton (approved):

fulmicoton:
I agree with your change (introducing a DateTime object)... Chrono was nice for this.

src/lib.rs
uklotzde: Just the timestamp `2019-10-12T05:20:50.52Z` or like `DateTime { 2019-10-12T05:20:50.52Z }`?

fulmicoton: Just the date is ok I think.

uklotzde: https://github.com/quickwit-oss/tantivy/pull/1307/commits/d6b46a3da14672fc9f2429115c53c06ce6f20f43

The documentation of the `DateTime` wrapper now mentions the internal UTC representation, because that leaks not only for debugging but also when serializing as a JSON value.

I have noticed that the query parser is not tested with explicit time zone offsets, only with UTC? But that's unrelated and a different task.

uklotzde:
> I agree with your change (introducing a DateTime object)... Chrono was nice for this.

Nevertheless using the `chrono` type directly was unsound because of the precision loss when storing those values.

fulmicoton (approved):

fulmicoton:
Thank you @uklotzde for the very solid work.

uklotzde:
@fulmicoton Please double check once again that I didn't miss anything. Especially, if the explicit truncation to seconds inside the newtype is viable and doesn't break existing use cases. Formerly this truncation code was scattered at various places. I am still not completely familiar with when and how values are transformed and stored.

fulmicoton:
It should be ok @uklotzde

PR #1315, opened 2022-03-17
Removes all usage of block_on, and use a oneshot channel instead.

fulmicoton:
Calling `block_on` panics in certain context.
For instance, it panics when it is called in a the context of another
call to block.

Using it in tantivy is unnecessary. We replace it by a thin wrapper
around a oneshot channel that supports both async/sync.

### DIFF
--- a/src/directory/watch_event_router.rs
+++ b/src/directory/watch_event_router.rs
@@ -1,7 +1,6 @@
 use std::sync::{Arc, RwLock, Weak};
 
-use futures::channel::oneshot;
-use futures::{Future, TryFutureExt};
+use crate::FutureResult;
 
 /// Cloneable wrapper for callbacks registered when watching files of a `Directory`.
 #[derive(Clone)]
@@ -74,12 +73,11 @@ impl WatchCallbackList {
     }
 
     /// Triggers all callbacks
-    pub fn broadcast(&self) -> impl Future<Output = ()> {
+    pub fn broadcast(&self) -> FutureResult<()> {
         let callbacks = self.list_callback();
-        let (sender, receiver) = oneshot::channel();
-        let result = receiver.unwrap_or_else(|_| ());
+        let (result, sender) = FutureResult::create("One of the callback panicked.");
         if callbacks.is_empty() {
-            let _ = sender.send(());
+            let _ = sender.send(Ok(()));
             return result;
         }
         let spawn_res = std::thread::Builder::new()
@@ -88,7 +86,7 @@ impl WatchCallbackList {
                 for callback in callbacks {
                     callback.call();
                 }
-                let _ = sender.send(());
+                let _ = sender.send(Ok(()));
             });
         if let Err(err) = spawn_res {
             error!(
@@ -106,8 +104,6 @@ mod tests {
     use std::sync::atomic::{AtomicUsize, Ordering};
     use std::sync::Arc;
 
-    use futures::executor::block_on;
-
     use crate::directory::{WatchCallback, WatchCallbackList};
 
     #[test]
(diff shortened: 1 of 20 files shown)


src/directory/watch_event_router.rs
```diff
@@ -96,7 +93,17 @@ impl WatchCallbackList {
                 err
             );
         }
-        result
+        broadcast_cb
+    }
+}
+
+pub struct BroadcastCallback {
```
fulmicoton: Remove this and only use ScheduledResult

src/indexer/segment_updater.rs
```diff
@@ -468,21 +456,20 @@ impl SegmentUpdater {
...
-        self.schedule_task(async move {
+        self.schedule_task(move || {
             let segment_entries = segment_updater.purge_deletes(opstamp)?;
             segment_updater.segment_manager.commit(segment_entries);
             segment_updater.save_metas(opstamp, payload)?;
-            let _ = garbage_collect_files(segment_updater.clone()).await;
-            segment_updater.consider_merge_options().await;
-            Ok(())
+            let _ = garbage_collect_files(segment_updater.clone());
+            segment_updater.consider_merge_options();
```
PSeitz, line 471: previously they were executed serially, now it's fire and forget

src/indexer/segment_updater.rs
```diff
@@ -349,39 +347,29 @@ impl SegmentUpdater {
...
-            return Err(crate::TantivyError::SystemError(
-                "Segment updater killed".to_string(),
-            ));
+            return crate::TantivyError::SystemError("Segment updater killed".to_string()).into();
         }
-        let (sender, receiver) = oneshot::channel();
-        self.pool.spawn_ok(async move {
-            let task_result = task.await;
+        let (scheduled_result, sender) =
+            FutureResult::create("A segment_updater future did not success. This  never happen.");
```
PSeitz:
```suggestion
            FutureResult::create("A segment_updater future did not succeed. This can never happen.");
```

src/future_result.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub fn wait(self) -> crate::Result<T> {
+        match self.inner {
+            Inner::FailedBeforeStart(err) => Err(err.unwrap()),
+            Inner::InProgress {
+                receiver,
+                error_msg_if_failure,
+            } => receiver.recv().unwrap_or_else(|_| {
+                Err(crate::TantivyError::SystemError(
+                    error_msg_if_failure.to_string(),
+                ))
```
PSeitz, line 61: can't we return the original error?

src/indexer/segment_updater.rs
fulmicoton: not really. previously the function were async for shitty reasons,. now they are blocking.

PSeitz: that's quite confusing, shouldn't there be a `wait()`?

fulmicoton: Ah yes the first one needs to be wait !

src/indexer/segment_updater.rs
```diff
@@ -468,21 +456,20 @@ impl SegmentUpdater {
...
         let segment_updater: SegmentUpdater = self.clone();
-        self.schedule_task(async move {
+        self.schedule_task(move || {
             let segment_entries = segment_updater.purge_deletes(opstamp)?;
             segment_updater.segment_manager.commit(segment_entries);
             segment_updater.save_metas(opstamp, payload)?;
-            let _ = garbage_collect_files(segment_updater.clone()).await;
-            segment_updater.consider_merge_options().await;
-            Ok(())
+            let _ = garbage_collect_files(segment_updater.clone());
```
fulmicoton, line 469:
```suggestion
            let _ = garbage_collect_files(segment_updater.clone()).wait();
```

src/indexer/segment_updater.rs
fulmicoton: ah no both are blocking. There is no issue here.
I think what is misleading you is that the method garbage_collect_files return a FutureResult. This one is a function that actually performs the GC.

src/future_result.rs
fulmicoton: We do if on the other side, we sent the error to the FutureResult.
This error is only used if the sender side was dropped.

Even if we were very careful it could happen on a panic for instance.

src/indexer/segment_updater.rs
PSeitz:
```suggestion
            FutureResult::create("A segment_updater future did not succeed. This should never happen.");
```

src/indexer/prepared_commit.rs
```diff
@@ -37,20 +35,18 @@ impl<'a> PreparedCommit<'a> {
...
+        self.commit_future().wait()
     }
 
     /// Proceeds to commit.
     ///
     /// Unfortunately, contrary to what `PrepareCommit` may suggests,
     /// this operation is not at all really light.
     /// At this point deletes have not been flushed yet.
-    pub async fn commit_async(self) -> crate::Result<Opstamp> {
+    pub fn commit_future(self) -> FutureResult<Opstamp> {
```
PSeitz, line 46: I think schedule_commit would be a better name

src/indexer/prepared_commit.rs
```diff
@@ -37,20 +35,18 @@ impl<'a> PreparedCommit<'a> {
     /// Proceeds to commit.
     /// See `.commit_async()`.
```
PSeitz: update comment

PSeitz (approved):

PR #1319, opened 2022-03-20
Expull: replace read_to_end with iterator over bytes

PSeitz:
Expotential Unrolled List
read_to_end in expull may consume a lot of memory.
Since it is used by the postinglist record, it contains all docids(+optional positions, term frequencies) for one term, replace copy with iterator

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -11,7 +11,10 @@ mod writer;
 
 pub use bitset::*;
 pub use serialize::{BinarySerializable, DeserializeFrom, FixedSize};
-pub use vint::{read_u32_vint, read_u32_vint_no_advance, serialize_vint_u32, write_u32_vint, VInt};
+pub use vint::{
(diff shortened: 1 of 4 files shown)


fulmicoton:
I prefer to work with a `&[u8]`. It makes it much easier to optimize things.

Did you observe a performance regression / improvement?
Did it shave off the memory peaks you observed before during indexing?

PSeitz:
> 
> Did you observe a performance regression / improvement? Did it shave off the memory peaks you observed before during indexing?

I didn't see an impact on indexing performance.

I noticed a big chunk (33.6MB) due to read_to_end, which was gone. This also makes sense, the posting lists can get huge for some terms.

> I prefer to work with a `&[u8]`. It makes it much easier to optimize things.

For decompression a single vint an iterator was already used.
I agree, on &[u8] is better to optimize, what I would prefer here is to complete one block on the correct vint bounds, so that we can create an vint iterator over the blocks.

PR #1324, opened 2022-03-27
Use rayon for indexing workers.

fmassot:


src/indexer/index_writer.rs
```diff
@@ -61,7 +60,7 @@ pub struct IndexWriter {
 
     memory_arena_in_bytes_per_thread: usize,
 
-    workers_join_handle: Vec<JoinHandle<crate::Result<()>>>,
+    workers_future_results: Vec<Mutex<FutureResult<()>>>,
```
fmassot, line 63: I had to use a Mutex to be able to clone the `IndexWriter`, useful for multiple producers use case. Not sure if there is a better solution :/

src/indexer/index_writer.rs
```diff
@@ -342,11 +353,12 @@ impl IndexWriter {
...
-        let former_workers_handles = std::mem::take(&mut self.workers_join_handle);
-        for join_handle in former_workers_handles {
-            join_handle
-                .join()
-                .map_err(|_| error_in_index_worker_thread("Worker thread panicked."))?
+        let former_workers_future_results = std::mem::take(&mut self.workers_future_results);
+        for future_result in former_workers_future_results.into_iter() {
+            future_result
+                .into_inner()
+                .map_err(|_| TantivyError::Poisoned)?
```
fmassot, line 360: When we consume the mutex, we may get an error if the mutex is poisoned, hence the `TantivyError::Poisoned`.

fmassot: There is still something that annoys me with Rayon: before we were able to track if a thread was panicking and raise consequently an `error_in_index_worker_thread`. But it's no more the case in the new version.

fmassot: In case of a panic in the closure, the process that spawned it will be killed. Cf. https://docs.rs/rayon/1.5.1/rayon/struct.ThreadPoolBuilder.html#method.panic_handler

src/indexer/index_writer.rs
fulmicoton: i don't think we allow cloning IndexWriter today. Do we?

fmassot: In the multiple producer example

fmassot: Indexwriter is in an arc mutex

src/indexer/index_writer.rs
ChillFish8: > When we consume the mutex, we may get an error if the mutex is poisoned, hence the TantivyError::Poisoned.

I wonder if using the parking_lot crate may be of benefit here? Mutexes are used a fair bit around tantivy (notably around the block cache) although it would be an additional dependency IIRC rayon (along with alot of other libs)  already uses parking_lot so I imagine cargo will be able to re-use.

PR #1325, opened 2022-03-29
fast field on string

PSeitz:
enables FAST on string fields, which creates a fastfield containing the term ordinals

### DIFF
--- a/src/fastfield/multivalued/reader.rs
+++ b/src/fastfield/multivalued/reader.rs
@@ -27,29 +27,13 @@ impl<Item: FastValue> MultiValuedFastFieldReader<Item> {
         }
     }
 
-    /// Returns `(start, stop)`, such that the values associated
-    /// to the given document are `start..stop`.
+    /// Returns `[start, end)`, such that the values associated
+    /// to the given document are `start..end`.
     #[inline]
     fn range(&self, doc: DocId) -> Range<u64> {
         let start = self.idx_reader.get(doc);
-        let stop = self.idx_reader.get(doc + 1);
-        start..stop
-    }
-
-    /// Returns `(start, stop)`, such that the values associated
-    /// to the given documents are `start..stop`.
-    #[inline]
-    fn range_multi(&self, doc: DocId, num: u32) -> Range<u64> {
-        let start = self.idx_reader.get(doc);
-        let stop = self.idx_reader.get(doc + num);
-        start..stop
-    }
-
-    /// Returns the array of values associated to the given `doc`s.
-    #[inline]
-    pub fn get_vals_multi(&self, doc: DocId, num: u32, vals: &mut Vec<Item>) {
-        let range = self.range_multi(doc, num);
-        self.get_vals_for_range(range, vals);
+        let end = self.idx_reader.get(doc + 1);
+        start..end
     }
 
     /// Returns the array of values associated to the given `doc`.
--- a/src/fastfield/multivalued/reader.rs
+++ b/src/fastfield/multivalued/reader.rs
@@ -36,13 +36,35 @@ impl<Item: FastValue> MultiValuedFastFieldReader<Item> {
         start..stop
     }
 
+    /// Returns `(start, stop)`, such that the values associated
+    /// to the given documents are `start..stop`.
+    #[inline]
+    fn range_multi(&self, doc: DocId, num: u32) -> Range<u64> {
+        let start = self.idx_reader.get(doc);
+        let stop = self.idx_reader.get(doc + num);
+        start..stop
+    }
+
+    /// Returns the array of values associated to the given `doc`s.
+    #[inline]
+    pub fn get_vals_multi(&self, doc: DocId, num: u32, vals: &mut Vec<Item>) {
+        let range = self.range_multi(doc, num);
+        self.get_vals_for_range(range, vals);
+    }
+
+    /// Returns the array of values associated to the given `doc`.
+    #[inline]
+    fn get_vals_for_range(&self, range: Range<u64>, vals: &mut Vec<Item>) {
+        let len = (range.end - range.start) as usize;
+        vals.resize(len, Item::make_zero());
+        self.vals_reader.get_range(range.start, &mut vals[..]);
+    }
+
     /// Returns the array of values associated to the given `doc`.
     #[inline]
     pub fn get_vals(&self, doc: DocId, vals: &mut Vec<Item>) {
         let range = self.range(doc);
-        let len = (range.end - range.start) as usize;
-        vals.resize(len, Item::make_zero());
-        self.vals_reader.get_range(range.start, &mut vals[..]);
+        self.get_vals_for_range(range, vals);
     }
 
     /// Returns the minimum value for this fast field.
--- a/src/fastfield/multivalued/writer.rs
+++ b/src/fastfield/multivalued/writer.rs
@@ -4,7 +4,7 @@ use fnv::FnvHashMap;
 use tantivy_bitpacker::minmax;
 
 use crate::fastfield::serializer::BitpackedFastFieldSerializerLegacy;
-use crate::fastfield::{value_to_u64, CompositeFastFieldSerializer};
+use crate::fastfield::{value_to_u64, CompositeFastFieldSerializer, FastFieldType};
 use crate::indexer::doc_id_mapping::DocIdMapping;
 use crate::postings::UnorderedTermId;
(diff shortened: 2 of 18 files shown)


src/schema/text_options.rs
```diff
@@ -14,6 +15,13 @@ pub struct TextOptions {
     indexing: Option<TextFieldIndexing>,
     #[serde(default)]
     stored: bool,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "is_false")]
```
fulmicoton:
```suggestion
```

fulmicoton: I think it is nice to avoid skipping it on serialize. (We want to handle its absence though)

src/fastfield/multivalued/reader.rs
```diff
@@ -36,15 +36,37 @@ impl<Item: FastValue> MultiValuedFastFieldReader<Item> {
         start..stop
     }
 
+    /// Returns `(start, stop)`, such that the values associated
```
fulmicoton:
```suggestion
    /// Returns `[start, end)`, such that the values associated
```

src/fastfield/multivalued/reader.rs
```diff
@@ -36,15 +36,37 @@ impl<Item: FastValue> MultiValuedFastFieldReader<Item> {
...
+    /// to the given documents are `start..stop`.
```
fulmicoton:
```suggestion
    /// to the given documents are `[start..end)`.
```

src/fastfield/multivalued/reader.rs
```diff
@@ -36,15 +36,37 @@ impl<Item: FastValue> MultiValuedFastFieldReader<Item> {
...
+    #[inline]
+    fn range_multi(&self, doc: DocId, num: u32) -> Range<u64> {
+        let start = self.idx_reader.get(doc);
+        let stop = self.idx_reader.get(doc + num);
+        start..stop
```
fulmicoton:
```suggestion
        let end = self.idx_reader.get(doc + num);
        start..end
```

src/fastfield/multivalued/reader.rs
```diff
@@ -36,15 +36,37 @@ impl<Item: FastValue> MultiValuedFastFieldReader<Item> {
...
+    }
+
+    /// Returns the array of values associated to the given `doc`s.
+    #[inline]
+    pub fn get_vals_multi(&self, doc: DocId, num: u32, vals: &mut Vec<Item>) {
```
fulmicoton: is that a range of docs?
Shouldn't we use the following prototype?

```
fn get_vals_doc_range(&self, doc_range: Range<DocId>, vals: &mut Vec<Item>)
```

I have no idea what is the point of this function though.

fulmicoton (changes requested):
See comment inline.
- point of the multi accessor?
- if it is relevant and we want to keep it, we should fix its doc and prototype.
- minor rustdoc changes.

src/fastfield/multivalued/reader.rs
PSeitz: I moved the code to the tests

src/schema/text_options.rs
PSeitz: absence would be false, should be fine/intuitive for fastfields on string?

fulmicoton: Oh yes it is intuitive, but it has no real benefit and hurts discoverability.

src/fastfield/multivalued/reader.rs
```diff
@@ -27,24 +27,30 @@ impl<Item: FastValue> MultiValuedFastFieldReader<Item> {
...
+        let end = self.idx_reader.get(doc + 1);
+        start..end
     }
 
     /// Returns the array of values associated to the given `doc`.
     #[inline]
-    pub fn get_vals(&self, doc: DocId, vals: &mut Vec<Item>) {
-        let range = self.range(doc);
+    fn get_vals_for_range(&self, range: Range<u64>, vals: &mut Vec<Item>) {
         let len = (range.end - range.start) as usize;
```
fulmicoton, line 42: I think there is a range.len()

src/fastfield/multivalued/writer.rs
```diff
@@ -38,16 +38,19 @@ pub struct MultiValuedFastFieldWriter {
...
     doc_index: Vec<u64>,
+    is_storing_term_ids: bool,
     is_facet: bool,
 }
 
 impl MultiValuedFastFieldWriter {
-    /// Creates a new `IntFastFieldWriter`
-    pub(crate) fn new(field: Field, is_facet: bool) -> Self {
+    /// Creates a new `MultiValuedFastFieldWriter`
+    pub(crate) fn new(field: Field, mut is_storing_term_ids: bool, is_facet: bool) -> Self {
```
fulmicoton: I wonder if it would be less confusing if we added a enum type

```
enum FastFieldType {
   Numerical,
   String,
   Facet,
}

File: src/fastfield/readers.rs [hunk 8]
```diff
@@ -39,6 +39,9 @@ pub(crate) fn type_and_cardinality(field_type: &FieldType) -> Option<(FastType,
             .get_fastfield_cardinality()
             .map(|cardinality| (FastType::Date, cardinality)),
         FieldType::Facet(_) => Some((FastType::U64, Cardinality::MultiValues)),
+        FieldType::Str(options) if options.is_fast() => {
+            Some((FastType::U64, Cardinality::MultiValues))
```
fulmicoton, line 43: Do you think we could handle Cardinality::SingleValue too?

src/postings/postings_writer.rs
```diff
@@ -145,6 +146,7 @@ pub(crate) trait PostingsWriter {
         term_buffer: &mut Term,
         ctx: &mut IndexingContext,
         indexing_position: &mut IndexingPosition,
+        mut term_id_fast_field_writer_opt: Option<&mut MultiValuedFastFieldWriter>,
```
fulmicoton, line 149: I think you can remove mut here

src/postings/postings_writer.rs
```diff
@@ -164,9 +166,14 @@ pub(crate) trait PostingsWriter {
             term_buffer.append_bytes(token.text.as_bytes());
             let start_position = indexing_position.end_position + token.position as u32;
             end_position = start_position + token.position_length as u32;
-            self.subscribe(doc_id, start_position, term_buffer, ctx);
+            let unordered_term_id = self.subscribe(doc_id, start_position, term_buffer, ctx);
+            if let Some(term_id_fast_field_writer) = term_id_fast_field_writer_opt.as_mut() {
```
fulmicoton, line 170: ... and remove as_mut here. If it does not compile, add `.clone()`

Right now the type of term_id_fast_field_writer is `&mut &mut T`

fulmicoton (changes requested):
See comments.

src/fastfield/multivalued/reader.rs
PSeitz: weirdly doesn't seem to be implemented for `Range<u64>`

src/postings/postings_writer.rs
PSeitz: clone doesn't work on that type

```
[rustc E0599] [E] the method `clone` exists for enum `std::option::Option<&mut MultiValuedFastFieldWriter>`, but its trait bounds were not satisfied
the following trait bounds were not satisfied:
`&mut MultiValuedFastFieldWriter: Clone`
```

src/fastfield/multivalued/reader.rs
fulmicoton: It's given via the [ExactSizeIterator](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html) trait

src/postings/postings_writer.rs
fulmicoton: Ah right that's interesting!

src/fastfield/multivalued/writer.rs
PSeitz: yep

src/fastfield/readers.rs
fulmicoton: discussed offline, @PSeitz  will try that in another PR.

src/postings/postings_writer.rs
fulmicoton: we cannot

fulmicoton (approved):

PR #1329, opened 2022-04-06
add term aggregation

PSeitz:


src/aggregation/agg_req_with_accessor.rs
```diff
@@ -27,11 +31,32 @@ impl AggregationsWithAccessor {
     }
 }
 
+#[derive(Clone)]
+pub(crate) enum FastFieldAccessor {
+    Multi(MultiValuedFastFieldReader<u64>),
+    Single(DynamicFastFieldReader<u64>),
+}
+impl FastFieldAccessor {
+    pub fn as_single(&self) -> &DynamicFastFieldReader<u64> {
```
fulmicoton: can we return an Option and panic on the call site?

src/aggregation/agg_req.rs
```diff
@@ -102,10 +102,25 @@ pub(crate) struct BucketAggregationInternal {
 impl BucketAggregationInternal {
     pub(crate) fn as_histogram(&self) -> &HistogramAggregation {
```
fulmicoton: can we return an Option and panic on the call site?

src/aggregation/agg_req.rs
```diff
@@ -102,10 +102,25 @@ pub(crate) struct BucketAggregationInternal {
...
         match &self.bucket_agg {
-            BucketAggregationType::Range(_) => panic!("unexpected aggregation"),
             BucketAggregationType::Histogram(histogram) => histogram,
+            _ => panic!("unexpected aggregation"),
         }
     }
+    pub(crate) fn as_term(&self) -> &TermsAggregation {
```
fulmicoton: can we return an Option and panic on the call site?

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -27,11 +31,32 @@ impl AggregationsWithAccessor {
...
+        match self {
+            FastFieldAccessor::Multi(_) => panic!("unexpected ff cardinality"),
+            FastFieldAccessor::Single(reader) => reader,
+        }
+    }
+    pub fn as_multi(&self) -> &MultiValuedFastFieldReader<u64> {
```
fulmicoton: can we return an Option and panic on the call site?

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -121,29 +158,36 @@ pub(crate) fn get_aggs_with_accessor_and_validate(
 fn get_ff_reader_and_validate(
     reader: &SegmentReader,
     field_name: &str,
-) -> crate::Result<(DynamicFastFieldReader<u64>, Type)> {
+    multi: bool,
```
fulmicoton: I don't understand the multi argument here. 

What happens if multi == true and Cardinality is single valued?
Is this ok?

Can we comment on the function?
Shouldl the bool be of the type Cardinality instead? and call it cardinality too?

fulmicoton (changes requested):
See minor commetns inline

src/aggregation/agg_req.rs
PSeitz: yep

src/aggregation/agg_req_with_accessor.rs
PSeitz: I refactored the function and it accepts cardinality as parameter, I think it should be clear now

src/aggregation/mod.rs
```diff
@@ -337,17 +338,79 @@ mod tests {
...
+                Term::from_field_text(text_field, term),
+                IndexRecordOption::Basic,
+            );
+
+            searcher.search(&term_query, &collector)?
+        } else {
+            searcher.search(&AllQuery, &collector)?
+        };
+
+        let res: Value = serde_json::from_str(&serde_json::to_string(&agg_res)?)?;
```
fulmicoton, line 367: Can you add a commment explaining why we serialize and deserialize?
Wouldn't serde_json::to_value do the job ?

src/aggregation/mod.rs
```diff
@@ -337,17 +338,79 @@ mod tests {
...
+        agg_req: Aggregations,
+        index: &Index,
+        query: Option<(&str, &str)>,
+    ) -> crate::Result<Value> {
+        let collector = AggregationCollector::from_aggs(agg_req);
+
+        let reader = index.reader()?;
+        let searcher = reader.searcher();
+        let agg_res = if let Some((field, term)) = query {
+            let text_field = reader.searcher().schema().get_field(field).unwrap();
```
fulmicoton, line 354: can we remove this unwrap?

I don't understand why it is justified. Assuming it is justified, that's a hidden contract.

fulmicoton: ah ok this is just for tests! don't worry then!

src/aggregation/bucket/range.rs
```diff
@@ -102,22 +103,69 @@ pub struct SegmentRangeCollector {
     field_type: Type,
 }
 
+#[derive(Clone, PartialEq)]
+pub(crate) struct SegmentRangeBucketEntry {
+    pub key: Key,
+    pub doc_count: u64,
+    pub sub_aggregation: Option<SegmentAggregationResultsCollector>,
+    /// The from range of the bucket. Equals f64::MIN when None.
+    pub from: Option<f64>,
```
fulmicoton, line 112: what kind of interval is this? Semi-open?

src/aggregation/agg_req_with_accessor.rs
fulmicoton: Is this pushed?

PSeitz: It should be now

src/aggregation/mod.rs
```diff
@@ -454,17 +533,18 @@ mod tests {
...
         } else {
             let collector = AggregationCollector::from_aggs(agg_req);
 
             let searcher = reader.searcher();
-            searcher.search(&term_query, &collector).unwrap()
+            searcher.search(&AllQuery, &collector).unwrap()
         };
 
         let res: Value = serde_json::from_str(&serde_json::to_string(&agg_res)?)?;
+        // println!("{}", serde_json::to_string_pretty(&res).unwrap());
```
```suggestion
```

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -78,10 +115,14 @@ impl MetricAggregationWithAccessor {
...
             | MetricAggregation::Stats(StatsAggregation { field: field_name }) => {
-                let (accessor, field_type) = get_ff_reader_and_validate(reader, field_name)?;
+                let (accessor, field_type) =
+                    get_ff_reader_and_validate(reader, field_name, Cardinality::SingleValue)?;
 
                 Ok(MetricAggregationWithAccessor {
-                    accessor,
+                    accessor: accessor
+                        .as_single()
+                        .expect("unexpected fast field cardinatility")
```
fulmicoton:
```suggestion
                        .expect("unexpected fast field cardinality")
```

src/aggregation/mod.rs
PSeitz: yes, serde_json::to_value is better, I changed it

PSeitz: ah no, I used that here to also test de/serialization roundtrip

fulmicoton (approved):

PR #1341, opened 2022-04-19
make TantivyError cloneable

PSeitz:


fulmicoton:
This removes too much context, and the case for cloneable errors is too weak.

PR #1342, opened 2022-04-19
support order by sub_aggregation

PSeitz:
support order property on term aggregations
order can be by doc_count, key, or a metric sub_aggregation

### DIFF
--- a/src/aggregation/agg_result.rs
+++ b/src/aggregation/agg_result.rs
@@ -7,7 +7,6 @@
 use std::cmp::Ordering;
 use std::collections::HashMap;
 
-use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 
 use super::agg_req::{Aggregations, AggregationsInternal, BucketAggregationInternal};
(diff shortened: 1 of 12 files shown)


src/aggregation/agg_result.rs
```diff
@@ -18,55 +17,77 @@ use super::intermediate_agg_result::{
...
 impl AggregationResults {
+    pub(crate) fn get_value_from_aggregation(
+        &self,
+        name: &str,
+        agg_property: &str,
+    ) -> crate::Result<Option<f64>> {
+        if let Some(agg) = self.0.get(name) {
+            agg.get_value_from_aggregation(name, agg_property)
+        } else {
+            // Should return an error here? Missing aggregation could be intentional to save
```
fulmicoton: you tell me :D

src/aggregation/agg_result.rs
```diff
@@ -18,55 +17,77 @@ use super::intermediate_agg_result::{
...
     ///
     /// Internal function, CollectorAggregations is used instead Aggregations, which is optimized
     /// for internal processing, by splitting metric and buckets into seperate groups.
     pub(crate) fn from_intermediate_and_req_internal(
         results: IntermediateAggregationResults,
         req: &AggregationsInternal,
-    ) -> Self {
-        let mut result = HashMap::default();
+    ) -> crate::Result<Self> {
+        // let mut result = HashMap::default();
```
fulmicoton:
```suggestion
```

fulmicoton (changes requested):
can you make that branch in quickwit-oss?

src/aggregation/agg_result.rs
PSeitz: Currently this can't happen. We do validation on the request for this.

In the future it could be intentional or a bug, so I'm not sure. Maybe be strict?

PSeitz: I changed it to return an error

PR #1346, opened 2022-04-22
term agg

PSeitz:
- support order property on term aggregations
- return error on missing sub aggregation
- fix is_fast for bytes field
- minor fixes

### DIFF
--- a/src/aggregation/agg_result.rs
+++ b/src/aggregation/agg_result.rs
@@ -7,7 +7,6 @@
 use std::cmp::Ordering;
 use std::collections::HashMap;
 
-use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 
 use super::agg_req::{Aggregations, AggregationsInternal, BucketAggregationInternal};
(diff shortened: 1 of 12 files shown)


src/aggregation/agg_result.rs
```diff
@@ -18,55 +17,76 @@ use super::intermediate_agg_result::{
...
-
+    ) -> crate::Result<Self> {
         // Important assumption:
         // When the tree contains buckets/metric, we expect it to have all buckets/metrics from the
         // request
-        if let Some(buckets) = results.buckets {
-            result.extend(buckets.into_iter().zip(req.buckets.values()).map(
-                |((key, bucket), req)| {
-                    (
+        let mut result: HashMap<_, _> = if let Some(buckets) = results.buckets {
```
fulmicoton: This is very hard to follow.
Can you:
- isolate the two bodies in independent functions?
- add an assert on the stuff being zipped being of the same size.
- use an iterative style maybe?

src/aggregation/bucket/mod.rs
```diff
@@ -11,8 +11,130 @@ mod histogram;
...
+        order: Order::Desc,
+    };
+
+    let order_str = serde_json::to_string(&order).unwrap();
+    assert_eq!(order_str, "{\"_key\":\"desc\"}");
+    let order_deser = serde_json::from_str(&order_str).unwrap();
+
+    assert_eq!(order, order_deser);
+
+    let order_deser: serde_json::Result<CustomOrder> = serde_json::from_str(&"{}");
```
fulmicoton:
```suggestion
    let order_deser: serde_json::Result<CustomOrder> = serde_json::from_str("{}");
```

src/aggregation/intermediate_agg_result.rs
```diff
@@ -218,7 +221,7 @@ impl IntermediateBucketResult {
                     })
                     .collect();
 
-                std::mem::swap(entries_left, &mut buckets);
+                std::mem::swap(buckets_left, &mut buckets);
```
fulmicoton: the use of mem::swap is useless here.

src/aggregation/intermediate_agg_result.rs
```diff
@@ -246,22 +255,75 @@ impl IntermediateTermBucketResult {
...
+                })
             })
-            .collect();
-        buckets.sort_unstable_by_key(|bucket| std::cmp::Reverse(bucket.doc_count));
+            .collect::<crate::Result<_>>()?;
+
+        let order = req.order.order;
+        match req.order.target {
+            OrderTarget::Key => {
+                buckets.sort_by(|bucket1, bucket2| {
```
fulmicoton: nitpick, prefer left and right than 1 / 2

src/aggregation/intermediate_agg_result.rs
fulmicoton: buckets_left = buckets;

src/aggregation/intermediate_agg_result.rs
```diff
@@ -246,22 +255,75 @@ impl IntermediateTermBucketResult {
...
+                    if req.order.order == Order::Desc {
+                        bucket1
```
fulmicoton: very local, but there is probably a nicer way to write this.

fulmicoton (approved):
Please address inline comments.

Also can you do a pass on the PR and see if the code can be simplified?

src/aggregation/intermediate_agg_result.rs
PSeitz: entries_left is a &mut binding from a match pattern and buckets is a temporary variable. I tried some things, but couldn't find a solution other than mem::swap

fulmicoton: `*entries_left = buckets` did not work?

src/aggregation/intermediate_agg_result.rs
PSeitz: I did a minor improvement on it

PR #1347, opened 2022-04-22
fix query parser error field not found

PSeitz:
fix query parser error field not found

return query on parse error


fulmicoton (approved):

fulmicoton:
Nice

PSeitz:
Fixes https://github.com/quickwit-oss/quickwit/issues/748
Related to https://github.com/quickwit-oss/quickwit/issues/1145

PR #1349, opened 2022-04-27
print whole query on syntax error

### LINKED ISSUE #748: Repeat query in syntax error
When there is a syntax error in the query parser, one important context is the query itself.

Since we pass it via an URL, it is not uncommon to mess up with the URL encoding. 
If Quickwit repeated the query it was trying to parse  in the error, that could help spot trivial URL encoding errors very fast.

PSeitz:
Fixes https://github.com/quickwit-oss/quickwit/issues/748


PSeitz:
Fixes https://github.com/quickwit-oss/quickwit/issues/748

fulmicoton (approved):

PR #1350, opened 2022-04-28
update edition

PSeitz:
update dependencies, update edition

### DIFF
--- a/src/core/executor.rs
+++ b/src/core/executor.rs
@@ -1,7 +1,7 @@
-use std::sync::atomic::AtomicUsize;
-
 use crossbeam::channel;
 use rayon::{ThreadPool, ThreadPoolBuilder};
+
+use crate::TantivyError;
 
 /// Search executor whether search request are single thread or multithread.
 ///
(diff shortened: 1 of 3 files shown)


src/core/executor.rs
```diff
@@ -47,16 +49,17 @@ impl Executor {
         match self {
             Executor::SingleThread => args.map(f).collect::<crate::Result<_>>(),
             Executor::ThreadPool(pool) => {
-                let args_with_indices: Vec<(usize, A)> = args.enumerate().collect();
-                let num_fruits = args_with_indices.len();
+                // let args_with_indices: Vec<(usize, A)> = args.enumerate().collect();
```
fulmicoton:
```suggestion
```

src/core/executor.rs
```diff
@@ -47,16 +49,17 @@ impl Executor {
...
+                let args: Vec<A> = args.collect();
+                let num_fruits = args.len();
                 let fruit_receiver = {
                     let (fruit_sender, fruit_receiver) = channel::unbounded();
                     pool.scope(|scope| {
-                        for arg_with_idx in args_with_indices {
+                        for (idx, arg) in args.into_iter().enumerate() {
```
fulmicoton, line 57: what is this AtomicUsize about?

fulmicoton (changes requested):
See comments

src/core/executor.rs
PSeitz: The way variables are captured seems to have changed.
I does not compile, because idx is not moved into the closure anymore.
Adding `move` does not work, because it also moves `f`, which we don't want to move.

fulmicoton: And the AtomicUsize helps!???

PSeitz: Yes, with the AtomicUsize, we avoid the lifetime issue. This is the error:

```
error[E0373]: closure may outlive the current function, but it borrows `idx`, which is owned by the current function
  --> src/core/executor.rs:60:41
   |
57 |                     pool.scope(|scope| {
   |                                 ----- has type `&rayon::Scope<'1>`
...
60 |                             scope.spawn(|_| {
   |                                         ^^^ may outlive borrowed value `idx`
61 |                                 let fruit = f(arg);
62 |                                 if let Err(err) = fruit_sender.send((idx, fruit)) {
   |                                                                      --- `idx` is borrowed here
```

fulmicoton (approved):

PR #1356, opened 2022-05-03
Minor changes in query parsing to allow quickwit#1334.

### LINKED ISSUE #1355: Making "." in quickwit field parsing spec possible in tantivy
See https://github.com/quickwit-oss/quickwit/issues/1334 and https://github.com/quickwit-oss/quickwit/issues/1338

We need to:
- allow for field names to contain a "\." in the query parser.
- fix the full path -> (field, path) resolving logic 
- Use the `full path -> (field, path)` resolving logic for regular literals.
- Return an error if the path is targeting a non JSON field but with a non empty json path.

fulmicoton:
Minor changes in query parsing for quickwit#1334.
    
    Quickwit's still heavily relies on generating field names
    containing a '.' for nested object, yet allows for
    user defined field names to contain a dot.
    
    In order to reuse tantivy query parser, we will end up
    using quickwit field names directly into tantivy.
    Only '.' will be escaped.
    
    This PR makes minor changes in how tantivy query parser parses
    a field name and resolves it to a field.
    Some of the new edge case behavior is hacky.
    
    Closes #1355

### DIFF
--- a/src/query/query_parser/query_parser.rs
+++ b/src/query/query_parser/query_parser.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeSet, HashMap};
+use std::collections::HashMap;
 use std::num::{ParseFloatError, ParseIntError};
 use std::ops::Bound;
 use std::str::FromStr;
@@ -30,7 +30,7 @@ pub enum QueryParserError {
     #[error("Unsupported query: {0}")]
     UnsupportedQuery(String),
     /// The query references a field that is not in the schema
-    #[error("Field does not exists: '{0:?}'")]
+    #[error("Field does not exists: '{0}'")]
     FieldDoesNotExist(String),
     /// The query contains a term for a `u64` or `i64`-field, but the value
     /// is neither.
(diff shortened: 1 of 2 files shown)


src/query/query_parser/query_parser.rs
```diff
@@ -1399,29 +1437,56 @@ mod test {
...
+            QueryParser::new(schema.clone(), Vec::new(), TokenizerManager::default());
         assert_eq!(
             query_parser.split_full_path("first.toto"),
-            ("first.toto", "")
+            Some((schema.get_field("first.toto").unwrap(), ""))
         );
         assert_eq!(
-            query_parser.split_full_path("first.titi"),
-            ("first", "titi")
+            query_parser.split_full_path("first.toto.bubu"),
```
PSeitz, line 1469: I don't understand this, split_full_path returns a field_name and json path. 
How can a text field have a json path?

fulmicoton: split_full_path just splits something into Field and json path. regardless of the field type.

An error is returned if the field is of type text yes has a non empty json path, but that happens downstream.

PR #1357, opened 2022-05-04
Expose helpers to generate json field writer terms 

### LINKED ISSUE #1302: Expose a public API for interacting with JSON field types outside of the QueryParser
# Expose a public API for interacting with JSON field types outside of the QueryParser

As of right now, there is no way to public way to build queries to interact with the new JSON field type outside of the QueryParser, 
due to the `JsonTermWriter` not being publically exposed, and also possibly slightly overkill.

Ideally a simple(ish) API allows users to produce a term to query with by supplying the Field, JSON Path and value itself.

saroh:
Added an easy "constructor" for given field + path to pass on to helper functions.
Adds two helper functions, unfortunately they cannot have the same signature as text types need a bit more work and generate more than one Term.

I'm not sure about the function naming.

closes #1302

### DIFF
--- a/src/indexer/json_term_writer.rs
+++ b/src/indexer/json_term_writer.rs
@@ -200,30 +200,30 @@ fn infer_type_from_str(text: &str) -> TextOrDateTime {
 }
 
 // Tries to infer a JSON type from a string
-pub(crate) fn infer_fast_value_term(
+pub(crate) fn convert_to_fast_value_and_get_term(
     json_term_writer: &mut JsonTermWriter,
     phrase: &str,
 ) -> Option<Term> {
     if let Ok(dt) = OffsetDateTime::parse(phrase, &Rfc3339) {
         let dt_utc = dt.to_offset(UtcOffset::UTC);
-        return Some(generate_term_from_json_writer(
+        return Some(set_fastvalue_and_get_term(
             json_term_writer,
             DateTime::from_utc(dt_utc),
         ));
     }
     if let Ok(u64_val) = str::parse::<u64>(phrase) {
-        return Some(generate_term_from_json_writer(json_term_writer, u64_val));
+        return Some(set_fastvalue_and_get_term(json_term_writer, u64_val));
     }
     if let Ok(i64_val) = str::parse::<i64>(phrase) {
-        return Some(generate_term_from_json_writer(json_term_writer, i64_val));
+        return Some(set_fastvalue_and_get_term(json_term_writer, i64_val));
     }
     if let Ok(f64_val) = str::parse::<f64>(phrase) {
-        return Some(generate_term_from_json_writer(json_term_writer, f64_val));
+        return Some(set_fastvalue_and_get_term(json_term_writer, f64_val));
     }
     None
 }
 // helper function to generate a Term from a json fastvalue
-pub(crate) fn generate_term_from_json_writer<T: FastValue>(
+pub(crate) fn set_fastvalue_and_get_term<T: FastValue>(
     json_term_writer: &mut JsonTermWriter,
     value: T,
 ) -> Term {
@@ -232,7 +232,7 @@ pub(crate) fn generate_term_from_json_writer<T: FastValue>(
 }
 
 // helper function to generate a list of terms with their positions from a textual json value
-pub(crate) fn generate_terms_from_json_writer(
+pub(crate) fn set_string_and_get_terms(
     json_term_writer: &mut JsonTermWriter,
     value: &str,
     text_analyzer: &TextAnalyzer,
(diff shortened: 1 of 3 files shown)


fulmicoton:
Thanks @saroh !
@PSeitz can you review?

src/indexer/json_term_writer.rs
```diff
@@ -199,12 +199,74 @@ fn infer_type_from_str(text: &str) -> TextOrDateTime {
...
+    if let Ok(i64_val) = str::parse::<i64>(phrase) {
+        return Some(generate_term_from_json_writer(json_term_writer, i64_val));
+    }
+    if let Ok(f64_val) = str::parse::<f64>(phrase) {
+        return Some(generate_term_from_json_writer(json_term_writer, f64_val));
+    }
+    None
+}
+// helper function to generate a Term from a json fastvalue
+pub(crate) fn generate_term_from_json_writer<T: FastValue>(
```
PSeitz: set_fastvalue_and_get_term

src/indexer/json_term_writer.rs
```diff
@@ -199,12 +199,74 @@ fn infer_type_from_str(text: &str) -> TextOrDateTime {
     }
 }
 
+// Tries to infer a JSON type from a string
+pub(crate) fn infer_fast_value_term(
```
PSeitz: I think the naming is hard to follow

convert_to_fast_value_and_get_term

src/indexer/json_term_writer.rs
```diff
@@ -199,12 +199,74 @@ fn infer_type_from_str(text: &str) -> TextOrDateTime {
...
+pub(crate) fn set_fastvalue_and_get_term<T: FastValue>(
+    json_term_writer: &mut JsonTermWriter,
+    value: T,
+) -> Term {
+    json_term_writer.set_fast_value(value);
+    json_term_writer.term().clone()
+}
+
+// helper function to generate a list of terms with their positions from a textual json value
+pub(crate) fn set_string_and_get_terms(
```
saroh, line 235: Thanks for the suggestions, FYI I renamed this one as well!

src/indexer/json_term_writer.rs
```diff
@@ -199,12 +199,74 @@ fn infer_type_from_str(text: &str) -> TextOrDateTime {
...
+}
+
 pub struct JsonTermWriter<'a> {
     term_buffer: &'a mut Term,
     path_stack: Vec<usize>,
 }
 
 impl<'a> JsonTermWriter<'a> {
+    // Prepares writing terms for a given field
+    pub fn initialize(field: Field, json_path: &str, term_buffer: &'a mut Term) -> Self {
```
PSeitz: initialize is quite uncommon for a constructor, typical is `new` or `from_field_and_json_path`.

Can we move ownership of the Term into the JsonTermWriter? Currently it's unclear why it's passed as mut ref (performance or are the mutated object contents used somewhere else?)

saroh: @PSeitz I'll rename it to `from_field_and_json_path`

It looks like we pass the mut ref for performance cf https://github.com/saroh/tantivy/blob/706fbd68868523c4e41ff3341f8a94f37f9f44f2/src/indexer/segment_writer.rs#L276-L288 , should I still move the ownership ?

saroh: I started moving ownership locally, but given It's hard for me to weigh pros and cons and it starts getting a bit consequent in terms of changes before I can run the tests, I've put it on hold. 

From an outsider point of view, IMHO it would simplify a lot the API to move the ownership to the `JsonTermWriter`. 

Maybe it could make sense to make `JsonTermWriter` private for the crate (unsure about the scope here), and expose another struct for general use which would simply wrap a `JsonTermWriter` with ownership of the term buffer ?

Also do we need to make sure the `json_path` is not empty string or is this to be validated elsewhere/ OK for the time being?

PSeitz: > Maybe it could make sense to make JsonTermWriter private for the crate (unsure about the scope here), and expose another struct for general use which would simply wrap a JsonTermWriter with ownership of the term buffer ?

Yeah it's similar to `Term`, which has some API weirdness, which may be improved when splitting in private and public.

> Also do we need to make sure the json_path is not empty string or is this to be validated elsewhere/ OK for the time being?

I don't see `path()` anywhere used in tantivy. If you see  a place where you can add an assertion feel free to add one.

PSeitz (approved):

src/indexer/json_term_writer.rs
saroh: After a bit of research, the specs states we should be OK as is and the check is unnecessary.
image
https://www.ecma-international.org/publications-and-standards/standards/ecma-404/

PR #1358, opened 2022-05-04
add alias shard_size to split_size for quickwit

PSeitz:
add alias shard_size split_size for quickwit

improve some docs

### DIFF
--- a/src/aggregation/bucket/term_agg.rs
+++ b/src/aggregation/bucket/term_agg.rs
@@ -97,11 +97,11 @@ pub struct TermsAggregation {
     /// doc_count returned by each shard. It’s the sum of the size of the largest bucket on
     /// each segment that didn’t fit into `shard_size`.
     ///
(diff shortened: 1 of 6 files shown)


fulmicoton (approved):

src/aggregation/bucket/term_agg.rs
```diff
@@ -81,6 +81,7 @@ pub struct TermsAggregation {
     ///
     /// Should never be smaller than size.
     #[serde(skip_serializing_if = "Option::is_none", default)]
+    #[serde(alias = "split_size")]
```
fulmicoton: nitpick... can we do it the other way around?

variable named split_size and alias = "shard_size" ?

fulmicoton (approved):

src/aggregation/bucket/term_agg.rs
PSeitz: yep, done

PR #1242, opened 2021-12-28
doc(termdict) expose structs

saroh:
also add merger doc + lint
refs #1232

PR #1359, opened 2022-05-04
Fix documentation regression

saroh:
This breaks the doc on doc.rs as the type seems to shadow the struct https://docs.rs/tantivy/latest/tantivy/termdict/type.TermDictionary.html
introduced by #1293 which may not have been up to date with what was done in #1242


fulmicoton:
Thanks and sorry for that!

fulmicoton (approved):

PR #1361, opened 2022-05-06
Shortens blocks' last_key in the SSTable block index.

fulmicoton:
Right now we store last key in the blocks of the SSTable index.
This PR replaces the last key by a shorter string that is greater or
equal and still lesser than the next key.
This property is sufficiently to ensure the block index
works properly.

### DIFF
--- a/src/termdict/sstable_termdict/sstable/mod.rs
+++ b/src/termdict/sstable_termdict/sstable/mod.rs
@@ -145,6 +145,12 @@ where
     }
 
     pub fn write_key(&mut self, key: &[u8]) {
+        // If this is the first key in the block, we use it to
+        // shorten the last term in block in the last block.
+        if self.first_ordinal_of_the_block == self.num_terms {
+            self.index_builder
+                .shorten_last_block_key_given_next_key(key);
+        }
         let keep_len = common_prefix_len(&self.previous_key, key);
         let add_len = key.len() - keep_len;
         let increasing_keys = add_len > 0 && (self.previous_key.len() == keep_len)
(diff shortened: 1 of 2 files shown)


src/termdict/sstable_termdict/sstable/sstable_index.rs
```diff
@@ -41,10 +44,40 @@ pub struct SSTableIndexBuilder {
...
+fn find_shorter_str_in_between(left: &mut Vec<u8>, right: &[u8]) {
+    assert!(&left[..] < right);
+    let common_len = common_prefix_len(&left, right);
+    if left.len() == common_len {
+        return;
+    }
+    // It is possible to do one character shorter in some case,
+    // but it is not worth the extra complexity
+    for pos in (common_len + 1)..left.len() {
+        if left[pos] != 255u8 {
```
PSeitz: what's the meaning of 255 here?

the comment says string, but we operate on bytes, what about utf-8?

fulmicoton: 255 is the max value for a u8.

It is not especially utf8. All of these are byte strings.
There are already u64 byte representation in there.

We do not expose the last term in block so it should be safe.

src/termdict/sstable_termdict/sstable/sstable_index.rs
```diff
@@ -41,10 +44,39 @@ pub struct SSTableIndexBuilder {
...
```
PSeitz:
```suggestion
        if left[pos] != u8::MAX {
```

src/termdict/sstable_termdict/sstable/mod.rs
```diff
@@ -145,6 +145,12 @@ where
     }
 
     pub fn write_key(&mut self, key: &[u8]) {
+        // If this is the first key in the block, we use it to
+        // shorten the last term in block in the last block.
```
PSeitz:
```suggestion
        // shorten the last term in the last block.
```

src/termdict/sstable_termdict/sstable/sstable_index.rs
```diff
@@ -32,7 +33,9 @@ pub struct BlockAddr {
 
 #[derive(Debug, Serialize, Deserialize)]
 struct BlockMeta {
-    pub last_key: Vec<u8>,
+    /// Any byte string that is lexicographically greater than the last key in the block,
```
PSeitz, line 36: greater than or equal to?

PSeitz (approved):

src/termdict/sstable_termdict/sstable/sstable_index.rs
fulmicoton: yes absolutely. fixing.

PR #1363, opened 2022-05-10
Add aggregation bucket limit

### LINKED ISSUE #1331: Aggregation Max Bucket Limit
Currently we don't have a max bucket limit like [elastic](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-settings.html#search-settings-max-buckets) after which we abort.

We should have one in order to avoid too expensive queries.

PSeitz:
Change SegmentCollector.collect to return a `Result`

Validation happens on different phases depending on the aggregation
Term: During segment collection
Histogram: At the end when converting in intermediate buckets (we preallocate empty buckets for the range) Revisit after https://github.com/quickwit-oss/tantivy/issues/1370
Range: When validating the request

Closes #1331

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Unreleased
 - Add [histogram](https://github.com/quickwit-oss/tantivy/pull/1306) aggregation (@PSeitz)
 - Add support for fastfield on text fields (@PSeitz)
 - Add terms aggregation (@PSeitz)
+- API Change: `SegmentCollector.collect` changed to return a `Result`. 
 
 Tantivy 0.17
 ================================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ Unreleased
 - Add [histogram](https://github.com/quickwit-oss/tantivy/pull/1306) aggregation (@PSeitz)
 - Add support for fastfield on text fields (@PSeitz)
 - Add terms aggregation (@PSeitz)
-- API Change: `SegmentCollector.collect` changed to return a `Result`. 
 
 Tantivy 0.17
 ================================
--- a/examples/custom_collector.rs
+++ b/examples/custom_collector.rs
@@ -102,11 +102,12 @@ struct StatsSegmentCollector {
 impl SegmentCollector for StatsSegmentCollector {
     type Fruit = Option<Stats>;
 
-    fn collect(&mut self, doc: u32, _score: Score) {
+    fn collect(&mut self, doc: u32, _score: Score) -> crate::Result<()> {
         let value = self.fast_field_reader.get(doc) as f64;
         self.stats.count += 1;
         self.stats.sum += value;
         self.stats.squared_sum += value * value;
+        Ok(())
     }
 
     fn harvest(self) -> <Self as SegmentCollector>::Fruit {
--- a/examples/custom_collector.rs
+++ b/examples/custom_collector.rs
@@ -102,7 +102,7 @@ struct StatsSegmentCollector {
 impl SegmentCollector for StatsSegmentCollector {
     type Fruit = Option<Stats>;
 
-    fn collect(&mut self, doc: u32, _score: Score) -> crate::Result<()> {
+    fn collect(&mut self, doc: u32, _score: Score) -> tantivy::Result<()> {
         let value = self.fast_field_reader.get(doc) as f64;
         self.stats.count += 1;
         self.stats.sum += value;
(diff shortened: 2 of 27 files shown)


examples/custom_collector.rs
```diff
@@ -102,11 +102,12 @@ struct StatsSegmentCollector {
 impl SegmentCollector for StatsSegmentCollector {
     type Fruit = Option<Stats>;
 
-    fn collect(&mut self, doc: u32, _score: Score) {
+    fn collect(&mut self, doc: u32, _score: Score) -> tantivy::Result<()> {
         let value = self.fast_field_reader.get(doc) as f64;
         self.stats.count += 1;
         self.stats.sum += value;
         self.stats.squared_sum += value * value;
+        Ok(())
```
fulmicoton: Can you run the bench before and after this change to check it does not impact the compiler perf? 
(Top + Count on "the" term query)

The MultiCollector is likely to be the trickiest because of the dynamic dispatch. Unfortunately, it is not part of the benchmark.

fulmicoton (changes requested):
I'm scared of the tantivy API change (Collector::collect returning a Result).

If I understand correctly, the goal is to detect that we have reached the collection limit and abort the query.

An alternative could be to detect the excess of buckets, and keep on the collection but avoid adding extra bucket, and return an error when harvesting the segment.
Would that be unreasonable?

examples/custom_collector.rs
PSeitz: Yep, but in my benches so far adding `Result` didn't add overhead, except for very tight and simple loops.

E.g. here `let value = self.fast_field_reader.get(doc) as f64;` probably considerably outweighs the Result overhead.

PSeitz:
> I'm scared of the tantivy API change (Collector::collect returning a Result).
> 
> If I understand correctly, the goal is to detect that we have reached the collection limit and abort the query.

Scared because of performance?

I checked the performance and the "the" TOP10_COUNT query is indeed slower. (TOP_10 and COUNT are the same speed though) I introduced a change which I considered for some time now, which is to collect hits and pass them as block. I need to add a benchmark for the `MultiCollector`, I think this would profit above average from that. Aggregation may also benefit.

In the current change, the collect_block is quite simple implemented, just fowarding to the underlying function, but I with manual unrolling, we could probably gain some more performance.

| Orig  | With Result | With Result collect_block
| ------------- | ------------- |------------- |
|1,632 μs | 1,847 μs +13.2 % | 1,664 μs +2.3 %
|336,574 docs|336,574 docs|336,574 docs

Currently untested which may help is to pass to `collect_block` always a fixed size array of e.g. 64.

Currently we always pass the score, but having some mechanism to only request docs without score could make make some gains on some scenarios.

> 
> An alternative could be to detect the excess of buckets, and keep on the collection but avoid adding extra bucket, and return an error when harvesting the segment. Would that be unreasonable?

I thought about that, but returning a result in collect is the right approach imo, since we could also want to read from other sources than infallible in-memory sources in the segment collection.

fulmicoton:
> Scared because of performance?

Performance and misuse yes.
I prefer the collect method to not return a Result.

Maybe a bool would do the trick?

COUNT and TOP_10 have a special code path. They are irrelevant here.

PSeitz:
> 
> Performance and misuse yes. I prefer the collect method to not return a Result.

Misuse how? Every tantivy user should know how to handle `Result`.

> 
> Maybe a bool would do the trick?

Like a C-style API, on false collect the error somewhere? It's likely faster, but also pretty ugly.


I added a bench with the MultiCollector (Count + Top10). It is faster with the collect_block approach.
|  Orig  |  With Result collect_block
| ------------- | ------------- |
|2,275 μs +26.1 % | 1,804 μs
|336,574 docs |  336,574 docs | 

`collect_block` has other potential performance gains, e.g. for TopN we could check upfront if the best score from the block would make it in the TopN, if not skip the whole block. That would probably work pretty well for large results, when the TopN is becoming saturated.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Unreleased
 - Add [histogram](https://github.com/quickwit-oss/tantivy/pull/1306) aggregation (@PSeitz)
 - Add support for fastfield on text fields (@PSeitz)
 - Add terms aggregation (@PSeitz)
+- API Change: `SegmentCollector.collect` changed to return a `Result`.
```
fulmicoton:
```suggestion
```

src/aggregation/bucket/range.rs
```diff
@@ -153,7 +154,7 @@ impl SegmentRangeCollector {
     ) -> crate::Result<IntermediateBucketResult> {
         let field_type = self.field_type;
 
-        let buckets = self
+        let buckets: FnvHashMap<SerializedKey, IntermediateRangeBucketEntry> = self
```
fulmicoton, line 157: that's super helpful! thanks

fulmicoton (approved):

PR #1364, opened 2022-05-10
Refactorize PhraseScorer::compute_phrase_match

saroh:
saves a redundant call to `positions` for slop queries.

### DIFF
--- a/src/query/phrase_query/phrase_scorer.rs
+++ b/src/query/phrase_query/phrase_scorer.rs
@@ -252,33 +252,27 @@ impl<TPostings: Postings> PhraseScorer<TPostings> {
                 .positions(&mut self.left);
         }
         let mut intersection_len = self.left.len();
-        let end_term = if self.has_slop() {
-            self.num_terms
-        } else {
-            self.num_terms - 1
-        };
-        for i in 1..end_term {
+        for i in 1..self.num_terms {
             {
                 self.intersection_docset
                     .docset_mut_specialized(i)
                     .positions(&mut self.right);
             }
-            intersection_len = if self.has_slop() {
-                intersection_with_slop(
+            if self.has_slop() {
+                intersection_len = intersection_with_slop(
                     &mut self.left[..intersection_len],
                     &self.right[..],
                     self.slop,
-                )
-            } else {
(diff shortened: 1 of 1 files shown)


src/query/phrase_query/phrase_scorer.rs
```diff
@@ -252,33 +252,27 @@ impl<TPostings: Postings> PhraseScorer<TPostings> {
...
+                intersection_len =
+                    intersection(&mut self.left[..intersection_len], &self.right[..]);
+            }
             if intersection_len == 0 {
                 return 0;
             }
         }
-        self.intersection_docset
-            .docset_mut_specialized(self.num_terms - 1)
-            .positions(&mut self.right);
```
saroh: Or do this if `if !self.has_slop()`

fulmicoton:
I understand the refactoring... The contract is even uglier than it was before.
Do you have an idea on how to refactor this in a way that makes the code more understandable?

Probably we should have less code reuse and have two different set of functions for 
stuff with slop and stuff without slop?

saroh:
I'll add it's probable we forget slop on the final term as the final intersection is ran w/o it, or we do a redundant `intersect` call ? Originaly it just caught my eye we weren't iterating on the same number of terms for slop'd and non slop'd and there was a dangling call at the end in one case. And I tried to understand things from here.

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -236,13 +190,22 @@ impl<TPostings: Postings> PhraseScorer<TPostings> {
...
 
     fn compute_phrase_count(&mut self) -> u32 {
-        let intersection_len = self.compute_phrase_match();
-        intersection_count(&self.left[..intersection_len], &self.right[..]) as u32
+        // there's a possible optim:
+        // compute_phrase_match until the penultimate term
+        // re implement the matching:
+        //   * match self.left with the ultimate term
+        //   * count without storing in left
+        self.compute_phrase_match() as u32
```
saroh: I can add the implems for both optims here, just wanted to know if this looked better now and if this was necessary or not :)

fulmicoton (approved):

fulmicoton:
@saroh Thanks for the hardwork!

PR #1372, opened 2022-05-16
refactor doc store

PSeitz:
refactor cache
expose cache statistics
remove unnecessary arc
unduplicate code

### DIFF
--- a/src/store/compressors.rs
+++ b/src/store/compressors.rs
@@ -68,7 +68,11 @@ impl Compressor {
         }
     }
     #[inline]
-    pub(crate) fn compress(&self, uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
+    pub(crate) fn compress_into(
+        &self,
+        uncompressed: &[u8],
+        compressed: &mut Vec<u8>,
+    ) -> io::Result<()> {
         match self {
             Self::None => {
                 compressed.clear();
@@ -118,8 +122,14 @@ impl Compressor {
         }
     }
 
+    pub(crate) fn decompress(&self, compressed_block: &[u8]) -> io::Result<Vec<u8>> {
+        let mut decompressed_block = Vec::with_capacity(compressed_block.len() * 2);
+        self.decompress_into(compressed_block, &mut decompressed_block)?;
+        Ok(decompressed_block)
+    }
+
     #[inline]
-    pub(crate) fn decompress(
+    pub(crate) fn decompress_into(
         &self,
         compressed: &[u8],
         decompressed: &mut Vec<u8>,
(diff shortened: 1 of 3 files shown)


src/store/reader.rs
```diff
@@ -21,17 +21,60 @@ const LRU_CACHE_CAPACITY: usize = 100;
...
+}
+
+pub struct CacheStats {
+    pub num_entries: usize,
+    pub cache_hits: usize,
+    pub cache_misses: usize,
+}
+
+impl BlockCache {
+    fn get_from_cache(&self, pos: &usize) -> Option<Block> {
```
fulmicoton: `&usize` ?

src/store/reader.rs
```diff
@@ -74,24 +130,30 @@ impl StoreReader {
...
-            return Ok(block.clone());
+        if let Some(block) = self.cache.get_from_cache(checkpoint.byte_range.start) {
+            return Ok(block);
         }
 
-        self.cache_misses.fetch_add(1, Ordering::SeqCst);
-
         let compressed_block = self.compressed_block(checkpoint)?;
+
+        self.decompress_block(compressed_block, checkpoint.byte_range.start)
```
fulmicoton: Can we put the put_cache logic in this function (as opposed to putting it into decompress_block)?

The idea is that the caching all happen in the same function.

fulmicoton (approved):

src/store/reader.rs
```diff
@@ -74,24 +130,30 @@ impl StoreReader {
...
+    }
+
+    fn decompress_block(
```
fulmicoton: see comment above. This applies to the other call site too.

`decompress_block` might be better as function (= not a method) after that change.

Or alternatively, it could be helper method with a blanket impl on the compressor side (maybe without the wrapping into OwnedBytes then).

PSeitz: Moved to compressor

src/store/compressors.rs
```diff
@@ -118,8 +122,14 @@ impl Compressor {
         }
     }
 
+    pub(crate) fn decompress(&self, compressed_block: &[u8]) -> io::Result<Vec<u8>> {
+        let mut decompressed_block = Vec::with_capacity(compressed_block.len() * 2);
```
fulmicoton: with_capacity is not a good idea here.

src/store/reader.rs
```diff
@@ -242,53 +301,34 @@ impl StoreReader {
...
-        self.compressor
-            .decompress(compressed_block.as_slice(), &mut decompressed_block)?;
-
-        let block = OwnedBytes::new(decompressed_block);
-        self.cache
-            .lock()
-            .unwrap()
-            .put(checkpoint.byte_range.start, block.clone());
-
+        let block = self.decompress_block(compressed_block, checkpoint.byte_range.start)?;
```
fulmicoton: i don't see this defined... is a file missing?

fulmicoton (approved):

PR #1373, opened 2022-05-16
Fix deps for unicode regex compiling

saroh:
following #1356 

Fixes `./example/json_field.rs` (which fails with v0.17)
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Syntax(
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regex parse error:
    \\(\+|\^|`|:|\{|\}|"|\[|\]|\(|\)|\~|!|\\|\*|\s)
                                                ^^
error: Unicode-aware Perl class not found (make sure the unicode-perl feature is enabled)
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
)', /home/saroh/.cargo/git/checkouts/tantivy-1e252d10ffb11ab7/3223bdf/query-grammar/src/query_grammar.rs:29:64
``` 
+doc lint

maybe we could add the example somewhere in the doc so they're tested/serve as tests ?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ base64 = "0.13.0"
 byteorder = "1.4.3"
 crc32fast = "1.3.2"
 once_cell = "1.10.0"
-regex ={ version = "1.5.5", default-features = false, features = ["std"] }
+regex = { version = "1.5.5", default-features = false, features = ["std", "unicode"] }
 tantivy-fst = "0.3.0"
 memmap2 = { version = "0.5.3", optional = true }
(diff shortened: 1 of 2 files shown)


fulmicoton (approved):

fulmicoton:
:heart:

PR #1374, opened 2022-05-18
Add Zstd compression support, Make block size configurable via IndexSettings

kryesh:
Made block_size a configurable option in IndexSettings - making this configurable rather than a constant is important because using larger block sizes allows for compression to benefit from using a dictionary for larger chunks of data, resulting in better compression ratios. 

Also added zstd support via the zstd crate - implementation is based on the lz4 one, just adapted to the interface provided by the zstd crate.

Compression ratios for a sample dataset ingesting syslog for an app of mine (highly compressible, ~22GB raw):
loggen -i -n 1000000 --size 2000 --active-connections 12 127.0.0.1 5514

Lz4, 16k blocks (current lz4 config): 744M
Lz4, 512k blocks: 526M
Zstd, 16k blocks: 573M
Zstd, 512k blocks: 379M

### DIFF
--- a/src/store/compression_zstd_block.rs
+++ b/src/store/compression_zstd_block.rs
@@ -9,7 +9,7 @@ pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()>
 
     let count_size = std::mem::size_of::<u64>();
 
-    let max_size: usize = zstd::compress_bound(uncompressed.len()) + count_size;
+    let max_size: usize = zstd::zstd_safe::compress_bound(uncompressed.len()) + count_size;
 
     compressed.resize(max_size, 0);
 
@@ -32,7 +32,7 @@ pub fn decompress(compressed: &[u8], decompressed: &mut Vec<u8>) -> io::Result<(
 
     let count_size = std::mem::size_of::<u64>();
 
-    let uncompressed_size_bytes: &[u8; count_size] = compressed
+    let uncompressed_size_bytes: &[u8; 8] = compressed
         .get(..count_size)
         .ok_or(io::ErrorKind::InvalidData)?
         .try_into()
@@ -40,11 +40,11 @@ pub fn decompress(compressed: &[u8], decompressed: &mut Vec<u8>) -> io::Result<(
 
     let uncompressed_size = u64::from_le_bytes(*uncompressed_size_bytes);
 
-    decompressed.resize(uncompressed_size, 0);
+    decompressed.resize(uncompressed_size as usize, 0);
     let decompressed_size = decompress_to_buffer(&compressed[count_size..], decompressed)
         .map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err.to_string()))?;
 
-    if decompressed_size != uncompressed_size {
+    if decompressed_size != uncompressed_size as usize {
         return Err(io::Error::new(
             io::ErrorKind::InvalidData,
             "doc store block not completely decompressed, data corruption".to_string(),
--- a/src/store/compression_zstd_block.rs
+++ b/src/store/compression_zstd_block.rs
@@ -3,25 +3,25 @@ use std::io::{self, Read, Write};
 use zstd::bulk::{compress_to_buffer, decompress_to_buffer};
 use zstd::DEFAULT_COMPRESSION_LEVEL;
 
-const USIZE_SIZE: usize = std::mem::size_of::<usize>();
-
 #[inline]
 pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
     compressed.clear();
 
-    let max_size: usize = uncompressed.len() + USIZE_SIZE;
+    let count_size = std::mem::size_of::<u64>();
+
+    let max_size: usize = zstd::compress_bound(uncompressed.len()) + count_size;
+
     compressed.resize(max_size, 0);
 
     let compressed_size = compress_to_buffer(
         uncompressed,
-        &mut compressed[USIZE_SIZE..],
+        &mut compressed[count_size..],
         DEFAULT_COMPRESSION_LEVEL,
     )
     .map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err.to_string()))?;
 
-    compressed[0..USIZE_SIZE].copy_from_slice(&uncompressed.len().to_le_bytes());
-
-    compressed.resize(compressed_size + USIZE_SIZE, 0);
+    compressed[0..count_size].copy_from_slice(&(uncompressed.len() as u64).to_le_bytes());
+    compressed.resize(compressed_size + count_size, 0);
 
     Ok(())
 }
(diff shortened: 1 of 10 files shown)


src/store/compression_zstd_block.rs
```diff
@@ -0,0 +1,54 @@
...
+pub fn decompress(compressed: &[u8], decompressed: &mut Vec<u8>) -> io::Result<()> {
+    decompressed.clear();
+
+    let uncompressed_size_bytes: &[u8; USIZE_SIZE] = compressed
+        .get(..USIZE_SIZE)
+        .ok_or(io::ErrorKind::InvalidData)?
+        .try_into()
+        .unwrap();
+
+    let uncompressed_size = usize::from_le_bytes(*uncompressed_size_bytes);
```
fulmicoton:
```suggestion
    let uncompressed_size = u64::from_le_bytes(*uncompressed_size_bytes);
```

src/store/compression_zstd_block.rs
```diff
@@ -0,0 +1,54 @@
+use std::io::{self, Read, Write};
+
+use zstd::bulk::{compress_to_buffer, decompress_to_buffer};
+use zstd::DEFAULT_COMPRESSION_LEVEL;
+
+const USIZE_SIZE: usize = std::mem::size_of::<usize>();
```
fulmicoton: that const is not helpful

src/store/compression_zstd_block.rs
```diff
@@ -0,0 +1,54 @@
...
+    compressed.resize(max_size, 0);
+
+    let compressed_size = compress_to_buffer(
+        uncompressed,
+        &mut compressed[USIZE_SIZE..],
+        DEFAULT_COMPRESSION_LEVEL,
+    )
+    .map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err.to_string()))?;
+
+    compressed[0..USIZE_SIZE].copy_from_slice(&uncompressed.len().to_le_bytes());
```
fulmicoton: usize is platform dependent.
Let's use u64 instead of usize for stuff that is serialized.

src/store/compression_zstd_block.rs
```diff
@@ -0,0 +1,54 @@
...
+use zstd::bulk::{compress_to_buffer, decompress_to_buffer};
+use zstd::DEFAULT_COMPRESSION_LEVEL;
+
+const USIZE_SIZE: usize = std::mem::size_of::<usize>();
+
+#[inline]
+pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
+    compressed.clear();
+
+    let max_size: usize = uncompressed.len() + USIZE_SIZE;
```
fulmicoton: I don't think this is correct.
It is not possible to create a non-destructive compression algorithm that creates only payload <= to their original size.

src/store/compression_zstd_block.rs
```diff
@@ -0,0 +1,54 @@
...
+    let uncompressed_size_bytes: &[u8; USIZE_SIZE] = compressed
+        .get(..USIZE_SIZE)
+        .ok_or(io::ErrorKind::InvalidData)?
+        .try_into()
+        .unwrap();
+
+    let uncompressed_size = usize::from_le_bytes(*uncompressed_size_bytes);
+
+    decompressed.resize(uncompressed_size, 0);
+    decompressed.resize(decompressed.capacity(), 0);
```
fulmicoton:
```suggestion
```

fulmicoton (changes requested):

fulmicoton (changes requested):
See comments inline

src/store/compression_zstd_block.rs
kryesh: Good catch - fixed with:
https://github.com/kryesh/tantivy/commit/d4e5b484378f80599d7e22cc3dc461ed72f2ca7d#diff-a1e070670997e81b916410e76f4f4bd6f663da072ad801314b080abb52d88289R12

src/store/compression_zstd_block.rs
kryesh: Removed the duplicate line

src/store/compression_zstd_block.rs
```diff
@@ -0,0 +1,55 @@
+use std::io::{self, Read, Write};
+
+use zstd::bulk::{compress_to_buffer, decompress_to_buffer};
+use zstd::DEFAULT_COMPRESSION_LEVEL;
+
+#[inline]
+pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
+    compressed.clear();
+
+    let count_size = std::mem::size_of::<u64>();
```
PSeitz: 4GB (u32) should be enough, same limitation as in lz4

src/store/compression_zstd_block.rs
```diff
@@ -20,31 +18,30 @@ pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()>
...
-        .try_into()
-        .unwrap();
-
-    let uncompressed_size = u64::from_le_bytes(*uncompressed_size_bytes);
+    decompressed.clear();
+    decompressed.resize(uncompressed_size, 0);
 
-    decompressed.resize(uncompressed_size as usize, 0);
     let decompressed_size = decompress_to_buffer(&compressed[count_size..], decompressed)
         .map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err.to_string()))?;
```
fulmicoton: It is already an io Result. No need to convert it.

PSeitz:
Can you add the feature to the tests in .github/workflows/test.yml?

src/store/mod.rs
```diff
@@ -139,11 +145,11 @@ pub mod tests {
         Ok(())
     }
 
-    fn test_store(compressor: Compressor) -> crate::Result<()> {
+    fn test_store(compressor: Compressor, blocksize: usize) -> crate::Result<()> {
```
fulmicoton, line 148: Can we add a test in our test_store suite that test for random payloads?

fulmicoton, line 148: Can we add a test in our test_store suite that test for random (high entropy) payloads?

We spotted a bug in the original form of this PR.

src/store/writer.rs
```diff
@@ -65,7 +65,7 @@ impl StoreWriter {
         VInt(doc_num_bytes as u64).serialize(&mut self.current_block)?;
         self.current_block.write_all(serialized_document)?;
         self.doc += 1;
-        if self.current_block.len() > BLOCK_SIZE {
```
fulmicoton, line 68: that's true for the other stores, and it does not matter much, but shouldn't we make that `>=`?.

This is almost philosophy at this point, but my mental model is 
"We close a block once it is full". "Being full means being greater or equal to BLOCK_SIZE"

fulmicoton, line 68: that's true for the other stores, and it does not matter much, but shouldn't we make that `>=`?.

This is almost philosophy at this point, but my mental model is 
"We close a block once it is full". "Being full means being greater or equal to BLOCK_SIZE"

fulmicoton:
@kryesh this is a great contribution thank you!

fulmicoton:
@kryesh this is a great contribution thank you!

src/store/mod.rs
kryesh: I can look into it later today - but just wanted to confirm, would the test be for the case where the input is incompressible to the point where the compressed output is larger than the input?
Just making sure the test is for the right thing

src/store/writer.rs
kryesh: I just went with what was already there, personally I don't think it makes much difference since blocks are allowed to overflow anyway;
Happy to go either way on this one

kryesh:
> @kryesh this is a great contribution thank you!

Thanks!
I stumbled across the block size thing while trying to figure out why an index that was compressed was getting compressed further when stored on a zfs filesystem with compression enabled, as for zstd I figured that while I was looking at it I might as well add it.
Hopefully it proves useful!

src/store/mod.rs
fulmicoton: yes

fulmicoton (approved):

fulmicoton:
Should we use a common dictionary for all of the blocks?

PR #1378, opened 2022-05-30
enable setting compression level

PSeitz:


src/store/writer.rs
```diff
@@ -34,10 +35,16 @@ impl StoreWriter {
     ///
     /// The store writer will writes blocks on disc as
     /// document are added.
-    pub fn new(writer: WritePtr, compressor: Compressor, block_size: usize) -> StoreWriter {
+    pub fn new(
+        writer: WritePtr,
+        compressor: Compressor,
```
fulmicoton: why isn't that part of the compressor?

fulmicoton (changes requested):
Why isn't compression part of the compressor?

If we have parameters that are compressor specific, it seems like an antipattern to have the union of these parameters in a list of options.

src/store/writer.rs
PSeitz: Compressor is currently constructed by deserializing from the compressor name. To pass more parameters we would need much more complicated deserialization.
And there are cases where the Compressor is only used for decompression.

PSeitz:
> Why isn't compression part of the compressor?

We have decompress only compressor
```
    /// Opens a store reader
    pub fn open(store_file: FileSlice) -> io::Result<StoreReader> {
        let (footer, data_and_offset) = DocStoreFooter::extract_footer(store_file)?;

        let (data_file, offset_index_file) = data_and_offset.split(footer.offset as usize);
        let index_data = offset_index_file.read_bytes()?;
        let space_usage = StoreSpaceUsage::new(data_file.len(), offset_index_file.len());
        let skip_index = SkipIndex::open(index_data);
        Ok(StoreReader {
            compressor: footer.compressor,
            data: data_file,
            cache: Arc::new(Mutex::new(LruCache::new(LRU_CACHE_CAPACITY))),
            cache_hits: Default::default(),
            cache_misses: Default::default(),
            skip_index: Arc::new(skip_index),
            space_usage,
        })
    }
```

> 
> If we have parameters that are compressor specific, it seems like an antipattern to have the union of these parameters in a list of options.

We could have `zstd_compression_level` or similar instead of`docstore_compression_level`, but I think we can use that parameter for multiple compressors. It's usually just a number.

fulmicoton:
@PSeitz Can you investigate if a refactoring to make that nicer?

So the compressor `enum` would contain these params.
It then needs to be serialized/deserialized in the meta.json, as well as in the footer.

The footer part is a bit lame, because the level is not necessary, but that's fine.

src/store/compression_zstd_block.rs
```diff
@@ -4,7 +4,11 @@ use zstd::bulk::{compress_to_buffer, decompress_to_buffer};
 use zstd::DEFAULT_COMPRESSION_LEVEL;
 
 #[inline]
-pub fn compress(uncompressed: &[u8], compressed: &mut Vec<u8>) -> io::Result<()> {
+pub fn compress(
+    uncompressed: &[u8],
+    compressed: &mut Vec<u8>,
+    compression_level: Option<i32>,
```
fulmicoton, line 10: can it be negative?

fulmicoton (approved):

src/store/compression_zstd_block.rs
PSeitz: Yes, for more speed

PR #1381, opened 2022-06-02
Add a lenient query parser

zoonage:
* #5 
* Question: Is there some way of consolidating the special chars with no spaces and one without?
* If a query cannot be passed all the special characters are removed from the query string. If that fails to be compiled into an AST then an empty query (which should return no results) will be returned.
* Add `unicode-perl` as a feature flag to `regex` due to
  ```
  ---- query_grammar::test::test_field_name_re stdout ----
  thread 'query_grammar::test::test_field_name_re' panicked at 'called `Result::unwrap()` on an `Err` value: Syntax(
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  regex parse error:
      \\(\+|\^|`|:|\{|\}|"|\[|\]|\(|\)|\~|!|\\|\*|\s)
                                                  ^^
  error: Unicode-aware Perl class not found (make sure the unicode-perl feature is enabled)
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  )', query-grammar/src/query_grammar.rs:573:82
  ```

### DIFF
--- a/query-grammar/src/lib.rs
+++ b/query-grammar/src/lib.rs
@@ -5,6 +5,7 @@ use combine::parser::Parser;
 
 pub use crate::occur::Occur;
 use crate::query_grammar::parse_to_ast;
+pub use crate::query_grammar::SPECIAL_CHARS_NO_SPACE;
 pub use crate::user_input_ast::{UserInputAst, UserInputBound, UserInputLeaf, UserInputLiteral};
 
 pub struct Error;
--- a/query-grammar/Cargo.toml
+++ b/query-grammar/Cargo.toml
@@ -14,4 +14,4 @@ edition = "2018"
 [dependencies]
 combine = {version="4", default-features=false, features=[] }
 once_cell = "1.7.2"
-regex ={ version = "1.5.4", default-features = false, features = ["std"] }
+regex ={ version = "1.5.4", default-features = false, features = ["std", "unicode-perl"] }
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -15,6 +15,9 @@ use crate::Occur;
 
 // Note: '-' char is only forbidden at the beginning of a field name, would be clearer to add it to
 // special characters.
+pub const SPECIAL_CHARS_NO_SPACE: &[char] = &[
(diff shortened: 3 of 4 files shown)


query-grammar/src/lib.rs
```diff
@@ -4,12 +4,55 @@ mod user_input_ast;
...
+/// characters and evaluating the query again.
+fn parse_query_leniently(query: &str) -> (UserInputAst, Option<Error>) {
+    match parse_query(query) {
+        Ok(uia) => (uia, None),
+        Err(e1) => match parse_query(&remove_special_chars_from_query(query)) {
+            Ok(uia) => (uia, Some(e1)),
+            // We should never reach this branch as the above query should always be valid
+            // with no special characters.
+            // TOOD: Is there a better way of handling this? Perhaps `query("")`?
+            Err(e2) => unreachable!(),
```
fulmicoton: How do will we prove or at least the right level of confidence that this is true (This is not true with the current PR)? Fuzzing?

zoonage:
Just realised that this PR is in the wrong bit of the code, moving it into the `QueryParser` bit

query-grammar/src/lib.rs
zoonage: I've changed it from an `unimplemented!()` to now use `""` as a query which _should_ return no results from my understanding. I was a bit stuck between defaulting to `""` or `"*"`, but I think `""` might be less misleading to and end user as an invalid query -> no results makes more sense then returning all results

lukethacoder:
Is this waiting for changes? Happy to help if I can to get this across the line

zoonage:
> Is this waiting for changes? Happy to help if I can to get this across the line

@lukethacoder feel free to take it from here, I've not got time to work on this unfortunately

fmassot:
@zoonage the issue #5 was reassigned to @trinity-1686a who is currently working on it.

PR #1382, opened 2022-06-02
Add boolean fields

boraarslan:
The first draft to add boolean field support. There is a very high chance I missed some stuff.

I will update the docs after the review (in case of something needs to change)

### DIFF
--- a/src/fastfield/multivalued/mod.rs
+++ b/src/fastfield/multivalued/mod.rs
@@ -239,7 +239,8 @@ mod tests {
         index_writer.add_document(doc!(bool_field=> true, bool_field => false))?;
         index_writer.add_document(doc!())?;
         index_writer.add_document(doc!(bool_field=> false))?;
-        index_writer.add_document(doc!(bool_field=> true, bool_field => true, bool_field => false))?;
+        index_writer
+            .add_document(doc!(bool_field=> true, bool_field => true, bool_field => false))?;
         index_writer.commit()?;
 
         let searcher = index.reader()?.searcher();
--- a/src/fastfield/multivalued/mod.rs
+++ b/src/fastfield/multivalued/mod.rs
@@ -229,23 +229,23 @@ mod tests {
     #[test]
     fn test_multivalued_bool() -> crate::Result<()> {
         let mut schema_builder = Schema::builder();
-        let field = schema_builder.add_bool_field(
+        let bool_field = schema_builder.add_bool_field(
             "multifield",
             NumericOptions::default().set_fast(Cardinality::MultiValues),
         );
         let schema = schema_builder.build();
         let index = Index::create_in_ram(schema);
         let mut index_writer = index.writer_for_tests()?;
-        index_writer.add_document(doc!(field=> true, field => false))?;
+        index_writer.add_document(doc!(bool_field=> true, bool_field => false))?;
         index_writer.add_document(doc!())?;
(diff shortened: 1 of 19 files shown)


src/schema/field_type.rs
```diff
@@ -46,6 +46,8 @@ pub enum Type {
     I64 = b'i',
     /// `f64`
     F64 = b'f',
+    /// `bool`
+    Bool = b'o',
```
PSeitz, line 50: you can use 'b' for bool

PSeitz:
Nice job. Checkout `test_operation_strategy` and extend it to cover boolean fields. It covers different scenarios with randomized input.

src/schema/field_type.rs
boraarslan: 'b' is reserved for `Bytes`

src/indexer/index_writer.rs
```diff
@@ -1522,11 +1532,18 @@ mod tests {
         // multivalue fast field tests
         for segment_reader in searcher.segment_readers().iter() {
             let ff_reader = segment_reader.fast_fields().u64s(multi_numbers).unwrap();
+            // let bool_ff_reader = segment_reader.fast_fields().bools(multi_bools).unwrap();
```
boraarslan: This line was panicking on every test but only in some Inputs with a panic message:
```rust
'called `Result::unwrap()` on an `Err` value: SchemaError("Field(multi_bools) data was not found")'
```
I think `MultiValue` bools work fine and the issue is with the `MultiValuedFastFieldReader` since the other tests don't fail. I will remove the comments when this is fixed.

fulmicoton:
@PSeitz @boraarslan Should we merge?

@PSeitz Do we have a mechanism to use the field type to change the fastfield codec logic? It would be nice to force a bitset codec actually... Maybe even drop the DynamicReader for booleans.

PSeitz:
> 
> @PSeitz Do we have a mechanism to use the field type to change the fastfield codec logic? It would be nice to force a bitset codec actually... Maybe even drop the DynamicReader for booleans.

Calling`new_u64_fast_field` instead `create_auto_detect_u64_fast_field_with_idx` always uses Bitpacked encoding. (I think we need to rename that)
But if we have large blocks of true or false, the codec estimator could offer better compression.

fulmicoton:
@boraarslan @PSeitz Any update?

PSeitz:
LGTM, setting the codec can imo be done in another PR

boraarslan:
I agree with @PSeitz but If codec change is a must have I can look into that in this PR.

src/fastfield/multivalued/mod.rs
```diff
@@ -226,6 +226,37 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_multivalued_bool() -> crate::Result<()> {
+        let mut schema_builder = Schema::builder();
+        let field = schema_builder.add_bool_field(
```
PSeitz:
```suggestion
        let bool_field = schema_builder.add_bool_field(
```

PSeitz:
```suggestion
        let bool_field = schema_builder.add_bool_field(
```

fulmicoton:
@PSeitz Can you have approve & merge the PR?

PSeitz (approved):

PR #1388, opened 2022-06-15
Extend FAQ

PSeitz:


PSeitz:
Not sure if this is the best place for this.

README.md
```diff
@@ -152,4 +152,13 @@ You can also find other bindings on [GitHub](https://github.com/search?q=tantivy
...
+- According to our [search latency benchmark](https://tantivy-search.github.io/bench/), Tantivy is approximately 2x faster than Lucene.
+
+### Does tantivy support incremental indexing?
+- Yes.
+
+### How can I edit documents?
+- Data in tantivy is immutable. To edit a document, the document needs to be deleted and reindexed.
+
+### When will my documents be searchable during indexing?
+- Documents will be searchable after a `commit`. A commit can be executed manually, or when reaching the memory budget during indexing. Comitting means the segment will be flushed to disk.
```
fulmicoton: That's incorrect. 

A commit can only be done manually.

Segment getting serialized to disk can happen when we reach the memory budget, but the data will not be made searchable until a manual commit is done.

It might be important to also add that existing reader's need to be reload for changes to be visible.

fulmicoton (changes requested):
see comment

README.md
maximkpa: Documents will be searchable after a `commit`, which can only be done manually. When reaching the memory budget during indexing, a segment can get serialized to disk, but the data will not be searchable until a commit is done (not necessary to repeat "manual commit" here imo). It is important to reload the existing reader for changes to be visible.

README.md
```diff
@@ -161,4 +161,4 @@ You can also find other bindings on [GitHub](https://github.com/search?q=tantivy
 - Data in tantivy is immutable. To edit a document, the document needs to be deleted and reindexed.
 
 ### When will my documents be searchable during indexing?
-- Documents will be searchable after a `commit` is called on `IndexWriter`. Comitting also causes the current segment to be serialized to disk. When reaching the memory budget during indexing, a segment also gets serialized to disk, but the data will not be searchable until `commit` is called. It is important to `reload` existing `IndexReader` for changes to be visible.
+- Documents will be searchable after a `commit` is called on an `IndexWriter`. Committing also causes the current segment to be serialized to a disk. When reaching the memory budget during indexing, a segment also gets serialized to the disk, but the data will not be searchable until a `commit` is called. It is important to `reload` the existing `IndexReader` for changes to be visible.
```
fulmicoton:
```suggestion
- Documents will be searchable after a `commit` is called on an `IndexWriter`. Existing `IndexReader`s will also need to be reloaded in order to reflect the changes. Finally, changes are only visible to newly acquired `Searcher`.
```

fulmicoton (approved):

PR #1389, opened 2022-06-16
use separate thread to compress block store

PSeitz:
Use seperate thread to compress block store for increased indexing performance. This allows to use slower compressors with higher compression ratio, with less or no perfomance impact (with enough cores).

A seperate thread is spawned to compress the docstore, which handles single blocks and stacking from other docstores.
The spawned compressor thread does not write, instead it sends back the compressed data. This is done in order to avoid writing multithreaded on the same file.

Small benchmark 1GB hdfs, zstd level 8

```
Pre
 Total Nowait Merge: 43.30 Mb/s
 Total Wait Merge: 43.28 Mb/s

Post
Total Nowait Merge: 67.69 Mb/s
Total Wait Merge: 67.69 Mb/s
```

### DIFF
--- a/src/store/writer.rs
+++ b/src/store/writer.rs
@@ -126,7 +126,7 @@ impl StoreWriter {
         // intermediary_buffer due to the borrow checker
         // a new buffer costs ~1% indexing performance
         let doc_num_bytes = self.intermediary_buffer.len();
-        VInt(doc_num_bytes as u64).serialize(&mut self.current_block)?;
+        VInt(doc_num_bytes as u64).serialize_into_vec(&mut self.current_block);
         self.current_block
             .write_all(&self.intermediary_buffer[..])?;
         self.num_docs_in_current_block += 1;
--- a/src/store/writer.rs
+++ b/src/store/writer.rs
@@ -92,6 +92,7 @@ impl StoreWriter {
         self.intermediary_buffer.capacity() + self.current_block.capacity()
     }
 
+    /// Checks if the current block is full, and if so, compresses and flushes it.
     fn check_flush_block(&mut self) -> io::Result<()> {
         if self.current_block.len() > self.block_size {
             let block = self.get_current_block();
--- a/src/store/writer.rs
+++ b/src/store/writer.rs
@@ -1,9 +1,8 @@
 use std::io::{self, Write};
-use std::sync::mpsc::{sync_channel, Receiver, SyncSender, TryRecvError};
+use std::sync::mpsc::{sync_channel, Receiver, SyncSender};
 use std::thread::{self, JoinHandle};
 
-use common::{BinarySerializable, VInt};
-use ownedbytes::OwnedBytes;
+use common::{BinarySerializable, CountingWriter, VInt};
 
 use super::compressors::Compressor;
 use super::footer::DocStoreFooter;
@@ -28,14 +27,11 @@ pub struct StoreWriter {
     intermediary_buffer: Vec<u8>,
     current_block: Vec<u8>,
 
-    writer: WritePtr,
-    // the channel to communicate with the compressor thread.
+    // the channel to send data to the compressor thread.
     compressor_sender: SyncSender<BlockCompressorMessage>,
-    // the channel to receive data to write from the compressor thread.
-    data_receiver: Receiver<OwnedBytes>,
 
     // the handle to check for errors on the thread
-    compressor_thread_handle: JoinHandle<io::Result<(DocStoreFooter, SkipIndexBuilder)>>,
+    compressor_thread_handle: JoinHandle<io::Result<()>>,
 }
 
 enum BlockCompressorMessage {
@@ -55,16 +51,13 @@ impl StoreWriter {
     ) -> io::Result<StoreWriter> {
         let thread_builder = thread::Builder::new().name("docstore compressor thread".to_string());
 
-        // Data channel to send fs writes, to write only from current thread
-        let (data_sender, data_receiver): (SyncSender<OwnedBytes>, Receiver<OwnedBytes>) =
-            sync_channel(3);
         // Channel to send uncompressed data to compressor channel
         let (block_sender, block_receiver): (
             SyncSender<BlockCompressorMessage>,
             Receiver<BlockCompressorMessage>,
         ) = sync_channel(3);
         let thread_join_handle = thread_builder.spawn(move || {
-            let mut block_compressor = BlockCompressor::new(compressor, data_sender);
+            let mut block_compressor = BlockCompressor::new(compressor, writer);
             while let Ok(packet) = block_receiver.recv() {
                 match packet {
                     BlockCompressorMessage::AddBlock(block) => {
@@ -84,10 +77,8 @@ impl StoreWriter {
             num_docs_in_current_block: 0,
             intermediary_buffer: Vec::new(),
             current_block: Vec::new(),
-            writer,
             compressor_sender: block_sender,
             compressor_thread_handle: thread_join_handle,
-            data_receiver,
         })
     }
 
@@ -106,32 +97,7 @@ impl StoreWriter {
             self.compressor_sender
                 .send(BlockCompressorMessage::AddBlock(block))
                 .map_err(|err| io::Error::new(io::ErrorKind::Other, err))?;
-            self.fetch_writes_from_channel()?;
         }
-        Ok(())
-    }
-
-    /// Try to empty the queue to write into the file.
-    ///
-    /// This is done in order to avoid writing from multiple threads into the file.
-    fn fetch_writes_from_channel(&mut self) -> io::Result<()> {
-        loop {
-            match self.data_receiver.try_recv() {
-                Ok(data) => {
-                    self.writer.write_all(data.as_slice())?;
-                }
-                Err(TryRecvError::Empty) => {
-                    break;
-                }
-                Err(TryRecvError::Disconnected) => {
-                    return Err(io::Error::new(
-                        io::ErrorKind::Other,
-                        "compressor data channel unexpected closed".to_string(),
-                    ));
-                }
-            }
-        }
-
         Ok(())
     }
 
@@ -204,21 +170,11 @@ impl StoreWriter {
         }
         drop(self.compressor_sender);
 
-        // Wait for remaining data on the channel to write
-        while let Ok(data) = self.data_receiver.recv() {
-            self.writer.write_all(data.as_slice())?;
-        }
-
-        // The compressor thread should have finished already, since data_receiver stopped
-        // receiving
-        let (docstore_footer, offset_index_writer) = self
-            .compressor_thread_handle
+        self.compressor_thread_handle
             .join()
             .map_err(|err| io::Error::new(io::ErrorKind::Other, format!("{:?}", err)))??;
 
-        offset_index_writer.serialize_into(&mut self.writer)?;
-        docstore_footer.serialize(&mut self.writer)?;
-        self.writer.terminate()
+        Ok(())
     }
 }
 
@@ -228,8 +184,7 @@ pub struct BlockCompressor {
     doc: DocId,
     offset_index_writer: SkipIndexBuilder,
     intermediary_buffer: Vec<u8>,
-    written_bytes: usize,
-    data_sender: SyncSender<OwnedBytes>,
+    writer: CountingWriter<WritePtr>,
 }
 
 struct DocumentBlock {
@@ -244,14 +199,13 @@ impl DocumentBlock {
 }
 
 impl BlockCompressor {
-    fn new(compressor: Compressor, data_sender: SyncSender<OwnedBytes>) -> Self {
+    fn new(compressor: Compressor, writer: WritePtr) -> Self {
         Self {
             compressor,
             doc: 0,
             offset_index_writer: SkipIndexBuilder::new(),
             intermediary_buffer: Vec::new(),
-            written_bytes: 0,
-            data_sender,
+            writer: CountingWriter::wrap(writer),
         }
     }
 
@@ -261,17 +215,13 @@ impl BlockCompressor {
         self.compressor
             .compress_into(&block.data[..], &mut self.intermediary_buffer)?;
 
-        let byte_range = self.written_bytes..self.written_bytes + self.intermediary_buffer.len();
-
-        self.data_sender
-            .send(OwnedBytes::new(self.intermediary_buffer.to_owned()))
-            .map_err(|err| io::Error::new(io::ErrorKind::Other, err))?;
-
-        self.written_bytes += byte_range.len();
+        let start_offset = self.writer.written_bytes() as usize;
+        self.writer.write_all(&self.intermediary_buffer)?;
+        let end_offset = self.writer.written_bytes() as usize;
 
         self.register_checkpoint(Checkpoint {
             doc_range: self.doc..self.doc + block.num_docs_in_block,
-            byte_range,
+            byte_range: start_offset..end_offset,
         });
         Ok(())
     }
@@ -290,14 +240,11 @@ impl BlockCompressor {
             self.compress_block(block)?;
         }
         let doc_shift = self.doc;
-        let start_shift = self.written_bytes;
+        let start_shift = self.writer.written_bytes() as usize;
 
         // just bulk write all of the block of the given reader.
-        self.data_sender
-            .send(store_reader.block_data()?)
-            .map_err(|err| io::Error::new(io::ErrorKind::Other, err))?;
-
-        self.written_bytes += store_reader.block_data()?.as_slice().len();
+        self.writer
+            .write_all(store_reader.block_data()?.as_slice())?;
 
         // concatenate the index of the `store_reader`, after translating
         // its start doc id and its start file offset.
@@ -310,12 +257,13 @@ impl BlockCompressor {
         }
         Ok(())
     }
-    fn close(self) -> io::Result<(DocStoreFooter, SkipIndexBuilder)> {
-        drop(self.data_sender);
+    fn close(mut self) -> io::Result<()> {
+        let header_offset: u64 = self.writer.written_bytes() as u64;
+        let docstore_footer =
+            DocStoreFooter::new(header_offset, Decompressor::from(self.compressor));
 
-        let header_offset: u64 = self.written_bytes as u64;
-        let footer = DocStoreFooter::new(header_offset, Decompressor::from(self.compressor));
-
-        Ok((footer, self.offset_index_writer))
+        self.offset_index_writer.serialize_into(&mut self.writer)?;
+        docstore_footer.serialize(&mut self.writer)?;
+        self.writer.terminate()
     }
 }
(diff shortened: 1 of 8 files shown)


src/store/writer.rs
```diff
@@ -56,21 +96,53 @@ impl StoreWriter {
...
+    /// Try to empty the queue to write into the file.
+    ///
+    /// This is done in order to avoid writing from multiple threads into the file.
+    fn fetch_writes_from_channel(&mut self) -> io::Result<()> {
+        loop {
+            match self.data_receiver.try_recv() {
+                Ok(data) => {
+                    self.writer.write_all(data.as_slice())?;
+                }
+                Err(err) => match err {
```
fulmicoton: Can you merge the match? 
```
match x {
   Ok(_) => 
   Err(TryRecvError::Empty) => 
   Err(TryRecvError::Disconnecto) =>
}

File: src/store/writer.rs [hunk 2]
```diff
@@ -21,30 +24,67 @@ use crate::DocId;
...
-    pub fn new(writer: WritePtr, compressor: Compressor, block_size: usize) -> StoreWriter {
-        StoreWriter {
+    pub fn new(
+        writer: WritePtr,
+        compressor: Compressor,
+        block_size: usize,
+    ) -> io::Result<StoreWriter> {
+        let thread_builder = thread::Builder::new().name("docstore compressor thread".to_string());
+
+        // Data channel to send fs writes, to write only from current thread
```
fulmicoton: interesting... why do we want to write only from current thread?

src/store/writer.rs
```diff
@@ -21,30 +24,67 @@ use crate::DocId;
...
+        let (data_sender, data_receiver) = sync_channel(3);
```
fulmicoton: can you put the type of the object?

src/store/writer.rs
```diff
@@ -21,30 +24,67 @@ use crate::DocId;
...
+        // Channel to send uncompressed data to compressor channel
+        let (block_sender, block_receiver) = sync_channel(3);
```
fulmicoton: can you put the type of the object?

src/store/writer.rs
PSeitz: Even though in tantivy we create a separate file, which is fine for a separate thread to write into, the `Directory` trait itself doesn't require explicitly thread safe writes. `TerminatingWrite` is not Send, so the current contract is that the writers stay on the same thread.

It's unlikely to be an issue, but it could be two threads write next to each other at the same page or cache line. So depending on the write target and buffer synchronization between the threads, it could be that they overwrite each others data. Since rust doesn't cover race conditions outside it's memory model, e.g. Files, I'm extra careful there.

src/store/writer.rs
PSeitz: yep

src/store/writer.rs
PSeitz: done

src/store/writer.rs
fulmicoton: That makes sense. Can we make `TerminatingWrite: Send` and simplify the code though?

fulmicoton: Ah so you just offload compression to a different thread. Writing is still done in the same place. Maybe it is clearer that way let me keep on reading on.

src/store/writer.rs
```diff
@@ -85,28 +159,145 @@ impl StoreWriter {
...
+            .join()
+            .map_err(|err| io::Error::new(io::ErrorKind::Other, format!("{:?}", err)))??;
+
+        offset_index_writer.serialize_into(&mut self.writer)?;
+        docstore_footer.serialize(&mut self.writer)?;
+        self.writer.terminate()
+    }
+}
+
+/// BlockCompressor is seperated from StoreWriter, to be run in an own thread
```
fulmicoton:
```suggestion
/// BlockCompressor is separated from StoreWriter, to be run in an own thread
```

fulmicoton:
A bit of analyis...

We write the docstore
1- when building a new segment
2a- when merging N segments. If we do no have any deletes in the segment being appended  we have an operation called stacking that makes it possible to more or less concatenate teh docstore (it is a tiny bit more complicate that but what ever)
2b- If we have deletes, then we need to rebuild the docstore. the operation is then close to what happens in 1.

Threading can help with 2 things
A docstore compression is heavy and does not need to happen on the same thread as inverted index building. We love offloading stuff like this because it increases the indexing throughput without producing smaller segments.
B we do not need to have the CPU wait on IO.


The current approach implementation will not help with B, because the IO still happens in the same thread as the original.
2a is all about B. This is probably more important when using the SSTable implementation.
1 and 2b is probably more about A. (we still io wait but more time is spent on CPU than on IO when building a segment. The gain is probably not negligible though).

2a and 2b are not as important because they do starve us on resource, and they do not impact time to search. (Finishing a merge quickly by using more cores is not very important). They could help resource management on quickwit however. (If all task take clearly one full thread, we can more easily rely on our own scheduling and size our task thread pool by the number of cores).

I think we want to at least do the IO on the thread that does the compression.
(That means moving the File into the thread).

fulmicoton:
Please test on a dataset that has trigger merges (wikipedia is fine), and rely on the sstable dict.

fulmicoton (changes requested):
See comments on the Conversation tab.
https://github.com/quickwit-oss/tantivy/pull/1389/

PSeitz:
I tested merge operation on hdfs 14GB(2,4GB index size, 4 segments) and wikipedia 8GB (5,7GB index size, 5 segment) with sstable. It seems to be marginal faster. Merge throughput is ~50MB/s hdfs and 75MB/s wikipedia (measured on index size, not input size), which is slower than most disks. On indexing, no noticeable speed was observed. I think the impact is too small to change the API (adding Send to TerminatingWrite). Although another upside would be that the code would be simpler.


```
➜  tantivy-cli git:(main) ✗ du -sh hdfs/
2,4G    hdfs/

➜  tantivy-cli git:(main) ✗ du -sh wikipedia
5,7G    wikipedia
```

| Run  		| Write on seperate thread | Write on one thread |
| ------------- 		| ------------- | -------------	 	|
| hdfs  1. run			| 50.43 secs  	| 52.63 secs  		|
| hdfs  2. run			| 48.38 secs  	| 50.57 secs 		|
| wikipedia  1. run		| 71.74 secs  	| 77.27 secs  		|
| wikipedia  2. run		| 79.46 secs	| 77.25 secs 		|
| wikipedia  3. run		| 81.64 secs	|				|

fulmicoton:
@PSeitz Thanks for investigating! That makes sense. We don't flush or anything, so the write are just pushing the data to OS buffer, and the actual write to disk will be done asynchronously by the OS, provided our throughput does not beat the hardware.
The docstore actually write a lot of data but after compression it is fine.
I suspect you would see different results with a lesser hard drive like EBS (gp2 is 250 MiB/s - to be split between writing a split and merging at the same time etc.).

Anyway, can you move the write to a different thread, if only to simplify the code?

src/store/writer.rs
```diff
@@ -21,30 +23,63 @@ use crate::DocId;
...
     /// The store writer will writes blocks on disc as
     /// document are added.
-    pub fn new(writer: WritePtr, compressor: Compressor, block_size: usize) -> StoreWriter {
-        StoreWriter {
+    pub fn new(
+        writer: WritePtr,
+        compressor: Compressor,
+        block_size: usize,
+    ) -> io::Result<StoreWriter> {
+        let thread_builder = thread::Builder::new().name("docstore compressor thread".to_string());
```
fulmicoton:
```suggestion
        let thread_builder = thread::Builder::new().name("docstore-compressor-thread".to_string());
```

src/store/writer.rs
```diff
@@ -21,30 +23,63 @@ use crate::DocId;
...
+                match packet {
+                    BlockCompressorMessage::AddBlock(block) => {
+                        block_compressor.compress_block(block)?;
+                    }
+                    BlockCompressorMessage::Stack((store_reader, block)) => {
+                        block_compressor.stack(block, store_reader)?;
+                    }
+                }
+            }
+            block_compressor.close()
```
fulmicoton:
```suggestion
            block_compressor.close()?;
            Ok(())
```

fulmicoton: nitpick.. I think the code is actually easier to read like this.

src/store/writer.rs
```diff
@@ -56,21 +91,26 @@ impl StoreWriter {
...
-            self.write_and_compress_block()?;
+            let block = self.get_current_block();
+            self.compressor_sender
+                .send(BlockCompressorMessage::AddBlock(block))
+                .map_err(|err| io::Error::new(io::ErrorKind::Other, err))?;
         }
         Ok(())
     }
 
+    fn get_current_block(&mut self) -> DocumentBlock {
```
fulmicoton: Rename required.
That method is named like an accesssor but it does way more than this...

src/store/writer.rs
```diff
@@ -21,30 +23,63 @@ use crate::DocId;
 pub struct StoreWriter {
     compressor: Compressor,
     block_size: usize,
-    doc: DocId,
-    first_doc_in_block: DocId,
-    offset_index_writer: SkipIndexBuilder,
-    writer: CountingWriter<WritePtr>,
+    num_docs_in_current_block: DocId,
```
fulmicoton, line 26: good rename! Thanks

src/store/writer.rs
```diff
@@ -85,76 +125,145 @@ impl StoreWriter {
...
+}
 
-    fn register_checkpoint(&mut self, checkpoint: Checkpoint) {
-        self.offset_index_writer.insert(checkpoint.clone());
-        self.first_doc_in_block = checkpoint.doc_range.end;
-        self.doc = checkpoint.doc_range.end;
+/// BlockCompressor is separated from StoreWriter, to be run in an own thread
+pub struct BlockCompressor {
+    compressor: Compressor,
+    doc: DocId,
```
fulmicoton: The previous name `first_doc_in_block` vs `doc` was actually clearer.
Why the rename? 

If it is motivated can you add a comment at the very least?

src/store/writer.rs
```diff
@@ -56,21 +92,26 @@ impl StoreWriter {
...
-    /// Store bytes of a serialized document.
-    ///
-    /// The document id is implicitely the current number
-    /// of documents.
-    pub fn store_bytes(&mut self, serialized_document: &[u8]) -> io::Result<()> {
-        let doc_num_bytes = serialized_document.len();
-        VInt(doc_num_bytes as u64).serialize(&mut self.current_block)?;
-        self.current_block.write_all(serialized_document)?;
-        self.doc += 1;
+    fn check_flush_block(&mut self) -> io::Result<()> {
```
fulmicoton, line 95: nice refactor

fulmicoton, line 95:
```suggestion
    /// Checks if the current block is full, and if so, compresses and flushes it.
    fn check_flush_block(&mut self) -> io::Result<()> {
```

src/store/writer.rs
```diff
@@ -21,30 +23,64 @@ use crate::DocId;
...
+    // the channel to send data to the compressor thread.
+    compressor_sender: SyncSender<BlockCompressorMessage>,
+
+    // the handle to check for errors on the thread
+    compressor_thread_handle: JoinHandle<io::Result<()>>,
+}
+
+enum BlockCompressorMessage {
+    AddBlock(DocumentBlock),
+    Stack((StoreReader, DocumentBlock)),
```
fulmicoton: The semantics of Stack are very strange.
Can you comment on it? (and probably inverse the two args)

src/store/writer.rs
```diff
@@ -85,76 +126,145 @@ impl StoreWriter {
...
+    fn register_checkpoint(&mut self, checkpoint: Checkpoint) {
+        self.offset_index_writer.insert(checkpoint.clone());
+        self.doc = checkpoint.doc_range.end;
+    }
+
+    /// Stacks a store reader on top of the documents written so far.
+    /// This method is an optimization compared to iterating over the documents
+    /// in the store and adding them one by one, as the store's data will
+    /// not be decompressed and then recompressed.
+    fn stack(&mut self, block: DocumentBlock, store_reader: StoreReader) -> io::Result<()> {
```
fulmicoton: this really needs a comment on what `block` is about.

src/store/writer.rs
fulmicoton: Why cannot it be  `Stack(StoreReader)` and have the client emit two messages instead of one?

src/store/writer.rs
```diff
@@ -85,76 +126,145 @@ impl StoreWriter {
...
     /// not be decompressed and then recompressed.
-    pub fn stack(&mut self, store_reader: &StoreReader) -> io::Result<()> {
-        if !self.current_block.is_empty() {
-            self.write_and_compress_block()?;
-        }
-        assert_eq!(self.first_doc_in_block, self.doc);
-        let doc_shift = self.doc;
-        let start_shift = self.writer.written_bytes() as usize;
+    pub fn stack(&mut self, store_reader: StoreReader) -> io::Result<()> {
+        self.check_flush_block()?;
```
fulmicoton: why do we need to `check_flush_block`? Any non empty block will have to be flushed anyway isn't it?

src/store/writer.rs
```diff
@@ -85,76 +126,145 @@ impl StoreWriter {
...
+            doc: 0,
+            offset_index_writer: SkipIndexBuilder::new(),
+            intermediary_buffer: Vec::new(),
+            writer: CountingWriter::wrap(writer),
+        }
     }
 
-    fn write_and_compress_block(&mut self) -> io::Result<()> {
-        assert!(self.doc > 0);
+    fn compress_block(&mut self, block: DocumentBlock) -> io::Result<()> {
```
fulmicoton: Should we rename `compress_and_write` ?

src/store/writer.rs
PSeitz: that was unintended in a refactoring

src/store/writer.rs
PSeitz: yes, should be removed

src/store/writer.rs
PSeitz: split into two

src/store/writer.rs
```diff
@@ -56,17 +91,26 @@ impl StoreWriter {
...
+        let block = DocumentBlock {
+            data: self.current_block.to_owned(),
+            num_docs_in_block: self.num_docs_in_current_block,
+        };
+        self.current_block.clear();
+        self.num_docs_in_current_block = 0;
+        if !block.is_empty() {
+            self.compressor_sender
+                .send(BlockCompressorMessage::AddBlock(block))
+                .map_err(|err| io::Error::new(io::ErrorKind::Other, err))?;
```
fulmicoton:
```suggestion
        // We don't do anything if the current block is empty to begin with.
        if self.current_block.is_empty() {
            return Ok(());
        }
        let block = DocumentBlock {
            data: self.current_block.to_owned(),
            num_docs_in_block: self.num_docs_in_current_block,
        };
        self.current_block.clear();
        self.num_docs_in_current_block = 0;
        self.compressor_sender
            .send(BlockCompressorMessage::AddBlock(block))
            .map_err(|err| io::Error::new(io::ErrorKind::Other, err))?;
```

src/store/writer.rs
```diff
@@ -56,17 +91,26 @@ impl StoreWriter {
...
+    fn check_flush_block(&mut self) -> io::Result<()> {
         if self.current_block.len() > self.block_size {
-            self.write_and_compress_block()?;
+            self.send_current_block_to_compressor()?;
+        }
+        Ok(())
+    }
+
+    /// Flushes current uncompressed block and sends to compressor.
+    fn send_current_block_to_compressor(&mut self) -> io::Result<()> {
```
fulmicoton, line 103: I think we can return the SendError directly here.
See discussion on call site.

src/store/writer.rs
```diff
@@ -85,76 +129,137 @@ impl StoreWriter {
...
-    pub fn stack(&mut self, store_reader: &StoreReader) -> io::Result<()> {
-        if !self.current_block.is_empty() {
-            self.write_and_compress_block()?;
-        }
-        assert_eq!(self.first_doc_in_block, self.doc);
-        let doc_shift = self.doc;
-        let start_shift = self.writer.written_bytes() as usize;
+    pub fn stack(&mut self, store_reader: StoreReader) -> io::Result<()> {
+        // We flush the current block first before stacking
+        self.send_current_block_to_compressor()?;
```
fulmicoton, line 159: Both errors are actually errors on the compressing thread.

If there is an error, could we join the companion thread and return 
- its io::Error if it has one
- a custome io::Error if it panicked.

The code that join/harvest the error could be factorized in an independant method.

src/store/writer.rs
```diff
@@ -85,76 +129,137 @@ impl StoreWriter {
...
+
+    /// Finalized the store writer.
+    ///
+    /// Compress the last unfinished block if any,
+    /// and serializes the skip list index on disc.
+    pub fn close(mut self) -> io::Result<()> {
+        self.send_current_block_to_compressor()?;
+        drop(self.compressor_sender);
+
+        self.compressor_thread_handle
```
fulmicoton, line 175: Same as above.

src/store/writer.rs
```diff
@@ -85,76 +129,137 @@ impl StoreWriter {
...
+    }
+
+    /// Store bytes of a serialized document.
+    ///
+    /// The document id is implicitely the current number
+    /// of documents.
+    pub fn store_bytes(&mut self, serialized_document: &[u8]) -> io::Result<()> {
+        let doc_num_bytes = serialized_document.len();
+        VInt(doc_num_bytes as u64).serialize(&mut self.current_block)?;
+        self.current_block.write_all(serialized_document)?;
```
fulmicoton:
```suggestion
        self.current_block.extend_from_slice(serialized_document);
```

src/store/writer.rs
```diff
@@ -85,76 +129,137 @@ impl StoreWriter {
...
+        Ok(())
+    }
+
+    /// Store bytes of a serialized document.
+    ///
+    /// The document id is implicitely the current number
+    /// of documents.
+    pub fn store_bytes(&mut self, serialized_document: &[u8]) -> io::Result<()> {
+        let doc_num_bytes = serialized_document.len();
+        VInt(doc_num_bytes as u64).serialize(&mut self.current_block)?;
```
fulmicoton:
```suggestion
        VInt(doc_num_bytes as u64).serialize_into_vec(&mut self.current_block);
```

src/store/writer.rs
```diff
@@ -85,76 +129,137 @@ impl StoreWriter {
         VInt(doc_num_bytes as u64).serialize(&mut self.current_block)?;
```
fulmicoton:
```suggestion
        VInt(doc_num_bytes as u64).serialize_into_vec(&mut self.current_block);
```

fulmicoton (approved):
Approved but please have a look at the change suggeston and the error handling suggestion. The latter is very optional or can be done later.

src/store/writer.rs
PSeitz: that's a good idea, I don't like the error handling here currently, it's not deterministic which error is returned .. but to join the thread we need to consume `self`. We could swap it with another handle or put it in an option, I don't really like either of those

PR #1393, opened 2022-06-19
add support for phrase slop in query language

### LINKED ISSUE #1390: Add slop phrase query to the query grammar.
@halvorboe added phrase queries with slop #1241 . This feature is interesting quickwit users. We want to expose it through the query parser.

Tilde should be used to define the slop.
`"foo bar"~4

saroh:
### Breaking changes to the existing query language: 
+ `~` is now allowed unescaped in the field name

The first commit still allows `~` in words, might be a bit flimsy.
Second one exposes distance for all leaves. Do we need to explicitly fail for leaves that do not handle slop/distance + implement it for those who do ? (Term maybe, I haven't checked)

Not realy sure on the name, slop is hard to understand, distance is maybe a little to vague.

closes #1390

### DIFF
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -49,10 +49,10 @@ fn word<'a>() -> impl Parser<&'a str, Output = String> {
     (
         satisfy(|c: char| {
             !c.is_whitespace()
-                && !['-', '^', '`', ':', '{', '}', '"', '[', ']', '(', ')'].contains(&c)
+                && !['-', '^', '`', ':', '{', '}', '"', '[', ']', '(', ')', '~'].contains(&c)
         }),
         many(satisfy(|c: char| {
-            !c.is_whitespace() && ![':', '^', '{', '}', '"', '[', ']', '(', ')'].contains(&c)
+            !c.is_whitespace() && ![':', '^', '{', '}', '"', '[', ']', '(', ')', '~'].contains(&c)
         })),
     )
         .map(|(s1, s2): (char, String)| format!("{}{}", s1, s2))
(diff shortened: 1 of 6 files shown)


query-grammar/src/query_grammar.rs
```diff
@@ -49,10 +49,10 @@ fn word<'a>() -> impl Parser<&'a str, Output = String> {
     (
         satisfy(|c: char| {
             !c.is_whitespace()
-                && !['-', '^', '`', ':', '{', '}', '"', '[', ']', '(', ')'].contains(&c)
+                && !['-', '^', '`', ':', '{', '}', '"', '[', ']', '(', ')', '~'].contains(&c)
```
PSeitz: I think we need to add '~' to SPECIAL_CHARS and ESCAPED_SPECIAL_CHARS_PATTERN to allow it in field names

query-grammar/src/query_grammar.rs
```diff
@@ -714,4 +727,19 @@ mod test {
         );
         test_is_parse_err("abc +    ");
     }
+
+    #[test]
+    fn test_matching_distance() {
+        assert!(parse_to_ast().parse("\"a b\"~").is_err());
```
PSeitz, line 737: can you add a test to search for literal `~`, e.g. "~/Documents"

PSeitz:
For naming I think matching distance is much better than slop

query-grammar/src/query_grammar.rs
saroh: It's currently in there which means we only allow it escaped. I've done the change to allow this.

So in terms of breaking changes to the existing query language: 
+ `~` is not allowed in bare words anymore and should be quoted
+ `~` is now allowed unescaped in the field name

saroh:
I gave a quick look, I should be able to plug in the fuzzy query for bare words with the `~` operator soon.

query-grammar/src/query_grammar.rs
```diff
@@ -120,22 +120,35 @@ fn date_time<'a>() -> impl Parser<&'a str, Output = String> {
...
+            phrase,
+            matching_distance,
+        },
+    )
+}
+
+fn matching_distance_val<'a>() -> impl Parser<&'a str, Output = u32> {
+    let matching_distance = (char('~'), many1(digit())).map(|(_, distance): (_, String)| distance);
+    optional(matching_distance).map(|distance| match distance {
+        Some(d) => d.parse::<u32>().unwrap(),
```
PSeitz: can we return an error instead of panic for `"a b"~a`

fulmicoton: I don't think panics in that case because it would not match the `many1(digit())`.

`"a   b"~1000000000000000000000000000000000000000` on the hand will panic and you are right we need to fix that!

fulmicoton:
@PSeitz @saroh I recommend we stick to slop. 

It has the merit of 
- being unambiguous with fuzzy matching.
- being in line with Lucene's vocabulary

saroh:
@fulmicoton @PSeitz so we keep slop in the `phrase_query` mod but it's ok to use matching distance for the grammar ?

Can we define "it" as being an `u8` in the query grammar as it is the type that was chosen for the fuzzy query distance and 255 seems to be good enough of a max distance whatever the `Query` impl for the time being ?

fulmicoton:
Right now let's not handle fuzzy query in the grammar, it would break quickwit.
For the naming, I'd stick to slop in the grammar for the moment.

query-grammar/src/query_grammar.rs
saroh: We probably need to add back the `~` in bare words ?

saroh:
I've got this PR https://github.com/saroh/tantivy/pull/1/files which I can push in here or open later. Adds support for "foo"~1, no changes to the query language.

PSeitz (approved):

PR #1396, opened 2022-06-22
Added datetime support

### LINKED ISSUE #1408: Add precision to datetime field and datetime field internally expressed in microsecs.
In order to have a better DateTime support , In Tantivy, we need to implement:
- Update the current DateTime type to hold unix-timestamp in microseconds
- The input parsing with support only for `rfc3339` format.
- The query parsing with support only for `rfc3339` format.
- DateTime fields will be output in unix-timestamp microsecond, This helps Tantivy users format the value in whatever format for display.
- Assume all DateTime received to be in UTC if not explicitly specified in the format

The DateTime field schema spec needs precision. 
- FastField: We will internally store all dates in `i64` using the unix-timestamp microseconds value. The precision option will be used to store what’s exactly needed in fast fields. As of now, we won’t support nanosecond precision.
- In docStore:  We store the timestamp as `i64` representing the unix-timestamp value in microseconds of the parsed input without any precision knowledge.

evanxg852000:
Implements the new DateTime field supporting feature in Tantivy.
Closes https://github.com/quickwit-oss/tantivy/issues/1408

### DIFF
--- a/examples/date_time_field.rs
+++ b/examples/date_time_field.rs
@@ -1,0 +1,69 @@
+// # DateTime field example
+//
+// This example shows how the DateTime field can be used
+
+use tantivy::collector::TopDocs;
+use tantivy::query::QueryParser;
+use tantivy::schema::{Cardinality, DateOptions, Schema, Value, INDEXED, STORED, STRING};
+use tantivy::Index;
+
+fn main() -> tantivy::Result<()> {
+    // # Defining the schema
+    let mut schema_builder = Schema::builder();
+    let opts = DateOptions::from(INDEXED)
+        .set_stored()
+        .set_fast(Cardinality::SingleValue)
+        .set_precision(tantivy::DatePrecision::Seconds);
+    let occurred_at = schema_builder.add_date_field("occurred_at", opts);
+    let event_type = schema_builder.add_text_field("event", STRING | STORED);
+    let schema = schema_builder.build();
+
+    // # Indexing documents
+    let index = Index::create_in_ram(schema.clone());
+
+    let mut index_writer = index.writer(50_000_000)?;
+    let doc = schema.parse_document(
+        r#"{
+        "occurred_at": "2022-06-22T12:53:50.53Z",
+        "event": "pull-request"
+    }"#,
+    )?;
+    index_writer.add_document(doc)?;
+    let doc = schema.parse_document(
+        r#"{
+        "occurred_at": "2022-06-22T13:00:00.22Z",
+        "event": "comment"
+    }"#,
+    )?;
+    index_writer.add_document(doc)?;
+    index_writer.commit()?;
+
+    let reader = index.reader()?;
+    let searcher = reader.searcher();
+
+    // # Default fields: event_type
+    let query_parser = QueryParser::for_index(&index, vec![event_type]);
+    {
+        let query = query_parser.parse_query("event:comment")?;
+        let count_docs = searcher.search(&*query, &TopDocs::with_limit(5))?;
+        assert_eq!(count_docs.len(), 1);
+    }
+    {
+        let query = query_parser
+            .parse_query(r#"occurred_at:[2022-06-22T12:58:00Z TO 2022-06-23T00:00:00Z}"#)?;
+        let count_docs = searcher.search(&*query, &TopDocs::with_limit(4))?;
+        assert_eq!(count_docs.len(), 1);
+        for (_score, doc_address) in count_docs {
+            let retrieved_doc = searcher.doc(doc_address)?;
+            assert!(matches!(
+                retrieved_doc.get_first(occurred_at),
+                Some(Value::Date(_))
+            ));
+            assert_eq!(
+                schema.to_json(&retrieved_doc),
+                r#"{"event":["comment"],"occurred_at":["2022-06-22T13:00:00.22Z"]}"#
+            );
+        }
+    }
+    Ok(())
+}
--- a/fastfield_codecs/src/bitpacked.rs
+++ b/fastfield_codecs/src/bitpacked.rs
@@ -14,7 +14,7 @@ pub struct BitpackedFastFieldReader {
     pub max_value_u64: u64,
 }
 
-impl<'data> FastFieldCodecReader for BitpackedFastFieldReader {
+impl FastFieldCodecReader for BitpackedFastFieldReader {
     /// Opens a fast field given a file.
     fn open_from_bytes(bytes: &[u8]) -> io::Result<Self> {
         let (_data, mut footer) = bytes.split_at(bytes.len() - 16);
--- a/src/collector/histogram_collector.rs
+++ b/src/collector/histogram_collector.rs
@@ -72,8 +72,7 @@ impl HistogramComputer {
             return;
         }
         let delta = value - self.min_value;
-        let delta_u64 = delta.to_u64();
-        let bucket_id: usize = self.divider.divide(delta_u64) as usize;
+        let bucket_id: usize = self.divider.divide(delta) as usize;
         if bucket_id < self.counts.len() {
             self.counts[bucket_id] += 1;
         }
@@ -287,7 +286,7 @@ mod tests {
             DateTime::from_primitive(
                 Date::from_calendar_date(1980, Month::January, 1)?.with_hms(0, 0, 0)?,
             ),
-            3600 * 24 * 365, // it is just for a unit test... sorry leap years.
+            3_600_000_000 * 24 * 365, // it is just for a unit test... sorry leap years.
             10,
         );
         let week_histogram = searcher.search(&all_query, &week_histogram_collector)?;
--- a/src/fastfield/mod.rs
+++ b/src/fastfield/mod.rs
@@ -52,11 +52,13 @@ pub trait MultiValueLength {
     fn get_total_len(&self) -> u64;
 }
 
-/// Trait for types that are allowed for fast fields: (u64, i64 and f64).
+/// Trait for types that are allowed for fast fields:
+/// (u64, i64 and f64, bool, DateTime).
 pub trait FastValue: Clone + Copy + Send + Sync + PartialOrd + 'static {
     /// Converts a value from u64
     ///
     /// Internally all fast field values are encoded as u64.
+    /// **Note: To be used for converting encoded Term, Posting values.**
     fn from_u64(val: u64) -> Self;
 
     /// Converts a value to u64.
@@ -189,24 +191,27 @@ impl FastValue for bool {
 }
 
 impl FastValue for DateTime {
-    fn from_u64(timestamp_u64: u64) -> Self {
-        let unix_timestamp = i64::from_u64(timestamp_u64);
-        Self::from_unix_timestamp(unix_timestamp)
+    /// Converts a timestamp microseconds into DateTime.
+    ///
+    /// **Note the timestamps is expected to be in microseconds.**
+    fn from_u64(timestamp_micros_u64: u64) -> Self {
+        let timestamp_micros = i64::from_u64(timestamp_micros_u64);
+        Self::from_timestamp_micros(timestamp_micros)
     }
 
     fn to_u64(&self) -> u64 {
-        self.into_unix_timestamp().to_u64()
+        common::i64_to_u64(self.into_timestamp_micros())
     }
 
     fn fast_field_cardinality(field_type: &FieldType) -> Option<Cardinality> {
         match *field_type {
-            FieldType::Date(ref integer_options) => integer_options.get_fastfield_cardinality(),
+            FieldType::Date(ref options) => options.get_fastfield_cardinality(),
             _ => None,
         }
     }
 
     fn as_u64(&self) -> u64 {
-        self.into_unix_timestamp().as_u64()
+        self.into_timestamp_micros().as_u64()
     }
 
     fn to_type() -> Type {
@@ -261,9 +266,9 @@ mod tests {
     use super::*;
     use crate::directory::{CompositeFile, Directory, RamDirectory, WritePtr};
     use crate::merge_policy::NoMergePolicy;
-    use crate::schema::{Document, Field, NumericOptions, Schema, FAST, STRING, TEXT};
+    use crate::schema::{Document, Field, Schema, FAST, STRING, TEXT};
     use crate::time::OffsetDateTime;
-    use crate::{Index, SegmentId, SegmentReader};
+    use crate::{DateOptions, DatePrecision, Index, SegmentId, SegmentReader};
 
     pub static SCHEMA: Lazy<Schema> = Lazy::new(|| {
         let mut schema_builder = Schema::builder();
@@ -559,8 +564,8 @@ mod tests {
     }
 
     #[test]
-    fn test_default_datetime() {
-        assert_eq!(0, DateTime::make_zero().into_unix_timestamp());
+    fn test_default_date() {
+        assert_eq!(0, DateTime::make_zero().into_timestamp_secs());
     }
 
     fn get_vals_for_docs(ff: &MultiValuedFastFieldReader<u64>, docs: Range<u32>) -> Vec<u64> {
@@ -766,10 +771,15 @@ mod tests {
     fn test_datefastfield() -> crate::Result<()> {
         use crate::fastfield::FastValue;
         let mut schema_builder = Schema::builder();
-        let date_field = schema_builder.add_date_field("date", FAST);
+        let date_field = schema_builder.add_date_field(
+            "date",
+            DateOptions::from(FAST).set_precision(DatePrecision::Microseconds),
+        );
         let multi_date_field = schema_builder.add_date_field(
             "multi_date",
-            NumericOptions::default().set_fast(Cardinality::MultiValues),
+            DateOptions::default()
+                .set_precision(DatePrecision::Microseconds)
+                .set_fast(Cardinality::MultiValues),
         );
         let schema = schema_builder.build();
         let index = Index::create_in_ram(schema);
@@ -797,23 +807,23 @@ mod tests {
         let dates_fast_field = fast_fields.dates(multi_date_field).unwrap();
         let mut dates = vec![];
         {
-            assert_eq!(date_fast_field.get(0u32).into_unix_timestamp(), 1i64);
+            assert_eq!(date_fast_field.get(0u32).into_timestamp_micros(), 1i64);
             dates_fast_field.get_vals(0u32, &mut dates);
             assert_eq!(dates.len(), 2);
-            assert_eq!(dates[0].into_unix_timestamp(), 2i64);
-            assert_eq!(dates[1].into_unix_timestamp(), 3i64);
+            assert_eq!(dates[0].into_timestamp_micros(), 2i64);
+            assert_eq!(dates[1].into_timestamp_micros(), 3i64);
         }
         {
-            assert_eq!(date_fast_field.get(1u32).into_unix_timestamp(), 4i64);
+            assert_eq!(date_fast_field.get(1u32).into_timestamp_micros(), 4i64);
             dates_fast_field.get_vals(1u32, &mut dates);
             assert!(dates.is_empty());
         }
         {
-            assert_eq!(date_fast_field.get(2u32).into_unix_timestamp(), 0i64);
+            assert_eq!(date_fast_field.get(2u32).into_timestamp_micros(), 0i64);
             dates_fast_field.get_vals(2u32, &mut dates);
             assert_eq!(dates.len(), 2);
-            assert_eq!(dates[0].into_unix_timestamp(), 5i64);
-            assert_eq!(dates[1].into_unix_timestamp(), 6i64);
+            assert_eq!(dates[0].into_timestamp_micros(), 5i64);
+            assert_eq!(dates[1].into_timestamp_micros(), 6i64);
         }
         Ok(())
     }
--- a/src/fastfield/writer.rs
+++ b/src/fastfield/writer.rs
@@ -7,12 +7,13 @@ use tantivy_bitpacker::BlockedBitpacker;
 
 use super::multivalued::MultiValuedFastFieldWriter;
 use super::serializer::FastFieldStats;
-use super::{FastFieldDataAccess, FastFieldType};
+use super::{FastFieldDataAccess, FastFieldType, FastValue};
 use crate::fastfield::{BytesFastFieldWriter, CompositeFastFieldSerializer};
 use crate::indexer::doc_id_mapping::DocIdMapping;
 use crate::postings::UnorderedTermId;
-use crate::schema::{Cardinality, Document, Field, FieldEntry, FieldType, Schema};
+use crate::schema::{Cardinality, Document, Field, FieldEntry, FieldType, Schema, Value};
 use crate::termdict::TermOrdinal;
+use crate::DatePrecision;
 
 /// The `FastFieldsWriter` groups all of the fast field writers.
 pub struct FastFieldsWriter {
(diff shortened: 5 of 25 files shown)


evanxg852000:
Questions: @fulmicoton
- I will handle Date field deprecation in a following PR. The idea is to make use of the new DateTime field underneath with defaults that correspond to deprecating Date field.   WDT?
- I currently have this `DateTimeOptions::set_default_input_timezone(mut self, timezone: Tz) -> DateTimeOptions `. I would like to replace Tz with &str so that we don't have to expose chrono_tz, unfortunately  `Tz::from_str(&time_zone_name)` returns a result, it looks kind of weird in the builder pattern. 
please see https://github.com/quickwit-oss/tantivy/pull/1396/files#diff-d23fd0e9872657894b5ebb4f510cfbd6d099eee1e200c96e1c727f7322af0016R168-R173   what's you take?
- Currently, I am implementing serialisation/deserialisation for types like `DateTimePrecison` and `DateTimeOptions` I feel like we can remove those and let tantivy users decide how they want to construct, and what mnemonic they want to use. Right? 
See: https://github.com/quickwit-oss/tantivy/pull/1396/files#diff-d23fd0e9872657894b5ebb4f510cfbd6d099eee1e200c96e1c727f7322af0016R425-R464
https://github.com/quickwit-oss/tantivy/pull/1396/files#diff-d230b222363da840c9989a567adc9d16a32bcafaafd6e62fb85afcd94d4189c6R235-R245

fulmicoton:
- I will handle Date field deprecation in a following PR. The idea is to make use of the new DateTime field underneath with defaults that correspond to deprecating Date field. WDT?

Yes if it is possible that would be great. If it is not possible let me know.

- I currently have this DateTimeOptions::set_default_input_timezone(mut self, timezone: Tz) -> DateTimeOptions . I would like to replace Tz with &str so that we don't have to expose chrono_tz, unfortunately Tz::from_str(&time_zone_name) returns a result, it looks kind of weird in the builder pattern.
please see https://github.com/quickwit-oss/tantivy/pull/1396/files#diff-d23fd0e9872657894b5ebb4f510cfbd6d099eee1e200c96e1c727f7322af0016R168-R173 what's you take?

If possible, I'd keep DateTime without timezone in tantivy and have a notion of timezone in quickwit... but I might not have thought this through so let me know what would be the obstacle. (Query parsing I assume?)

evanxg852000:
> If possible, I'd keep DateTime without timezone in tantivy and have a notion of timezone in quickwit... but I might not have thought this through so let me know what would be the obstacle. (Query parsing I assume?)

We can leave timezone thing for quickwit, this just means we will let the Tantivy users know that when a timezone is not specified in the input, we assume UTC. I think that's also good for a library

examples/date_time_field.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use tantivy::collector::TopDocs;
+use tantivy::query::QueryParser;
+use tantivy::schema::{Cardinality, DateTimeOptions, Schema, Value, INDEXED, STORED, STRING};
+use tantivy::{DateTimeFormat, DateTimePrecision, Index};
+
+fn main() -> tantivy::Result<()> {
+    // # Defining the schema
+    let mut schema_builder = Schema::builder();
+    let mut date_formats = HashSet::new();
```
fulmicoton: surely this needs to be a ordered. HashSet is a very bad idea here.

src/schema/field_type.rs
```diff
@@ -58,13 +65,14 @@ pub enum Type {
...
 }
 
-const ALL_TYPES: [Type; 9] = [
+const ALL_TYPES: [Type; 10] = [
     Type::Str,
     Type::U64,
     Type::I64,
     Type::F64,
     Type::Bool,
     Type::Date,
```
fulmicoton: The old Date type was actually a `DateTime` (yes the name sucks).

Is the plan to add DateTime and deprecate the older Date, or to replace right away?

src/schema/date_time_options.rs
```diff
@@ -0,0 +1,606 @@
...
+    indexed: bool,
+    // This boolean has no effect if the field is not marked as indexed true.
+    fieldnorms: bool,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    fast: Option<Cardinality>,
+    stored: bool,
+
+    // Accepted input format for parsing input as DateTime.
+    #[serde(default = "default_input_formats")]
+    input_formats: HashSet<DateTimeFormat>,
```
fulmicoton: This cannot be an HashSet. Order matters here.

src/schema/date_time.rs
```diff
@@ -0,0 +1,350 @@
...
+}
+
+impl DerefMut for PreciseDateTime {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+impl fmt::Debug for PreciseDateTime {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
fulmicoton:
```suggestion
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```

src/schema/date_time.rs
```diff
@@ -0,0 +1,350 @@
...
+            OffsetDateTime::from_unix_timestamp_nanos((timestamp as i128) * 1_000_000)
+        }
+        DateTimePrecision::Microseconds => {
+            OffsetDateTime::from_unix_timestamp_nanos((timestamp as i128) * 1000)
+        }
+        DateTimePrecision::Nanoseconds => {
+            OffsetDateTime::from_unix_timestamp_nanos(timestamp as i128)
+        }
+    }
+    .map_err(|error| error.to_string())
```
fulmicoton:
```suggestion
    .map_err(ToString::to_string)
```

src/schema/date_time.rs
```diff
@@ -0,0 +1,350 @@
...
+        self.timestamp
+    }
+
+    /// Returns the underlying precision.
+    pub fn get_precision(&self) -> DateTimePrecision {
+        self.precision
+    }
+
+    /// Convert to UTC `OffsetDateTime`
+    pub fn into_utc(self) -> OffsetDateTime {
```
fulmicoton: good idea

src/schema/date_time.rs
```diff
@@ -0,0 +1,350 @@
...
+
+    /// Converts to the underlying timestamp.
+    /// TODO: kept for backward compatibility, should we remove it?
+    pub const fn into_timestamp(self) -> i64 {
+        let Self { timestamp, .. } = self;
+        timestamp
+    }
+
+    /// Returns the underlying timestamp.
+    pub fn get_timestamp(&self) -> i64 {
```
fulmicoton: The name does not make it clear that this is with the given precision...
`get_truncated_timestamp` `get_inner_timestamp`...

src/schema/date_time.rs
```diff
@@ -0,0 +1,350 @@
...
+        match precision {
+            DateTimePrecision::Seconds => timestamp,
+            DateTimePrecision::Milliseconds => timestamp / 1_000,
+            DateTimePrecision::Microseconds => timestamp / 1_000_000,
+            DateTimePrecision::Nanoseconds => timestamp / 1_000_000_000,
+        }
+    }
+
+    /// Converts to the underlying timestamp.
+    /// TODO: kept for backward compatibility, should we remove it?
```
fulmicoton, line 103: but that's not really backward compatible anyway is it?

src/schema/date_time.rs
```diff
@@ -0,0 +1,350 @@
...
+    pub const fn into_timestamp(self) -> i64 {
+        let Self { timestamp, .. } = self;
```
fulmicoton: self.timestamp would have worked too

src/schema/date_time.rs
```diff
@@ -0,0 +1,350 @@
...
+pub struct DateTime {
+    /// The timestamp component.
+    pub(crate) timestamp: i64,
+    /// The precision component to denote if the timestamp
+    /// value is in seconds, milliseconds, microseconds or nanoseconds.
+    pub(crate) precision: DateTimePrecision,
+}
+
+impl DateTime {
+    /// Create new instance from UNIX timestamp.
```
fulmicoton:
```suggestion
    /// Create new instance from UNIX timestamp (in seconds).
```

(I agree UNIX timestamp generally suggests seconds, but we might as well be explicit here)

src/schema/date_time_options.rs
```diff
@@ -0,0 +1,606 @@
...
+
+    // Internal storage precision, used to avoid storing
+    // very large numbers when not needed. This optimizes compression.
```
fulmicoton: Does it apply to 
- indexing
- stored
- fast 
- index sorting
or only
- fast
or only 
-fast
-indexing?

It is not an obvious question and it should be detailed here.

src/schema/date_time_options.rs
```diff
@@ -0,0 +1,606 @@
...
+
+    /// Parses number input.
+    pub fn parse_number(&self, value: i64) -> Result<OffsetDateTime, String> {
+        (self.number_parser)(value)
+    }
+}
+
+impl From<DateTimeOptions> for DateTimeParsersHolder {
+    fn from(opts: DateTimeOptions) -> Self {
+        let mut string_parsers: VecDeque<StringDateTimeParser> = VecDeque::new();
```
fulmicoton: why not Vec?

src/schema/date_time_options.rs
```diff
@@ -0,0 +1,606 @@
...
+    }
+}
+
+impl DateTimeParsersHolder {
+    /// Parses string input.
+    pub fn parse_string(&mut self, value: String) -> Result<OffsetDateTime, String> {
+        for (index, parser) in self.string_parsers.iter().enumerate() {
+            if let Ok(date_time) = parser(&value) {
+                // Move successful parser in front of queue.
+                self.string_parsers.swap(0, index);
```
fulmicoton: Ah I see... 
I think Vec has swap too.

I am not sure this is a good idea.

So with the HashSet defintion, the output of quickwit right now is not really deterministic at compile time.
Date format can be ambiguous ( think about MM-DD-YYY vs DD-MM-YYYY  and 01/03/2021).
Which means that our output depends on the order of our date formats.

By using a HashSet, you make the output of quickwit dependent on the implementation of hashset.
(A change in the rust standard library fro instance, would break unit tests.) 
We need to stay away from HashSet.

I would also stay away from the reordering logic, but I wonder...
So it has two functions I believe:
- a) performance.
- b) best effort to deal with the ambiguity described above.

a) can be partially addressed by extracting ez to check predicate for each format.
For instance on the date string length.
- If the date is not exactly 10chrs. skip format XXXXX.

b) is tricky. With your solution we can "eventually" find the right format when we reach a date that is unambiguous (e.g. 24/02/2021). On the other hand, I think we should give all the tools to the user to tell use the right thing. I'd rather not do the swapping logic and just rely on the priority logic.

src/schema/date_time_options.rs
```diff
@@ -0,0 +1,606 @@
...
+        DateTimeParsersHolder {
+            string_parsers,
+            number_parser,
+        }
+    }
+}
+
+/// Parses datetime strings using RFC3339 formatting.
+fn rfc3339_parser(value: &str) -> Result<OffsetDateTime, String> {
+    OffsetDateTime::parse(value, &Rfc3339).map_err(|error| error.to_string())
```
fulmicoton, line 334: nitpick: I think you would be better off keepign the original error in these helper functions and convert them on the caller side.

src/schema/date_time_options.rs
```diff
@@ -0,0 +1,606 @@
...
+
+// Parser function types
+type StringDateTimeParser = Arc<dyn Fn(&str) -> Result<OffsetDateTime, String> + Sync + Send>;
+type NumberDateTimeParser = Arc<dyn Fn(i64) -> Result<OffsetDateTime, String> + Sync + Send>;
+
+/// A struct for holding datetime parsing functions.
+#[derive(Clone)]
+pub struct DateTimeParsersHolder {
+    /// Functions for parsing string input as datetime.
+    string_parsers: VecDeque<StringDateTimeParser>,
```
fulmicoton: Vec?

src/schema/value.rs
```diff
@@ -26,6 +26,8 @@ pub enum Value {
     Bool(bool),
     /// Date/time with second precision
     Date(DateTime),
+    /// DateTime with timestamp precision
+    DateTime(PreciseDateTime),
```
fulmicoton, line 30: DateTimeWithPrecision?

src/schema/date_time.rs
```diff
@@ -0,0 +1,350 @@
...
+/// A wrapper around DateTime to denote it has custom precision.
+// It only exists to provide support the underlying fast
+// field value for both `[Value::Date]` and `[Value::DateTime]` variants.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct PreciseDateTime(
+    /// The inner DateTime
+    pub DateTime,
+);
+
+impl Deref for PreciseDateTime {
```
fulmicoton: But the inner DateTime is incorrect right? The precision here is arbitrary.
We don't want to expose it.

src/schema/date_time.rs
```diff
@@ -0,0 +1,350 @@
...
+        }
+    }
+    .map_err(|error| error.to_string())
+}
+
+/// A wrapper around DateTime to denote it has custom precision.
+// It only exists to provide support the underlying fast
+// field value for both `[Value::Date]` and `[Value::DateTime]` variants.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct PreciseDateTime(
```
fulmicoton: Rename to TruncatedTimestamp maybe.

src/schema/date_time.rs
```diff
@@ -0,0 +1,350 @@
...
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+impl fmt::Debug for PreciseDateTime {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("DateTime")
+            .field("timestamp", &self.0.timestamp)
+            .field("precision", &self.0.precision)
```
fulmicoton: But the precision is wrong isn't it? Also we don't want to print `DateTime` here.

fulmicoton:
I don't understand the logic around PreciseDateTime.

In the FastField we store truncated values I assume? (e.g. a milliseconds precision timestamp for instance).
Then I understand that you want a type on which you can implement FastValue. This means you create the type by taking the int without really knowing the precision. The schema is then needed to deal with this.
Then why do we expose the inner DateTime? That seems error-prone.

evanxg852000:
So PreciseDateTime is the new date time with actual precision, currently, DateTime API was enhanced to conform to the old DateTime that's why I kept the old API and added few things. This allowed using DateTime like before and not breaking existing code. PreciseDateTime is the one that actually makes use of the new API. As I said in the comment it only exist because I could implement both fastfield types for the same DateTime. I don't want to rip something without us agreeing so DateTime behaves as the way it use to with a precison -> seconds and the timestamp is second. when using the DateTime field, precison can change to millis or ... .
We always store the computed timestamp value then when extracting the date  the precision is used.
In both `DateTime` (seconds precision) and `PreciseDateTime` (any precision) the precison is always correct. I think I need to find a better name.
By exposing the inner DateTime, user only needs to worry about that type that already existed.

src/schema/field_type.rs
evanxg852000: So all these two types `DateTime` and `PreciseDateTime` exist because my plan is to keep Type::Date working with a deprecated label. If we can drop the old DateTime type and accept the breaking change let me know your opinion.

src/schema/date_time.rs
evanxg852000: Precision is always correct, whether you use `DateTime::from_unix_timestamp(u64)` or `DateTime::from_utc_with_precision(utc_dt, precison)` constructor. Maybe the names should be:
- DateTime ->  DateTimeWithSecondPrecision (old DateTime)
- PreciseDateTime ->  DateTimeWithCustomPrecision (for new fast field)

fulmicoton:
> In both DateTime (seconds precision) and PreciseDateTime (any precision) the precison is always correct. 

@evanxg852000  I don't see how this is possible... In the FastValue of PreciseDateTime, from assumes that we have seconds in the fast field, and the conversion to a fastvalue is taking whatever truncated value we had.

src/indexer/json_term_writer.rs
```diff
@@ -325,7 +327,7 @@ impl<'a> JsonTermWriter<'a> {
         self.close_path_and_set_type(T::to_type());
         self.term_buffer
             .as_mut()
-            .extend_from_slice(val.to_u64().to_be_bytes().as_slice());
+            .extend_from_slice(val.to_u64(None).to_be_bytes().as_slice());
```
evanxg852000: @fulmicoton I am not applying the precision here to DateTime. my reasoning is that this is all indexing though we are pushing fast values. Am I wrong?

examples/date_time_field.rs
```diff
@@ -0,0 +1,69 @@
...
+//
+// This example shows how the DateTime field can be used
+
+use tantivy::collector::TopDocs;
+use tantivy::query::QueryParser;
+use tantivy::schema::{Cardinality, DateOptions, Schema, Value, INDEXED, STORED, STRING};
+use tantivy::Index;
+
+fn main() -> tantivy::Result<()> {
+    // # Defining the schema
```
fulmicoton, line 11: awesome!

fastfield_codecs/src/bitpacked.rs
```diff
@@ -14,7 +14,7 @@ pub struct BitpackedFastFieldReader {
     pub max_value_u64: u64,
 }
 
-impl<'data> FastFieldCodecReader for BitpackedFastFieldReader {
```
fulmicoton, line 17: thx

src/lib.rs
```diff
@@ -171,17 +195,31 @@ impl DateTime {
...
+        self.timestamp_micros / 1_000
+    }
+
+    /// Convert to UNIX timestamp in microseconds.
+    pub const fn into_timestamp_micros(self) -> i64 {
+        self.timestamp_micros
+    }
+
+    /// Convert to UNIX timestamp in microseconds.
+    pub const fn into_timestamp_nanos(self) -> i64 {
```
fulmicoton: i think we should ditch the nanos one. It could suggest we have nanos precision.

src/lib.rs
```diff
@@ -145,21 +145,45 @@ use crate::time::{OffsetDateTime, PrimitiveDateTime, UtcOffset};
...
+    pub const fn from_timestamp_micros(microseconds: i64) -> Self {
+        Self {
+            timestamp_micros: microseconds,
+        }
+    }
+
+    /// Create new from UNIX timestamp in nanoseconds.
+    pub const fn from_timestamp_nanos(nanoseconds: i128) -> Self {
+        Self {
+            timestamp_micros: (nanoseconds as i64 / 1_000),
```
guilload: We should divide into an `i128` first and then convert to `i64`. [See for yourself](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f86402599d1d1ac71742503ffb37d0a9). Can you add the corresponding unit test?

src/lib.rs
```diff
@@ -145,21 +145,45 @@ use crate::time::{OffsetDateTime, PrimitiveDateTime, UtcOffset};
...
+
+    /// Create new from UNIX timestamp in microseconds.
+    pub const fn from_timestamp_micros(microseconds: i64) -> Self {
+        Self {
+            timestamp_micros: microseconds,
+        }
+    }
+
+    /// Create new from UNIX timestamp in nanoseconds.
+    pub const fn from_timestamp_nanos(nanoseconds: i128) -> Self {
```
fulmicoton: I would remove that one (and inline the implem in `from_utc`)

src/fastfield/mod.rs
```diff
@@ -62,7 +63,9 @@ pub trait FastValue: Clone + Copy + Send + Sync + PartialOrd + 'static {
     /// Converts a value to u64.
     ///
     /// Internally all fast field values are encoded as u64.
-    fn to_u64(&self) -> u64;
```
fulmicoton: I understand you use the field type for truncation on datetime, but it seems too convoluted. Let's not change the `to_u64`.

Instead, let's put the truncation at the fast field writer level.

src/fastfield/writer.rs
```diff
@@ -305,7 +335,7 @@ impl IntFastFieldWriter {
     pub fn add_document(&mut self, doc: &Document) {
         match doc.get_first(self.field) {
             Some(v) => {
-                self.add_val(super::value_to_u64(v));
+                self.add_val(super::value_to_u64(v, &self.field_type));
```
fulmicoton: We can put the truncation logic here.

```
let mut val = 
    super::value_to_u64(v, self.field_type);
if let Some(precision) = self.precision_opt {
    val = precision.truncate(val);
}
```

What's a bit lame is that we will have to do it elsewhere for indexing.

src/indexer/json_term_writer.rs
fulmicoton: Great question!

`FastValue` is a poorly named trait. 
It is not only used for fast values, it is also used for the inverted index.

In reality it offers a strictly monotonic mapping from the type to U64 (and its reciprocal). 

We could be named `MonoMapU64` or something like that.
Can you fill a ticket to do this rename?

fulmicoton: Back to your question... in the `json_term_writer` we should apply some truncation on dates.

I suggest we always truncate with second precision?

We should probably truncate to seconds always for regular (=not Json) date fields too; and only use the precision setting for fast fields.

Besides this is the current behavior of tantivy in a sense.

src/indexer/segment_writer.rs
```diff
@@ -248,7 +248,7 @@ impl SegmentWriter {
                 FieldType::Date(_) => {
                     for value in values {
                         let date_val = value.as_date().ok_or_else(make_schema_error)?;
-                        term_buffer.set_u64(date_val.to_u64());
+                        term_buffer.set_u64(date_val.to_u64(None));
```
fulmicoton: See discussion in json_term_writer.
We want some truncation here.

I suggest we truncate to the second all of the time.

src/fastfield/writer.rs
```diff
@@ -230,6 +258,7 @@ impl FastFieldsWriter {
 /// using `common::i64_to_u64` and `common::f64_to_u64`.
 pub struct IntFastFieldWriter {
     field: Field,
+    field_type: FieldType,
```
fulmicoton, line 261: I think we coudl keep the precision only.

src/schema/date_time_options.rs
```diff
@@ -0,0 +1,287 @@
+use std::ops::BitOr;
+
+use serde::{Deserialize, Serialize};
+
+use super::Cardinality;
+use crate::schema::flags::{FastFlag, IndexedFlag, SchemaFlagList, StoredFlag};
+
+/// DateTime Precision
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub enum DatePrecision {
```
fulmicoton, line 11: `"precision": "Milliseconds" `
does not look great in a JSON.

`"precision": "milliseconds" would look better IMHO

can you add attriubte to make the serde serialization look nicer?
(serde lower should do the job.)

src/lib.rs
evanxg852000: inlined it and added a test for it.

src/collector/histogram_collector.rs
```diff
@@ -72,7 +72,7 @@ impl HistogramComputer {
             return;
         }
         let delta = value - self.min_value;
-        let delta_u64 = delta.to_u64();
+        let delta_u64 = delta.to_u64(None);
```
fulmicoton: we can remove that line anyway I think.

`delta_u64 = delta`

fulmicoton (changes requested):
see comments inline

evanxg852000:
@fulmicoton In order to support the special precision settings truncation for fast values and the seconds truncation for others, I have introduced `FastField::from_u64_fast` and `FastField::to_u64_fast` methods to use them in fast field handling. It was becoming hard to treat `Type::Date` in places especially the query parsing gets involved by constructing Terms from the same `FastField` traits.

src/schema/date_time_options.rs
```diff
@@ -0,0 +1,288 @@
...
+    stored: bool,
+    // Internal storage precision, used to optimize storage
+    // compression on fast fields.
+    #[serde(default)]
+    precision: DatePrecision,
+}
+
+impl Default for DateOptions {
+    fn default() -> Self {
+        Self {
```
fulmicoton: You have clippy warnings here. Either use derive(Default) or add an annotation to make sure Clippy keeps quiet.
I'm ok with both solutions.

fulmicoton (approved):

src/schema/date_time_options.rs
evanxg852000: used Default, shorter & the recommended Clippy way

fulmicoton (approved):

fulmicoton:
Good job!

twe4ked:
Thanks for this! Any chance of a release with this included soon? 🙏

PSeitz:
@twe4ked The next release will be a breaking release, so we try to add more stuff into it. Maybe a release in around 1 month

PR #1401, opened 2022-06-28
Add slop to quoted bare words

saroh:
Following #1393 
The fact that phrase queries are tokenized does not make this ideal TBH. It'd probably be better to have delimiters in the query language for a non tokenized string and have a slop param for it.

### DIFF
--- a/src/query/query_parser/logical_ast.rs
+++ b/src/query/query_parser/logical_ast.rs
@@ -8,6 +8,7 @@ use crate::Score;
 #[derive(Clone)]
 pub enum LogicalLiteral {
     Term(Term),
+    TermWithSlop(Term, u8),
     Phrase(Vec<(usize, Term)>, u32),
     Range {
         field: Field,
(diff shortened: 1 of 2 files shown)


src/query/query_parser/logical_ast.rs
```diff
@@ -74,6 +75,14 @@ impl fmt::Debug for LogicalLiteral {
     fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
         match *self {
             LogicalLiteral::Term(ref term) => write!(formatter, "{:?}", term),
+            LogicalLiteral::TermWithSlop(ref term, slop) => {
+                write!(formatter, "\"{:?}\"", term)?;
+                if slop > 0 {
+                    write!(formatter, "~{:?}", slop)
```
fulmicoton, line 81:
```suggestion
                    write!(formatter, "~{}", slop)
```

fulmicoton:
It is too dangerous to add if it is not soemthing we can disable via conf.

It is very easy to craft a fuzzy query that is very expensive, so any service with tantivy in prod would be vulnerable.

saroh:
@fulmicoton yeah I'm closing this, imo it requires a little bit of design, we can't both tokenize and not tokenize within `""`.

PR #1403, opened 2022-07-01
expose doc store cache size

PSeitz:
expose lru doc store cache size
optimize doc store cache size

### DIFF
--- a/src/core/searcher.rs
+++ b/src/core/searcher.rs
@@ -6,7 +6,7 @@ use crate::core::{Executor, SegmentReader};
 use crate::query::Query;
 use crate::schema::{Document, Schema, Term};
 use crate::space_usage::SearcherSpaceUsage;
-use crate::store::StoreReader;
+use crate::store::{CacheStats, StoreReader};
 use crate::{DocAddress, Index, Opstamp, SegmentId, TrackedObject};
 
 /// Identifies the searcher generation accessed by a [Searcher].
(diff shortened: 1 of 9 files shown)


src/core/searcher.rs
```diff
@@ -77,11 +77,17 @@ impl Searcher {
...
         generation: TrackedObject<SearcherGeneration>,
+        doc_store_cache_size: usize,
     ) -> io::Result<Searcher> {
         let store_readers: Vec<StoreReader> = segment_readers
             .iter()
             .map(SegmentReader::get_store_reader)
             .collect::<io::Result<Vec<_>>>()?;
+
+        for store_reader in &store_readers {
+            store_reader.set_cache_size(doc_store_cache_size);
```
fulmicoton: why could not this be done at the SegmentReader creation?

src/store/reader.rs
```diff
@@ -112,6 +144,11 @@ impl StoreReader {
         self.cache.stats()
     }
 
+    /// Set lru cache size for decompressed blocks. Defaults to 100 (LRU_CACHE_CAPACITY).
+    pub(crate) fn set_cache_size(&self, size: usize) {
+        self.cache.set_size(size)
```
fulmicoton: why is this a setter?

fulmicoton:
that PR is allowing to set the cache size on the docstore. The PR does not really explain why.
My understanding is that the benefit is to allow a cache of size 1 when merging?

Could this be done on segment reader creation? 
Or not done at all actually? Are you thinking of enabling a large LRU size?

src/core/searcher.rs
PSeitz: yes I changed it

PSeitz:
> that PR is allowing to set the cache size on the docstore. The PR does not really explain why. My understanding is that the benefit is to allow a cache of size 1 when merging?
> 
> Could this be done on segment reader creation? Or not done at all actually? Are you thinking of enabling a large LRU size?

One benefit is to reduce the cache size for merging. The other benefit is to reduce memory consumption for quickwit fetch_docs. The cache size is much more important with the larger blocks now.

fulmicoton (approved):

PR #1411, opened 2022-07-07
remove searcher pool and make Searcher cloneable

### LINKED ISSUE #1410: Remove searcher pool
The searcher pool was originally provided, since searcher methods were mutable. Since there are no mutable methods anymore, we can remove the pool and simply make the searcher cloneable.

PSeitz:
closes #1410


src/core/searcher.rs
```diff
@@ -222,17 +218,61 @@ impl Searcher {
...
+            inner: Arc::new(inner),
+        }
+    }
+}
+
+/// Holds a list of `SegmentReader`s ready for search.
+///
+/// It guarantees that the `Segment` will not be removed before
+/// the destruction of the `Searcher`.
+pub struct SearcherInner {
```
fulmicoton: should be private I believe.

src/reader/mod.rs
```diff
@@ -195,41 +204,59 @@ impl InnerIndexReader {
...
+            *old_searcher = searcher;
+        }
+
         Ok(())
     }
 
-    fn searcher(&self) -> LeasedItem<Searcher> {
-        self.searcher_pool.acquire()
+    fn searcher(&self) -> Searcher {
+        self.searcher.lock().unwrap().clone()
```
fulmicoton: The lock could be a problem for high QPS use cases.
(we recommend fetching a new searcher for every request).

Could you check if crossbeam::AtomicCell are lockless and perf well?
Alternatively ArcSwap but i htink the latter is not part of the current deps.

fulmicoton (changes requested):
see comment inline.

src/reader/mod.rs
shikhar: `AtomicCell::load()` only works for `Copy` types, so `ArcSwap` is probably the way to go!

(a naive question from me here: https://github.com/vorner/arc-swap/issues/73)

fulmicoton: Ah ah :) it's a small world.

src/collector/tests.rs
```diff
@@ -263,7 +265,7 @@ impl SegmentCollector for BytesFastFieldSegmentCollector {
     }
 }
 
-fn make_test_searcher() -> crate::Result<Searcher> {
```
fulmicoton: Doesn't Searcher always wraps an Arc?

fulmicoton (commented):
Why did you switch to `Arc<Searcher>`? It has pros and cons but I like the usability of just having a `Searcher` wrapping a private  `Arc<InnerSearcher>`

@PSeitz

PSeitz:
> Why did you switch to `Arc<Searcher>`? It has pros and cons but I like the usability of just having a `Searcher` wrapping a private `Arc<InnerSearcher>`
> 
> @PSeitz

@fulmicoton 
I wasn't sure about that, but there this were the pros/cons I considered:

1. Either we would have a double Arc in `InnerIndexReader`, which is redundant. Or we would have `Arc<InnerSearcher>` in InnerIndexReader and always wrap it into Searcher (probably noop). Having InnerSearcher in InnerIndexReader and Searcher is a little weird.
2. The user knows the cost of cloning in comparison to hiding the Arc in the Searcher.
3. The API change is quite soft, since most APIs accept a `&Searcher`, which works fine with `&Arc<Searcher>`

fulmicoton:
I prefer 1. but I did not understand your last sentence.
> Having InnerSearcher in InnerIndexReader and Searcher is a little weird.

So the arcswap would contain a `Arc<Inner...>` and we would wrap it (no cost) into a Searcher object.

PSeitz:
> I prefer 1. but I did not understand your last sentence.
> 
> > Having InnerSearcher in InnerIndexReader and Searcher is a little weird.
> 
> So the arcswap would contain a `Arc<Inner...>` and we would wrap it (no cost) into a Searcher object.

The Searcher itself has a `Searcher{ inner: Arc<InnerSearcher>}`, but also the InnerIndexReader `InnerIndexReader{ ...inner: Arc<InnerSearcher>}`

fulmicoton (approved):

PR #1414, opened 2022-07-12
Phrase query: scale score on used slop

### LINKED ISSUE #1392: Include slop in the scoring formula for the phrasescorer
Specs are TBD

saroh:
A few points to address:
*  the actual scoring function:
    *  I've gone for ~~sloppy maths~~ a heuristic:
        https://github.com/quickwit-oss/tantivy/pull/1414/files#diff-6778ebdb19b3c516dabd923da7eff56076630ce5e4cd7804f28dc43265751661R139-R149
        the score is scaled by $1 - {\text{used slop} \over (\text{slop}+1)*\text{phrase count}}$ which garantees a scaling factor in $]0;1]$
    *  there are probably a lot of other options. I haven't looked into the BM25 implem for instance. The resulting intersection after `phrase_match` is executed is the vector of matching positions which could also be of use.
* the implem: 
    * the first commit is a quick hack/POC
    * the second one is a *tentative* to have a better implem, It creates a `PhraseScorer` trait as well as two Implems `ExactPhraseScorer` and `SlopPhraseScorer`

closes #1392


src/query/phrase_query/exact_phrase_scorer.rs
```diff
@@ -0,0 +1,291 @@
+use std::cmp::Ordering;
```
saroh, line 1: This file is non slop code that was in we still have refs to slop here but they should be gone :)

src/query/phrase_query/exact_phrase_scorer.rs
```diff
@@ -0,0 +1,291 @@
...
+    intersection_docset: Intersection<PostingsWithOffset<TPostings>, PostingsWithOffset<TPostings>>,
+    num_terms: usize,
+    left: Vec<u32>,
+    right: Vec<u32>,
+    phrase_count: u32,
+    fieldnorm_reader: FieldNormReader,
+    similarity_weight: Bm25Weight,
+    scoring_enabled: bool,
+    slop: u32,
+    used_slop: u32,
```
saroh, line 21: slop specific attributes ;) This WIP commit would make us get rid of these in case of exact phrase scoring + get rid of the `fn has_slop()` .

src/query/phrase_query/exact_phrase_scorer.rs
```diff
@@ -0,0 +1,291 @@
...
+            .positions(&mut self.right);
+        intersection_len
+    }
+
+    fn has_slop(&self) -> bool {
+        self.slop > 0
+    }
+}
+
+impl<TPostings: Postings> DocSet for ExactPhraseScorer<TPostings> {
```
saroh, line 113: This `Docset` could be "blanket implemented" for all `PhraseScorer` I guess ?

src/query/phrase_query/phrase_weight.rs
```diff
@@ -72,13 +73,53 @@ impl PhraseWeight {
...
             self.slop,
-        )))
+        ))))
+    }
+
+    fn phrase_scorer_as_scorer(
+        &self,
+        reader: &SegmentReader,
+        boost: Score,
+    ) -> crate::Result<Option<Box<dyn Scorer>>> {
```
saroh, line 89: This is a copy/paste of `fn phrase_scorer`, juste the returned boxed dyn trait changes.

src/query/phrase_query/phrase_weight.rs
```diff
@@ -72,13 +73,53 @@ impl PhraseWeight {
...
+                    .inverted_index(term.field())?
+                    .read_postings_no_deletes(term, IndexRecordOption::WithFreqsAndPositions)?
+                {
+                    term_postings_list.push((offset, postings));
+                } else {
+                    return Ok(None);
+                }
+            }
+        }
+        Ok(Some(Box::new(ExactPhraseScorer::new(
```
saroh, line 116: Obviously here you'd have if self.slop > 0 -> ExactPhraseScorer, else SlopPhraseScorer

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -244,242 +171,32 @@ fn intersection_exists_with_slop(left: &[u32], right: &[u32], slop: u32) -> bool
...
-                .positions(&mut self.left);
-        }
-        let mut intersection_len = self.left.len();
-        for i in 1..self.num_terms - 1 {
-            {
-                self.intersection_docset
-                    .docset_mut_specialized(i)
-                    .positions(&mut self.right);
-            }
-            intersection_len = if self.has_slop() {
```
saroh, line 335: Second commit is here to remove these `if self.has_slop` + the unnecessary slop attributes in `ExactPhraseScorer`. It seems like a lot of code to do so though, the way I've done it at least ;)

src/query/phrase_query/exact_phrase_scorer.rs
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    fn doc(&self) -> DocId {
+        self.intersection_docset.doc()
+    }
+
+    fn size_hint(&self) -> u32 {
+        self.intersection_docset.size_hint()
+    }
+}
```
saroh, line 120: Both structs have the exact same impl for `DocSet` but there's allready a blanket implem for `DocSet`. Unsure how to workaround this (I could create common functions but it seems a bit overkill).

src/query/phrase_query/phrase_weight.rs
```diff
@@ -45,10 +46,66 @@ impl PhraseWeight {
...
+                        similarity_weight,
+                        fieldnorm_reader,
+                        self.scoring_enabled,
+                        self.slop,
+                    ))))
+                }
+            }
+            _ => Ok(None),
+        }
+    }
```
saroh: I did not find an elegant way to factorize this code which only differs on the return type.

src/query/phrase_query/slop_phrase_scorer.rs
```diff
@@ -0,0 +1,340 @@
...
+#[cfg(test)]
+mod tests {
+    use super::{intersection_count_with_slop, intersection_with_slop};
+
+    fn test_intersection_aux(left: &[u32], right: &[u32], expected: &[u32], slop: u32) {
+        let mut left_vec = Vec::from(left);
+        let left_mut = &mut left_vec[..];
+        let mut right_vec = Vec::from(right);
+        let right_mut = &mut right_vec[..];
+        // TODO test count that wasn't done previously :)
```
saroh: also both mods are missing tests on `intersection_exists` if I'm not mistaken.

src/query/phrase_query/phrase_weight.rs
```diff
@@ -88,7 +143,7 @@ impl PhraseWeight {
 
 impl Weight for PhraseWeight {
     fn scorer(&self, reader: &SegmentReader, boost: Score) -> crate::Result<Box<dyn Scorer>> {
-        if let Some(scorer) = self.phrase_scorer(reader, boost)? {
+        if let Some(scorer) = self.phrase_scorer_as_scorer(reader, boost)? {
             Ok(Box::new(scorer))
```
fmassot: You have already a box here, so you can only return `Ok(scorer)`

src/query/phrase_query/phrase_weight.rs
```diff
@@ -88,7 +143,7 @@ impl PhraseWeight {
 
 impl Weight for PhraseWeight {
     fn scorer(&self, reader: &SegmentReader, boost: Score) -> crate::Result<Box<dyn Scorer>> {
```
fmassot, line 117: why not having only this `scorer` function and get rid of `phrase_scorer_as_scorer`?

```rust
fn scorer(&self, reader: &SegmentReader, boost: Score) -> crate::Result<Box<dyn Scorer>> {
    match self.get_scorer_params(reader, boost)? {
        Some((term_postings_list, similarity_weight, fieldnorm_reader)) => {
            if self.slop == 0 {
                Ok(Box::new(ExactPhraseScorer::new(
                    term_postings_list,
                    similarity_weight,
                    fieldnorm_reader,
                    self.scoring_enabled,
                )))
            } else {
                Ok(Box::new(SlopPhraseScorer::new(
                    term_postings_list,
                    similarity_weight,
                    fieldnorm_reader,
                    self.scoring_enabled,
                    self.slop,
                )))
            }
        }
        _ => Ok(Box::new(EmptyScorer)),
    }
}
```

src/query/phrase_query/exact_phrase_scorer.rs
```diff
@@ -0,0 +1,259 @@
...
+
+pub struct ExactPhraseScorer<TPostings: Postings> {
+    intersection_docset: Intersection<PostingsWithOffset<TPostings>, PostingsWithOffset<TPostings>>,
+    num_terms: usize,
+    left: Vec<u32>,
+    right: Vec<u32>,
+    phrase_count: u32,
+    fieldnorm_reader: FieldNormReader,
+    similarity_weight: Bm25Weight,
+    scoring_enabled: bool,
```
fmassot, line 20: using the name `scoring_enabled` seems weird as it suggests we can skip the scorer and thus use the empty scorer instead.

fmassot:
You may have already read it, just in case, here is one of the lucene implementation: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/SloppyPhraseMatcher.java

src/query/phrase_query/exact_phrase_scorer.rs
saroh: It's an optim used by the `Query` trait https://github.com/saroh/tantivy/blob/1392-phrase-slop-scoring/src/query/query.rs#L47-L48
Here it allows us to return fast on successful match for our last iteration in the phrase scorer's case. If I inderstand correctly in this case the `BM25Weight` receives a `tf` of 0. So there will be no ranking nor count but docs will still match. Where as `EmptyScorer` will make no docs match apparently.
So we still score but I'd say all documents will have a zero score/match count.

saroh:
> You may have already read it, just in case, here is one of the lucene implementation: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/SloppyPhraseMatcher.java

I gave a quick look.
We're doing things differently it seems cf their doc: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/SloppyPhraseMatcher.java#L36-L52. 
They seem to score on the inverse of largest phrase match size. I haven't looked enough in the implem to 100% confirm how their Matcher works but it does not look like they take into account multiple matches givent the comments? 

In our case because we count multiple matches, we could score on $\text{phrase count}\over\text{match length}$.
I think we should be able to modify the implem to do so-> quickly put $\text{match length} = \text{phrase count} * \text{num terms} + \text{used slop}$.

src/query/phrase_query/exact_phrase_scorer.rs
saroh: added tests to reflect this https://github.com/quickwit-oss/tantivy/pull/1414/files#diff-b46d170c28bc9075958011ee0f1c172d00e2a39169d470250f10151ba7588d3bR190-R263 > `_no_scoring()` tests

src/query/phrase_query/mod.rs
```diff
@@ -1,10 +1,14 @@
+mod exact_phrase_scorer;
 mod phrase_query;
 mod phrase_scorer;
 mod phrase_weight;
+mod slop_phrase_scorer;
 
+pub use self::exact_phrase_scorer::ExactPhraseScorer;
 pub use self::phrase_query::PhraseQuery;
 pub use self::phrase_scorer::PhraseScorer;
```
saroh, line 9: Maybe remove this as it's a trait now ?

saroh:
@fulmicoton little ping on this one 🙏

src/query/phrase_query/phrase_weight.rs
```diff
@@ -88,10 +115,26 @@ impl PhraseWeight {
...
+                } else {
+                    Ok(Box::new(SlopPhraseScorer::new(
+                        term_postings_list,
+                        similarity_weight,
+                        fieldnorm_reader,
+                        self.scoring_enabled,
+                        self.slop,
+                    )))
+                }
+            }
```
saroh, line 136: duplicates https://github.com/quickwit-oss/tantivy/pull/1414/files#diff-b46d170c28bc9075958011ee0f1c172d00e2a39169d470250f10151ba7588d3bR50-R67 which is in teh same file but we need to return a different Boxed trait

PR #1415, opened 2022-07-14
FileHandle: Change from boxed to Arc.

pier-oliviert:
Changing from a Box<dyn FileHandle> to an Arc<dyn FileHandle> would
allow for a user of tantivy to manage file handles outside of tantivy
and be able to manage their life cycle.

I'm currently working with tantivy and one of the thing I wanted to do was to manage file access outside of tantivy in a way that file access is abstracted completely from tantivy. I got everything working on my end, but I needed to make this change to tantivy because I needed shared ownership of a given FileHandle.

I believe this change is safe to other user of tantivy, but I am new to both Tantivy & Rust, so maybe there's something else that needs to change to make this happen.

### DIFF
--- a/src/directory/directory.rs
+++ b/src/directory/directory.rs
@@ -111,7 +111,7 @@ pub trait Directory: DirectoryClone + fmt::Debug + Send + Sync + 'static {
     ///
     /// Users of `Directory` should typically call `Directory::open_read(...)`,
     /// while `Directory` implementor should implement `get_file_handle()`.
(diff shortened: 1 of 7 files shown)


fulmicoton (approved):

fulmicoton:
@pier-oliviert Your formatting settings seems incorrect.

`make fmt` should run the right settings.

pier-oliviert:
@fulmicoton just force-pushed what I hope is the final fix for the linter. Let me know if you need anything else from me.

PR #1418, opened 2022-07-20
apply gcd on fastfield as preprocessing

PSeitz:


fastfield_codecs/src/bitpacked.rs
```diff
@@ -105,9 +105,9 @@ impl FastFieldCodecSerializer for BitpackedFastFieldSerializer {
     /// It requires a `min_value` and a `max_value` to compute
     /// compute the minimum number of bits required to encode
     /// values.
-    fn serialize(
-        write: &mut impl Write,
-        _fastfield_accessor: &impl FastFieldDataAccess,
+    fn serialize<W: Write>(
```
fulmicoton: how is it different?

fastfield_codecs/src/lib.rs
```diff
@@ -40,13 +40,15 @@ pub trait FastFieldCodecSerializer {
     /// Serializes the data using the serializer into write.
     /// There are multiple iterators, in case the codec needs to read the data multiple times.
     /// The iterators should be preferred over using fastfield_accessor for performance reasons.
-    fn serialize(
-        write: &mut impl Write,
-        fastfield_accessor: &impl FastFieldDataAccess,
+    fn serialize<W>(
+        write: &mut W,
```
fulmicoton: how is it different? You use it with different W?

src/fastfield/bytes/mod.rs
```diff
@@ -11,7 +11,7 @@ mod tests {
     use crate::{DocAddress, DocSet, Index, Searcher, Term};
 
     #[test]
-    fn test_bytes() -> crate::Result<()> {
+    fn test_bytes2() -> crate::Result<()> {
```
fulmicoton: ?

src/fastfield/mod.rs
```diff
@@ -505,10 +513,15 @@ mod tests {
         permutation
     }
 
-    #[test]
-    fn test_intfastfield_permutation() -> crate::Result<()> {
+    // Warning: this generates the same permutation at each call
+    pub fn generate_permutation_gcd() -> Vec<u64> {
```
fulmicoton, line 530: does this need to be public?

src/fastfield/mod.rs
```diff
@@ -527,15 +540,27 @@ mod tests {
             let data = fast_fields_composite.open_read(*FIELD).unwrap();
             let fast_field_reader = DynamicFastFieldReader::<u64>::open(data)?;
 
-            let mut a = 0u64;
-            for _ in 0..n {
+            for a in 0..n {
                 assert_eq!(fast_field_reader.get(a as u32), permutation[a as usize]);
-                a = fast_field_reader.get(a as u32);
```
fulmicoton, line 533: yeah that's dumb. I suspect it was copy pasted from something in a bunch, or that it was trying to test that accessing in any order was working.

src/fastfield/serializer/mod.rs
```diff
@@ -33,8 +36,45 @@ use crate::schema::Field;
...
 
+impl From<FastFieldCodecName> for FastFieldCodecEnableCheck {
+    fn from(codec_name: FastFieldCodecName) -> Self {
+        FastFieldCodecEnableCheck {
+            enabled_codecs: vec![codec_name],
+        }
+    }
+}
+
+pub const FF_HEADER_MAGIC_NUMBER: u8 = 123u8;
```
fulmicoton: why 123? (as opposed to 255)

src/fastfield/reader.rs
```diff
@@ -61,6 +62,34 @@ pub trait FastFieldReader<Item: FastValue>: Clone {
...
+        let min_value = bytes.read_u64();
+        FFHeader {
+            field_id,
+            gcd,
+            min_value,
+        }
+    } else {
+        // old version
+        FFHeader {
+            field_id: magic_number_or_field_id,
```
fulmicoton: wait so today we cannot have more than 256 fields!?

src/fastfield/reader.rs
```diff
@@ -61,6 +62,34 @@ pub trait FastFieldReader<Item: FastValue>: Clone {
...
+            gcd: 1,
+            min_value: 0,
```
fulmicoton: can we make that header version so it does not byte us again?

src/fastfield/reader.rs
```diff
@@ -75,29 +104,35 @@ pub enum DynamicFastFieldReader<Item: FastValue> {
...
 impl<Item: FastValue> DynamicFastFieldReader<Item> {
     /// Returns correct the reader wrapped in the `DynamicFastFieldReader` enum for the data.
-    pub fn open(file: FileSlice) -> crate::Result<DynamicFastFieldReader<Item>> {
-        let mut bytes = file.read_bytes()?;
-        let id = bytes.read_u8();
-
+    pub fn open_from_id(
+        bytes: OwnedBytes,
+        id: u8,
+        gcd: u64,
```
fulmicoton: hmmm so you choose to make gcd somewhat special... Could it have been just another Codec? (well wrapping another codec)

src/fastfield/reader.rs
```diff
@@ -238,7 +290,7 @@ impl<Item: FastValue, C: FastFieldCodecReader + Clone> FastFieldReader<Item>
     /// deleted document, and should be considered as an upper bound
     /// of the actual maximum value.
     fn min_value(&self) -> Item {
-        Item::from_u64(self.reader.min_value())
+        Item::from_u64(self.reader.min_value() * self.gcd + self.min_value)
```
fulmicoton, line 293: Hmmm... Interesting.

So you went for 
` gcd * (min_value + encoded_val)`.

At the same cost, `min_value + gcd * encoded_val` was slightly more general.

(it does not matter though)

fulmicoton:
```suggestion
        Item::from_u64(self.min_value + self.reader.min_value() * self.gcd)
```

src/fastfield/reader.rs
```diff
@@ -247,7 +299,7 @@ impl<Item: FastValue, C: FastFieldCodecReader + Clone> FastFieldReader<Item>
     /// deleted document, and should be considered as an upper bound
     /// of the actual maximum value.
     fn max_value(&self) -> Item {
-        Item::from_u64(self.reader.max_value())
+        Item::from_u64((self.reader.max_value() * self.gcd) + self.min_value)
```
fulmicoton:
```suggestion
        Item::from_u64(self.min_value + self.reader.max_value() * self.gcd)
```

src/fastfield/serializer/gcd.rs
```diff
@@ -0,0 +1,226 @@
...
+use gcd::Gcd;
+use libdivide::Divider;
+
+pub const GCD_DEFAULT: u64 = 1;
+
+fn compute_gcd(vals: &[u64], base: u64) -> u64 {
+    let mut gcd = (vals[0] - base).gcd(vals[1] - base);
+
+    for el in vals.iter().map(|el| el - base) {
+        gcd = gcd.gcd(el);
```
fulmicoton: TIL about the binary GCD algorithm

fulmicoton:
```rust
if gcd == 1 {
    return 1;
}
```

fulmicoton: Is it something we might want to make faster?

In addition to the short cut above, another easy win is:
```
if el % gcd == 0 { //< that one should rely on libdivide.
    continue;
}
```

With that gcd should be called only a couple of times...

src/fastfield/serializer/gcd.rs
```diff
@@ -0,0 +1,226 @@
...
+    sample_values.push(stats.max_value);
+    sample_values
+}
+
+pub(crate) fn find_gcd_from_samples(
+    samples: &[u64],
+    vals: impl Iterator<Item = u64>,
+    base: u64,
+) -> Option<u64> {
+    let estimate_gcd = compute_gcd(samples, base);
```
fulmicoton: ah right so you do it here... my bad.

I think you can fuse `compute_gcd` and `is_valid_gcd` in one loop. I know I mentionned sampling, but it does not make sense here.

src/fastfield/serializer/gcd.rs
```diff
@@ -0,0 +1,226 @@
...
+        None
+    }
+}
+
+pub(crate) fn find_gcd(
+    fastfield_accessor: &impl FastFieldDataAccess,
+    stats: FastFieldStats,
+    vals: impl Iterator<Item = u64>,
+) -> Option<u64> {
+    if stats.num_vals == 0 {
```
fulmicoton: In other words, I think we can get away with simply.

```rust
fn euclid_gcd(mut left: u64, mut right: u64) -> u64 {
    while right != 0 {
        (left, right) = (right, left % right);
    }
    left
}

fn find_gcd(mut numbers: impl Iterator<Item=u64>) -> Option<u64> {
    let mut gcd = numbers.next()?;
    let mut gcd_divider = Divider::new(gcd);
    for el in numbers {
        if gcd == 1 {
            return 1;
        }
        // Ideally remainder would be ...
        // https://lemire.me/blog/2019/02/08/faster-remainders-when-the-divisor-is-a-constant-beating-compilers-and-libdivide/
        // But we don't have that in rust. libdivide will do.
        let rem = gcd_divider.remainder(gcd);
        if rem == 0 {
            continue;
        }
        gcd = compute_gcd(gcd, rem);
        gcd_divider = Divider::new(gcd);
    }
    Some(gcd)
}
```

src/fastfield/reader.rs
PSeitz: terrible naming, it should be `magic_number_or_codec_id`, I think we use u16 for field ids

src/fastfield/reader.rs
PSeitz: Yes, currently it's its own layer, e.g.

Create: i64 -> convert to u64 -> find/apply gcd -> detect/serialize codec
Read: read u64 from codec -> apply gcd -> convert to i64

When we move it into the codec, every codec needs to be gcd aware.

On the read side, the codecs are already wrapped in `FastFieldReaderCodecWrapper`. We could have a separate wrapper like `FastFieldReaderCodecGCDWrapper`, but that would be a performance optimization.

src/fastfield/serializer/mod.rs
PSeitz: I think that 123 is less probable on data/memory corruption

src/fastfield/mod.rs
PSeitz: yes, but it's only `#[cfg(test)]` code

fastfield_codecs/src/lib.rs
PSeitz: Initially I wanted to reduce the number of generic parameters by moving the two `impl Iterator` into one generic, but that wasn't possible, because two closures are always different types, even when identical. The `W` is a residue of that refactor I forgot to rollback.

src/fastfield/reader.rs
PSeitz: I don't understand, we have now a header version in the metadata that allow us to parse different header versions, but it's unused currently since there's only one header version.

fastfield_codecs/src/multilinearinterpol.rs
```diff
@@ -75,6 +75,7 @@ impl BinarySerializable for Function {
         self.positive_val_offset.serialize(write)?;
         self.slope.serialize(write)?;
         self.num_bits.serialize(write)?;
+
```
fulmicoton:
```suggestion
```

src/fastfield/mod.rs
```diff
@@ -276,10 +276,17 @@ mod tests {
...
+        schema_builder.add_i64_field("field", FAST);
+        schema_builder.build()
+    });
+
     pub static FIELD: Lazy<Field> = Lazy::new(|| SCHEMA.get_field("field").unwrap());
+    pub static FIELDI64: Lazy<Field> = Lazy::new(|| SCHEMAI64.get_field("field").unwrap());
 
     #[test]
-    pub fn test_fastfield() {
+    pub fn test_fastfield2() {
```
fulmicoton: why the rename?

src/fastfield/reader.rs
fulmicoton: Ah that's the `header_version`? Ok good.

src/fastfield/serializer/gcd.rs
```diff
@@ -0,0 +1,176 @@
+use fastdivide::DividerU64;
+use gcd::Gcd;
+
+pub const GCD_DEFAULT: u64 = 1;
+
+pub fn find_gcd(mut numbers: impl Iterator<Item = u64>, base: u64) -> Option<u64> {
```
fulmicoton: Can we remove `base` from here and put that in the caller?

The contract of this function would make more sense.

src/fastfield/serializer/gcd.rs
```diff
@@ -0,0 +1,176 @@
...
+    let mut num1 = 0;
+    let mut num2 = 0;
+    loop {
+        let num = numbers.next()? - base;
```
fulmicoton: what is this loop doing? is it picking the two first non null integers?

src/fastfield/serializer/gcd.rs
```diff
@@ -0,0 +1,176 @@
...
+            num1 = num;
+        }
+        if num2 == 0 {
+            num2 = num;
+        }
+        if num1 != 0 && num2 != 0 {
+            break;
+        }
+    }
+    let mut gcd = (num1).gcd(num2);
```
fulmicoton: I'm not sure what the gcd crate returns on value 0, but we want `gcd(0, n) = n`

src/fastfield/serializer/gcd.rs
```diff
@@ -0,0 +1,176 @@
...
+    if gcd == 0 {
+        return None;
```
fulmicoton: So this branch happens if all numbers are zero?

What about we simply start the function by 

```
let mut non_null_numbers = numbers.filter(|n| *n != 0);
```

and remove 12 lines of code?
```

File: src/fastfield/serializer/mod.rs [hunk 21]
```diff
@@ -52,60 +92,130 @@ fn codec_estimation<T: FastFieldCodecSerializer, A: FastFieldDataAccess>(
...
+            min_value: u64,
+            gcd: u64,
+        }
+        impl<T: FastFieldDataAccess> FastFieldDataAccess for WrappedFFAccess<T> {
+            fn get_val(&self, position: u64) -> u64 {
+                (self.fastfield_accessor.get_val(position) - self.min_value) / self.gcd
+            }
+        }
+        // let gcd = find_gcd(&fastfield_accessor, stats.clone(),
+        // iter_gen()).unwrap_or(GCD_DEFAULT);
```
fulmicoton:
```suggestion
```

src/fastfield/serializer/gcd.rs
PSeitz: yep I changed it

src/fastfield/serializer/gcd.rs
PSeitz: yep

src/fastfield/serializer/gcd.rs
PSeitz: yes that's better

src/fastfield/serializer/gcd.rs
PSeitz: with the 0-filter this case is covered

src/fastfield/serializer/gcd.rs
```diff
@@ -0,0 +1,160 @@
+use fastdivide::DividerU64;
+use gcd::Gcd;
+
+pub const GCD_DEFAULT: u64 = 1;
+
+// Find GCD for iterator of numbers
+pub fn find_gcd(numbers: impl Iterator<Item = u64>) -> Option<u64> {
+    let mut numbers = numbers.filter(|n| *n != 0);
+    let mut gcd = (numbers.next()?).gcd(numbers.next()?);
```
fulmicoton: `gcd = numbers.next()?;`

is sufficient

actually the current version is incorrect on `0, 10` for instance.

src/fastfield/serializer/gcd.rs
```diff
@@ -0,0 +1,160 @@
...
+    if gcd == 1 {
```
fulmicoton: if you fix the above you can remove this 3 lines.

src/fastfield/serializer/gcd.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    let mut gcd_divider = DividerU64::divide_by(gcd);
+    for val in numbers {
+        let remainder = val - (gcd_divider.divide(val)) * gcd;
+        if remainder == 0 {
+            continue;
+        }
+        gcd = gcd.gcd(val);
+        if gcd == 1 {
+            return None;
```
fulmicoton: ```return Some(1);```

I think the `Option<_>` is mostly nice to return None when there are no element. L/ike min for instance.

src/fastfield/serializer/gcd.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        gcd_divider = DividerU64::divide_by(gcd);
+    }
+    Some(gcd)
+}
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashMap;
+    use std::path::Path;
```
fulmicoton: add unit test for gcd....

in particular, what happens with 

```
[0] -> Some(1)
[0, 10]-> Some(10)
[10, 0]-> Some(10)
[] -> None
[15,30,5,10] -> 5
[15,16,10] -> 1
[0,5,5,5] -> 5
```

src/fastfield/reader.rs
fulmicoton: No I was suggesting moving it as one codec.
```
- codec1
- codec2
- gcd<C: Codec>
```
I assumed you would do this because it seemed like the natural solution to have: 
- one dynamic dispatch 
as opposed to 
- having two (dynamic dispatch or if statement).

fulmicoton: on the read path that is.

src/fastfield/reader.rs
```diff
@@ -158,27 +201,35 @@ impl<Item: FastValue, C: FastFieldCodecReader> FastFieldReaderCodecWrapper<Item,
...
             bytes,
             _phantom: PhantomData,
         })
     }
     #[inline]
     pub(crate) fn get_u64(&self, doc: u64) -> Item {
-        Item::from_u64(self.reader.get_u64(doc, self.bytes.as_slice()))
+        let mut data = self.reader.get_u64(doc, self.bytes.as_slice());
+        if self.gcd != 1 {
+            data *= self.gcd;
```
fulmicoton: yeah so GCD<Codec> as a codec remove this if statement, and avoid making GCD too special.

src/fastfield/serializer/gcd.rs
PSeitz: I would keep the `gcd == 1` check, or we scan the whole fastfield when it begins with 1

fulmicoton: the same test exists in the loop.

PSeitz: yes, but it's after the remainder check. we can have it at the beginning, but then we would check it more often than needed

src/fastfield/gcd.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    fn test_fastfield_gcd_i64_with_codec(
+        codec_name: FastFieldCodecName,
+        num_vals: usize,
+    ) -> crate::Result<()> {
+        let path = Path::new("test");
+        let mut docs = vec![];
+        for i in 1..=num_vals {
+            let val = i as i64 * 1000i64;
+            docs.push(doc!(*FIELDI64=>val));
```
fulmicoton, line 130: great you had a i64 test.

src/fastfield/serializer/mod.rs
```diff
@@ -52,60 +79,154 @@ fn codec_estimation<T: FastFieldCodecSerializer, A: FastFieldDataAccess>(
...
+    ) -> io::Result<()>
+    where
+        F: Fn() -> I,
+        I: Iterator<Item = u64>,
+    {
         let field_write = self.composite_write.for_field_with_idx(field, idx);
+        let gcd = find_gcd(iter_gen().map(|val| val - stats.min_value)).unwrap_or(GCD_DEFAULT);
+
+        if gcd == 1 {
+            Self::create_auto_detect_u64_fast_field_with_idx_gcd(
```
fulmicoton: early return.

src/fastfield/serializer/mod.rs
```diff
@@ -52,60 +79,154 @@ fn codec_estimation<T: FastFieldCodecSerializer, A: FastFieldDataAccess>(
...
+
+    /// Serialize data into a new u64 fast field. The best compression codec will be chosen
+    /// automatically.
+    pub fn create_auto_detect_u64_fast_field_with_idx_gcd<W: Write>(
+        codec_enable_checker: FastFieldCodecEnableCheck,
+        field: Field,
+        field_write: &mut CountingWriter<W>,
+        stats: FastFieldStats,
+        fastfield_accessor: impl FastFieldDataAccess,
+        iter1: impl Iterator<Item = u64>,
```
fulmicoton, line 204: can you think of a way to ditch iter1/iter2 stuff as well as the Fn() -> impl Iterator stuff? (can be done in a separate ticket)

Would 

`impl IntoIterator for &SomeType`

work? (the reference is Copy)...
or maybe can We afford materializing a Vec?

fulmicoton (approved):

PR #1420, opened 2022-07-21
Replace deprecated std package's constants on floats and integers

k-yomo:
We can use associated constants on floats and integers directly from rust [1.4.3](https://github.com/rust-lang/rust/blob/master/RELEASES.md#libraries-19), and imported constants are deprecated now.

This PR replaces those deprecated constants.


fulmicoton (approved):
thank you

PR #1424, opened 2022-07-25
Add support for keyed parameter in range and histgram aggregations

### LINKED ISSUE #1336: Aggregation: Support `keyed` parameter
Currently the `keyed` parameter in the aggregations is not yet supported, but it's part of elasticsearch API.
E.g. in the `Range` aggregation:
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-range-aggregation.html#_keyed_response_4

k-yomo:
Resolves #1336 

This PR adds support for `keyed` parameter in range and histogram aggregations in the same way elasticsearch does.


src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -117,6 +117,9 @@ pub struct HistogramAggregation {
     /// Cannot be set in conjunction with min_doc_count > 0, since the empty buckets from extended
     /// bounds would not be returned.
     pub extended_bounds: Option<HistogramBounds>,
+    /// Whether to return the buckets as a hash map
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub keyed: Option<bool>,
```
PSeitz: why `Option<bool>` instead of bool

src/aggregation/bucket/range.rs
```diff
@@ -51,13 +49,16 @@ use crate::{DocId, TantivyError};
...
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
 pub struct RangeAggregation {
     /// The field to aggregate on.
     pub field: String,
     /// Note that this aggregation includes the from value and excludes the to value for each
     /// range. Extra buckets will be created until the first to, and last from, if necessary.
     pub ranges: Vec<RangeAggregationRange>,
+    /// Whether to return the buckets as a hash map
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub keyed: Option<bool>,
```
PSeitz: why `Option<bool>,` instead of `bool`

src/aggregation/bucket/histogram/histogram.rs
k-yomo: Used Option following the other optional params and to make it easier to omit by `Default::default()` but no strong reason.
Should we use `bool`?

src/aggregation/agg_result.rs
```diff
@@ -137,6 +137,17 @@ impl BucketResult {
...
 
+/// This is the wrapper of buckets entries, which can be vector or hashmap
+/// depending on if it's keyed or not.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum BucketEntries<T> {
+    /// Vector format bucket entries
+    Vec(Vec<T>),
+    /// HashMap format bucket entries
+    HashMap(HashMap<String, T>),
```
PSeitz: FnvHashMap would be better here imo

src/aggregation/intermediate_agg_result.rs
```diff
@@ -281,6 +281,21 @@ impl IntermediateBucketResult {
...
                         .total_cmp(&right.from.unwrap_or(f64::MIN))
                 });
+
+                let is_keyed = req
+                    .as_range()
+                    .expect("unexpected aggregation, expected range aggregation")
+                    .keyed
+                    .is_some();
+                let buckets = if is_keyed {
+                    let mut bucket_map = HashMap::new();
```
PSeitz: we can use with_capacity here

PSeitz:
Awesome thank you! 
One thing that is missing are custom keys in ranges. Can be done in this PR or we create a separate issue for that.

```json
GET sales/_search
{
  "aggs": {
    "price_ranges": {
      "range": {
        "field": "price",
        "keyed": true,
        "ranges": [
          { "key": "cheap", "to": 100 },
          { "key": "average", "from": 100, "to": 200 },
          { "key": "expensive", "from": 200 }
        ]
      }
    }
  }
}

```

src/aggregation/bucket/histogram/histogram.rs
PSeitz: `Default::default()` should also work on bool (defaults to false). Since `None` equals `false`, I think it's better to use bool directly

k-yomo:
@PSeitz 
Thanks for the review!!  I'll add custom keys support in this PR🙏

src/aggregation/bucket/histogram/histogram.rs
k-yomo: Oh I see, didn't know that...!  Let me change those to `bool`!

k-yomo:
@PSeitz 
Oh sorry, I thought custom keys are already there and not working with `keyed` but it's actually not implemented yet? 

If so, it would be great if we could have a separate issue🙏
I would be happy to work on it though!

src/aggregation/bucket/range.rs
k-yomo: fixed in https://github.com/quickwit-oss/tantivy/pull/1424/commits/5ab5f070edb5ae7e2f00daeb626e517e8290d83a

src/aggregation/bucket/histogram/histogram.rs
k-yomo: fixed in https://github.com/quickwit-oss/tantivy/pull/1424/commits/5ab5f070edb5ae7e2f00daeb626e517e8290d83a

src/aggregation/agg_result.rs
k-yomo: I see!  updated! https://github.com/quickwit-oss/tantivy/pull/1424/commits/80a1418284b140a833fe35aba688d0b14197a9b2

src/aggregation/intermediate_agg_result.rs
k-yomo: fixed! https://github.com/quickwit-oss/tantivy/commit/80a1418284b140a833fe35aba688d0b14197a9b2

PSeitz:
> @PSeitz Oh sorry, I thought custom keys are already there and not working with `keyed` but it's actually not implemented yet?
> 
> If so, it would be great if we could have a separate issue I would be happy to work on it though!

Okay cool, I created an issue for that here: https://github.com/quickwit-oss/tantivy/issues/1425

src/aggregation/agg_req.rs
```diff
@@ -36,7 +37,8 @@
 //!       "ranges": [
 //!         { "from": 3.0, "to": 7.0 },
 //!         { "from": 7.0, "to": 20.0 }
-//!       ]
+//!       ],
+//!       "keyed": false
```
PSeitz: this is the default, I think we can just omit it

src/aggregation/mod.rs
```diff
@@ -881,7 +891,8 @@ mod tests {
         { "from": 7.0, "to": 19.0 },
         { "from": 19.0, "to": 20.0 },
         { "from": 20.0 }
-      ]
+      ],
+      "keyed": false
```
PSeitz: do we have to add this? the default behaviour should be the same

src/aggregation/agg_req.rs
k-yomo: It didn't pass the tests...
https://github.com/quickwit-oss/tantivy/runs/7517376126?check_suite_focus=true

Ah I see, I removed `serde(default)` by mistake, let me fix

k-yomo: Ah I okey, it seems I just need to use `#[serde(default)]`🙏

k-yomo: fixed in https://github.com/quickwit-oss/tantivy/pull/1424/commits/6444516a8252f7dba9602d276f9c67b955e1200a

src/aggregation/mod.rs
```diff
@@ -132,6 +132,7 @@
 //!             bucket_agg: BucketAggregationType::Range(RangeAggregation{
 //!                 field: "score".to_string(),
 //!                 ranges: vec![(3f64..7f64).into(), (7f64..20f64).into()],
+//!                 keyed: false,
```
PSeitz, line 135:
```suggestion
```

src/aggregation/mod.rs
```diff
@@ -518,12 +521,14 @@ mod tests {
...
                 "offset": 3.0,
+                "keyed": false
             },
             "aggs": {
                 "bucketsL2": {
                     "histogram": {
                         "field": "score",
-                        "interval":  70.0
+                        "interval":  70.0,
+                        "keyed": false
```
PSeitz:
```suggestion
```

src/aggregation/mod.rs
```diff
@@ -536,7 +541,8 @@ mod tests {
...
```
PSeitz:
```suggestion
```

src/aggregation/mod.rs
k-yomo: @PSeitz 
Could we do that?
Just removing the field will cause compiler error.
```
missing field `keyed` in initializer of `RangeAggregation`
```

src/aggregation/mod.rs
k-yomo: Oh sorry missed those, fixed in https://github.com/quickwit-oss/tantivy/pull/1424/commits/9b6b60cc2bd66c825fc8cbcc01851cba01601027

src/aggregation/mod.rs
PSeitz: ah okay there's no `..Default::default`, we can keep it here

PSeitz (approved):

PR #1427, opened 2022-07-28
handle empty segments for merge

### LINKED ISSUE #1189: Investigate empty segment merge
An issue occurred where merge after commit selected only empty segments.  

```
thread 'merge_thread1' panicked at 'Unexpected error, empty readers in IndexMerger', src/indexer/merger.rs:330:16
thread 'merge_thread1' panicked at 'You forgot to flush "00004481000000000000000000000000.term" before its writter got Drop. Do not rely on drop. This also occurs when the indexer crashed, so you may want to
 check the logs for the root cause.', src/directory/ram_directory.rs:49:13
stack backtrace:
```

Investigate if checking for non-empty segments before triggering a merge is sufficient. There are also different possible scenarios, when we end up with an empty segment:

- fresh new segments ends up with no docs after processing the deletes
- old segments ends up with no docs after processing deletes
- merged segment ends up with no docs after processing the deletes

**To Reproduce**

Change `test_functional_indexing_sorted` to run with 15threads

`let mut index_writer = index.writer_with_num_threads(15, 150_000_000)?;`

Maybe run in a loop or multiple times.

PSeitz:
Optionally create a segment on merge
Create no segment when all segments to merge are empty

fixes #1189

### DIFF
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -510,10 +510,12 @@ impl IndexWriter {
         Ok(self.committed_opstamp)
     }
 
-    /// Merges a given list of segments
+    /// Merges a given list of segments.
+    ///
+    /// If all segments are empty no new segment will be created.
     ///
     /// `segment_ids` is required to be non-empty.
-    pub fn merge(&mut self, segment_ids: &[SegmentId]) -> FutureResult<()> {
+    pub fn merge(&mut self, segment_ids: &[SegmentId]) -> FutureResult<Option<SegmentMeta>> {
         let merge_operation = self.segment_updater.make_merge_operation(segment_ids);
         let segment_updater = self.segment_updater.clone();
         segment_updater.start_merge(merge_operation)
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -513,7 +513,7 @@ impl IndexWriter {
     /// Merges a given list of segments
     ///
     /// `segment_ids` is required to be non-empty.
-    pub fn merge(&mut self, segment_ids: &[SegmentId]) -> FutureResult<SegmentMeta> {
+    pub fn merge(&mut self, segment_ids: &[SegmentId]) -> FutureResult<()> {
         let merge_operation = self.segment_updater.make_merge_operation(segment_ids);
         let segment_updater = self.segment_updater.clone();
         segment_updater.start_merge(merge_operation)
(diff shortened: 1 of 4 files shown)


src/indexer/merger.rs
```diff
@@ -200,6 +200,12 @@ impl IndexMerger {
                 readers.push(reader);
             }
         }
+
+        if readers.is_empty() {
+            let err_msg = format!("Can't merge, all segments are empty {:?}.", segments);
+            return Err(crate::TantivyError::InvalidArgument(err_msg));
```
fulmicoton: Can we do something better?
Instead of returning an error, we want such a merge operation to succeed, not produce any new segment but remove the previous segments.

PSeitz: I considered that, but `merge` seems to be the wrong operation that is triggered. Instead we could have another operation, like `remove` that is called when all segments are empty.
We should still safeguard the merge code, or allow it to be run on empty data.

fulmicoton: @PSeitz but  triggerring a merge over several segment and discoving that they all contain no docs is a legit case.

I don't think it needs to be a separate operation.

PSeitz: I think that merge by definition creates a new segment, so should we create an empty segment?

src/indexer/index_writer.rs
```diff
@@ -515,7 +515,7 @@ impl IndexWriter {
     /// Merges a given list of segments
     ///
     /// `segment_ids` is required to be non-empty.
-    pub fn merge(&mut self, segment_ids: &[SegmentId]) -> FutureResult<SegmentMeta> {
```
fulmicoton, line 516: Can we return `FutureResult<Option<SegmentMeta>>` ?

It can be very useful for people building tooling.

src/indexer/index_writer.rs
```diff
@@ -515,7 +515,7 @@ impl IndexWriter {
     /// Merges a given list of segments
     ///
     /// `segment_ids` is required to be non-empty.
```
fulmicoton, line 517: Document that this function may not result in the creation of a new segment.

src/indexer/segment_updater.rs
```diff
@@ -573,35 +585,42 @@ impl SegmentUpdater {
...
-                                merge_operation.segment_ids(),
-                                advance_deletes_err
-                            );
-                            assert!(!cfg!(test), "Merge failed.");
-
-                            // ... cancel merge
-                            // `merge_operations` are tracked. As it is dropped, the
-                            // the segment_ids will be available again for merge.
-                            return Err(advance_deletes_err);
+                if let Some(mut after_merge_segment_entry) = after_merge_segment_entry.as_mut() {
```
fulmicoton: clippy... it does need to be mut because it is a mut reference.

src/indexer/segment_updater.rs
```diff
@@ -573,35 +585,42 @@ impl SegmentUpdater {
     fn end_merge(
         &self,
         merge_operation: MergeOperation,
-        mut after_merge_segment_entry: SegmentEntry,
-    ) -> crate::Result<SegmentMeta> {
+        mut after_merge_segment_entry: Option<SegmentEntry>,
+    ) -> crate::Result<Option<SegmentMeta>> {
         let segment_updater = self.clone();
-        let after_merge_segment_meta = after_merge_segment_entry.meta().clone();
+        let after_merge_segment_meta = after_merge_segment_entry
```
fulmicoton, line 591: after_merge_segment_meta is not used below...

fulmicoton (approved):
See comments inline

src/indexer/segment_updater.rs
PSeitz: yes it's returned

PR #1432, opened 2022-08-01
add gcd test for DateTime

PSeitz:
add gcd test for DateTime

### DIFF
--- a/src/fastfield/gcd.rs
+++ b/src/fastfield/gcd.rs
@@ -194,7 +194,6 @@ mod tests {
         let file = directory.open_read(path).unwrap();
         let composite_file = CompositeFile::open(&file)?;
         let file = composite_file.open_read(*FIELD).unwrap();
-        assert_eq!(file.len(), 100);
         let fast_field_reader = DynamicFastFieldReader::<u64>::open(file)?;
         assert_eq!(fast_field_reader.get(0), 1000u64);
         assert_eq!(fast_field_reader.get(1), 2000u64);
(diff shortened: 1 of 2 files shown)


fastfield_codecs/src/linearinterpol.rs
```diff
@@ -102,7 +102,11 @@ fn get_slope(first_val: u64, last_val: u64, num_vals: u64) -> f32 {
 
 #[inline]
 fn get_calculated_value(first_val: u64, pos: u64, slope: f32) -> u64 {
-    first_val + (pos as f32 * slope) as u64
+    if slope < 0.0 {
+        (first_val - (pos as f32 * slope) as u64) as u64
```
fulmicoton: i don't understand... why would that be correct?

PSeitz: It's not, it should be
`(pos as f32 * -slope)`

PSeitz: I moved this to the bugfix PR https://github.com/quickwit-oss/tantivy/pull/1439
This PR is now only adding a test and exposing the gcd

fulmicoton: Why is it * -slope?

fulmicoton (changes requested):
Let's not expose `gcd` and just check on the compressed len in the unit test...

src/fastfield/gcd.rs
```diff
@@ -14,10 +14,11 @@ pub const GCD_CODEC_ID: u8 = 4;
 /// Holds the data and the codec to the read the data.
 #[derive(Clone)]
 pub struct GCDFastFieldCodec<CodecReader> {
-    gcd: u64,
+    pub(crate) gcd: u64,
```
fulmicoton:
```suggestion
    gcd: u64,
```

src/fastfield/gcd.rs
```diff
@@ -228,6 +229,76 @@ mod tests {
...
+        let mut schema_builder = Schema::builder();
+        let date_options = DateOptions::default()
+            .set_fast(Cardinality::SingleValue)
+            .set_precision(precision);
+        let field = schema_builder.add_date_field("field", date_options);
+        let schema = schema_builder.build();
+
+        let docs = vec![doc!(field=>time1), doc!(field=>time2), doc!(field=>time3)];
+
+        let directory = get_index(&docs, &schema, codec_name.clone().into())?;
```
fulmicoton:
```suggestion
        let directory = get_index(&docs, &schema, codec_name.into())?;
```

fulmicoton (approved):

PR #1434, opened 2022-08-03
Add support for json-field snippet generation

### LINKED ISSUE #1431: Can Tantivy generate snippets from JSON field?
I found an example of how generate snippets from TEXT field. Then I have tried to do the same with JSON field but I didn't get anything. Can I do it with JSON field?

k-yomo:
Resolves #1431

### DIFF
--- a/src/snippet/mod.rs
+++ b/src/snippet/mod.rs
@@ -245,6 +245,8 @@ impl SnippetGenerator {
             }
             let term_str = if let Some(term_str) = term.as_str() {
                 term_str
+            } else if let Some(term_str) = term.as_json_str() {
+                term_str
             } else {
                 continue;
             };
@@ -280,7 +282,17 @@ impl SnippetGenerator {
     pub fn snippet_from_doc(&self, doc: &Document) -> Snippet {
         let text: String = doc
             .get_all(self.field)
-            .flat_map(Value::as_text)
+            .flat_map(|v| match v {
+                Value::Str(text) => Some(vec![text.as_str()]),
+                Value::JsonObject(json_object) => Some(
+                    json_object
+                        .values()
+                        .flat_map(|v| v.as_str())
+                        .collect::<Vec<&str>>(),
+                ),
+                _ => None,
+            })
+            .flatten()
             .collect::<Vec<&str>>()
             .join(" ");
         self.snippet(&text)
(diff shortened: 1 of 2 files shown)


src/snippet/mod.rs
```diff
@@ -566,4 +580,42 @@ Survey in 2016, 2017, and 2018."#;
...
+        let query = query_parser
+            .parse_query("json.test:rust json.test:design")
+            .unwrap();
+        let snippet_generator = SnippetGenerator::create(&searcher, &*query, json_field).unwrap();
+        {
+            let doc = doc!(json_field => json_val.clone());
+            let snippet = snippet_generator.snippet_from_doc(&doc);
+            assert_eq!(
+                snippet.to_html(),
+                "this is the <b>design</b> of <b>rust</b>"
```
lavrd, line 616: Does this method should generate `{"test": "this is the <b>design</b> of <b>rust</b>"}`?

k-yomo: Ah I see..!

src/snippet/mod.rs
```diff
@@ -280,7 +282,17 @@ impl SnippetGenerator {
     pub fn snippet_from_doc(&self, doc: &Document) -> Snippet {
         let text: String = doc
             .get_all(self.field)
-            .flat_map(Value::as_text)
+            .flat_map(|v| match v {
+                Value::Str(text) => Some(vec![text.as_str()]),
+                Value::JsonObject(json_object) => Some(
+                    json_object
+                        .values()
+                        .flat_map(|v| v.as_str())
```
fulmicoton, line 290: This only works over one level of nesting.

k-yomo: Oh that's true...!  Let me rework

src/snippet/mod.rs
k-yomo: @lavrd @fulmicoton 
Hi, sorry for the late response.  When the nest is deep, should it return
```
{"a": {"b": {"c": "this is the <b>design</b> of <b>rust</b>"}}}
```

or 
```
{"c": "this is the <b>design</b> of <b>rust</b>"}
```

lavrd: Hi, I think it is more a question for maintainers, but I expect and want to have first variant with whole json. Also, if you can implement both variants, maybe you can add an option to chose deep level? Because maybe other users want to have another deep level 😅

daphnenhuch-at:
Is there any chance this would get revived and/or picked back up?

PR #1436, opened 2022-08-08
Expose inner file slice for fieldnorms

boraarslan:
We need the inner file slice to fetch fieldnorms in Quickwit.
PR is very trivial, just exposing a couple of stuff.

### DIFF
--- a/src/core/segment_reader.rs
+++ b/src/core/segment_reader.rs
@@ -128,7 +128,8 @@ impl SegmentReader {
         })
     }
 
-    pub(crate) fn fieldnorms_readers(&self) -> &FieldNormReaders {
+    #[doc(hidden)]
+    pub fn fieldnorms_readers(&self) -> &FieldNormReaders {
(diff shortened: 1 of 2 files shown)


src/core/segment_reader.rs
```diff
@@ -128,7 +128,8 @@ impl SegmentReader {
         })
     }
 
-    pub(crate) fn fieldnorms_readers(&self) -> &FieldNormReaders {
+    #[doc(hidden)]
+    pub fn fieldnorms_readers(&self) -> &FieldNormReaders {
```
PSeitz, line 132: Is `get_fieldnorms_reader` not sufficent?

boraarslan: I need to access the file slice to download relevant parts asynchronously. Unfortunately, I can't use `FieldNormReader` for that.

PSeitz (approved):

PR #1439, opened 2022-08-11
fix get calculated value

### LINKED ISSUE #1440: Occasional subtract-with-overflow panic in u64 fastfield access
**Describe the bug**
Sometimes an `attempt to subtract with overflow` panic occurs when reading an u64-fastfield. The panic specifically occurs on [this line](https://github.com/quickwit-oss/tantivy/blob/main/fastfield_codecs/src/multilinearinterpol.rs#L166) where `interpolation.positive_val_offset` would have an unexpectedly large value.
The values observed are:
```
calculated_value = 1432057903
diff = 184630655
interpolation.positive_val_offset = 18446744073709551557
```

As suggested by @PSeitz on discord the issue probably comes from a wrongfull cast to `u64` in the `get_calculated_value` function which causes `interpolation.positive_val_offset` to be set to a value very close to `u64::MAX` when serialized (from [here](https://github.com/quickwit-oss/tantivy/blob/main/fastfield_codecs/src/multilinearinterpol.rs#L240-L256)).

`tantivy = "0.18.0"`

**To Reproduce**
I haven't been able to consistently reproduce the issue.  @PSeitz suggested a solution which might have solved the issue, but since I wasn't able to consistently reproduce the issue before the change, this is very hard to tell. I will open a pull request with the proposed solution

PSeitz:
fix get calculated value by delaying cast
Fixes #1440

### DIFF
--- a/fastfield_codecs/src/linearinterpol.rs
+++ b/fastfield_codecs/src/linearinterpol.rs
@@ -91,7 +91,7 @@ impl FastFieldCodecReader for LinearInterpolFastFieldReader {
 pub struct LinearInterpolFastFieldSerializer {}
 
 #[inline]
-fn get_slope(first_val: u64, last_val: u64, num_vals: u64) -> f32 {
+pub(crate) fn get_slope(first_val: u64, last_val: u64, num_vals: u64) -> f32 {
     if num_vals <= 1 {
         return 0.0;
     }
@@ -101,7 +101,7 @@ fn get_slope(first_val: u64, last_val: u64, num_vals: u64) -> f32 {
 }
 
 #[inline]
-fn get_calculated_value(first_val: u64, pos: u64, slope: f32) -> u64 {
+pub(crate) fn get_calculated_value(first_val: u64, pos: u64, slope: f32) -> u64 {
     first_val + (pos as f32 * slope) as u64
 }
 
--- a/fastfield_codecs/src/linearinterpol.rs
+++ b/fastfield_codecs/src/linearinterpol.rs
@@ -97,7 +97,16 @@ pub(crate) fn get_slope(first_val: u64, last_val: u64, num_vals: u64) -> f32 {
     }
     //  We calculate the slope with f64 high precision and use the result in lower precision f32
     //  This is done in order to handle estimations for very large values like i64::MAX
-    ((last_val as f64 - first_val as f64) / (num_vals as u64 - 1) as f64) as f32
+    let diff = diff(last_val, first_val);
+    (diff / (num_vals as u64 - 1) as f64) as f32
+}
+
+fn diff(val1: u64, val2: u64) -> f64 {
+    if val1 >= val2 {
+        (val1 - val2) as f64
+    } else {
+        (val2 - val1) as f64 * -1.0
+    }
 }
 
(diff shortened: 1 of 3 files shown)


fastfield_codecs/src/linearinterpol.rs
```diff
@@ -91,17 +91,26 @@ impl FastFieldCodecReader for LinearInterpolFastFieldReader {
...
-fn get_slope(first_val: u64, last_val: u64, num_vals: u64) -> f32 {
+pub(crate) fn get_slope(first_val: u64, last_val: u64, num_vals: u64) -> f32 {
     if num_vals <= 1 {
         return 0.0;
     }
     //  We calculate the slope with f64 high precision and use the result in lower precision f32
     //  This is done in order to handle estimations for very large values like i64::MAX
-    ((last_val as f64 - first_val as f64) / (num_vals as u64 - 1) as f64) as f32
+    let diff = diff(last_val, first_val);
+    (diff / (num_vals as u64 - 1) as f64) as f32
```
boraarslan: <sub>a little nitpick </sub> 
Casting `num_vals` to u64 has no effect.

boraarslan (approved):

sachaarbonel:
Hey, I'm still having the same error but in this line "src/postings/recorder.rs:280:46"
EDIT: Opened a PR #1447 which fixes it for me

fastfield_codecs/src/linearinterpol.rs
```diff
@@ -91,17 +91,26 @@ impl FastFieldCodecReader for LinearInterpolFastFieldReader {
...
         return 0.0;
     }
     //  We calculate the slope with f64 high precision and use the result in lower precision f32
     //  This is done in order to handle estimations for very large values like i64::MAX
-    ((last_val as f64 - first_val as f64) / (num_vals as u64 - 1) as f64) as f32
+    let diff = diff(last_val, first_val);
+    (diff / (num_vals - 1) as f64) as f32
+}
+
+fn diff(val1: u64, val2: u64) -> f64 {
```
fulmicoton, line 115: Can you add a comment?

For which values is this dance useful?

fulmicoton:
@sachaarbonel can you provide a unit test that highlights the remaining problem?

fulmicoton (changes requested):
Can we remove the floating point logic? It is error-prone and I suspect it is slow too.
As an alternative, I suggest we compute and encode the slope as an integer with a bitshift of 32 bit.

(Of course we still need to test and take special care of extreme values when computing the slope value.)

```
pub fn slope_multiple_using_float_arithmetic(pos: u64, slope: f32) -> u64 {
    (pos as f32 * slope) as u64
}
Estimated 12 cycles on UiCA


pub fn slope_multiple_using_int_arithmetic(pos: u64, slope: u64) -> u64 {
    (pos *slope) >> 32 as u64
}

Estimated 1.5 cycle on UiCA
```

fastfield_codecs/src/linearinterpol.rs
PSeitz: I added a comment

fulmicoton (approved):

PR #1442, opened 2022-08-11
Add missing `seek` method to RequiredOptionalScorer

k-yomo:
Now `RequiredOptionalScorer` doesn't implement `seek`, so it uses fallbacked below implementation.
https://github.com/quickwit-oss/tantivy/blob/f4d76213708011c7394861add5902193fc99d4fb/src/docset.rs#L38-L45

It can be more efficient sometimes to use `req_scorer`'s seek.


fulmicoton:
@k-yomo Thank you!

PR #1444, opened 2022-08-12
Support for SnippetGenerator in async context

### LINKED ISSUE #1443: Needs an async version of `Searcher.doc_freq`
In order to support snippeting in Quickwit, we need to have an async version of `Searcher.doc_freq`, currently we cannot build a SnippetGenerator in an async context.

evanxg852000:
Added few methods to allow creating `SinppetGenerator` in an async context outside of Tantivy.
- added `doc_freq_async` to `InvertedIndexReader` 
- added `new` method to `SnippetGenerator`

Closes https://github.com/quickwit-oss/tantivy/issues/1443

### DIFF
--- a/src/core/inverted_index_reader.rs
+++ b/src/core/inverted_index_reader.rs
@@ -230,4 +230,13 @@ impl InvertedIndexReader {
         }
         Ok(())
     }
+
+    /// Returns the number of documents containing the term asynchronously.
+    pub async fn doc_freq_async(&self, term: &Term) -> crate::AsyncIoResult<u32> {
+        Ok(self
+            .get_term_info_async(term)
+            .await?
+            .map(|term_info| term_info.doc_freq)
+            .unwrap_or(0u32))
+    }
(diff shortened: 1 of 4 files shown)


PSeitz (approved):

src/core/inverted_index_reader.rs
```diff
@@ -230,4 +230,13 @@ impl InvertedIndexReader {
         }
         Ok(())
     }
+
+    /// Returns the number of documents containing the term asynchronously.
+    pub async fn doc_freq_async(&self, term: &Term) -> crate::AsyncIoResult<u32> {
```
PSeitz, line 235: In the future, rust may support to be generic over async. That would make things much easier for us

https://blog.rust-lang.org/inside-rust/2022/07/27/keyword-generics.html

evanxg852000: Yes indeed this is needed here :+1:

PR #1447, opened 2022-08-16
Fix another "attempt to subtract with overflow" error

sachaarbonel:
Related to this issue #1440 but in the implementation of Record for TfAndPositionRecorder serialize method

### DIFF
--- a/src/postings/recorder.rs
+++ b/src/postings/recorder.rs
@@ -277,7 +277,7 @@ impl Recorder for TfAndPositionRecorder {
                         break;
                     }
                     Some(position_plus_one) => {
-                        let delta_position = position_plus_one - prev_position_plus_one;
+                        let delta_position = diff(position_plus_one, prev_position_plus_one);
                         buffer_positions.push(delta_position);
                         prev_position_plus_one = position_plus_one;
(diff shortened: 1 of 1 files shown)


src/postings/recorder.rs
```diff
@@ -304,6 +304,14 @@ impl Recorder for TfAndPositionRecorder {
     }
 }
 
+fn diff(val1: u32, val2: u32) -> u32 {
```
adamreichold, line 307: This is not comparable to #1439 which just changed how the value is computed but produced the same value in any case (due to the `* -1.0` in the second branch). This function here would yield a different value for `delta_position` than intended.

I suspect the fix is in deeper in the stack if `VInt32Reader` is expected to yield monotonically increasing positions but does not.

PSeitz:
@sachaarbonel thanks for the PR. Although it looks similar to #1440, this is completely unrelated. Positions should always be in order. Do you have a minimal use case to reproduce that?

sachaarbonel:
Ah, I didn't pay attention to the order of the results. I'll try to have a minimal example (the JSON I'm indexing is 7MB right now).

PR #1448, opened 2022-08-17
Check for the special case of the root facet as prefix of other facets.

adamreichold:
I noticed that the root facet is currently not considered a prefix of other facets in terms of `Facet::is_prefix_of` and made this PR to clarify that this is intentional or modify the implementation to include that case if not.


fulmicoton:
@adamreichold. This was a bug indeed! Thanks for the contribution!

fulmicoton (approved):

PR #1452, opened 2022-08-20
clippy

fulmicoton:


query-grammar/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::derive_partial_eq_without_eq)]
```
adamreichold, line 1: Does it really make sense to allow this instead of just adding the derive or is this mainly about the code noise of doing that?

fulmicoton: The lint feels pointless to me.

PR #1456, opened 2022-08-22
fix issue in composite file

PSeitz:
The file offsets were recorded incorrectly in some cases, e.g. when the recording looked like this [(Field 1, Index 0, Offset 0), (Field 1, Index 1, Offset 14), (Field 0, Index 0, Offset 14)]. The last file is offset 14 to end of file for field 0. But the data was converted to a vec and sorted, which changes the last file to Field 1.

### DIFF
--- a/src/directory/composite_file.rs
+++ b/src/directory/composite_file.rs
@@ -38,7 +38,7 @@ impl BinarySerializable for FileAddr {
 /// A `CompositeWrite` is used to write a `CompositeFile`.
 pub struct CompositeWrite<W = WritePtr> {
     write: CountingWriter<W>,
-    offsets: HashMap<FileAddr, u64>,
+    offsets: Vec<(FileAddr, u64)>,
 }
 
 impl<W: TerminatingWrite + Write> CompositeWrite<W> {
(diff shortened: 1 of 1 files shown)


src/directory/composite_file.rs
```diff
@@ -106,6 +98,14 @@ pub struct CompositeFile {
     offsets_index: HashMap<FileAddr, Range<usize>>,
 }
 
+impl std::fmt::Debug for CompositeFile {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
fulmicoton, line 102:
```suggestion
    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
```

fulmicoton:
What an horrible bug. Thanks for spotting!

fulmicoton (approved):

src/directory/composite_file.rs
```diff
@@ -60,8 +60,8 @@ impl<W: TerminatingWrite + Write> CompositeWrite<W> {
     pub fn for_field_with_idx(&mut self, field: Field, idx: usize) -> &mut CountingWriter<W> {
         let offset = self.write.written_bytes();
         let file_addr = FileAddr::new(field, idx);
-        assert!(!self.offsets.contains_key(&file_addr));
-        self.offsets.insert(file_addr, offset);
+        assert!(!self.offsets.iter().any(|el| el.0 == file_addr));
```
adamreichold, line 63: Should this really be an `assert` instead of a `debug_assert`, i.e. is it necessary for memory safety? This does have linear instead of amortized constant runtime in the number of offsets now, so avoiding the check in release builds would be nice if possible IMHO.

PR #1460, opened 2022-08-22
move iter to FastFieldDataAccess

PSeitz:


fastfield_codecs/src/lib.rs
```diff
@@ -74,12 +76,19 @@ impl<'a> FastFieldDataAccess for &'a [u64] {
     fn get_val(&self, position: u64) -> u64 {
         self[position as usize]
     }
+
+    fn iter<'b>(&'b self) -> Box<dyn Iterator<Item = u64> + 'b> {
+        Box::new((self as &[u64]).iter().cloned())
+    }
 }
 
 impl FastFieldDataAccess for Vec<u64> {
```
fulmicoton, line 85: Do we need an impl for Vec?

fulmicoton: Assuming we do, could we just call the slice implementation?

fastfield_codecs/src/bitpacked.rs
```diff
@@ -113,13 +113,11 @@ impl FastFieldCodecSerializer for BitpackedFastFieldSerializer {
         write: &mut impl Write,
         _fastfield_accessor: &dyn FastFieldDataAccess,
```
fulmicoton:
```suggestion
        fastfield_accessor: &dyn FastFieldDataAccess,
```

src/indexer/merger.rs
```diff
@@ -764,27 +763,30 @@ impl IndexMerger {
...
             }
+
+            fn iter<'b>(&'b self) -> Box<dyn Iterator<Item = u64> + 'b> {
+                Box::new(
+                    self.doc_id_mapping
+                        .iter_old_doc_addrs()
+                        .flat_map(|old_doc_addr| {
+                            let ff_reader =
+                                &self.fast_field_readers[old_doc_addr.segment_ord as usize];
+                            let mut vals = Vec::new();
```
fulmicoton, line 774: Can we remove this alloc?

fulmicoton (approved):

fastfield_codecs/src/lib.rs
PSeitz: We can remove it
Edit: it's used in the tests as a convenience

src/indexer/merger.rs
PSeitz: It should be possible, but it's out of scope of this PR I think. Maybe by returning an iterator over the elements instead of filling an vec

PR #1462, opened 2022-08-22
Fix possibility of inconsistency between `Searcher` and `SearcherGeneration`

### LINKED ISSUE #1461: `Searcher` and its `SearcherGeneration` may reference a different set of segments
# Bug

With tantivy @ d24f31f965543e7ad300b4ce9965b27aef8f3be1,

ran into a situation where Gen 1 below of the Searcher has an inconsistent set of segment IDs referenced in its `SearcherGeneration` object. 

## Gen 0: good

```
Searcher(
	[
		Seg("2b9d1870"),
		Seg("de008a03"),
		Seg("1443006d"),
		Seg("ce5a247e"),
		Seg("70af8500"),
		Seg("7c542ba6")
	]
)

SearcherGeneration {
	segments: {
		Seg("1443006d"): Some(103364896),
		Seg("2b9d1870"): Some(103364896),
		Seg("70af8500"): Some(103364896),
		Seg("7c542ba6"): None,
		Seg("ce5a247e"): Some(103364896),
		Seg("de008a03"): Some(103364896),
	},
	generation_id: 0
}
```

## Gen 1: bad

```
Searcher(
	[
		Seg("2b9d1870"),
		Seg("de008a03"),
		Seg("1443006d"),
		Seg("ce5a247e"),
		Seg("70af8500"),
		Seg("68356e6c”),
	]
)

SearcherGeneration {
	segments: {
		Seg("1240c4ed"): None,
		Seg("1443006d"): Some(103367743),
		Seg("2b9d1870"): Some(103367743),
		Seg("5a475796"): None,
		Seg("7c542ba6"): Some(103367743),
		Seg("70af8500"): Some(103367743),
		Seg("ce5a247e"): Some(103367743),
		Seg("de008a03"): Some(103367743)
	},
	generation_id: 1,
}

```

# Reason

Looks like this may have been a regression introduced in 23fe73a6c0a0a9f9b0a263f7fa31b278a9a6101f cc @PSeitz 

In between the below calls to `open_segment_readers()` is potential for a race condition.

https://github.com/quickwit-oss/tantivy/blob/23fe73a6c0a0a9f9b0a263f7fa31b278a9a6101f/src/reader/mod.rs#L239
https://github.com/quickwit-oss/tantivy/blob/23fe73a6c0a0a9f9b0a263f7fa31b278a9a6101f/src/reader/mod.rs#L224

The solution seems to be to only call it once when creating a `Searcher`, this used to be the case before. And maybe we can add an assertion that the set of `SegmentId`s is identical.

shikhar:
Fixes #1461

### DIFF
--- a/src/core/searcher.rs
+++ b/src/core/searcher.rs
@@ -247,6 +247,15 @@ impl SearcherInner {
         generation: TrackedObject<SearcherGeneration>,
         doc_store_cache_size: usize,
     ) -> io::Result<SearcherInner> {
+        assert_eq!(
+            &segment_readers
+                .iter()
+                .map(|reader| (reader.segment_id(), reader.delete_opstamp()))
+                .collect::<BTreeMap<_, _>>(),
+            generation.segments(),
+            "Set of segments referenced by this Searcher and its SearcherGeneration must match"
+        );
+
         let store_readers: Vec<StoreReader> = segment_readers
             .iter()
             .map(|segment_reader| segment_reader.get_store_reader(doc_store_cache_size))
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -11,7 +11,7 @@ use self::warming::WarmingState;
 use crate::core::searcher::{SearcherGeneration, SearcherInner};
 use crate::directory::{Directory, WatchCallback, WatchHandle, META_LOCK};
 use crate::store::DOCSTORE_CACHE_CAPACITY;
-use crate::{Index, Inventory, Searcher, SegmentReader, TrackedObject};
+use crate::{Index, Inventory, Searcher, SegmentReader};
 
 /// Defines when a new version of the index should be reloaded.
 ///
@@ -167,18 +167,12 @@ impl InnerIndexReader {
         searcher_generation_inventory: Inventory<SearcherGeneration>,
     ) -> crate::Result<Self> {
         let searcher_generation_counter: Arc<AtomicU64> = Default::default();
-        let segment_readers = Self::open_segment_readers(&index)?;
-        let searcher_generation = Self::create_new_searcher_generation(
-            &segment_readers,
-            &searcher_generation_counter,
-            &searcher_generation_inventory,
-        );
         let searcher = Self::create_searcher(
             &index,
-            segment_readers,
             doc_store_cache_size,
             &warming_state,
-            searcher_generation,
+            searcher_generation_counter.load(atomic::Ordering::Acquire),
+            &searcher_generation_inventory,
         )?;
         Ok(InnerIndexReader {
             doc_store_cache_size,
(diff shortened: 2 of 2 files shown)


src/core/searcher.rs
```diff
@@ -247,6 +247,15 @@ impl SearcherInner {
         generation: TrackedObject<SearcherGeneration>,
         doc_store_cache_size: usize,
     ) -> io::Result<SearcherInner> {
+        assert_eq!(
```
adamreichold, line 250: Since the assertion does seem to protect memory safety and is quite costly (allocating a new `BTreeMap`), I think it would be better if this was a `debug_assert_eq`.

src/reader/mod.rs
```diff
@@ -236,17 +223,15 @@ impl InnerIndexReader {
     }
 
     fn reload(&self) -> crate::Result<()> {
-        let segment_readers = Self::open_segment_readers(&self.index)?;
-        let searcher_generation = Self::create_new_searcher_generation(
-            &segment_readers,
-            &self.searcher_generation_counter,
-            &self.searcher_generation_inventory,
-        );
+        let generation_id = 1 + self
```
PSeitz, line 226: why +1 and fetch_add?

adamreichold: [`fetch_add`](https://doc.rust-lang.org/stable/std/sync/atomic/struct.AtomicUsize.html#method.fetch_add) returns the previous value before doing the addition, hence the `+1` is necessary to obtain the result that is stored in the atomic.

(GCC for example provides both `__atomic_add_fetch` and `__atomic_fetch_add` which differ only in which value is returned, c.f https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html)

src/core/searcher.rs
fulmicoton: This only happens on an Index reload.

adamreichold: Fair enough, I think that makes this a nit. I would still suggest the change since the loss of robustness is minimal and it avoids adding "distributed fat" to release builds.

fulmicoton: I usually like to keep that kind of assert in prod. They help me sleep at night.

One strong argument is that if we haven't spotted an edge case that breaks this invariant, the assert will eventually spot it, and the system running it will "only" crash (as opposed to do bad things). 

@shikhar that's your call... but I'm in team `assert!` here.

shikhar: Practically: if it were a debug assertion, that may have helped as would have surfaced when trying to reproduce the issue we were seeing locally, but we also test on k8s where only release-with-debug-symbols builds are used and a debug assertion wouldn't have trigger then. So a preference for keeping this a regular assertion from me as well!

shikhar:
#1464 replaces this

PR #1463, opened 2022-08-23
`ConstScoreQuery`

shikhar:
Based this on `BoostQuery`, was able to reuse the existing `ConstScorer`.


fulmicoton (changes requested):
Can we move the ConstScorer in there too?

src/query/const_score_query.rs
```diff
@@ -0,0 +1,126 @@
...
+/// The document set matched by the `ConstScoreQuery` is strictly the same as the underlying query.
+/// The configured score is used for each document.
+pub struct ConstScoreQuery {
+    query: Box<dyn Query>,
+    score: Score,
+}
+
+impl ConstScoreQuery {
+    /// Builds a const score query.
+    pub fn new(query: Box<dyn Query>, score: Score) -> ConstScoreQuery {
```
fulmicoton, line 19: I think the `ConstScoreQuery` symbol is not accessible from client in the library.
(hence the clippy warning below)

Can you re-export it from `tantivy::query`?

src/query/const_score_query.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    fn explain(&self, reader: &SegmentReader, doc: u32) -> crate::Result<Explanation> {
+        let mut scorer = self.scorer(reader, 1.0)?;
+        if scorer.seek(doc) != doc {
+            return Err(TantivyError::InvalidArgument(format!(
+                "Document #({}) does not match",
+                doc
+            )));
+        }
+        let mut explanation =
```
shikhar: Should we avoid calling the child scorer here too?

```rust
let mut explanation = Explanation::new("Const", self.score);
```

fulmicoton: Either way works for me.

fulmicoton (changes requested):
please fix clippy

src/query/const_score_query.rs
shikhar: Removing the inner score from here, it gets included in detail anyway

fulmicoton (approved):

PR #1464, opened 2022-08-23
fixes race condition in Searcher

### LINKED ISSUE #1461: `Searcher` and its `SearcherGeneration` may reference a different set of segments

PSeitz:
fixes #1461
fixes a race condition in Searcher, by avoiding repeated calls to open_segment_readers and passing them instead as argument

### DIFF
--- a/src/core/searcher.rs
+++ b/src/core/searcher.rs
@@ -247,6 +247,14 @@ impl SearcherInner {
         generation: TrackedObject<SearcherGeneration>,
         doc_store_cache_size: usize,
     ) -> io::Result<SearcherInner> {
+        debug_assert_eq!(
+            &segment_readers
+                .iter()
+                .map(|reader| (reader.segment_id(), reader.delete_opstamp()))
+                .collect::<BTreeMap<_, _>>(),
+            generation.segments(),
+            "Set of segments referenced by this Searcher and its SearcherGeneration must match"
+        );
         let store_readers: Vec<StoreReader> = segment_readers
             .iter()
             .map(|segment_reader| segment_reader.get_store_reader(doc_store_cache_size))
--- a/src/core/searcher.rs
+++ b/src/core/searcher.rs
@@ -247,7 +247,7 @@ impl SearcherInner {
         generation: TrackedObject<SearcherGeneration>,
         doc_store_cache_size: usize,
     ) -> io::Result<SearcherInner> {
-        debug_assert_eq!(
+        assert_eq!(
             &segment_readers
                 .iter()
                 .map(|reader| (reader.segment_id(), reader.delete_opstamp()))
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -206,7 +206,7 @@ impl InnerIndexReader {
         searcher_generation_counter: &Arc<AtomicU64>,
         searcher_generation_inventory: &Inventory<SearcherGeneration>,
     ) -> TrackedObject<SearcherGeneration> {
-        let generation_id = searcher_generation_counter.fetch_add(1, atomic::Ordering::Relaxed);
+        let generation_id = searcher_generation_counter.fetch_add(1, atomic::Ordering::AcqRel);
         let searcher_generation =
             SearcherGeneration::from_segment_readers(segment_readers, generation_id);
         searcher_generation_inventory.track(searcher_generation)
@@ -219,11 +219,11 @@ impl InnerIndexReader {
         searcher_generation_counter: &Arc<AtomicU64>,
         searcher_generation_inventory: &Inventory<SearcherGeneration>,
     ) -> crate::Result<Arc<SearcherInner>> {
-        let segment_readers = Self::open_segment_readers(&index)?;
+        let segment_readers = Self::open_segment_readers(index)?;
         let searcher_generation = Self::track_segment_readers_in_inventory(
             &segment_readers,
-            &searcher_generation_counter,
-            &searcher_generation_inventory,
+            searcher_generation_counter,
+            searcher_generation_inventory,
         );
 
         let schema = index.schema();
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -164,21 +164,18 @@ impl InnerIndexReader {
         doc_store_cache_size: usize,
         index: Index,
         warming_state: WarmingState,
+        // The searcher_generation_inventory is not used as source, but as target to track the
+        // loaded segments.
         searcher_generation_inventory: Inventory<SearcherGeneration>,
     ) -> crate::Result<Self> {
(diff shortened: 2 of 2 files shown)


fulmicoton:
@shikhar can you review https://github.com/quickwit-oss/tantivy/pull/1464

src/core/searcher.rs
```diff
@@ -247,6 +247,14 @@ impl SearcherInner {
         generation: TrackedObject<SearcherGeneration>,
         doc_store_cache_size: usize,
     ) -> io::Result<SearcherInner> {
+        debug_assert_eq!(
```
fulmicoton: As discussed in the other thread, I prefer having an actually assert here. (as opposed to debug_assert)

shikhar (commented):
@PSeitz aside from clippy nits, LGTM

I did have one unrelated change I was sneaking into my PR, that I wanted to ask about. 

Is `Relaxed` memory order safe here? Both a load and store are happening, so `AcqRel` seems safer to me. And this is by no means a hot path so being conservative seems fair.

https://github.com/quickwit-oss/tantivy/blob/998b1263f6a28696ed35f685376cb612287fe840/src/reader/mod.rs#L212

src/reader/mod.rs
```diff
@@ -204,7 +201,7 @@ impl InnerIndexReader {
         Ok(segment_readers)
     }
 
-    fn create_new_searcher_generation(
+    fn track_segment_readers_in_inventory(
```
shikhar, line 204: The previous name seems better to me, since we are creating a `SearcherGeneration` and making sure to do it in a way that we track it int the inventory, and the inventory tracks `SearcherGeneration` not segment readers.

PSeitz: my issue with the previous name was that it's not clear that inventory is used to track and not the source of the generation

PSeitz:
> @PSeitz aside from clippy nits, LGTM
> 
> I did have one unrelated change I was sneaking into my PR, that I wanted to ask about.
> 
> Is `Relaxed` memory order safe here? Both a load and store are happening, so `AcqRel` seems safer to me. And this is by no means a hot path so being conservative seems fair.
> 
> https://github.com/quickwit-oss/tantivy/blob/998b1263f6a28696ed35f685376cb612287fe840/src/reader/mod.rs#L212

Yeah, I think there could be scenarios where simultaneous reloads are triggered at the same time from multiple threads.

fulmicoton:
@shikhar The operation is still Atomic... I think we just want to ensure that the id delivered are unique, and Relaxed is sufficient for that. AcqRel stuff is about instruction reordering.

PSeitz:
> @shikhar I think we just want to ensure that the id delivered are unique, and Relaxed is sufficient for that... That being said it is not a hot part of the code. We can use SeqCst and not sweat it.

I think it could be an issue when two threads increment, but due to weak fencing, we get two times the same value back. If now one of those callers has a new segment in their list, we have the same id to different `SearcherGeneration`.

It's very unlikely though, and on many platforms Relaxed is already AcqRel.

adamreichold:
> I think it could be an issue when two threads increment, but due to weak fencing, we get two times the same value back.

The memory orderings mostly determine how _other_ memory accesses are ordered w.r.t. the atomic accesses, they do not affect the atomicity of an operation like `fetch_add`, i.e. multiple threads doing `fetch_add(1, Ordering::Relaxed)` will never see identical values (ignoring overflow).

However, their accesses to other memory locations that are shared with other threads will not be ordered w.r.t. the ID generation, e.g. their accesses to `searcher_generation_inventory` are do not necessarily reflect their accesses to `searcher_generation_counter`, meaning that a thread with lower ID might see the stores to the inventory of a thread with a higher ID.

While I suspect that this is not an issue (this would have involve unsafe code if I understand things correctly), my personal choice would be `Ordering::AcqRel` as well just to capture any implicit dependencies which I do not see ATM. (I would suggest avoiding `Ordering::SeqCst` as mixing those two orderings can have unexpected effects and sticking to `AcqRel` is often preferable if global ordering is not required as it is simpler to  reason about the explicit pairings between acquire and release operations to the same memory address.)

fulmicoton:
> It's very unlikely though, and on many platforms Relaxed is already AcqRel.

This is a very bad argument :). Also ARM is pretty popular.

> we get two times the same value back

You cannot get two times the same value back... That's not what ordering is about.
The explanation here is ok:
https://en.cppreference.com/w/cpp/atomic/memory_order

...


Anyway let's defensively go for `Ordering::AcqRel`...

fulmicoton (approved):

shikhar:
Thanks folks, I need to learn more about memory orderings :)

PR #1466, opened 2022-08-23
Extend facet deserialization to handle owned in addition to borrowed strings.

adamreichold:
Noticed this when trying to deserialize `Facet` from URL query parameters. This also avoids the panic when deserializing (potentially untrusted) content.


fulmicoton (approved):
Thank you! Error handling is much better.

PR #1469, opened 2022-08-23
Cache dependencies in CI to speed up build times.

adamreichold:
I am not sure how effective this is due mixing stable and nightly within a workflow and whether it was tried before, but I have had good experiences using the [`rust-cache`](https://github.com/Swatinem/rust-cache) action myself.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -13,6 +13,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Install Rust
         run: rustup toolchain install nightly --component llvm-tools-preview
+      - uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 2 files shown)


adamreichold:
An unscientific experiment pushing the same commit with a different timestamp and hence hash leads to the following numbers:

| job | cold | warm |
| - | - | - |
| coverage | 22.5m | 12m |
| test |  13.5m | 14m |
| test (cargo-nextest) | | 11m |

So a significant win for the coverage job, but no improvement for the test job.

EDIT: [cargo-nextest](https://nexte.st/) seems to help a wee bit with the test job but I am not sure whether it is worth the additional complexity (especially since the test runner does not support doctests yet).

fulmicoton:
@adamreichold Great win! Thank you!

fulmicoton (approved):

PR #1471, opened 2022-08-24
Split test into check and test CI jobs

adamreichold:
Sorry for not making this part of the earlier CI PR but I did not get this far in time.

~~In contrast to the comment in the workflow definition, the nightly toolchain seems to have been used only for formatting but not for testing in the CI. Hence, this PR drops it to reduce the toolchain installation step. (There can also be only one override toolchain per directory.)~~

EDIT: Still installing a minimal nightly toolchain to keep the existing formatting...

In addition, it proposes splitting up the test job into a check (Rustfmt, Clippy) and test job (unit tests and doc tests) to gain a bit of parallelism in the CI. This seems especially useful to get early feedback for commits which fail due to formatting or linting.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Install Rust
-        run: rustup toolchain install nightly --component llvm-tools-preview
+        run: rustup toolchain install nightly --profile minimal --component llvm-tools-preview
       - uses: Swatinem/rust-cache@v2
       - uses: taiki-e/install-action@cargo-llvm-cov
       - name: Generate code coverage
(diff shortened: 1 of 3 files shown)


adamreichold:
I see this won't fly due to nightly and stable not agreeing on the formatting...

adamreichold:
Not sure where the unit test failure is coming from as I cannot reproduce that locally...

adamreichold:
> Not sure where the unit test failure is coming from as I cannot reproduce that locally...

Seems like cargo-nextest is not picking up the override and still needs the `+stable` flag.

adamreichold:
> > Not sure where the unit test failure is coming from as I cannot reproduce that locally...
> 
> Seems like cargo-nextest is not picking up the override and still needs the `+stable` flag.

Flakiness might also be related to #1472

PSeitz (approved):
Cool, seems to be faster than before

PSeitz:
@adamreichold Thanks for your work on that, the new CI setup is much nicer to work with

PR #1473, opened 2022-08-24
Collapse overlapped highlighted ranges

akr4:
Hi. This fixes a problem that occurs when calling `Snippet.to_html()` and the `snippet.highlighted` are overlapped ranges like this:
```
[0..2, 1..3]
```
This type of ranges is generated by `NgramTokenizer`, for example.

### Error
In that case, there will be an error on calling `Snippet.to_html()`, which is caused by slicing `String` with invalid indices.
```
begin <= end (2 <= 1) when slicing `abc`
thread 'snippet::tests::test_snippet_with_overlapped_highlighted_ranges' panicked at 'begin <= end (2 <= 1) when slicing `abc`', library/core/src/str/mod.rs:111:5
stack backtrace:
...
   9: tantivy::snippet::Snippet::to_html
             at ./src/snippet/mod.rs:84:44
```

### Fix
This PR collapses overlapped ranges into non-overlapped ones to fix the problem like this:
```
[0..2, 1..3] -> [0..3]
```

### DIFF
--- a/src/snippet/mod.rs
+++ b/src/snippet/mod.rs
@@ -79,7 +79,7 @@ impl Snippet {
         let mut html = String::new();
         let mut start_from: usize = 0;
 
-        for item in self.highlighted.iter() {
+        for item in collapse_overlapped_ranges(&self.highlighted).iter() {
             html.push_str(&encode_minimal(&self.fragment[start_from..item.start]));
             html.push_str(HIGHLIGHTEN_PREFIX);
             html.push_str(&encode_minimal(&self.fragment[item.clone()]));
@@ -184,6 +184,28 @@ fn select_best_fragment_combination(fragments: &[FragmentCandidate], text: &str)
             highlighted: vec![],
         }
     }
+}
+
+/// Returns ranges that are collapsed into non-overlapped ranges.
+fn collapse_overlapped_ranges(ranges: &Vec<Range<usize>>) -> Vec<Range<usize>> {
+    if ranges.is_empty() {
+        return Vec::new();
+    }
+
+    let mut result = Vec::new();
+    let mut current = ranges[0].clone();
+
+    for range in ranges.iter().skip(1) {
+        if current.end > range.start {
+            current = current.start..range.end;
+        } else {
+            result.push(current);
+            current = range.clone();
+        }
+    }
+
+    result.push(current);
+    result
 }
 
 /// `SnippetGenerator`
@@ -320,10 +342,10 @@ mod tests {
 
     use maplit::btreemap;
 
-    use super::{search_fragments, select_best_fragment_combination};
+    use super::{search_fragments, select_best_fragment_combination, collapse_overlapped_ranges};
     use crate::query::QueryParser;
     use crate::schema::{IndexRecordOption, Schema, TextFieldIndexing, TextOptions, TEXT};
-    use crate::tokenizer::SimpleTokenizer;
+    use crate::tokenizer::{SimpleTokenizer, NgramTokenizer, Tokenizer};
     use crate::{Index, SnippetGenerator};
 
     const TEST_TEXT: &str = r#"Rust is a systems programming language sponsored by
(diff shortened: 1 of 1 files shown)


src/snippet/mod.rs
```diff
@@ -186,6 +186,28 @@ fn select_best_fragment_combination(fragments: &[FragmentCandidate], text: &str)
...
 }
 
+/// Returns ranges that are collapsed into non-overlapped ranges.
+fn collapse_overlapped_ranges(ranges: &Vec<Range<usize>>) -> Vec<Range<usize>> {
+    if ranges.is_empty() {
+        return Vec::new();
+    }
+
+    let mut result = Vec::new();
+    let mut current = ranges[0].clone();
```
adamreichold: I am not sure but maybe it would be nicer to iterate only once, e.g.

```rust
let mut result = Vec::new();

let mut ranges = ranges.iter();

let mut current = match ranges.next() {
  Some(range) => range.clone(),
  None => return result,
};

for range in ranges {
  ...

Comment (akr4) [reply src/snippet/mod.rs]: Sounds good. I changed so. Thanks.

File: src/snippet/mod.rs [hunk 2]
```diff
@@ -186,6 +186,29 @@ fn select_best_fragment_combination(fragments: &[FragmentCandidate], text: &str)
     }
 }
 
+/// Returns ranges that are collapsed into non-overlapped ranges.
```
fulmicoton, line 189: Please add your example in the comment.

src/snippet/mod.rs
```diff
@@ -186,6 +186,29 @@ fn select_best_fragment_combination(fragments: &[FragmentCandidate], text: &str)
...
+fn collapse_overlapped_ranges(ranges: &Vec<Range<usize>>) -> Vec<Range<usize>> {
```
fulmicoton:
```suggestion
fn collapse_overlapped_ranges(ranges: &[Range<usize>]) -> Vec<Range<usize>> {
```

&Vec is almost never a good idea.

src/snippet/mod.rs
```diff
@@ -186,6 +186,29 @@ fn select_best_fragment_combination(fragments: &[FragmentCandidate], text: &str)
...
+    let mut result = Vec::new();
```
fulmicoton, line 202: This assumes the ranges are sorted by start.
That's a hidden contract.

Please put that in the comment, and ideally check that with an assert.

src/snippet/mod.rs
```diff
@@ -186,6 +186,29 @@ fn select_best_fragment_combination(fragments: &[FragmentCandidate], text: &str)
...
+    let mut ranges = ranges.iter();
+
+    let mut current = match ranges.next() {
+        Some(range) => range.clone(),
+        None => return result,
+    };
+
+    for range in ranges {
+        if current.end > range.start {
+            current = current.start..range.end;
```
fulmicoton: nitpick... I feel the following is easier but that's your call.
```suggestion
            current.end = range.end;
```

src/snippet/mod.rs
fulmicoton: Another solution would be for you to pass an iterator directly.

`impl Iterator<Item=Range<usize>>`

src/snippet/mod.rs
```diff
@@ -186,6 +186,29 @@ fn select_best_fragment_combination(fragments: &[FragmentCandidate], text: &str)
     }
 }
 
+/// Returns ranges that are collapsed into non-overlapped ranges.
+fn collapse_overlapped_ranges(ranges: &Vec<Range<usize>>) -> Vec<Range<usize>> {
+    let mut result = Vec::new();
+    let mut ranges = ranges.iter();
```
fulmicoton: nitpick... I'm not fond of shadowing. 
It is sometime useful to prevent the usage of a variable.

Here I would either call it `ranges_it`, or pass an Iterator as argument.

fulmicoton:
@akr4 Awesome job spotting the job and fixing it! I had a bunch of rather minor comments.

akr4:
@fulmicoton Thank you very much for your review. I will make changes tomorrow.

akr4:
Committed some changes. The main ones are changing the argument from `&Vec` to `&[ ]` and adding a comment. What do you think of this?

fulmicoton (approved):

fulmicoton (approved):
Approved! Thank you @akr4

fulmicoton:
@akr4 Can you run `make fmt`? (It requires rust nightly)

akr4:
@fulmicoton Done.

PR #1478, opened 2022-08-24
GCD wrapper uses DividerU64

fulmicoton:


src/fastfield/serializer/mod.rs
```diff
@@ -142,17 +143,19 @@ impl CompositeFastFieldSerializer {
...
+                self.gcd
+                    .divide(self.fastfield_accessor.get_val(position) - self.base_value)
             }
-            fn iter<'b>(&'b self) -> Box<dyn Iterator<Item = u64> + 'b> {
+            fn iter(&self) -> Box<dyn Iterator<Item = u64> + '_> {
                 Box::new(
                     self.fastfield_accessor
                         .iter()
-                        .map(|val| (val - self.base_value) / self.gcd),
+                        .map(|val| self.gcd.divide(val - self.base_value)),
```
PSeitz, line 158: if you release a new version of `fastdivide` with the `Div` impl, the code could stay the same, which is more readable

PSeitz (approved):

PR #1480, opened 2022-08-24
fix overflow issue in interpolation

### LINKED ISSUE #1479: attempt to subtract with overflow
There's a flaky test `histogram_test_min_value_positive_force_merge_segments`, which is very similar to https://github.com/quickwit-oss/tantivy/issues/1440

```
 --- STDERR:                            tantivy aggregation::bucket::histogram::histogram::tests::histogram_test_min_value_positive_force_merge_segments ---
thread 'merge_thread_3' panicked at 'attempt to subtract with overflow', /home/runner/work/tantivy/tantivy/fastfield_codecs/src/linear.rs:131:9
Error: SystemError("Merge operation failed.")
thread 'aggregation::bucket::histogram::histogram::tests::histogram_test_min_value_positive_force_merge_segments' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/test/src/lib.rs:184:5
```

PSeitz:
use saturating_sub and saturating_add to cover edge cases with values close to u64::MAX or 0 in combination with imprecise computation

fixes #1479

### DIFF
--- a/fastfield_codecs/src/lib.rs
+++ b/fastfield_codecs/src/lib.rs
@@ -227,6 +227,11 @@ mod tests {
         data_and_names.push((vec![5, 50, 3, 13, 1, 1000, 35], "rand small"));
         data_and_names.push((vec![10], "single value"));
 
+        data_and_names.push((
+            vec![1572656989877777, 1170935903116329, 720575940379279, 0],
+            "overflow error",
+        ));
+
         data_and_names
     }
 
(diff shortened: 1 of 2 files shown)


fulmicoton (approved):

fulmicoton:
@pseitz approved... but let's switch to integer arithmetics.

PR #1482, opened 2022-08-25
Int based linear interpol

fulmicoton:


fastfield_codecs/src/line.rs
```diff
@@ -0,0 +1,204 @@
...
+        // = - shift * abs_slope
+        // ```
+        u64::MAX - abs_slope
+    }
+}
+
+impl Line {
+    #[inline(always)]
+    pub fn eval(&self, x: u64) -> u64 {
+        let linear_part = (x.wrapping_mul(self.slope) >> 32) as i32 as u64;
```
PSeitz, line 66: why cast first to i32?

fulmicoton: you want negative slopes to be interpreted as negative... 
After the bitshift, you want stuff to be interpreted as negative is the 32th bit is set to 1, not the 64th.
There is probably a nicer a way to express it by expressing the slope as a i64.

PR #1485, opened 2022-08-26
merge FastFieldCodecReader wit FastFieldDataAccess

PSeitz:
- num_vals to FastFieldCodecReader
- split open_from_bytes to own trait
- rename get_u64 to get_val
- merge FastFieldCodecReader and FastFieldDataAccess traits

### DIFF
--- a/fastfield_codecs/src/bitpacked.rs
+++ b/fastfield_codecs/src/bitpacked.rs
@@ -42,7 +42,7 @@ impl FastFieldCodecDeserializer for BitpackedReader {
 }
 impl FastFieldCodecReader for BitpackedReader {
     #[inline]
-    fn get_u64(&self, doc: u64) -> u64 {
+    fn get_val(&self, doc: u64) -> u64 {
         self.min_value_u64 + self.bit_unpacker.get(doc, &self.data)
     }
     #[inline]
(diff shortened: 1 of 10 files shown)


fastfield_codecs/src/bitpacked.rs
```diff
@@ -16,27 +16,32 @@ pub struct BitpackedReader {
     bit_unpacker: BitUnpacker,
     pub min_value_u64: u64,
     pub max_value_u64: u64,
+    pub num_vals: u64,
 }
 
-impl FastFieldCodecReader for BitpackedReader {
+impl FastFieldCodecDeserializer for BitpackedReader {
```
fulmicoton, line 22: I'd put that in the `XXXXSerializer`s and rename the trait `XXXXCodec`, and add an associated trait over the reader.

The XXXXSerializer and the XXXXDeserializer are not really independent.

fulmicoton (approved):

fastfield_codecs/src/bitpacked.rs
PSeitz: I think don't fully understand. I split the Deserializer from the `FastFieldDataAccess`, because e.g. in the merge case we provide access without deserializing.

PR #1492, opened 2022-08-29
fastfield return iterator instead fill vec

PSeitz:
return iterator from get_vals method. This will allow to save on unnecessary vec allocations.

### DIFF
--- a/fastfield_codecs/src/blockwise_linear.rs
+++ b/fastfield_codecs/src/blockwise_linear.rs
@@ -300,13 +300,9 @@ impl FastFieldCodec for BlockwiseLinearCodec {
         // If this doesn't overflow the algorithm should be fine
         let theorethical_maximum_offset =
             fastfield_accessor.max_value() - fastfield_accessor.min_value();
-        if fastfield_accessor
+        fastfield_accessor
             .max_value()
-            .checked_add(theorethical_maximum_offset)
-            .is_none()
-        {
-            return None;
-        }
+            .checked_add(theorethical_maximum_offset)?;
 
         let first_val_in_first_block = fastfield_accessor.get_val(0);
         let last_elem_in_first_chunk = CHUNK_SIZE.min(fastfield_accessor.num_vals());
--- a/fastfield_codecs/src/column.rs
+++ b/fastfield_codecs/src/column.rs
@@ -1,3 +1,5 @@
+use std::ops::Range;
+
 pub trait Column<T = u64> {
     /// Return the value associated to the given idx.
     ///
@@ -8,23 +10,14 @@ pub trait Column<T = u64> {
     /// May panic if `idx` is greater than the column length.
     fn get_val(&self, idx: u64) -> T;
 
-    /// Fills an output buffer with the fast field values
-    /// associated with the `DocId` going from
-    /// `start` to `start + output.len()`.
-    ///
-    /// Regardless of the type of `Item`, this method works
-    /// - transmuting the output array
-    /// - extracting the `Item`s as if they were `u64`
-    /// - possibly converting the `u64` value to the right type.
+    /// Returns an iterator over given doc range.
     ///
     /// # Panics
     ///
-    /// May panic if `start + output.len()` is greater than
+    /// May panic if `range.end()` is greater than
     /// the segment's `maxdoc`.
-    fn get_range(&self, start: u64, output: &mut [T]) {
-        for (out, idx) in output.iter_mut().zip(start..) {
-            *out = self.get_val(idx);
-        }
+    fn get_range(&self, range: Range<u64>) -> Box<dyn Iterator<Item = T> + '_> {
+        Box::new(range.map(|idx| self.get_val(idx)))
     }
 
     /// Returns the minimum value for this fast field.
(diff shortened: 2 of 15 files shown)


fastfield_codecs/src/blockwise_linear.rs
```diff
@@ -300,13 +300,9 @@ impl FastFieldCodec for BlockwiseLinearCodec {
...
             fastfield_accessor.max_value() - fastfield_accessor.min_value();
-        if fastfield_accessor
+        fastfield_accessor
             .max_value()
-            .checked_add(theorethical_maximum_offset)
-            .is_none()
-        {
-            return None;
-        }
+            .checked_add(theorethical_maximum_offset)?;
```
fulmicoton, line 305: I prefer the earlier version... By far actually.

In the new version the intent is hidden.

fastfield_codecs/src/column.rs
```diff
@@ -8,23 +10,14 @@ pub trait Column<T = u64> {
...
-    /// Regardless of the type of `Item`, this method works
-    /// - transmuting the output array
-    /// - extracting the `Item`s as if they were `u64`
-    /// - possibly converting the `u64` value to the right type.
+    /// Returns an iterator over given doc range.
     ///
     /// # Panics
     ///
-    /// May panic if `start + output.len()` is greater than
+    /// May panic if `range.end()` is greater than
```
fulmicoton, line 19:
```suggestion
    /// May panic if `range.end` is greater than
```

src/fastfield/facet_reader.rs
```diff
@@ -76,7 +76,8 @@ impl FacetReader {
 
     /// Return the list of facet ordinals associated to a document.
     pub fn facet_ords(&self, doc: DocId, output: &mut Vec<u64>) {
-        self.term_ords.get_vals(doc, output);
```
fulmicoton, line 79: so here, this is a net loss.

fastfield_codecs/src/linear.rs
```diff
@@ -204,13 +204,9 @@ impl FastFieldCodec for LinearCodec {
...
             fastfield_accessor.max_value() - fastfield_accessor.min_value();
-        if fastfield_accessor
+        fastfield_accessor
             .max_value()
-            .checked_add(theorethical_maximum_offset)
-            .is_none()
-        {
-            return None;
-        }
+            .checked_add(theorethical_maximum_offset)?;
```
fulmicoton, line 209: same as before. I prefer the earlier version.

fulmicoton:
Do you have benchmark to confirm we don't have a regression in search?

The `fill` to buffer indirection point is that it get rids of dynamic dispatch.

fastfield_codecs/src/blockwise_linear.rs
PSeitz: it's from a clippt lint, I also think the previous version is more clear. We could use [overflowing_add](https://doc.rust-lang.org/std/primitive.i8.html#method.overflowing_add) instead and check the bool, or disable the lint

fulmicoton: What about disabling the lint locally?

PSeitz:
> Do you have benchmark to confirm we don't have a regression in search?
> 
> The `fill` to buffer indirection point is that it get rids of dynamic dispatch.

I removed the dynamic dispatch and added a bench. It is still slower, which I don't understand yet. 

```
Vec version - Most time spent in MultiValuedFastFieldReader<Item>::get_vals
running 1 test
test fastfield::bench::bench_multi_value_fflookup                             ... bench:     758,148 ns/iter (+/- 13,857)

+   93,36%    15,12%  tantivy-cbbeafa  tantivy-cbbeafab3d8c2ac9      [.] test::bench::Bencher::iter                                                                ▒
+   78,22%    78,20%  tantivy-cbbeafa  tantivy-cbbeafab3d8c2ac9      [.] tantivy::fastfield::multivalued::reader::MultiValuedFastFieldReader<Item>::get_vals       ▒


Iter version Static dispatch - Most time spent in sum
running 1 test
test fastfield::bench::bench_multi_value_fflookup                             ... bench:   1,382,918 ns/iter (+/- 32,6

+   96,21%    32,31%  tantivy-ea620f5  tantivy-ea620f5d7d4bf2df      [.] test::bench::Bencher::iter                                                               ▒
+   50,75%    50,73%  tantivy-ea620f5  tantivy-ea620f5d7d4bf2df      [.] <u64 as core::iter::traits::accum::Sum>::sum                                             ▒
+   13,12%    13,10%  tantivy-ea620f5  tantivy-ea620f5d7d4bf2df      [.] tantivy::fastfield::multivalued::reader::MultiValuedFastFieldReader<Item>::get_vals      ▒



Iter version dynamic dispatch
running 1 test
test fastfield::bench::bench_multi_value_fflookup                             ... bench:   1,550,563 ns/iter (+/- 31,509

```

fulmicoton:
@PSeitz Let's stick to the Vec filling solution then. It is much easier to reason with and allocation is usually not an issue (we reuse the same buffer over and over).

By using the iterator solution we are making the job of the CPU and the compiler very hard.

fulmicoton:
@PSeitz Closing this. (Feel free to reopen if I missed something.)

PR #1493, opened 2022-08-29
remove Column impl on Vec

PSeitz:
remove Column impl on Vec to avoid function shadowing


evanxg852000 (approved):
LGTM :+1: , maybe you can check those Clippy warnings  (not related to these changes)

PR #1496, opened 2022-08-31
Adding dragon API to build index without any thread.

fulmicoton:


common/src/writer.rs
```diff
@@ -62,7 +62,7 @@ impl<W: TerminatingWrite> TerminatingWrite for CountingWriter<W> {
 pub struct AntiCallToken(());
 
 /// Trait used to indicate when no more write need to be done on a writer
-pub trait TerminatingWrite: Write + Send {
+pub trait TerminatingWrite: Write + Send + Sync {
```
PSeitz, line 65: Is this a requirement of `SimpleIndexWriter`?

fulmicoton: It is required to make `SimpleWriter: Send + Sync` which itself is required in Quickwit.

src/core/index_simple_writer.rs
```diff
@@ -0,0 +1,47 @@
+use crate::indexer::operation::AddOperation;
+use crate::indexer::segment_updater::save_metas;
+use crate::indexer::SegmentWriter;
+use crate::{Directory, Document, Index, IndexMeta, Opstamp, Segment};
+
+#[doc(hidden)]
+pub struct IndexSimpleWriter {
```
PSeitz: `SingleSegmentIndexWriter` or `IndexSingleSegmentWriter` would be better I think

fulmicoton: I renamed it SingleSegmentIndexWriter

src/core/index.rs
```diff
@@ -877,4 +899,28 @@ mod tests {
         );
         Ok(())
     }
+
+    #[test]
+    fn test_simple_writer() -> crate::Result<()> {
```
shikhar: `test_single_segment_writer`

PSeitz (approved):

PR #1497, opened 2022-08-31
custom num strategy, faster test

### LINKED ISSUE #1486: proptest strategy with better extreme edge cases
The default strategy in proptest does not cover well extreme edge cases, e.g.

` proptest::collection::vec(any::<u64>(), 1..5))`

does very seldom yield something like: 

`
vec![1572656989877777, 1170935903116329, 720575940379279, 0]
`

We should create a custom `Strategy`, to more often pass extreme values, to cover edge cases
```
    fn num_strategy() -> impl Strategy<Value = u64> {
        prop_oneof![
            (u64::MAX-10..64::MAX).prop_map(|num| num as u64),
            (0..10u64).prop_map(|num| num as u64),
            (0..1000u64).prop_map(|_| any::<u64>() )),
        ]
    }
```

PSeitz:
closes #1486
faster test with rand values

### DIFF
--- a/fastfield_codecs/src/lib.rs
+++ b/fastfield_codecs/src/lib.rs
@@ -121,8 +121,9 @@ impl<'a> From<&'a [u64]> for VecColum<'a> {
 
 #[cfg(test)]
 mod tests {
-    use proptest::arbitrary::any;
-    use proptest::proptest;
+    use proptest::prelude::*;
+    use proptest::strategy::Strategy;
+    use proptest::{prop_oneof, proptest};
 
     use crate::bitpacked::BitpackedCodec;
     use crate::blockwise_linear::BlockwiseLinearCodec;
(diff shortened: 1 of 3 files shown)


fulmicoton (approved):

fastfield_codecs/src/lib.rs
```diff
@@ -153,21 +154,32 @@ mod tests {
...
+        fn test_proptest_large(data in proptest::collection::vec(num_strategy(), 1..6000)) {
             create_and_validate::<LinearCodec>(&data, "proptest linearinterpol");
             create_and_validate::<BlockwiseLinearCodec>(&data, "proptest multilinearinterpol");
             create_and_validate::<BitpackedCodec>(&data, "proptest bitpacked");
         }
 
     }
+    fn num_strategy() -> impl Strategy<Value = u64> {
+        prop_oneof![
+            1 => prop::num::u64::ANY.prop_map(|num| u64::MAX - (num % 10) ),
```
fulmicoton, line 178: that's awesome!

PR #1504, opened 2022-09-02
Move to fastfield codec

fulmicoton:


fastfield_codecs/src/serialize.rs
```diff
@@ -0,0 +1,245 @@
...
+fn codec_estimation<C: FastFieldCodec, D: Column>(
+    fastfield_accessor: &D,
+    estimations: &mut Vec<(f32, FastFieldCodecType)>,
+) {
+    if let Some(ratio) = C::estimate(fastfield_accessor) {
+        estimations.push((ratio, C::CODEC_TYPE));
+    }
+}
+
+/// Serialize data into a new u64 fast field. The best compression codec will be chosen
```
PSeitz: the comment doesn't fit the function

fastfield_codecs/src/serialize.rs
```diff
@@ -0,0 +1,245 @@
...
+    let gcd_params_opt = if codecs.contains(&FastFieldCodecType::Gcd) {
+        gcd_params(&column)
+    } else {
+        None
+    };
+
+    let gcd_params = if let Some(gcd_params) = gcd_params_opt {
+        gcd_params
+    } else {
+        return serialize_without_gcd(column, output, codecs);
```
PSeitz, line 143: we could alternatively remove gcd from the codecs and call self

fastfield_codecs/src/serialize.rs
```diff
@@ -0,0 +1,245 @@
...
+        }
+    }
+    output.flush()?;
+    Ok(())
+}
+
+pub fn serialize_and_load<T: MonotonicallyMappableToU64 + Ord + Default>(
+    column: &[T],
+) -> Arc<dyn Column<T>> {
+    let mut buffer = Vec::new();
```
PSeitz, line 218: call `serialize_and_load_given_codecs` with ALL_CODEC_TYPES

fastfield_codecs/src/serialize.rs
fulmicoton: I agree! I think we need to remove gcd from the codecs and treat it differently. (somewhat like you use to do).

Have a footer common to all codecs with num_vals, min_val, max_val, and gcd. If gcd == 1 then we return a reader without any gcd mapper.

fastfield_codecs/src/serialize.rs
fulmicoton: ah right. Actuall serialize_and_load_given_codecs is not used so I removed it.

PSeitz (approved):

PR #1505, opened 2022-09-05
Refact fast field

fulmicoton:
Since there is an estimation part when initially creating multi value fastfields I added a bench to track the regression

```
now
test fastfield::multivalued::bench::bench_multi_value_ff_creation                                                        ... bench:  89,557,762 ns/iter (+/- 14,987,022)
test fastfield::multivalued::bench::bench_multi_value_ff_creation_with_sorting                                           ... bench: 107,875,930 ns/iter (+/- 4,337,126)
test fastfield::multivalued::bench::bench_multi_value_fflookup                                                           ... bench:   1,096,212 ns/iter (+/- 15,691)


before
running 3 tests
test fastfield::multivalued::bench::bench_multi_value_ff_creation                                                        ... bench:  61,929,371 ns/iter (+/- 4,473,032)
test fastfield::multivalued::bench::bench_multi_value_ff_creation_with_sorting                                           ... bench:  60,657,527 ns/iter (+/- 3,043,559)
test fastfield::multivalued::bench::bench_multi_value_fflookup                                                           ... bench:   1,239,103 ns/iter (+/- 22,236)

```

### DIFF
--- a/fastfield_codecs/src/bitpacked.rs
+++ b/fastfield_codecs/src/bitpacked.rs
@@ -1,9 +1,9 @@
 use std::io::{self, Write};
 
-use common::BinarySerializable;
 use ownedbytes::OwnedBytes;
 use tantivy_bitpacker::{compute_num_bits, BitPacker, BitUnpacker};
 
+use crate::serialize::NormalizedHeader;
 use crate::{Column, FastFieldCodec, FastFieldCodecType};
 
 /// Depending on the field type, a different
@@ -12,80 +12,25 @@ use crate::{Column, FastFieldCodec, FastFieldCodecType};
 pub struct BitpackedReader {
     data: OwnedBytes,
     bit_unpacker: BitUnpacker,
-    min_value_u64: u64,
-    max_value_u64: u64,
-    num_vals: u64,
+    normalized_header: NormalizedHeader,
 }
 
 impl Column for BitpackedReader {
     #[inline]
     fn get_val(&self, doc: u64) -> u64 {
-        self.min_value_u64 + self.bit_unpacker.get(doc, &self.data)
+        self.bit_unpacker.get(doc, &self.data)
     }
     #[inline]
     fn min_value(&self) -> u64 {
-        self.min_value_u64
+        0
     }
     #[inline]
     fn max_value(&self) -> u64 {
-        self.max_value_u64
+        self.normalized_header.max_value
     }
     #[inline]
     fn num_vals(&self) -> u64 {
-        self.num_vals
-    }
-}
-pub struct BitpackedSerializerLegacy<'a, W: 'a + Write> {
-    bit_packer: BitPacker,
-    write: &'a mut W,
-    min_value: u64,
-    num_vals: u64,
-    amplitude: u64,
-    num_bits: u8,
-}
-
-impl<'a, W: Write> BitpackedSerializerLegacy<'a, W> {
-    /// Creates a new fast field serializer.
-    ///
-    /// The serializer in fact encode the values by bitpacking
-    /// `(val - min_value)`.
-    ///
-    /// It requires a `min_value` and a `max_value` to compute
-    /// compute the minimum number of bits required to encode
-    /// values.
-    pub fn open(
-        write: &'a mut W,
-        min_value: u64,
-        max_value: u64,
-    ) -> io::Result<BitpackedSerializerLegacy<'a, W>> {
-        assert!(min_value <= max_value);
-        let amplitude = max_value - min_value;
-        let num_bits = compute_num_bits(amplitude);
-        let bit_packer = BitPacker::new();
-        Ok(BitpackedSerializerLegacy {
-            bit_packer,
-            write,
-            min_value,
-            num_vals: 0,
-            amplitude,
-            num_bits,
-        })
-    }
-    /// Pushes a new value to the currently open u64 fast field.
-    #[inline]
-    pub fn add_val(&mut self, val: u64) -> io::Result<()> {
-        let val_to_write: u64 = val - self.min_value;
-        self.bit_packer
-            .write(val_to_write, self.num_bits, &mut self.write)?;
-        self.num_vals += 1;
-        Ok(())
-    }
-    pub fn close_field(mut self) -> io::Result<()> {
-        self.bit_packer.close(&mut self.write)?;
-        self.min_value.serialize(&mut self.write)?;
-        self.amplitude.serialize(&mut self.write)?;
-        self.num_vals.serialize(&mut self.write)?;
-        Ok(())
+        self.normalized_header.num_vals
     }
 }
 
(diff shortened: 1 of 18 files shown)


fastfield_codecs/src/serialize.rs
```diff
@@ -187,25 +209,25 @@ fn serialize_without_gcd(
     // removing nan values for codecs with broken calculations, and max values which disables
     // codecs
     estimations.retain(|estimation| !estimation.0.is_nan() && estimation.0 != f32::MAX);
-    estimations.sort_by(|a, b| a.0.partial_cmp(&b.0).unwrap());
-    let (_ratio, codec_type) = estimations[0];
+    estimations
+        .sort_by(|(score_left, _), (score_right, _)| score_left.partial_cmp(&score_right).unwrap());
```
PSeitz, line 221:
```suggestion
        .sort_by(|(score_left, _), (score_right, _)| score_left.total_cmp(&score_right));
```

fastfield_codecs/src/serialize.rs
```diff
@@ -187,25 +209,25 @@ fn serialize_without_gcd(
...
+    Some(estimations.first()?.1)
+}
 
-    write_header(codec_type, output)?;
+fn serialize_given_header(
```
PSeitz:
```suggestion
fn serialize_with_codec(
```

PSeitz: there's no header parameter

src/fastfield/multivalued/writer.rs
```diff
@@ -200,24 +184,119 @@ impl MultiValuedFastFieldWriter {
...
+        if let Some((min, max)) = self.min_max_opt.lock().unwrap().clone() {
+            return (min, max);
+        }
+        let (min, max) = tantivy_bitpacker::minmax(self.iter()).unwrap_or((0u64, 0u64));
+        *self.min_max_opt.lock().unwrap() = Some((min, max));
+        (min, max)
+    }
+}
+impl<'a, C: Column> Column for MultivalueStartIndex<'a, C> {
+    fn get_val(&self, idx: u64) -> u64 {
```
PSeitz, line 247: Estimations of the codecs sample in 5% steps and call `get_val`, so this is quite expensive (10x full-scan in total). But this is much less error prone than the approach I used elsewhere with the reverse computation. We could handle `get_vals` the same everywhere like here and precompute the sampled values once via `iter` and pass them to the codec estimates.

fastfield_codecs/src/serialize.rs
fulmicoton: renamed.

src/fastfield/multivalued/writer.rs
fulmicoton: I think before we were always picking the bitpacker... which is a bit sad considering the index of  multivalued stuff can benefit from the linear codec the most.

PSeitz: Only in the initial creation, the merge code had autodetection

src/fastfield/multivalued/writer.rs
```diff
@@ -200,24 +184,119 @@ impl MultiValuedFastFieldWriter {
...
+    }
+
+    fn iter<'b>(&'b self) -> Box<dyn Iterator<Item = u64> + 'b> {
+        Box::new(MultivalueStartIndexIter {
+            column: &self.column,
+            doc_id_map: self.doc_id_map,
+            new_doc_id: 0,
+            offset: 0,
+        })
+    }
```
PSeitz, line 284: We can use iter map here
```suggestion
        let mut offset: u64 = 0;
        Box::new(
            std::iter::once(0_u64).chain(self.doc_id_map.iter_old_doc_ids().map(
                move |old_docid| {
                    let start_pos = self.column.get_val(old_docid as u64);
                    let end_pos = if old_docid as u64 + 1 >= self.column.num_vals() {
                        self.column.num_vals() as u64
                    } else {
                        self.column.get_val(old_docid as u64 + 1)
                    };
                    offset += end_pos - start_pos;
                    offset
                },
            )),
        )

```

fastfield_codecs/src/bitpacked.rs
```diff
@@ -12,80 +12,25 @@ use crate::{Column, FastFieldCodec, FastFieldCodecType};
...
 impl Column for BitpackedReader {
     #[inline]
     fn get_val(&self, doc: u64) -> u64 {
-        self.min_value_u64 + self.bit_unpacker.get(doc, &self.data)
+        self.bit_unpacker.get(doc, &self.data)
     }
     #[inline]
     fn min_value(&self) -> u64 {
-        self.min_value_u64
+        0
```
PSeitz, line 26: why is that correct?

fastfield_codecs/src/serialize.rs
```diff
@@ -46,125 +45,149 @@ fn codec_estimation<C: FastFieldCodec, D: Column>(
     }
 }
 
-fn write_header<W: io::Write>(codec_type: FastFieldCodecType, output: &mut W) -> io::Result<()> {
-    codec_type.to_code().serialize(output)?;
-    Ok(())
+#[derive(Debug, Copy, Clone)]
+pub struct NormalizedHeader {
```
PSeitz, line 49: GCDNormalizedHeader

fastfield_codecs/src/bitpacked.rs
fulmicoton: Great catch

fastfield_codecs/src/serialize.rs
```diff
@@ -21,107 +21,153 @@ use std::io;
...
     bytes: OwnedBytes,
+    header: &Header,
 ) -> io::Result<Arc<dyn Column<Item>>> {
-    let reader = open_gcd_from_bytes::<C>(bytes)?;
-    Ok(Arc::new(monotonic_map_column(reader, Item::from_u64)))
+    let normalized_header = header.normalized();
+    let reader = C::open_from_bytes(bytes, normalized_header)?;
+    let min_value = header.min_value;
+    if let Some(gcd) = header.gcd {
+        let monotonic_mapping = move |val: u64| Item::from_u64(min_value + val * gcd.get());
```
PSeitz, line 149: how the stats like `min_value` are handled is a little opaque now. We could use the common header for every fastfield as datasource, instead of wrapping and computation

src/fastfield/multivalued/writer.rs
```diff
@@ -200,24 +184,169 @@ impl MultiValuedFastFieldWriter {
...
+                    doc_id_map: self.doc_id_map,
+                    new_doc_id: 0,
+                    offset: 0u64,
+                },
+                seek_next_id: 0u64,
+            };
+        }
+        let to_skip = idx - random_seeker_lock.seek_next_id;
+        random_seeker_lock.seek_next_id = idx + 1;
+        random_seeker_lock.seek_head.nth(to_skip as usize).unwrap()
```
PSeitz, line 262: I think assuming that get_val calls always come in order is too dangerous. Multiple codec estimations would already violate that assumption
Update: I missed the reset part above

PSeitz (approved):

PR #1509, opened 2022-09-07
refactor, fix api

PSeitz:
refactor
fix clippy
fix docs
remove unused code
fix bytesfield index api

### DIFF
--- a/src/fastfield/serializer/mod.rs
+++ b/src/fastfield/serializer/mod.rs
@@ -1,6 +1,6 @@
 use std::io::{self, Write};
 
-use common::{BinarySerializable, CountingWriter};
+use common::CountingWriter;
 pub use fastfield_codecs::{Column, FastFieldStats};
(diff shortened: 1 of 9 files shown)


src/fastfield/serializer/mod.rs
```diff
@@ -84,13 +71,14 @@ impl CompositeFastFieldSerializer {
         Ok(())
     }
 
-    /// Start serializing a new [u8] fast field
-    pub fn new_bytes_fast_field_with_idx(
+    /// Start serializing a new [u8] fast field.
+    ///
+    /// The bytes will be stored as is, no compression will be applied.
+    pub fn new_bytes_fast_field(
```
fulmicoton: Nothing in the method name explains that it is return the thing that writes the concatenate fast bytes value, 
and that an index must be written separately?

Could you at least explain that in a comment and also ditch `FastBytesFieldSerializer` and just return
the CountingWriter?

fulmicoton (approved):
Approved with one minor comment

PR #1510, opened 2022-09-07
Allow for a same-thread doc compressor.

fulmicoton:
In addition, it isolates the doc compressor logic, better reports io::Result.

In the case of the same-thread doc compressor,
the blocks are also not copied.

### DIFF
--- a/src/core/index_meta.rs
+++ b/src/core/index_meta.rs
@@ -235,6 +235,14 @@ impl InnerSegmentMeta {
     }
 }
 
+fn return_true() -> bool {
+    true
+}
+
+fn is_true(val: &bool) -> bool {
+    *val
+}
+
 /// Search Index Settings.
 ///
 /// Contains settings which are applied on the whole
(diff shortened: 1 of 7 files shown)


src/core/index_meta.rs
```diff
@@ -235,6 +235,14 @@ impl InnerSegmentMeta {
     }
 }
 
+fn return_true() -> bool {
```
PSeitz, line 238:
```suggestion
fn docstore_compress_dedicated_thread() -> bool {
```

src/core/index_meta.rs
```diff
@@ -248,6 +256,12 @@ pub struct IndexSettings {
     /// The `Compressor` used to compress the doc store.
     #[serde(default)]
     pub docstore_compression: Compressor,
+    /// If set to true, docstore compression will happen on a dedicated thread.
+    /// (defaults: true)
+    #[doc(hidden)]
+    #[serde(default = "return_true")]
```
PSeitz, line 262:
```suggestion
    #[serde(default = "docstore_compress_dedicated_thread")]
```

src/store/store_compressor.rs
```diff
@@ -0,0 +1,267 @@
...
+use crate::store::{Compressor, Decompressor, StoreReader};
+use crate::DocId;
+
+pub struct BlockCompressor(Impls);
+
+// The struct wrapping an enum is just here to keep the
+// impls private.
+enum Impls {
+    SameThread(BlockCompressorImpl),
+    DedicatedThread(DedicatedThreadBlockCompressorImpl),
```
PSeitz: I think the naming could be improved here a little
```suggestion
enum BlockCompressorVariants {
    SameThread(SameThreadBlockCompressor),
    DedicatedThread(DedicatedThreadBlockCompressor),
```

PSeitz (approved):
LGTM except some naming comments

Issue #999, opened 2021-03-28
Storing arbitrary attributes for segments


ppodolsky:
In my project search index files are deployed through something like torrent technology. New replicas are asking by broadcast who can seed segments identified by X, Y, Z etc and then leeches all these segments.

Merging removes old segments so I have to choose between enabled merging and possibility to seed old segments.

My proposal is to give a way to store some extra key-value data for segments. It would allow to mark some segments as i.e. "frozen" = true and to omit these segments while calcualting merge policies.

Probably there is a better way. What is you thought about it? I could try to implement by myself so feel free to choose from options that are really suited better for further development.

fulmicoton:
Attaching meta data to segments and exposing them to the merge policy seems like a great idea.

Can you work on an actual spec?

ppodolsky:
Sure, but it seems we need to work out a solution in #971 firstly

ppodolsky:
@fulmicoton I've come back here to the issue as I'm in need of this now:)
I've taken a look over various ways and here are some initial thoughts:
- `meta.json` already stores some attributes for each segment
- Another way is to store attributes in separate file within segment

I'm slighly shifted toward having a separate file (something like `.props`). But it could be confusing taking into account that some attributes are already stored in `meta.json`

The main advantage of moving all attributes (`max_doc`, `deletes` and others which we would like to have in the future) to a separate file is that segments could be distributed between servers in easier way. Having all segment metadata inside the `.props` file means that you could attach new segment to the index by merely putting files into index directory and adding `segment_id` to the list of actual index segments in `meta.json`

PSeitz:
One thing that would be interesting for the quickwit use case is to store values in fields per segment if you have a limited number of values.

This information can be used for pruning or in merge options that prefer to merge same values together. This wouldn't be arbitrary information, but should probably stored in the same file.

PR #1512, opened 2022-09-07
[feat] SegmentAttributes

ppodolsky:
Related to #999 
This PR is a PoC. I've implemented following approach:
- `SegmentAttribute` enum, different options represent different types + different strategies of merging while merging segments
- Every segment has `HashMap<String, SegmentAttribute>` aka `SegmentAttributes` in its metadata
- `IndexSettings` has `SegmentAttributesConfig` that is used to create `SegmentAttributes` for inserting into every segment on segment creation.
- While merging we additionally merge all `SegmentAttributes` and set them to a new segment.

@PSeitz @fulmicoton Could you review? I haven't implemented tests or helper functions because not sure if Tantivy still needs this feature at all.


fmassot:
Thank you @ppodolsky for this PR! We were a bit overwhelmed these days :)
I don't have enough knowledge to say if such a feature can be added so I let @fulmicoton and @PSeitz answer on this.

But when looking a the code, I was not very fond of the enum, why not using a Trait for segment attributes?

ppodolsky:
Hi @fmassot ! Thank you for reply!

I can rewrite it to traits but I'd prefer to stabilize the approach verbally at first if the feature is still needed. 
Further I'd like to confirm that putting segment attributes config in `IndexMeta` and segment attributes in meta.json (not in separate file) is OK.

src/core/segment_attributes.rs
```diff
@@ -0,0 +1,194 @@
...
+    DisjunctiveBool(bool),
+    StringList(Vec<String>),
+}
+
+impl SegmentAttribute {
+    pub fn merge(&mut self, other: &SegmentAttribute) {
+        match self {
+            SegmentAttribute::ConjunctiveBool(a) => match other {
+                SegmentAttribute::ConjunctiveBool(b) => *a = *a && *b,
+                _ => unreachable!(),
```
fulmicoton: surely unreachable is incorrect.

We cannot call use unreachable for something that is enforced by the user writing consistent correct JSON configuration.

ppodolsky: Sure, I can reimplement in traits, we've discussed it above with @fmassot  
Is the overall approach ok? I mean storing attributes in meta.json and introducing concept of attribute types containing implementation of their merge behavior?

ppodolsky:
@fmassot
I have tried to implement it via traits and didn't like the way it looks. The main issue is that we need to store type. Enums are well suited for this purpose because they are serializable and naturally aligned with serialization format like 
`"<type_name>": <value>`.
Trait objects should be serialized in non-trivial way and we have to map and track trait and its serializable name for deserialization purposes

src/core/segment_attributes.rs
ppodolsky: I've left implementation using enums and fixed `unreachable!`

fmassot:
@ppodolsky I read again your PR and yes enum is fine per attribute... My thought about trait was on a higher level, I mean on the SegmentAttributes level. I don't have a genuine opinion on how to implement correctly this feature, and I did not want to add some noise here, sorry.

Concerning serialization/deserialization of traits, this can done with https://github.com/dtolnay/typetag, I believe it is just using an enum behind the scene.

PSeitz:
Currently the behaviors are provided by tantivy. As an alternative, we could have a registry.
```
global_registry.register_attribute(name, &str, attribute: Attribute); // name has to be unique

trait Attribute{
    fn merge(attrs: Vec<serde_json::Value>) -> serde_json::Value;
}
```

is_frozen attribute, ConjunctiveBool could be provided by tantivy.
```
{
     "is_frozen": true
}


register_attribute("is_frozen", ConjunctiveBool)
```

Not sure if this too unwieldy.

ppodolsky:
@PSeitz Yes, I came to this design while trying to reimplement it with traits. I was slightly confused by using serde_json::Value as value for segment attribute, not sure if tying to specific serialization format is worth to do. And without serde_json::Value there will be a need to create a Value representation which would be enum again.

What do you think?

PSeitz:
> @PSeitz Yes, I came to this design while trying to reimplement it with traits. I was slightly confused by using serde_json::Value as value for segment attribute, not sure if tying to specific serialization format is worth to do. And without serde_json::Value there will be a need to create a Value representation which would be enum again.
> 
> What do you think?

The enum would look like serde_json::Value, so we can just use that. serde_json::Value is not tied to json de/serialization, it's implementing the De/Serialization traits.

ppodolsky:
Ok but still have questions. Is global registry ok? It will bind a set of available attributes to the compiled version of application. In the approach implemented here indices can manage a set of attributes by their own.

ppodolsky:
Also, trait objects are not very friendly with OnceCell or whatever would be used for the global registry. It will require synchronization too.

ppodolsky:
Finally, safe trait objects won't be able to incapsulate default values for attributes. Putting Value inside trait object will make trait non-safe, it [would require](https://stackoverflow.com/questions/50021897/how-to-implement-serdeserialize-for-a-boxed-trait-object) manual implemenation of Serialize/Deserialize traits and so on. I honestly tried to implement it and might miss something but implementation looks rather messy

PSeitz:
> Ok but still have questions. Is global registry ok? It will bind a set of available attributes to the compiled version of application. In the approach implemented here indices can manage a set of attributes by their own.

True, the registry should probably have some attribute_name attribute association. The predefined attribute names would be reserved. So the current json format could stay the same, but `conjunctive_bool` would be an attribute_name.

 `{"is_frozen:{conjunctive_bool:false}}`

> Finally, safe trait objects won't be able to incapsulate default values for attributes. Putting Value inside trait object will make trait non-safe, it [would require](https://stackoverflow.com/questions/50021897/how-to-implement-serdeserialize-for-a-boxed-trait-object) manual implemenation of Serialize/Deserialize traits and so on. I honestly tried to implement it and might miss something but implementation looks rather messy

For de/serialization, you can just use the serde_json::Value. 

> Putting Value inside trait object will make trait non-safe

Why put Value inside the trait object, it's just merge behavior.

```rust
trait Attribute{
    fn merge(attrs: Vec<serde_json::Value>) -> serde_json::Value; // There's no self
}
```

> Finally, safe trait objects won't be able to incapsulate default values for attributes. 

Why do you need default, only for merge right?

```rust
trait Attribute{
    // handle default internally
    fn merge(attrs: Vec<Option<serde_json::Value>>) -> serde_json::Value;
}
```

fulmicoton:
No global registry please.

ppodolsky:
> Why do you need default, only for merge right?

Default is needed for any newly created segment, not just merge.
For example, this `is_frozen` attribute should be set to false for new segments, manually turned to true during specific event (in my case it is index publication that freezes segments and should prevent them from further merging) and again, set to false after any merge (can still happen if run it manually). Here the default is false.

fulmicoton:
@ppodolsky The idea to have the serialized object embed the logic of how things should be merged is not interesting, but I don't think we want that here.

Here is a counter proposition...

What about you have a public trait.
```
pub trait SegmentAttribute: Default + Serializer + Deserializer {
     fn merge(segment_attributes: &Self);
}
```

Internally to tantivy, you could then have a `SegmentAttributeMergerImpl`.

```
struct SegmentAttributeMergerImpl<S: SegmentAttribute>;

```

And a boxable trait (also private to tantivy)
```
trait SegmentAttributeMerger {
    fn merge_json(&self, segment_attributes_json: &[serde_json::Value]) -> serde_json::Value;
}
```

The index then would have public method

```
fn set_segment_attr::<S: SegmentAttribute>() {
    let typed_segment_attr_merger: SegmentAttributeMergerImpl<S> = SegmentAttributeMergerImpl;
    self.segment_attribute_merger = Some(Box::new(typed_segment_attr_merger));
}
```

Note that we are not making Index a generic here.

ppodolsky:
@fulmicoton Done :)

ppodolsky:
Friendly bump ⬆️
@fulmicoton

ppodolsky:
Bump:)

ppodolsky:
Regular bump @PSeitz @fulmicoton

src/core/segment_attributes.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+
+/// SegmentAttributes implementation owns custom segment attributes and its merging behavior
+pub trait SegmentAttributes: Default + Serialize + DeserializeOwned + Send + Sync + Clone {
+    /// Must be implemented for defining how to merge `SegmentAttributes` from
+    /// different segments
+    fn merge(segments_attributes: Vec<Self>) -> Self;
+}
```
fulmicoton:
```suggestion
```

src/core/segment_attributes.rs
```diff
@@ -0,0 +1,93 @@
...
+    fn box_clone(&self) -> Box<dyn SegmentAttributesMerger> {
+        Box::new((*self).clone())
+    }
+}
+
+impl Clone for Box<dyn SegmentAttributesMerger> {
+    fn clone(&self) -> Box<dyn SegmentAttributesMerger> {
+        self.box_clone()
+    }
+}
```
fulmicoton:
```suggestion
```

src/core/mod.rs
```diff
@@ -21,6 +22,9 @@ pub use self::index_meta::{
 pub use self::inverted_index_reader::InvertedIndexReader;
 pub use self::searcher::{Searcher, SearcherGeneration};
 pub use self::segment::Segment;
+pub use self::segment_attributes::{
+    SegmentAttributes, SegmentAttributesMerger, SegmentAttributesMergerImpl,
```
fulmicoton:
```suggestion
    SegmentAttributes, SegmentAttributesMerger```

File: src/core/index.rs [hunk 5]
```diff
@@ -401,9 +404,21 @@ impl Index {
             tokenizers: TokenizerManager::default(),
             executor: Arc::new(Executor::single_thread()),
             inventory,
+            segment_attributes_merger: None,
         }
     }
 
+    /// Create and set an instance of SegmentAttributes
+    pub fn set_segment_attributes_merger<S: SegmentAttributes + 'static>(&mut self) {
```
```suggestion
    pub fn set_segment_attributes_merger(&mut self, segment_attribute_merger: Box<dyn SegmentAttributeMerger>) {
```

src/core/index.rs
```diff
@@ -401,9 +404,21 @@ impl Index {
...
+        let typed_segment_attributes_merger = SegmentAttributesMergerImpl::<S>::new();
+        self.segment_attributes_merger = Some(Box::new(typed_segment_attributes_merger));
+    }
+
+    /// Accessor for SegmentAttributes
+    pub fn segment_attributes_merger(&self) -> &Option<Box<dyn SegmentAttributesMerger>> {
```
```suggestion
    pub fn segment_attributes_merger(&self) -> Option<&dyn SegmentAttributesMerger> {
```

src/indexer/merge_operation.rs
```diff
@@ -46,17 +46,20 @@ pub struct MergeOperation {
 pub(crate) struct InnerMergeOperation {
     target_opstamp: Opstamp,
     segment_ids: Vec<SegmentId>,
+    segment_attributes: Option<serde_json::Value>,
```
fulmicoton: Is it the `merged_segment_attributes`? Can we rename it that way?

src/indexer/segment_updater.rs
```diff
@@ -91,6 +91,7 @@ fn merge(
     index: &Index,
     mut segment_entries: Vec<SegmentEntry>,
     target_opstamp: Opstamp,
+    segment_attributes: Option<serde_json::Value>,
```
fulmicoton: I don't understand what this is thing is. Is it to override the merge segment attributes?

src/indexer/index_writer.rs
```diff
@@ -513,7 +513,22 @@ impl IndexWriter {
     ///
     /// `segment_ids` is required to be non-empty.
     pub fn merge(&mut self, segment_ids: &[SegmentId]) -> FutureResult<Option<SegmentMeta>> {
-        let merge_operation = self.segment_updater.make_merge_operation(segment_ids);
+        self.merge_with_attributes(segment_ids, None)
+    }
+
+    /// Merges a given list of segments and set attributes in merged segment.
```
fulmicoton, line 516: This is unclear what segment_attributes is, and what happens if it is set to None.

ppodolsky:
@fulmicoton Done, sorry for being late.
Also, had to turn Box into Arc because you suggested to remove `impl Clone for Box<...>` etc.

src/indexer/index_writer.rs
ppodolsky: Fixed

src/indexer/segment_updater.rs
ppodolsky: Fixed

src/indexer/merge_operation.rs
ppodolsky: Renamed

src/core/index.rs
ppodolsky: Applier

PR #1513, opened 2022-09-08
add ip codec

PSeitz:


common/src/vint.rs
```diff
@@ -5,6 +5,75 @@ use byteorder::{ByteOrder, LittleEndian};
...
+        } else {
+            output.push(next_byte);
+        }
+    }
+}
+
+/// Deserializes a u128 number
+///
+/// Returns the number and the slice after the vint
+pub fn deserialize_vint_u128(data: &[u8]) -> io::Result<(u128, &[u8])> {
```
fulmicoton, line 25: can we use generic to have vint for u64 and u128?

fulmicoton: (if it is too much work don't worry about it)

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+    }
+}
+impl PartialOrd for BlankSizeAndPos {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        self.blank_size.partial_cmp(&other.blank_size)
+    }
+}
+
+#[test]
+fn test_delta_and_pos_sort() {
```
fulmicoton: unit tests in a guarded module please

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+/// A blank is an unoccupied space in the data.
+/// E.g. [100, 201] would have a `BlankAndPos{ pos: 0, blank_size: 100}`
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+struct BlankSizeAndPos {
+    blank_size: u128,
+    /// Position in the sorted data.
+    pos: usize,
+}
+impl BlankSizeAndPos {
+    fn new(blank: u128, pos: usize) -> Self {
```
fulmicoton: Constructor with two nums are really a pain to read and are error prone.

Can you just use the struct construction maybe?

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+impl PartialOrd for BlankSizeAndPos {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        self.blank_size.partial_cmp(&other.blank_size)
+    }
+}
+
+#[test]
+fn test_delta_and_pos_sort() {
+    let mut deltas: BinaryHeap<BlankSizeAndPos> = BinaryHeap::new();
+    deltas.push(BlankSizeAndPos::new(10, 1));
```
fulmicoton: if the goal is to test Ord, let's test Ord. What is the point of the Binary heap here?

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+    assert_eq!(deltas.pop().unwrap().blank_size, 100);
+    assert_eq!(deltas.pop().unwrap().blank_size, 10);
+}
+
+/// Put the deltas for the sorted values into a binary heap
+fn get_deltas(values_sorted: &[u128]) -> BinaryHeap<BlankSizeAndPos> {
+    let mut prev_opt = None;
+    let mut deltas: BinaryHeap<BlankSizeAndPos> = BinaryHeap::new();
+    for (pos, value) in values_sorted.iter().cloned().enumerate() {
+        let delta = if let Some(prev) = prev_opt {
```
fulmicoton: rename `blank_size` and change the formula.

Right now it is fed into the `BlankSizeAndPos` struct, but semantically `delta` is not a blank size.

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+}
+impl BlankCollector {
+    fn new() -> Self {
+        Self {
+            blanks: vec![],
+            staged_blanks_sum: 0,
+        }
+    }
+    fn stage_blank(&mut self, blank: BlankSizeAndPos) {
+        self.staged_blanks_sum += blank.blank_size - 1;
```
fulmicoton: fix the semantics of blank_size.

If you have a value of 10 and a value of 12, the blank_size should be 1.

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+    }
+    fn num_blanks(&self) -> usize {
+        self.blanks.len()
+    }
+}
+
+/// Will collect blanks and add them to compact space if more bits are saved than cost from
+/// metadata.
+fn get_compact_space(values_sorted: &[u128], cost_per_blank: usize) -> CompactSpace {
+    let max_val = *values_sorted.last().unwrap_or(&0u128) + 1;
```
fulmicoton: that's not max_val. (tantivy uses this weird naming in max_doc to align itself with Lucene)

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+    let mut deltas = get_deltas(values_sorted);
+    let mut amplitude_compact_space = max_val;
+    let mut amplitude_bits: u8 = (amplitude_compact_space as f64).log2().ceil() as u8;
```
fulmicoton: no floating points arithmetics

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+            compact_space.add_blank(blank_start..=blank_end);
+        }
+    }
+    compact_space.add_blank(max_val..=u128::MAX);
+
+    compact_space.finish()
+}
+
+#[test]
+fn compact_space_test() {
```
fulmicoton: please move the unit test at the bottom

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+    let ips = vec![
```
fulmicoton:
```suggestion
    let ips = &[
```

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+        2u128, 4u128, 1000, 1001, 1002, 1003, 1004, 1005, 1008, 1010, 1012, 1260,
+    ];
+    let compact_space = get_compact_space(&ips, 11);
+    assert_eq!(compact_space.null_value, 5);
+    let amplitude = compact_space.amplitude_compact_space();
+    assert_eq!(amplitude, 20);
+    assert_eq!(2, compact_space.to_compact(2).unwrap());
+    assert_eq!(compact_space.to_compact(100).unwrap_err(), 0);
+
+    let mut output = vec![];
```
fulmicoton:
```suggestion
    let mut output: Vec<u8> = Vec::new();
```

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+impl CompactSpaceBuilder {
+    /// Creates a new compact space builder which will initially cover the whole space.
+    fn new() -> Self {
+        Self {
+            covered_space: vec![0..=u128::MAX],
+        }
+    }
+
+    /// Will extend the first range and assign the null value to it.
+    fn assign_and_return_null(&mut self) -> u128 {
```
fulmicoton: Is there an extra cost 0 to using 0 as the null value?

I suspect it might help the simplicity of the code.

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+        *self.covered_space[0].end()
+    }
+
+    /// Assumes that repeated add_blank calls don't overlap, which will be the case on sorted
+    /// values.
+    fn add_blank(&mut self, blank: RangeInclusive<u128>) {
+        let position = self
+            .covered_space
+            .iter()
+            .position(|range| range.start() <= blank.start() && range.end() >= blank.end());
```
fulmicoton:
```suggestion
            .position(|range| range.contains(&blank.start()));
```

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+        if let Some(position) = position {
```
fulmicoton:
```suggestion
        if let Some(position) = position {
             assert!(self.covered_space[position].end() >= blank.end());
```

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+            let old_range = self.covered_space.remove(position);
+            // Exact match, just remove
+            if old_range == blank {
+                return;
+            }
+            let new_range_end = blank.end().saturating_add(1)..=*old_range.end();
+            if old_range.start() == blank.start() {
+                self.covered_space.insert(position, new_range_end);
```
fulmicoton: Can we use a `BTreeMap<u128, u128>` ?

key: start_offset, end: end_offset?

(the point is to fix the quadratic complexity)

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+            }
+            self.covered_space.insert(position, new_range_end);
+            self.covered_space.insert(position, new_range_start);
+        }
+    }
+    fn finish(mut self) -> CompactSpace {
+        let null_value = self.assign_and_return_null();
+
+        let mut compact_start: u64 = 0;
+        let mut ranges_and_compact_start = vec![];
```
fulmicoton: capacity.

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+        for cov in self.covered_space {
+            let covered_range_len = cov.end() - cov.start();
```
fulmicoton: why not `cov.end() - cov.start() + 1` ?
Range have a len via ExactIterator.

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,783 @@
...
+            ranges_and_compact_start.push((cov, compact_start));
+            compact_start += covered_range_len as u64 + 1;
```
fulmicoton: Why do you put the +1 here?

fulmicoton (changes requested):
See comment inline.

fastfield_codecs/src/compact_space.rs
PSeitz: If we choose 0 we need to shift all numbers by one, but we can't shift if U128::MAX is occupied.

fastfield_codecs/src/compact_space.rs
PSeitz: I wanted to test that `heap.pop` gives descending blank_size. renamed the test

fastfield_codecs/src/compact_space.rs
PSeitz: done

fastfield_codecs/src/compact_space.rs
PSeitz: yeah should be part of covered_range_len

fastfield_codecs/src/compact_space.rs
PSeitz: ExactIterator is not implemented for u128

fastfield_codecs/src/compact_space.rs
PSeitz: I'll replace the algorithm with something better

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,943 @@
...
+/// Put the blanks for the sorted values into a binary heap
+fn get_blanks(values_sorted: &[u128]) -> BinaryHeap<BlankRange> {
+    let mut blanks: BinaryHeap<BlankRange> = BinaryHeap::new();
+    let mut add_range = |blank_range: RangeInclusive<u128>| {
+        let blank_range: Result<BlankRange, _> = blank_range.try_into();
+        if let Ok(blank_range) = blank_range {
+            blanks.push(blank_range);
+        }
+    };
+    for values in values_sorted.windows(2) {
```
fulmicoton: Itertools has tuple_windows.

```
for (previous, next) in values_sorted.tuple_windows() {

}
```

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,943 @@
...
+/// The cost per blank is quite hard actually, since blanks are delta encoded, the actual cost of
+/// blanks depends on the number of blanks.
+///
+/// The number is taken by looking at a real dataset. It is optimized for larger datasets.
+const COST_PER_BLANK_IN_BITS: usize = 36;
+
+/// The range of a blank in value space.
+///
+/// A blank is an unoccupied space in the data.
+/// Ordered by size
```
fulmicoton: Can you document the condition you are enforcing here?

(Ranges have a len greater or equal to 3)

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,943 @@
...
+        if blank_size < 2 {
+            Err("invalid range")
+        } else {
+            Ok(BlankRange { blank_range: range })
+        }
+    }
+}
+impl BlankRange {
+    fn blank_size(&self) -> u128 {
+        self.blank_range.end() - self.blank_range.start()
```
fulmicoton: still wrong.

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,943 @@
...
+        let blank_range: Result<BlankRange, _> = blank_range.try_into();
+        if let Ok(blank_range) = blank_range {
+            blanks.push(blank_range);
+        }
+    };
+    for values in values_sorted.windows(2) {
+        let blank_range = values[0] + 1..=values[1] - 1;
+        add_range(blank_range);
+    }
+    if let Some(first_val) = values_sorted.first().filter(|first_val| **first_val != 0) {
```
fulmicoton: nitpick... 

The use of filter here is correct but makes it difficult to read.

```
if let Some(first_val) = values_sorted.first() {
    if first_val > 0 {
       // ...
    }
}
```

would be easier.

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,943 @@
...
+        let blank_range = 0..=first_val - 1;
+        add_range(blank_range);
+    }
+
+    if let Some(last_val) = values_sorted
+        .last()
+        .filter(|last_val| **last_val != u128::MAX)
```
fulmicoton: same as above.

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,943 @@
...
+        Self {
+            blanks: vec![],
+            staged_blanks_sum: 0,
+        }
+    }
+    fn stage_blank(&mut self, blank: BlankRange) {
+        self.staged_blanks_sum += blank.blank_size();
+        self.blanks.push(blank);
+    }
+    fn drain(&mut self) -> std::vec::Drain<'_, BlankRange> {
```
fulmicoton: nitpick: `impl Iterator` is quite nice for that kind of method.

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,683 @@
...
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct CompactSpace {
+    ranges_and_compact_start: Vec<(RangeInclusive<u128>, u64)>,
+    pub null_value: u128,
+}
+
+impl BinarySerializable for CompactSpace {
+    fn serialize<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
+        VIntU128(self.null_value).serialize(writer)?;
+        VIntU128(self.ranges_and_compact_start.len() as u128).serialize(writer)?;
```
fulmicoton: why u128 for the len?

fastfield_codecs/src/compact_space.rs
PSeitz: yes I had to switch to it because of BTreeSet

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,683 @@
...
+
+        Ok(Self {
+            null_value,
+            ranges_and_compact_start,
+        })
+    }
+}
+
+impl CompactSpace {
+    fn amplitude_compact_space(&self) -> u128 {
```
fulmicoton: I am not sure how you defined amplitude here so I cannot readproof this.

Can you add a comment here.

e.g.
```
Amplitude is a number such that the compact space consists including the sentinel value used to identify null values.
 
The compact space is `0..=amplitude` .
```

fastfield_codecs/src/compact_space.rs
PSeitz: done

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,683 @@
...
+        let last_range = &self.ranges_and_compact_start[self.ranges_and_compact_start.len() - 1];
+        last_range.1 as u128 + (last_range.0.end() - last_range.0.start()) + 1
```
fulmicoton: why don't we use the `null_value` directly here?

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,683 @@
...
+    fn compact_to_u128(&self, compact: u64) -> u128 {
+        self.params.compact_space.unpack(compact)
+    }
+
+    /// Comparing on compact space: 1.2 GElements/s
+    ///
+    /// Comparing on original space: .06 GElements/s (not completely optimized)
+    pub fn get_range(&self, mut range: RangeInclusive<u128>) -> Vec<u64> {
+        if range.start() > range.end() {
+            range = *range.end()..=*range.start();
```
fulmicoton: The current code makes no sense.
```suggestion
    return Vec::new();
```

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,683 @@
...
+        }
+        let from_value = *range.start();
+        let to_value = *range.end();
+        assert!(to_value >= from_value);
+        let compact_from = self.to_compact(from_value);
+        let compact_to = self.to_compact(to_value);
+
+        // Quick return, if both ranges fall into the same non-mapped space, the range can't cover
```
fulmicoton: It is either a bug or a mistake....

Do we always map small values? What if values  are like this:

```
|<  big blank space >|   ...a lot of values ...|< big blank space >|
```

If we don't always map small values, what if the range asked is 0..u128::MAX

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,683 @@
...
+            }
+        }
+
+        positions
+    }
+
+    #[inline]
+    fn iter_compact(&self) -> impl Iterator<Item = u64> + '_ {
+        (0..self.params.num_vals)
+            .map(move |idx| self.params.bit_unpacker.get(idx as u64, &self.data) as u64)
```
fulmicoton: we can probably optimize this.

fastfield_codecs/src/compact_space.rs
PSeitz: just convenience, I change it to `VInt`

fastfield_codecs/src/compact_space.rs
PSeitz: `null_value` is not the same as `null_value_compact_space`. null_value_compact_space is not the last value if u128::MAX is used in the data

fastfield_codecs/src/compact_space.rs
PSeitz: We don't map always 0, that's not what the comment says.

`|<  big blank space >|   ...a lot of values ...|< big blank space >|`

That's the actual data, but incoming requests can cover the whole space. If a range request is incoming for the same `big blank space`, we can early exit.

fastfield_codecs/src/compact_space.rs
PSeitz: loop unrolling or do have something else in mind?

fastfield_codecs/src/compact_space.rs
PSeitz: Added a comment, which is pretty much what you proposed

fastfield_codecs/src/compact_space.rs
PSeitz: fixed

fastfield_codecs/src/compact_space.rs
PSeitz: I did some unrolling, it's 3 times faster. 

Currently the compact space is u64. We could switch to u32. It's unlikely that we hit the u32 boundary, but it can happen (around 1 billion elements in a segment with worse case number spreading).

To apply some simd, we would need to copy the data blockwise into an array. I suspect it will not help though. I never tried to intersperse simd with normal code.

In the bitpacker there is a check which we should get rid of in the loop
```
        if self.num_bits == 0 {
            return 0u64;
        }
```

fastfield_codecs/src/compact_space/blank_range.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+impl Ord for BlankRange {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.blank_size().cmp(&other.blank_size())
+    }
+}
+impl PartialOrd for BlankRange {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        self.blank_size().partial_cmp(&other.blank_size())
```
fulmicoton, line 41: Some(self.blank_size().cmp(&other.blank_size))

fulmicoton:
```suggestion
        Some(self.blank_size().cmp(&other.blank_size))
```
```

File: fastfield_codecs/src/compact_space/build_compact_space.rs [hunk 32]
```diff
@@ -0,0 +1,242 @@
...
+    total_num_values: usize,
+    cost_per_blank: usize,
+) -> CompactSpace {
+    let mut blanks = get_blanks(values_deduped_sorted);
+    let mut amplitude_compact_space = u128::MAX;
+    let mut amplitude_bits: u8 = num_bits(amplitude_compact_space);
+
+    let mut compact_space = CompactSpaceBuilder::new();
+    if values_deduped_sorted.is_empty() {
+        return compact_space.finish();
```
fulmicoton, line 87: Is this special case useful?

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/// Will collect blanks and add them to compact space if more bits are saved than cost from
+/// metadata.
+pub fn get_compact_space(
+    values_deduped_sorted: &BTreeSet<u128>,
+    total_num_values: usize,
+    cost_per_blank: usize,
+) -> CompactSpace {
+    let mut blanks = get_blanks(values_deduped_sorted);
```
fulmicoton:
```suggestion
    let mut blanks: BinaryHeap<BlankRange> = get_blanks(values_deduped_sorted);
```

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+        // when amplitude_new_bits changes
+        let cost = blank_collector.num_blanks() * cost_per_blank;
+        if cost >= saved_bits {
+            // Continue here, since although we walk over the blanks by size,
+            // we can potentially save a lot at the last bits, which are smaller blanks
+            //
+            // E.g. if the first range reduces the compact space by 1000 from 2000 to 1000, which
+            // saves 11-10=1 bit and the next range reduces the compact space by 950 to
+            // 50, which saves 10-6=4 bit
+            continue;
```
fulmicoton, line 110: I don't understand why this is not a `break` here.

The comment above is probably great, but I still don't get it.

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+    // Binary heap to process the gaps by their size
+    while let Some(blank_range) = blanks.pop() {
+        blank_collector.stage_blank(blank_range);
+
+        let staged_spaces_sum: u128 = blank_collector.staged_blanks_sum();
+        // +1 for later added null value
+        let amplitude_new_compact_space = amplitude_compact_space - staged_spaces_sum + 1;
+        let amplitude_new_bits = num_bits(amplitude_new_compact_space);
+        if amplitude_bits == amplitude_new_bits {
+            continue;
```
fulmicoton, line 97: That's tricky... 
There is ladder effect that makes it so that it is sometimes beneficial to add a blank even if it is not changing the amplitude.

Also I don't understand how continue makes sense here. 

For both continue in this loop, is it possible that we end up adding a blankrange after going through one of the continue?

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    // special case, when we don't collected any blanks because:
+    // * the data is empty (early exit)
+    // * the algorithm did decide it's not worth the cost, which can be the case for single values
+    //
+    // We drain one collected blank unconditionally, so the empty case is reserved for empty
+    // data, and therefore empty compact_space means the data is empty and no data is covered
+    // (conversely to all data) and we can assign null to it.
+    if compact_space.is_empty() {
+        compact_space.add_blanks(
```
fulmicoton: I don't understand the corner case here.

I don't think it is possible to go through this.

fulmicoton: Woudn't 
```rust
assert!(!compact_space.is_empty());
```
do the job?

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    fn is_empty(&self) -> bool {
+        self.blanks.is_empty()
+    }
+
+    /// Convert blanks to covered space and assign null value
+    fn finish(mut self) -> CompactSpace {
+        // sort by start. ranges are not allowed to overlap
+        self.blanks.sort_by_key(|blank| *blank.start());
```
fulmicoton: sort unstable.

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+        let mut covered_space = self
+            .blanks
+            .iter()
+            .tuple_windows()
+            .map(|(left, right)| {
+                assert!(
+                    left.end() < right.start(),
+                    "overlapping or adjacent ranges detected"
+                );
+                *left.end() + 1..=*right.start() - 1
```
fulmicoton: Wouldn't your code be generally simpler if you used Range instead of RangeInclusive?

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+    let mut add_range = |blank_range: RangeInclusive<u128>| {
+        let blank_range: Result<BlankRange, _> = blank_range.try_into();
+        if let Ok(blank_range) = blank_range {
+            blanks.push(blank_range);
+        }
+    };
+    for (first, second) in values_sorted.iter().tuple_windows() {
+        // Correctness Overflow: the values are deduped and sorted (BTreeSet property), that means
+        // there's always space between two values.
+        let blank_range = first + 1..=second - 1;
```
fulmicoton, line 16: `RangeInclusive` suck at expressing empty interval.
You have to use `end < start`.

`Range` woudl have been nicer here.

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    let mut blank_collector = BlankCollector::new();
+    // We will stage blanks until they reduce the compact space by 1 bit.
+    // Binary heap to process the gaps by their size
+    while let Some(blank_range) = blanks.pop() {
+        blank_collector.stage_blank(blank_range);
+
+        let staged_spaces_sum: u128 = blank_collector.staged_blanks_sum();
+        // +1 for later added null value
```
fulmicoton: thx for the comment!

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+                assert!(
+                    left.end() < right.start(),
+                    "overlapping or adjacent ranges detected"
+                );
+                *left.end() + 1..=*right.start() - 1
+            })
+            .collect::<Vec<_>>();
+
+        // Outside the blanks
+        if let Some(first_blank_start) = self.blanks.first().map(RangeInclusive::start) {
```
fulmicoton, line 175: nitpick: 
If you put that logic above, you can get rid of the large copy implied by the insert.

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+            // in case the first covered space ends at u128::MAX, assign null to the beginning
+            if *first_covered_space.end() == u128::MAX {
+                *first_covered_space = first_covered_space.start() - 1..=*first_covered_space.end();
+                *first_covered_space.start()
+            } else {
+                *first_covered_space = *first_covered_space.start()..=first_covered_space.end() + 1;
+                *first_covered_space.end()
+            }
+        } else {
+            covered_space.push(0..=0); // empty data case
```
fulmicoton, line 199: I really don't understand the edge cases... I'm not sure if this is needlessy complicated or if it is intrinsically difficult.

Maybe we should discuss it directly.

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+        // Correctness Overflow: the values are deduped and sorted (BTreeSet property), that means
+        // there's always space between two values.
+        let blank_range = first + 1..=second - 1;
+        add_range(blank_range);
+    }
+
+    // Replace after stabilization of https://github.com/rust-lang/rust/issues/62924
+    // Add preceeding range if values don't start at 0
+    if let Some(first_val) = values_sorted.iter().next() {
+        if *first_val != 0 {
```
fulmicoton: I like this!

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+            let blank_range = 0..=first_val - 1;
+            add_range(blank_range);
+        }
+    }
+
+    // Add succeeding range if values don't end at u128::MAX
+    if let Some(last_val) = values_sorted.iter().last() {
+        if *last_val != u128::MAX {
```
fulmicoton: I like this!

fastfield_codecs/src/compact_space/build_compact_space.rs
fulmicoton: Maybe put that at the top of the function?

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+        // Outside the blanks
+        if let Some(first_blank_start) = self.blanks.first().map(RangeInclusive::start) {
+            if *first_blank_start != 0 {
+                covered_space.insert(0, 0..=first_blank_start - 1);
+            }
+        }
+
+        if let Some(last_blank_end) = self.blanks.last().map(RangeInclusive::end) {
+            if *last_blank_end != u128::MAX {
+                covered_space.push(last_blank_end + 1..=u128::MAX);
```
fulmicoton, line 194: ah right... so that's why RangeInclusive is actually useful. My bad.

fastfield_codecs/src/compact_space/build_compact_space.rs
fulmicoton:
```suggestion
        self.blanks.sort_unstable_by_key(|blank| *blank.start());
```

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+            }
+        }
+
+        // Extend the first range and assign the null value to it.
+        let null_value = if let Some(first_covered_space) = covered_space.first_mut() {
```
fulmicoton: I have no clue what this is about. `null_value` is part of the covered space!?

fulmicoton: Could we take `null_value = 0` in the compact space and call it a day?

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,242 @@
...
+                *first_covered_space = *first_covered_space.start()..=first_covered_space.end() + 1;
+                *first_covered_space.end()
+            }
+        } else {
+            covered_space.push(0..=0); // empty data case
+            0u128
+        };
+
+        let mut compact_start: u64 = 0;
+        let mut ranges_mapping = Vec::with_capacity(covered_space.len());
```
fulmicoton:
```suggestion
        let mut ranges_mapping: Vec<RangeMapping> = Vec::with_capacity(covered_space.len());
```

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,719 @@
...
+///
+/// It will find spaces in the number range. For example:
+///
+/// 100, 101, 102, 103, 104, 50000, 50001
+/// could be mapped to
+/// 100..104 -> 0..4
+/// 50000..50001 -> 5..6
+///
+/// Compact space 0..=6 requires much less bits than 100..=50001
+///
```
fulmicoton: `compact_space/mod.rs` please.

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,719 @@
...
+    }
+}
+
+impl CompactSpace {
+    /// Amplitude is the value range of the compact space including the sentinel value used to
+    /// identify null values. The compact space is 0..=amplitude .
+    ///
+    /// It's only used to verify we don't exceed u64 number space, which would indicate a bug.
+    fn amplitude_compact_space(&self) -> u128 {
+        let last_range = &self.ranges_mapping[self.ranges_mapping.len() - 1];
```
fulmicoton: so self.ranges_mapping cannot be empty?

fastfield_codecs/src/compact_space.rs
fulmicoton: I still don't understand... What does `self.to_compact` returning Err() means?
Doesn't it mean the value is not part of the covered value?
If so isn't the match on Err(_), Err(_) below wrong?

fulmicoton: Ah got it... I had missed pos1 == pos2 below

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,719 @@
...
+            .binary_search_by(|probe| {
+                let value_range = &probe.value_range;
+                if value_range.contains(&value) {
+                    return Ordering::Equal;
+                } else if value < *value_range.start() {
+                    return Ordering::Greater;
+                } else if value > *value_range.end() {
+                    return Ordering::Less;
+                }
+                panic!("not covered all ranges in check");
```
fulmicoton:
```suggestion
                if value < *value_range.start() {
                    Ordering::Greater
                } else if value > *value_range.end() {
                    Ordering::Less
                } else {
                    Ordering::Equal
                }
```

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,719 @@
...
+            let blank_delta_end = value_range.end() - prev_value;
+            VIntU128(blank_delta_end).serialize(writer)?;
+            prev_value = *value_range.end();
+        }
+
+        Ok(())
+    }
+
+    fn deserialize<R: io::Read>(reader: &mut R) -> io::Result<Self> {
+        let null_value = VIntU128::deserialize(reader)?.0;
```
fulmicoton: The null_value is not a compact space thing? Why?

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,719 @@
...
+            range_mapping.compact_start + range_mapping.range_length()
+        });
+
+        let range = compact_from..=compact_to;
+        let mut positions = Vec::new();
+
+        let step_size = 4;
+        let cutoff = self.params.num_vals - self.params.num_vals % step_size;
+
+        let mut check_add = |idx, val| {
```
fulmicoton: Can you find a better name than `check_add`

`add_if_in_range` or something?

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,719 @@
...
+            if range.contains(&val) && val != self.params.null_value_compact_space {
```
fulmicoton: Why is it needed to check for the null value!?

If we put the null value in compact at the extremity, 0 or MAX+1 it should not be necessary to do that check,

fastfield_codecs/src/compact_space.rs
```diff
@@ -0,0 +1,719 @@
...
+    /// error with the index of the next range.
+    fn to_compact(&self, value: u128) -> Result<u64, usize> {
+        self.params.compact_space.to_compact(value)
+    }
+
+    fn compact_to_u128(&self, compact: u64) -> u128 {
+        self.params.compact_space.unpack(compact)
+    }
+
+    /// Comparing on compact space: 1.08 GElements/s, which equals a throughput of 17,3 Gb/s
```
fulmicoton: Cool! I would have expected this to be 3x slower.

fastfield_codecs/src/compact_space.rs
fulmicoton: The bitunpacking code is inefficient, and the filtering is inefficient.
I thought we could do better by fill a buffer of size 128 or something, and use simd to postfilter....
but 1Gel/s is already quite nice.

fastfield_codecs/src/compact_space.rs
PSeitz: No, the `CompactSpaceBuilder.finish` will ensure that. If there are no blanks, we add the full range to be covered. A blank that covers the whole space can't happen in the algorithm. But I think we can remove that assumption here, since it's needless

fastfield_codecs/src/compact_space.rs
PSeitz: yes that's better and will be faster. For reference I'll copy the discord message here

> In the old version, the user had to replace None with the null value the compressor chose, because the API was taking Vec<u128>. So the null value was not just a compact space thing. But this api sucked, in the new version we accept Vec<Option<u128>>. So this null value dance can actually be changed, because null is now only a  compact space thing. This should simplify things

fastfield_codecs/src/compact_space/build_compact_space.rs
PSeitz: There's a difference between staging and adding blanks. We stage blanks until they are have an positive effect and then add them.

If `num_bits(amplitude)` didn't change we continue staging blanks. There's no effect on the compression yet.

If one of the continue is triggered, we don't add the staged blanks but keep staging more blanks.

fastfield_codecs/src/compact_space/build_compact_space.rs
PSeitz: break would end the loop, but we have to check all the blanks so we `continue`. As the comment says there may be large gains in saved bits for the last blanks.

The cost check is only for the currently staged blanks, not all blanks

fastfield_codecs/src/compact_space/build_compact_space.rs
PSeitz: It's possible, I did run into this. If the cost for metadata is higher, we don't add blanks. That can happen for single values or very few values (depends on the `cost_per_blank` )

I think it was this test case:
```
    #[test]
    fn test_bug2() {
        let vals = &[340282366920938463463374607431768211455u128];
        let _data = test_aux_vals(vals);
    }

```

fastfield_codecs/src/compact_space/build_compact_space.rs
PSeitz: yep

fastfield_codecs/src/compact_space/build_compact_space.rs
PSeitz: yes that's better

fastfield_codecs/src/compact_space/build_compact_space.rs
fulmicoton: ah got it!

fastfield_codecs/src/compact_space/build_compact_space.rs
fulmicoton: thx

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,227 @@
...
+        // +1 for later added null value
+        let amplitude_new_compact_space = amplitude_compact_space - staged_spaces_sum + 1;
+        let amplitude_new_bits = num_bits(amplitude_new_compact_space);
+        if amplitude_bits == amplitude_new_bits {
+            continue;
+        }
+        let saved_bits = (amplitude_bits - amplitude_new_bits) as usize * total_num_values;
+        // TODO: Maybe calculate exact cost of blanks and run this more expensive computation only,
+        // when amplitude_new_bits changes
+        let cost = blank_collector.num_blanks() * cost_per_blank;
```
fulmicoton: Should this be `num_staged_blanks`?

fastfield_codecs/src/compact_space/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+            })
+            .map(|pos| {
+                let range_mapping = &self.ranges_mapping[pos];
+                let pos_in_range = (value - range_mapping.value_range.start()) as u64;
+                range_mapping.compact_start + pos_in_range
+            })
+    }
+
+    /// Unpacks a value from compact space u64 to u128 space
+    fn unpack(&self, compact: u64) -> u128 {
```
fulmicoton:
```suggestion
    fn from_compact(&self, compact: u64) -> u128 {
```

fastfield_codecs/src/compact_space/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+            range_mapping.compact_end()
+        });
+
+        let range = compact_from..=compact_to;
+        let mut positions = Vec::new();
+
+        let step_size = 4;
+        let cutoff = self.params.num_vals - self.params.num_vals % step_size;
+
+        let mut add_if_in_range = |idx, val| {
```
fulmicoton: can you rename `push_if_in_range` or something?

fastfield_codecs/src/compact_space/build_compact_space.rs
PSeitz: yep

fastfield_codecs/src/compact_space/mod.rs
PSeitz: Clippy complains about this one
```
warning: methods called `from_*` usually take no `self`
   --> fastfield_codecs/src/compact_space/mod.rs:156:21
    |
156 |     fn from_compact(&self, compact: u64) -> u128 {
    |                     ^^^^^
    |
    = note: `#[warn(clippy::wrong_self_convention)]` on by default
    = help: consider choosing a less ambiguous name
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention
```

fastfield_codecs/src/compact_space/build_compact_space.rs
```diff
@@ -0,0 +1,227 @@
...
+        // when amplitude_new_bits changes
+        let cost = blank_collector.num_staged_blanks() * cost_per_blank;
+        if cost >= saved_bits {
+            // Continue here, since although we walk over the blanks by size,
+            // we can potentially save a lot at the last bits, which are smaller blanks
+            //
+            // E.g. if the first range reduces the compact space by 1000 from 2000 to 1000, which
+            // saves 11-10=1 bit and the next range reduces the compact space by 950 to
+            // 50, which saves 10-6=4 bit
+            continue;
```
fulmicoton, line 110: now that the staged_blanks thing is fixed, isn't break ok here?

fastfield_codecs/src/compact_space/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+            compact_start += range_length as u64;
+        }
+
+        Ok(Self { ranges_mapping })
+    }
+}
+
+impl CompactSpace {
+    /// Amplitude is the value range of the compact space including the sentinel value used to
+    /// identify null values. The compact space is 0..=amplitude .
```
fulmicoton, line 109: I think something is incorrect here. With the current implementation, the compact space is `0..amplitude`

fulmicoton, line 109:
```suggestion
    /// identify null values. The compact space is 0..amplitude .
```

fastfield_codecs/src/compact_space/mod.rs
fulmicoton: That makes sense, but still can you find a set of two names that are somewhat symmetric?

e.g.
```
.value_to_compact(...)
.compact_to_value(..)
```

or whatever

fastfield_codecs/src/compact_space/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+    max_value: u128,
+    num_vals: u64,
+    num_bits: u8,
+}
+
+impl CompactSpaceCompressor {
+    /// Taking the vals as Vec may cost a lot of memory. It is used to sort the vals.
+    pub fn train_from(
+        vals: impl Iterator<Item = u128>,
+        total_num_values_incl_nulls: usize,
```
fulmicoton:
```suggestion
```

fulmicoton: This is a werid prototype. You will consume the iterator anyway.

Make sure the prototype of your function are as clear as possible.

fastfield_codecs/src/compact_space/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+    let compact_space = get_compact_space(values_sorted, total_num_values, COST_PER_BLANK_IN_BITS);
+    let amplitude_compact_space = compact_space.amplitude_compact_space();
+
+    assert!(
+        amplitude_compact_space <= u64::MAX as u128,
+        "case unsupported."
+    );
+
+    let num_bits = tantivy_bitpacker::compute_num_bits(amplitude_compact_space as u64);
+    let min_value = *values_sorted.iter().next().unwrap_or(&0);
```
fulmicoton:
```suggestion
    let min_value = values_sorted.first().copied().unwrap_or(0);
```

fastfield_codecs/src/compact_space/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+    let max_value = *values_sorted.iter().last().unwrap_or(&0);
```
fulmicoton:
```suggestion
    let max_value = values_sorted.last().copied().unwrap_or(0);
```

fastfield_codecs/src/compact_space/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+    assert!(
```
fulmicoton: `assert_eq` with the right definition... That kind of assert is cool to make sure there are no off-by-one errors.

fastfield_codecs/src/compact_space/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+    }
+
+    /// Converting to compact space for the decompressor is more complex, since we may get values
+    /// which are outside the compact space. e.g. if we map
+    /// 1000 => 5
+    /// 2000 => 6
+    ///
+    /// and we want a mapping for 1005, there is no equivalent compact space. We instead return an
+    /// error with the index of the next range.
+    fn to_compact(&self, value: u128) -> Result<u64, usize> {
```
fulmicoton:
```suggestion
    fn u128_to_compact(&self, value: u128) -> Result<u64, usize> {
```

fulmicoton: and you can use this pair of names in the compact_space too.

fastfield_codecs/src/lib.rs
```diff
@@ -388,6 +392,31 @@ mod bench {
         });
     }
 
+    #[bench]
+    fn bench_intfastfield_jumpy_fflookup_u128(b: &mut Bencher) {
```
fulmicoton: jumpy is probably not the most interesting one. Ideally something with a random stride with an average jump of 5 or so is nice.

fastfield_codecs/src/main.rs
```diff
@@ -1,9 +1,139 @@
...
+    println!(
+        "Num Bits per elem {:.2}",
+        (data.len() * 8) as f32 / dataset.len() as f32
+    );
+
+    let decompressor = CompactSpaceDecompressor::open(OwnedBytes::new(data)).unwrap();
+    // Sample some ranges
+    for value in dataset.iter().take(1110).skip(1100).cloned() {
+        print_time!("get range");
+        let doc_values = decompressor.get_range(value..=value);
```
fulmicoton: value..=value!?

fastfield_codecs/src/main.rs
```diff
@@ -1,9 +1,139 @@
...
+        .collect();
+
+    println!("IpAddrsAny\t{}", ip_addrs.len());
+    println!("IpAddrsV4\t{}", ip_addr_v4);
+
+    ip_addrs
+}
+
+fn bench_ip() {
+    let dataset = ip_dataset();
```
fulmicoton, line 86: make sure the dataset you are playing with is not sorted.

fulmicoton:
Can you add a bench (= not in the main) for the ranges?

Worst case scenario is shuffled IP with a hit rate of 50%.

fastfield_codecs/src/compact_space/build_compact_space.rs
PSeitz: This part is pretty much unchanged. saved bits may outrun cost later on

fastfield_codecs/src/compact_space/mod.rs
PSeitz: first and last are nightly only

fastfield_codecs/src/compact_space/mod.rs
PSeitz: first and last are nightly only

fastfield_codecs/src/main.rs
PSeitz: yes, to get all docs with this value

fastfield_codecs/src/main.rs
PSeitz: it's not sorting the dataset, it's loading the data from stdin

fastfield_codecs/src/compact_space/mod.rs
PSeitz: true, I switched to taking a `ColumnV2`

fastfield_codecs/src/compact_space/mod.rs
PSeitz: I changed it to be `0..=amplitude`

fastfield_codecs/src/compact_space/mod.rs
PSeitz: The `assert_eq` was not possible with the current algorithm, so I changed it a little bit

The reason it did not work was that we start by collecting `BlankRange`, which implies a minimum size of 3. That's fine for all cases except the borders, since it does not imply a cost by creating a covered range.
So if we had as max value `u128::MAX -1`, we did not add the blank range.

fastfield_codecs/src/main.rs
fulmicoton: yes but that's not really challenging the branch predictor is it?

PSeitz: depends on the the occurrences of the sampled ip

fulmicoton: yes... but so does the 1GB el/s result hold?

fastfield_codecs/src/compact_space/mod.rs
fulmicoton: You can stick to `Iterator<Item=u128>`, or even `ExactSizeIterator` if you don't want to write a loop

fulmicoton:
What is the actual plan with ColumnV2? Can we remove that from the current PR and discuss it independently?
(and keep the compact_code =0 reserved)

fastfield_codecs/src/main.rs
PSeitz: for the real world test data set yes, for shuffled ips mostly not

In most cases the number of bits used is below 32, so we could add a simd version for that case

PSeitz:
I added some benches, but the chosen distribution is maybe not ideal.

> What is the actual plan with ColumnV2? Can we remove that from the current PR and discuss it independently? (and keep the compact_code =0 reserved)

Main difference is that the column returns Option<T> as part on how we handle null: `    fn get_val(&self, idx: u64) -> Option<T>;`

The plan would be to migrate all codecs to it and start with the ip_codec as test case.

fastfield_codecs/src/compact_space/build_compact_space.rs
fulmicoton: yes you are correct!

fastfield_codecs/src/lib.rs
```diff
@@ -28,6 +29,7 @@ mod serialize;
 pub use self::bitpacked::BitpackedCodec;
 pub use self::blockwise_linear::BlockwiseLinearCodec;
 pub use self::column::{monotonic_map_column, Column, VecColumn};
+pub use self::compact_space::{ip_to_u128, CompactSpaceCompressor, CompactSpaceDecompressor};
```
fulmicoton: We don't need these to be public.

fastfield_codecs/src/compact_space/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+    /// error with the index of the next range.
+    fn u128_to_compact(&self, value: u128) -> Result<u64, usize> {
+        self.params.compact_space.u128_to_compact(value)
+    }
+
+    fn compact_to_u128(&self, compact: u64) -> u128 {
+        self.params.compact_space.compact_to_u128(compact)
+    }
+
+    /// Comparing on compact space: 1.08 GElements/s, which equals a throughput of 17,3 Gb/s
```
fulmicoton: I am getting 200millions elements per seconds on my laptop.
Can you update the figure with whatever hardware you have?

fastfield_codecs/src/column.rs
```diff
@@ -52,6 +53,12 @@ pub trait Column<T = u64>: Send + Sync {
     }
 }
 
+/// Extend Column Api
+pub trait ColumnExt<T = u64>: Column<T> {
+    /// Return the positions of values which are in the provided range.
+    fn get_between_vals(&self, range: RangeInclusive<T>) -> Vec<u64>;
```
fulmicoton: Should we make this part of `Column` maybe?

fulmicoton (approved):

fulmicoton (approved):

PR #1518, opened 2022-09-13
Make `Column: Send + Sync`

shikhar:
`Arc<dyn Column<V>>` will only implement `Send` if `Column` is both.


fulmicoton (approved):

PSeitz:
@shikhar You probably already noticed, but the next tantivy version will be a breaking change

PR #1519, opened 2022-09-14
Perf fix on the MonotonicMapping column

fulmicoton:
The Monotonic mapping was using the default implementation for `get_range` and `.iter`.

As a result, some of the column used in merge (e.g. multivalued fast fields) were exhibiting a very strong performance regression.

### DIFF
--- a/fastfield_codecs/src/column.rs
+++ b/fastfield_codecs/src/column.rs
@@ -17,15 +17,11 @@ pub trait Column<T = u64>: Send + Sync {
     /// associated with the `DocId` going from
     /// `start` to `start + output.len()`.
     ///
-    /// Regardless of the type of `Item`, this method works
-    /// - transmuting the output array
-    /// - extracting the `Item`s as if they were `u64`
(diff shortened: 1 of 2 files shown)


fastfield_codecs/src/column.rs
```diff
@@ -165,6 +173,23 @@ where
...
         self.from_column.num_vals()
     }
+
+    fn iter<'a>(&'a self) -> Box<dyn Iterator<Item = Output> + 'a> {
+        Box::new(self.from_column.iter().map(&self.monotonic_mapping))
+    }
+
+    fn get_range(&self, mut start: u64, output: &mut [Output]) {
+        let mut intermediary: MaybeUninit<[Input; 16]> = MaybeUninit::uninit();
+        let intermediary_ref = unsafe { intermediary.assume_init_mut() };
```
PSeitz: I would prefer default initialization with a constraint on `Input: Default` . This should probably be optional, if output is >= 16 ?

PSeitz (approved):

PR #1522, opened 2022-09-16
Hiding useless symbols and removing code.

fulmicoton:


fastfield_codecs/src/lib.rs
```diff
@@ -132,14 +133,6 @@ pub const ALL_CODEC_TYPES: [FastFieldCodecType; 3] = [
     FastFieldCodecType::Linear,
 ];
 
-#[derive(Debug, Clone)]
-/// Statistics are used in codec detection and stored in the fast field footer.
-pub struct FastFieldStats {
```
PSeitz, line 137: I was thinking about going the other way and replace all the methods with returning the stats

fulmicoton: You can do a follow up PR if you see an opportunity to simplify the code.

PSeitz (approved):

PR #1527, opened 2022-09-18
docs: Remove mentions of stream_field method.

waywardmonkeys:
This method doesn't exist, so no need to mention it.


PSeitz (approved):
Nice, thanks!

PR #1528, opened 2022-09-19
fix benches

PSeitz:


fastfield_codecs/src/lib.rs
```diff
@@ -27,10 +27,10 @@ mod column;
 mod gcd;
 mod serialize;
 
-use self::bitpacked::BitpackedCodec;
-use self::blockwise_linear::BlockwiseLinearCodec;
+pub use self::bitpacked::BitpackedCodec;
```
fulmicoton: I'd rather not have this public.

You need it for benches right? Can you put that under a feature flag, or 
make sure that the bench use the public API? (that will benchmarks using the static impl however)

PSeitz: fixed, I moved some benches into lib.rs

fulmicoton (approved):

PR #1531, opened 2022-09-19
Improvements to doc linking, grammar, etc.

waywardmonkeys:
More fixes to remove most of the direct HTML links (remaining ones are problematic and will be addressed separately).

Also uses backticks more to make formatting more consistent.

### DIFF
--- a/src/docset.rs
+++ b/src/docset.rs
@@ -3,10 +3,10 @@ use std::borrow::{Borrow, BorrowMut};
 use crate::fastfield::AliveBitSet;
 use crate::DocId;
 
-/// Sentinel value returned when a DocSet has been entirely consumed.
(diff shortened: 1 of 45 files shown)


src/docset.rs
```diff
@@ -20,21 +20,21 @@ pub trait DocSet: Send {
     /// assert_eq!(doc, docset.doc());
     /// ```
     ///
-    /// If we reached the end of the DocSet, TERMINATED should be returned.
+    /// If we reached the end of the `DocSet`, [`TERMINATED`] should be returned.
```
PSeitz, line 23: Why [`TERMINATED`]. It looks like an array is returned

PSeitz: ah nvm it's referencing the type

PSeitz (approved):

PSeitz:
Thanks!

PR #1535, opened 2022-09-19
add support for deleting all documents matching query

### LINKED ISSUE #1494: Enrich delete query
Currently, deletes are defined by a term query.

For quickwit, we will need to execute two kinds of delete queries:
- A general purpose delete query defined by terms, boolean operators, etc.
- A delete query defined by a field and a list of values (typically IDs, potentially hundreds of thousands of them).

I don't really know how to implement that in tantivy so I leave this to you @PSeitz @fulmicoton :)

trinity-1686a:
fix half of #1494

### DIFF
--- a/src/indexer/operation.rs
+++ b/src/indexer/operation.rs
@@ -1,35 +1,25 @@
-use itertools::Either;
-
 use crate::query::Weight;
 use crate::schema::{Document, Term};
 use crate::Opstamp;
 
 /// Timestamped Delete operation.
-//#[derive(Clone, Debug)]
 pub struct DeleteOperation {
     pub opstamp: Opstamp,
-    pub term: Either<Term, Box<dyn Weight>>,
+    pub target: DeleteTarget,
 }
 
-impl DeleteOperation {
-    pub fn new(opstamp: Opstamp, term: Term) -> Self {
-        DeleteOperation {
-            opstamp,
-            term: Either::Left(term),
-        }
-    }
-
-    pub fn new_for_query(opstamp: Opstamp, weight: Box<dyn Weight>) -> Self {
-        DeleteOperation {
-            opstamp,
-            term: Either::Right(weight),
-        }
-    }
+/// Target of a Delete operation
+pub enum DeleteTarget {
+    Term(Term),
+    Query(Box<dyn Weight>),
 }
 
 impl Default for DeleteOperation {
     fn default() -> Self {
-        DeleteOperation::new(0, Term::new())
+        DeleteOperation {
+            opstamp: 0,
+            target: DeleteTarget::Term(Term::new().into()),
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/indexer/operation.rs
```diff
@@ -1,22 +1,38 @@
...
 use crate::schema::{Document, Term};
 use crate::Opstamp;
 
 /// Timestamped Delete operation.
-#[derive(Clone, Eq, PartialEq, Debug)]
+//#[derive(Clone, Debug)]
 pub struct DeleteOperation {
     pub opstamp: Opstamp,
-    pub term: Term,
+    pub term: Either<Term, Box<dyn Weight>>,
```
adamreichold: Using `Either` here seems a bit "type blind" since it does not explain what the alternatives are. Why not create a new enum like

```rust
pub enum DeleteTarget {
  Term(Term),
  Query(Box<dyn Weight),
}
```

I think this would also avoid the need for the constructor methods as the literal syntax would stay readable, e.g.

```rust
DeleteOperation {
  opstamp: i as u64,
  target: DeleteTarget::Term(Term::from_field_u64(field, i as u64)),
}
```

adamreichold: (Alternatively, I would suggest `DeleteOperation::from_term` or `DeleteOperation::from_query` as consistent names for the constructors.)

src/indexer/operation.rs
```diff
@@ -1,18 +1,24 @@
...
-#[derive(Clone, Eq, PartialEq, Debug)]
 pub struct DeleteOperation {
     pub opstamp: Opstamp,
-    pub term: Term,
+    pub target: DeleteTarget,
+}
+
+/// Target of a Delete operation
+pub enum DeleteTarget {
+    Term(Term),
```
fulmicoton: Shouldn't we just use `Box<dyn Weight>` for `TermWeight` too?

trinity-1686a: I did not do that because it would require removing `Default` from `DeleteOperation`, but it looks like it's not used anywhere

fulmicoton:
It only partially fixes https://github.com/quickwit-oss/tantivy/issues/1494

We also need to add a query that matches a dicitonary. Can you create a separate issue maybe?

src/indexer/operation.rs
fulmicoton: yeah I was a rustrookie when I wrote that. The Default boundary is probably nonsense.

trinity-1686a: Hum, it might be trickier than I thought. [`test_deletequeue`](https://github.com/quickwit-oss/tantivy/pull/1535/files#diff-8ebb1eab4107a255705c22aac2a218ffbcaf3c7c6223d27ac9a9063eaebe2b8eR260) doesn't have access to a `Searcher`. I'm not sure how to make that test compile without creating a dummy index in ram.

fulmicoton (approved):

PR #1536, opened 2022-09-20
Introducing ColumnReader.

fulmicoton:
Introducing a ColumnReader trait and .reader() to Column, hence removing the dreaded Mutex in the `MultiValueStartIndex` thingy.

### DIFF
--- a/fastfield_codecs/src/column.rs
+++ b/fastfield_codecs/src/column.rs
@@ -4,6 +4,11 @@ use std::ops::RangeInclusive;
 use tantivy_bitpacker::minmax;
 
 pub trait Column<T: PartialOrd = u64>: Send + Sync {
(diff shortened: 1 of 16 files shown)


fastfield_codecs/src/column.rs
```diff
@@ -58,10 +66,70 @@ pub trait Column<T: PartialOrd = u64>: Send + Sync {
...
+        self.get()
+    }
+
+    fn advance(&mut self) -> bool {
+        self.idx = self.idx.wrapping_add(1);
+        self.idx < self.len
+    }
+
+    fn get(&self) -> T {
+        self.column.get_val(self.idx)
```
PSeitz, line 132: Isn't the main reason for `ColumnReader` to remove `get_val`? Or is this just a temp hack?

PR #1539, opened 2022-09-21
Add query matching terms in a set

trinity-1686a:
fix the 2nd half of #1494

### DIFF
--- a/src/query/set_query.rs
+++ b/src/query/set_query.rs
@@ -1,56 +1,73 @@
+use std::collections::HashMap;
+
 use tantivy_fst::raw::CompiledAddr;
 use tantivy_fst::{Automaton, Map};
 
-use crate::query::{AutomatonWeight, Query, Weight};
+use crate::query::score_combiner::DoNothingCombiner;
+use crate::query::{AutomatonWeight, BooleanWeight, Occur, Query, Weight};
 use crate::schema::Field;
 use crate::{Searcher, Term};
 
 /// A Term Set Query matches all of the documents containing any of the Term provided
-///
-/// Terms not using the right Field are discared.
 #[derive(Debug, Clone)]
 pub struct TermSetQuery {
-    field: Field,
-    terms: Vec<Term>,
+    terms_map: HashMap<Field, Vec<Term>>,
 }
 
 impl TermSetQuery {
     /// Create a Term Set Query
-    pub fn new<T: IntoIterator<Item = Term>>(field: Field, terms: T) -> Self {
-        let mut terms: Vec<_> = terms
+    pub fn new<T: IntoIterator<Item = Term>>(terms: T) -> Self {
+        let mut terms_map: HashMap<_, Vec<_>> = terms
             .into_iter()
-            .filter(|key| key.field() == field)
-            .collect();
-        terms.sort_unstable();
-        terms.dedup();
-        TermSetQuery { field, terms }
+            .map(|term| (term.field(), term))
+            .fold(HashMap::new(), |mut map, (field, term)| {
+                map.entry(field).or_default().push(term);
+                map
+            });
+        for terms in terms_map.values_mut() {
+            terms.sort_unstable();
+            terms.dedup();
+        }
+        TermSetQuery { terms_map }
     }
 
     fn specialized_weight(
         &self,
         searcher: &Searcher,
-    ) -> crate::Result<AutomatonWeight<SetDfaWrapper>> {
-        let field_entry = searcher.schema().get_field_entry(self.field);
-        let field_type = field_entry.field_type();
-        if !field_type.is_indexed() {
-            let error_msg = format!("Field {:?} is not indexed.", field_entry.name());
-            return Err(crate::TantivyError::SchemaError(error_msg));
+    ) -> crate::Result<BooleanWeight<DoNothingCombiner>> {
+        let mut sub_queries: Vec<(_, Box<dyn Weight>)> = Vec::with_capacity(self.terms_map.len());
+
+        for (&field, sorted_terms) in self.terms_map.iter() {
+            let field_entry = searcher.schema().get_field_entry(field);
+            let field_type = field_entry.field_type();
+            if !field_type.is_indexed() {
+                let error_msg = format!("Field {:?} is not indexed.", field_entry.name());
+                return Err(crate::TantivyError::SchemaError(error_msg));
+            }
+            let field_type = field_type.value_type();
+
+            // In practice this won't fail because:
+            // - we are writing to memory, so no IoError
+            // - Terms are ordered
+            let map = Map::from_iter(
+                sorted_terms
+                    .iter()
+                    .filter(|key| key.typ() == field_type)
+                    .map(|key| (key.value_bytes(), 0)),
+            )
+            .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
+
+            sub_queries.push((
+                Occur::Should,
+                Box::new(AutomatonWeight::new(field, SetDfaWrapper(map))),
+            ));
         }
-        let field_type = field_type.value_type();
 
-        // In practice this won't fail because:
-        // - we are writing to memory, so no IoError
-        // - BTreeSet are ordered, and we limit ourselves to values with a fixed prefix (which we
-        // strip), so Map::from_iter get values in order
-        let map = Map::from_iter(
-            self.terms
-                .iter()
-                .filter(|key| key.typ() == field_type)
-                .map(|key| (key.value_bytes(), 0)),
-        )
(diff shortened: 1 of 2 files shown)


src/query/set_query.rs
```diff
@@ -0,0 +1,198 @@
...
+/// Terms not using the right Field are discared.
+#[derive(Debug, Clone)]
+pub struct TermSetQuery {
+    field: Field,
+    terms: BTreeSet<Term>,
+}
+
+impl TermSetQuery {
+    /// Create a Term Set Query
+    pub fn new(field: Field, terms: BTreeSet<Term>) -> Self {
```
adamreichold: `BTreeSet` is a good data structure for a set that is continuously changing while always staying sorted and using the sorting to speed element access. In this case, it seems that the set is accessed only once to produce a weight.

Maybe it would be nicer to simply require `terms: T where T: IntoIterator<Item=Term>`, collect this into a `Vec` (where `terms.into_iter().collect::<Vec<_>>()` would not allocate if the iterator is created from a `Vec`) and sort and deduplicate this `Vec` once in this constructor?

I think this could yield nicer API and simpler and hence faster code, but then again it could be insignificant and thereby not worth it especially if one expects the terms to be already presented as a `BTreeSet`.

fulmicoton: I like the BTreeSet personally.

adamreichold: > I like the BTreeSet personally.

I would not say that I dislike it (or any other data structure for that matter), just that it brings more to the table than is required here. I do think it is arguably more complex than say `Vec` which is why I tried to suggest the simplest possible data structure for this particular task.

trinity-1686a: What should I do then? Personally I prefer using a `BTreeSet`, but I understand an `IntoIterator` is easier to provide in general

shikhar: As a potential user of this query`*`, if the type was a `BTreeSet` we would have to create one just for this purpose, and the query just needs immutable sorted & deduped data for which `Vec` suffices. I would vote for `IntoIterator`.


`*` currently doing union boolean query over tens of IDs -- looks like this should be much more efficient!

fulmicoton: IntoIterator it is then...

and then within the function:
`IntoIterator -> HashSet -> Vec -> Sort.`

src/query/set_query.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+impl TermSetQuery {
+    /// Create a Term Set Query
+    pub fn new<T: IntoIterator<Item = Term>>(field: Field, terms: T) -> Self {
+        let mut terms: Vec<_> = terms
+            .into_iter()
+            .filter(|key| key.field() == field)
+            .collect();
+        terms.sort_unstable();
```
fulmicoton: Ok for `Iterator -> Vec -> sort -> Dedup`.

Note it is slower  `Iterator -> HashSet -> Vec -> sort`  if the average number of occurence is high, but we can hope users are not feeding us with random stuff here.

src/query/set_query.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+impl Query for TermSetQuery {
+    fn weight(
+        &self,
+        searcher: &Searcher,
+        _scoring_enabled: bool,
+    ) -> crate::Result<Box<dyn Weight>> {
+        Ok(Box::new(self.specialized_weight(searcher)?))
```
fulmicoton, line 74: good job spotting `specialized_weight`

src/query/set_query.rs
```diff
@@ -0,0 +1,201 @@
...
+        let field_type = field_entry.field_type();
+        if !field_type.is_indexed() {
+            let error_msg = format!("Field {:?} is not indexed.", field_entry.name());
+            return Err(crate::TantivyError::SchemaError(error_msg));
+        }
+        let field_type = field_type.value_type();
+
+        // In practice this won't fail because:
+        // - we are writing to memory, so no IoError
+        // - BTreeSet are ordered, and we limit ourselves to values with a fixed prefix (which we
```
fulmicoton:
```suggestion
        // - Terms are ordered, and we limit ourselves to values with a fixed prefix (which we
```

src/query/set_query.rs
```diff
@@ -0,0 +1,201 @@
...
+use crate::schema::Field;
+use crate::{Searcher, Term};
+
+/// A Term Set Query matches all of the documents containing any of the Term provided
+///
+/// Terms not using the right Field are discared.
+#[derive(Debug, Clone)]
+pub struct TermSetQuery {
+    field: Field,
+    terms: Vec<Term>,
```
fulmicoton:
```suggestion
    sorted_terms: Vec<Term>,
```

src/query/set_query.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+impl TermSetQuery {
+    /// Create a Term Set Query
+    pub fn new<T: IntoIterator<Item = Term>>(field: Field, terms: T) -> Self {
```
fulmicoton: The API is unclear here.
`field` is redundant in the sense that terms contain a field too...
You did not document what happens if the terms are targetting different field either.

Can you at the very least document this?

Alternatives:
You can also drop the field argument and use the first term to detect what the field should be.
That still forces you to document the behavior.

You could also support terms with different fields. That will require more work... but not that much.
You can rely on the `BooleanWeight` to get the union of several `AutomatonWeight`.

I think it does the obvious optimization if there is only one of them.

I suspect we need to handle more than one field in quickwit anyway.

fulmicoton (commented):
pleae have a look at the clippy stuff too.

src/query/set_query.rs
```diff
@@ -0,0 +1,201 @@
...
+    fn start(&self) -> Self::State {
+        Some(self.0.as_ref().root().addr())
+    }
+    fn is_match(&self, state: &Self::State) -> bool {
+        state
+            .map(|s| self.0.as_ref().node(s).is_final())
+            .unwrap_or(false)
+    }
+    fn accept(&self, state: &Self::State, byte: u8) -> Self::State {
+        state.and_then(|state| {
```
fulmicoton: In quickwit and tantivy, avoid `and_then` if it is not helpful.
Also be careful about variable naming.

Here you used "state" for 3 different types...
One of which had very little with state. 

Using `Self::State` is making things needlessly difficult to read here too. 

```rust
  fn accept(&self, state_opt: &Option<CompiledAddr>, byte: u8) -> Option<CompiledAddr> {
       let state = state_opt?;
       let node = self.0.as_ref().node(state);
       let transition = node.find_input(byte)?;
       Some(state.transition_addr(transition))
  }
```

src/query/set_query.rs
```diff
@@ -0,0 +1,201 @@
...
+struct SetDfaWrapper(Map<Vec<u8>>);
+
+impl Automaton for SetDfaWrapper {
+    type State = Option<CompiledAddr>;
+
+    fn start(&self) -> Self::State {
+        Some(self.0.as_ref().root().addr())
+    }
+    fn is_match(&self, state: &Self::State) -> bool {
+        state
```
fulmicoton: That one is a personal nitpick but I tend to find the `if-let` style
more readable.

```rust
if let Some(state) = state_opt {
    self.0.as_ref().node(state).is_final()
} else {
    false
}
```

trinity-1686a:
Should I depreciate [`BooleanQuery::new_multiterms_query`](https://docs.rs/tantivy/latest/tantivy/query/struct.BooleanQuery.html#method.new_multiterms_query)? It does the same thing, but is probably slower as soon as there is more than one term per field being searched

src/query/set_query.rs
```diff
@@ -0,0 +1,229 @@
...
+/// A Term Set Query matches all of the documents containing any of the Term provided
+#[derive(Debug, Clone)]
+pub struct TermSetQuery {
+    terms_map: HashMap<Field, Vec<Term>>,
+}
+
+impl TermSetQuery {
+    /// Create a Term Set Query
+    pub fn new<T: IntoIterator<Item = Term>>(terms: T) -> Self {
+        let mut terms_map: HashMap<_, Vec<_>> = terms
```
fulmicoton: let mut terms_map = HashMap::default();
for term in terms {
   terms_map.entry(field).or_default().push(term);
}
```

is much easier to read IMHO

File: src/query/set_query.rs [hunk 10]
```diff
@@ -0,0 +1,228 @@
...
+            }
+            let field_type = field_type.value_type();
+
+            // In practice this won't fail because:
+            // - we are writing to memory, so no IoError
+            // - Terms are ordered
+            let map = Map::from_iter(
+                sorted_terms
+                    .iter()
+                    .filter(|key| key.typ() == field_type)
```
fulmicoton:
```suggestion
```

fulmicoton (approved):
Great job!

PR #1543, opened 2022-09-22
fix multivalue ff indexing regression

PSeitz:
fixes multivalue ff regression by avoiding using `get_val`. Line::train calls repeatedly get_val, but get_val implementation on Column for multivalues is very slow. The fix is to use the iterator instead. Longterm fix should be to remove get_val access in serialization.
```
Old Code 8e773ade7715210fb18a9a3b870329eed35dd1ad

test fastfield::bench::bench_multi_value_ff_merge_few_segments                                                           ... bench:  46,103,960 ns/iter (+/- 2,066,083)
test fastfield::bench::bench_multi_value_ff_merge_many_segments                                                          ... bench:  83,073,036 ns/iter (+/- 4,373,615)
est fastfield::bench::bench_multi_value_ff_merge_many_segments_log_merge                                                ... bench:  64,178,576 ns/iter (+/- 1,466,700)

Current

running 3 tests
test fastfield::multivalued::bench::bench_multi_value_ff_merge_few_segments                                              ... bench:  57,379,523 ns/iter (+/- 3,220,787)
test fastfield::multivalued::bench::bench_multi_value_ff_merge_many_segments                                             ... bench:  90,831,688 ns/iter (+/- 1,445,486)
test fastfield::multivalued::bench::bench_multi_value_ff_merge_many_segments_log_merge                                   ... bench: 158,313,264 ns/iter (+/- 28,823,250)

With Fix

running 3 tests
test fastfield::multivalued::bench::bench_multi_value_ff_merge_few_segments                                              ... bench:  57,635,671 ns/iter (+/- 2,707,361)
test fastfield::multivalued::bench::bench_multi_value_ff_merge_many_segments                                             ... bench:  91,468,712 ns/iter (+/- 11,393,581)
test fastfield::multivalued::bench::bench_multi_value_ff_merge_many_segments_log_merge                                   ... bench:  73,909,138 ns/iter (+/- 15,846,097)
```

### DIFF
--- a/fastfield_codecs/src/line.rs
+++ b/fastfield_codecs/src/line.rs
@@ -69,11 +69,17 @@ impl Line {
 
     // Same as train, but the intercept is only estimated from provided sample positions
     pub fn estimate(ys: &dyn Column, sample_positions: &[u64]) -> Self {
-        Self::train_from(ys, sample_positions.iter().cloned())
+        Self::train_from(
+            ys,
+            sample_positions
+                .iter()
+                .cloned()
+                .map(|pos| (pos, ys.get_val(pos))),
+        )
     }
 
     // Intercept is only computed from provided positions
-    fn train_from(ys: &dyn Column, positions: impl Iterator<Item = u64>) -> Self {
+    fn train_from(ys: &dyn Column, positions_and_values: impl Iterator<Item = (u64, u64)>) -> Self {
         let num_vals = if let Some(num_vals) = NonZeroU64::new(ys.num_vals() - 1) {
             num_vals
         } else {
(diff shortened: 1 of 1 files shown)


fastfield_codecs/src/line.rs
```diff
@@ -69,11 +69,17 @@ impl Line {
 
     // Same as train, but the intercept is only estimated from provided sample positions
     pub fn estimate(ys: &dyn Column, sample_positions: &[u64]) -> Self {
-        Self::train_from(ys, sample_positions.iter().cloned())
+        Self::train_from(
+            ys,
+            sample_positions
+                .iter()
+                .cloned()
+                .map(|pos| (pos, ys.get_val(pos))),
```
fulmicoton, line 77: It would be much more efficient to remove the `get_val` call and use the iterator here.

let step = /* ..  */;
let index = (0..n).step_by(step);
let vals = column.step_by(step);
let positions_and_values = index.zip(vals);

fulmicoton: (that requires removing `Line::estimate` and call `Line::train_from` directly in the codec.)

PSeitz: This are only 20 samples, but span the whole Column, so I'm not so sure this is more efficient. It depends on the iterator implementation if it can skip efficiently.

PSeitz: The default impl calls `next()` https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#284-289, that's probably much slower for large indexes.

fulmicoton (approved):

PR #1545, opened 2022-09-22
clippy: Remove borrows that the compiler will do.

waywardmonkeys:
This started showing up with clippy in rust 1.64.


PSeitz (approved):
Thanks!

PR #1548, opened 2022-09-23
use binary search instead of linear for get_val in merge

PSeitz:


src/indexer/sorted_doc_id_multivalue_column.rs
```diff
@@ -72,13 +72,18 @@ impl<'a> SortedDocIdMultiValueColumn<'a> {
...
+        let mut new_doc_id: DocId = self
             .offsets
-            .iter()
-            .position(|&offset| offset > pos)
-            .expect("pos is out of bounds") as DocId
-            - 1u32;
+            .binary_search_by(|&offset| offset.cmp(&pos))
+            .unwrap_or_else(|pos| pos - 1) as DocId; // Offsets start at 0, so -1 is safe
+
+        // There are duplicates in the data, but we want the one, where the next offset is not the
```
adamreichold: Wouldn't it be possible to get this behaviour directly by using [`partition_point`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.partition_point) instead of `binary_search_by`?

adamreichold: I was thinking of something like

```rust
let new_doc_id = self.offsets.partition_point(|&offset| offset > pos) as DocId - 1;
```

which should point at last entry with `offset <= pos` immediately without the second loop, shouldn't it?

PSeitz: cool that method is rater new, but I had to flip the condition to make the partition work (`[true, true, .., false, false, ..]`)

adamreichold: Ah, I am sorry, I got confused. Yes, you did flip the condition, but should the extra while loop be unnecessary now because you already get the last index which fulfils the original condition?

PSeitz: ah yes, I forgot to remove that

adamreichold: I think there are the mutable binding and the extra braces inside the closure left from the original implementation.

fulmicoton (approved):

src/indexer/sorted_doc_id_multivalue_column.rs
fulmicoton: thank you @adamreichold

src/indexer/sorted_doc_id_multivalue_column.rs
```diff
@@ -72,13 +72,9 @@ impl<'a> SortedDocIdMultiValueColumn<'a> {
...
-        // the offsets are strictly increasing so we can do a simple search on it.
-        let new_doc_id: DocId = self
-            .offsets
-            .iter()
-            .position(|&offset| offset > pos)
-            .expect("pos is out of bounds") as DocId
-            - 1u32;
+        // the offsets are strictly increasing so we can do a binary search on it.
+
+        let new_doc_id: DocId = self.offsets.partition_point(|&offset| offset <= pos) as DocId - 1; // Offsets start at 0, so -1 is safe
```
fulmicoton, line 77: great comment!

PR #1551, opened 2022-09-25
reuse samples, add EstimateColumn

PSeitz:
estimations can be expensive since the samples span the whole column
and depending on the implementation get_val can not be easily computed
without an index.
EstimateColumn adds a view over the column which limits num_vals
to 100_000.


fulmicoton:
Closing as it was adressed by #1587

PR #1553, opened 2022-09-26
ip field

PSeitz:
- add u128 multivalue reader and writer
- add ip to schema
- add ip writers, handle merge

### DIFF
--- a/fastfield_codecs/benches/bench.rs
+++ b/fastfield_codecs/benches/bench.rs
@@ -100,7 +100,8 @@ mod tests {
 
     fn get_u128_column_from_data(data: &[u128]) -> Arc<dyn Column<u128>> {
         let mut out = vec![];
-        serialize_u128(VecColumn::from(&data), &mut out).unwrap();
+        let iter = || data.iter().cloned();
+        serialize_u128(iter, data.len() as u64, &mut out).unwrap();
         let out = OwnedBytes::new(out);
         open_u128::<u128>(out).unwrap()
     }
--- a/fastfield_codecs/benches/bench.rs
+++ b/fastfield_codecs/benches/bench.rs
@@ -100,8 +100,8 @@ mod tests {
 
     fn get_u128_column_from_data(data: &[u128]) -> Arc<dyn Column<u128>> {
         let mut out = vec![];
-        let iter = || data.iter().cloned();
-        serialize_u128(iter, data.len() as u64, &mut out).unwrap();
+        let iter_gen = || data.iter().cloned();
+        serialize_u128(iter_gen, data.len() as u64, &mut out).unwrap();
         let out = OwnedBytes::new(out);
         open_u128::<u128>(out).unwrap()
     }
--- a/fastfield_codecs/benches/bench.rs
+++ b/fastfield_codecs/benches/bench.rs
@@ -102,7 +102,7 @@ mod tests {
         let mut out = vec![];
         serialize_u128(VecColumn::from(&data), &mut out).unwrap();
         let out = OwnedBytes::new(out);
-        open_u128(out).unwrap()
+        open_u128::<u128>(out).unwrap()
     }
 
     #[bench]
--- a/fastfield_codecs/src/column.rs
+++ b/fastfield_codecs/src/column.rs
@@ -3,6 +3,8 @@ use std::ops::RangeInclusive;
 
 use tantivy_bitpacker::minmax;
 
+use crate::monotonic_mapping::StrictlyMonotonicFn;
+
 pub trait Column<T: PartialOrd = u64>: Send + Sync {
     /// Return the value associated with the given idx.
     ///
@@ -137,15 +139,14 @@ where V: AsRef<[T]> + ?Sized
     }
 }
 
-struct MonotonicMappingColumn<C, T, U, Input> {
+struct MonotonicMappingColumn<C, T, Input> {
     from_column: C,
     monotonic_mapping: T,
-    monotonic_mapping_inv: U,
     _phantom: PhantomData<Input>,
 }
 
-/// Creates a view of a column transformed by a monotonic mapping.
-/// E.g. apply a gcd monotonic_mapping([100, 200, 300]) == [1, 2, 3]
+/// Creates a view of a column transformed by a strictly monotonic mapping. See
+/// [`StrictlyMonotonicFn`]. E.g. apply a gcd monotonic_mapping([100, 200, 300]) == [1, 2, 3]
 /// The provided mappings need to be the inverse of each other.
 ///
 /// The inverse of the mapping is required for:
@@ -156,49 +157,44 @@ struct MonotonicMappingColumn<C, T, U, Input> {
 /// Note that when opening a codec, the monotonic_mapping should be the inverse of the mapping
 /// during serialization. And therefore the monotonic_mapping_inv when opening is the same as
 /// monotonic_mapping during serialization.
-pub fn monotonic_map_column<C, T, U, Input: PartialOrd, Output: PartialOrd + Clone>(
+pub fn monotonic_map_column<C, T, Input, Output>(
     from_column: C,
     monotonic_mapping: T,
-    monotonic_mapping_inv: U,
 ) -> impl Column<Output>
 where
     C: Column<Input>,
-    T: Fn(Input) -> Output + Send + Sync,
-    U: Fn(Output) -> Input + Send + Sync,
-    Input: Send + Sync,
-    Output: Send + Sync,
+    T: StrictlyMonotonicFn<Input, Output> + Send + Sync,
+    Input: PartialOrd + Send + Sync + Clone,
+    Output: PartialOrd + Send + Sync + Clone,
 {
     MonotonicMappingColumn {
         from_column,
         monotonic_mapping,
-        monotonic_mapping_inv,
         _phantom: PhantomData,
     }
 }
 
-impl<C, T, U, Input: PartialOrd, Output: PartialOrd> Column<Output>
-    for MonotonicMappingColumn<C, T, U, Input>
+impl<C, T, Input, Output> Column<Output> for MonotonicMappingColumn<C, T, Input>
 where
     C: Column<Input>,
-    T: Fn(Input) -> Output + Send + Sync,
-    U: Fn(Output) -> Input + Send + Sync,
-    Input: Send + Sync,
-    Output: Send + Sync + Clone,
+    T: StrictlyMonotonicFn<Input, Output> + Send + Sync,
+    Input: PartialOrd + Send + Sync + Clone,
+    Output: PartialOrd + Send + Sync + Clone,
 {
     #[inline]
     fn get_val(&self, idx: u64) -> Output {
         let from_val = self.from_column.get_val(idx);
-        (self.monotonic_mapping)(from_val)
+        self.monotonic_mapping.mapping(from_val)
     }
 
     fn min_value(&self) -> Output {
         let from_min_value = self.from_column.min_value();
-        (self.monotonic_mapping)(from_min_value)
+        self.monotonic_mapping.mapping(from_min_value)
     }
 
     fn max_value(&self) -> Output {
         let from_max_value = self.from_column.max_value();
-        (self.monotonic_mapping)(from_max_value)
+        self.monotonic_mapping.mapping(from_max_value)
     }
 
     fn num_vals(&self) -> u64 {
@@ -206,13 +202,17 @@ where
     }
 
     fn iter(&self) -> Box<dyn Iterator<Item = Output> + '_> {
-        Box::new(self.from_column.iter().map(&self.monotonic_mapping))
+        Box::new(
+            self.from_column
+                .iter()
+                .map(|el| self.monotonic_mapping.mapping(el)),
+        )
     }
 
     fn get_between_vals(&self, range: RangeInclusive<Output>) -> Vec<u64> {
         self.from_column.get_between_vals(
-            (self.monotonic_mapping_inv)(range.start().clone())
-                ..=(self.monotonic_mapping_inv)(range.end().clone()),
+            self.monotonic_mapping.inverse(range.start().clone())
+                ..=self.monotonic_mapping.inverse(range.end().clone()),
         )
     }
 
@@ -259,20 +259,20 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::MonotonicallyMappableToU64;
+    // use crate::MonotonicallyMappableToU64;
 
-    #[test]
-    fn test_monotonic_mapping() {
-        let vals = &[1u64, 3u64][..];
-        let col = VecColumn::from(vals);
-        let mapped = monotonic_map_column(col, |el| el + 4, |_el| unimplemented!());
-        assert_eq!(mapped.min_value(), 5u64);
-        assert_eq!(mapped.max_value(), 7u64);
-        assert_eq!(mapped.num_vals(), 2);
-        assert_eq!(mapped.num_vals(), 2);
-        assert_eq!(mapped.get_val(0), 5);
-        assert_eq!(mapped.get_val(1), 7);
-    }
+    //#[test]
+    // fn test_monotonic_mapping() {
+    // let vals = &[1u64, 3u64][..];
+    // let col = VecColumn::from(vals);
+    // let mapped = monotonic_map_column(col, |el| el + 4, |_el| unimplemented!());
+    // assert_eq!(mapped.min_value(), 5u64);
+    // assert_eq!(mapped.max_value(), 7u64);
+    // assert_eq!(mapped.num_vals(), 2);
+    // assert_eq!(mapped.num_vals(), 2);
+    // assert_eq!(mapped.get_val(0), 5);
+    // assert_eq!(mapped.get_val(1), 7);
+    //}
 
     #[test]
     fn test_range_as_col() {
@@ -281,35 +281,34 @@ mod tests {
         assert_eq!(col.max_value(), 99);
     }
 
-    #[test]
-    fn test_monotonic_mapping_iter() {
-        let vals: Vec<u64> = (-1..99).map(i64::to_u64).collect();
-        let col = VecColumn::from(&vals);
-        let mapped =
-            monotonic_map_column(col, |el| i64::from_u64(el) * 10i64, |_| unimplemented!());
-        let val_i64s: Vec<i64> = mapped.iter().collect();
-        for i in 0..100 {
-            assert_eq!(val_i64s[i as usize], mapped.get_val(i));
-        }
-    }
+    //#[test]
+    // fn test_monotonic_mapping_iter() {
+    // let vals: Vec<u64> = (-1..99).map(i64::to_u64).collect();
+    // let col = VecColumn::from(&vals);
+    // let mapped =
+    // monotonic_map_column(col, |el| i64::from_u64(el) * 10i64, |_| unimplemented!());
+    // let val_i64s: Vec<i64> = mapped.iter().collect();
+    // for i in 0..100 {
+    // assert_eq!(val_i64s[i as usize], mapped.get_val(i));
+    //}
 
-    #[test]
-    fn test_monotonic_mapping_get_range() {
-        let vals: Vec<u64> = (-1..99).map(i64::to_u64).collect();
-        let col = VecColumn::from(&vals);
-        let mapped =
-            monotonic_map_column(col, |el| i64::from_u64(el) * 10i64, |_| unimplemented!());
-        assert_eq!(mapped.min_value(), -10i64);
-        assert_eq!(mapped.max_value(), 980i64);
-        assert_eq!(mapped.num_vals(), 100);
-        let val_i64s: Vec<i64> = mapped.iter().collect();
-        assert_eq!(val_i64s.len(), 100);
-        for i in 0..100 {
-            assert_eq!(val_i64s[i as usize], mapped.get_val(i));
-            assert_eq!(val_i64s[i as usize], i64::from_u64(vals[i as usize]) * 10);
-        }
-        let mut buf = [0i64; 20];
-        mapped.get_range(7, &mut buf[..]);
-        assert_eq!(&val_i64s[7..][..20], &buf);
-    }
+    //#[test]
+    // fn test_monotonic_mapping_get_range() {
+    // let vals: Vec<u64> = (-1..99).map(i64::to_u64).collect();
+    // let col = VecColumn::from(&vals);
+    // let mapped =
+    // monotonic_map_column(col, |el| i64::from_u64(el) * 10i64, |_| unimplemented!());
+    // assert_eq!(mapped.min_value(), -10i64);
+    // assert_eq!(mapped.max_value(), 980i64);
+    // assert_eq!(mapped.num_vals(), 100);
+    // let val_i64s: Vec<i64> = mapped.iter().collect();
+    // assert_eq!(val_i64s.len(), 100);
+    // for i in 0..100 {
+    // assert_eq!(val_i64s[i as usize], mapped.get_val(i));
+    // assert_eq!(val_i64s[i as usize], i64::from_u64(vals[i as usize]) * 10);
+    //}
+    // let mut buf = [0i64; 20];
+    // mapped.get_range(7, &mut buf[..]);
+    // assert_eq!(&val_i64s[7..][..20], &buf);
+    //}
 }
--- a/fastfield_codecs/src/column.rs
+++ b/fastfield_codecs/src/column.rs
@@ -139,16 +139,27 @@ where V: AsRef<[T]> + ?Sized
 
 struct MonotonicMappingColumn<C, T, U, Input> {
     from_column: C,
-    monotonic_mapping_to_output: T,
-    monotonic_mapping_to_input: U,
+    monotonic_mapping: T,
+    monotonic_mapping_inv: U,
     _phantom: PhantomData<Input>,
 }
 
 /// Creates a view of a column transformed by a monotonic mapping.
+/// E.g. apply a gcd monotonic_mapping([100, 200, 300]) == [1, 2, 3]
+/// The provided mappings need to be the inverse of each other.
+///
+/// The inverse of the mapping is required for:
+/// `fn get_between_vals(&self, range: RangeInclusive<T>) -> Vec<u64> `
+/// The user provides the original value range and we need to monotonic map them in the same way the
+/// serialization does before calling the underlying column.
+///
+/// Note that when opening a codec, the monotonic_mapping should be the inverse of the mapping
+/// during serialization. And therefore the monotonic_mapping_inv when opening is the same as
+/// monotonic_mapping during serialization.
 pub fn monotonic_map_column<C, T, U, Input: PartialOrd, Output: PartialOrd + Clone>(
     from_column: C,
-    monotonic_mapping_to_output: T,
-    monotonic_mapping_to_input: U,
+    monotonic_mapping: T,
+    monotonic_mapping_inv: U,
 ) -> impl Column<Output>
 where
     C: Column<Input>,
@@ -159,8 +170,8 @@ where
 {
     MonotonicMappingColumn {
         from_column,
-        monotonic_mapping_to_output,
-        monotonic_mapping_to_input,
+        monotonic_mapping,
+        monotonic_mapping_inv,
         _phantom: PhantomData,
     }
 }
@@ -177,17 +188,17 @@ where
     #[inline]
     fn get_val(&self, idx: u64) -> Output {
         let from_val = self.from_column.get_val(idx);
-        (self.monotonic_mapping_to_output)(from_val)
+        (self.monotonic_mapping)(from_val)
     }
 
     fn min_value(&self) -> Output {
         let from_min_value = self.from_column.min_value();
-        (self.monotonic_mapping_to_output)(from_min_value)
+        (self.monotonic_mapping)(from_min_value)
     }
 
     fn max_value(&self) -> Output {
         let from_max_value = self.from_column.max_value();
-        (self.monotonic_mapping_to_output)(from_max_value)
+        (self.monotonic_mapping)(from_max_value)
     }
 
     fn num_vals(&self) -> u64 {
@@ -195,17 +206,13 @@ where
     }
 
     fn iter(&self) -> Box<dyn Iterator<Item = Output> + '_> {
-        Box::new(
-            self.from_column
-                .iter()
-                .map(&self.monotonic_mapping_to_output),
-        )
+        Box::new(self.from_column.iter().map(&self.monotonic_mapping))
     }
 
     fn get_between_vals(&self, range: RangeInclusive<Output>) -> Vec<u64> {
         self.from_column.get_between_vals(
-            (self.monotonic_mapping_to_input)(range.start().clone())
-                ..=(self.monotonic_mapping_to_input)(range.end().clone()),
+            (self.monotonic_mapping_inv)(range.start().clone())
+                ..=(self.monotonic_mapping_inv)(range.end().clone()),
         )
     }
 
@@ -258,7 +265,7 @@ mod tests {
     fn test_monotonic_mapping() {
         let vals = &[1u64, 3u64][..];
         let col = VecColumn::from(vals);
-        let mapped = monotonic_map_column(col, |el| el + 4, |el| el);
+        let mapped = monotonic_map_column(col, |el| el + 4, |_el| unimplemented!());
         assert_eq!(mapped.min_value(), 5u64);
         assert_eq!(mapped.max_value(), 7u64);
         assert_eq!(mapped.num_vals(), 2);
@@ -278,7 +285,8 @@ mod tests {
     fn test_monotonic_mapping_iter() {
         let vals: Vec<u64> = (-1..99).map(i64::to_u64).collect();
         let col = VecColumn::from(&vals);
-        let mapped = monotonic_map_column(col, |el| i64::from_u64(el) * 10i64, i64::to_u64);
+        let mapped =
+            monotonic_map_column(col, |el| i64::from_u64(el) * 10i64, |_| unimplemented!());
         let val_i64s: Vec<i64> = mapped.iter().collect();
         for i in 0..100 {
             assert_eq!(val_i64s[i as usize], mapped.get_val(i));
@@ -289,7 +297,8 @@ mod tests {
     fn test_monotonic_mapping_get_range() {
         let vals: Vec<u64> = (-1..99).map(i64::to_u64).collect();
         let col = VecColumn::from(&vals);
-        let mapped = monotonic_map_column(col, |el| i64::from_u64(el) * 10i64, i64::to_u64);
+        let mapped =
+            monotonic_map_column(col, |el| i64::from_u64(el) * 10i64, |_| unimplemented!());
         assert_eq!(mapped.min_value(), -10i64);
         assert_eq!(mapped.max_value(), 980i64);
         assert_eq!(mapped.num_vals(), 100);
--- a/fastfield_codecs/src/column.rs
+++ b/fastfield_codecs/src/column.rs
@@ -126,7 +126,8 @@ impl<'a, T: Copy + PartialOrd + Send + Sync> Column<T> for VecColumn<'a, T> {
 }
 
 impl<'a, T: Copy + Ord + Default, V> From<&'a V> for VecColumn<'a, T>
-where V: AsRef<[T]> + ?Sized
+where
+    V: AsRef<[T]> + ?Sized,
 {
     fn from(values: &'a V) -> Self {
         let values = values.as_ref();
@@ -223,7 +224,8 @@ where
 pub struct IterColumn<T>(T);
 
 impl<T> From<T> for IterColumn<T>
-where T: Iterator + Clone + ExactSizeIterator
+where
+    T: Iterator + Clone + ExactSizeIterator,
 {
     fn from(iter: T) -> Self {
         IterColumn(iter)
@@ -259,20 +261,23 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    // use crate::MonotonicallyMappableToU64;
+    use crate::monotonic_mapping::{
+        StrictlyMonotonicMappingInverter, StrictlyMonotonicMappingToInternalBaseval,
+        StrictlyMonotonicMappingToInternalGCDBaseval,
+    };
 
-    //#[test]
-    // fn test_monotonic_mapping() {
-    // let vals = &[1u64, 3u64][..];
-    // let col = VecColumn::from(vals);
-    // let mapped = monotonic_map_column(col, |el| el + 4, |_el| unimplemented!());
-    // assert_eq!(mapped.min_value(), 5u64);
-    // assert_eq!(mapped.max_value(), 7u64);
-    // assert_eq!(mapped.num_vals(), 2);
-    // assert_eq!(mapped.num_vals(), 2);
-    // assert_eq!(mapped.get_val(0), 5);
-    // assert_eq!(mapped.get_val(1), 7);
-    //}
+    #[test]
+    fn test_monotonic_mapping() {
+        let vals = &[3u64, 5u64][..];
+        let col = VecColumn::from(vals);
+        let mapped = monotonic_map_column(col, StrictlyMonotonicMappingToInternalBaseval::new(2));
+        assert_eq!(mapped.min_value(), 1u64);
+        assert_eq!(mapped.max_value(), 3u64);
+        assert_eq!(mapped.num_vals(), 2);
+        assert_eq!(mapped.num_vals(), 2);
+        assert_eq!(mapped.get_val(0), 1);
+        assert_eq!(mapped.get_val(1), 3);
+    }
 
     #[test]
     fn test_range_as_col() {
@@ -281,34 +286,44 @@ mod tests {
         assert_eq!(col.max_value(), 99);
     }
 
-    //#[test]
-    // fn test_monotonic_mapping_iter() {
-    // let vals: Vec<u64> = (-1..99).map(i64::to_u64).collect();
-    // let col = VecColumn::from(&vals);
-    // let mapped =
-    // monotonic_map_column(col, |el| i64::from_u64(el) * 10i64, |_| unimplemented!());
-    // let val_i64s: Vec<i64> = mapped.iter().collect();
-    // for i in 0..100 {
-    // assert_eq!(val_i64s[i as usize], mapped.get_val(i));
-    //}
+    #[test]
+    fn test_monotonic_mapping_iter() {
+        let vals: Vec<u64> = (10..110u64).map(|el| el * 10).collect();
+        let col = VecColumn::from(&vals);
+        let mapped = monotonic_map_column(
+            col,
+            StrictlyMonotonicMappingInverter::from(
+                StrictlyMonotonicMappingToInternalGCDBaseval::new(10, 100),
+            ),
+        );
+        let val_i64s: Vec<u64> = mapped.iter().collect();
+        for i in 0..100 {
+            assert_eq!(val_i64s[i as usize], mapped.get_val(i));
+        }
+    }
 
-    //#[test]
-    // fn test_monotonic_mapping_get_range() {
-    // let vals: Vec<u64> = (-1..99).map(i64::to_u64).collect();
-    // let col = VecColumn::from(&vals);
-    // let mapped =
-    // monotonic_map_column(col, |el| i64::from_u64(el) * 10i64, |_| unimplemented!());
-    // assert_eq!(mapped.min_value(), -10i64);
-    // assert_eq!(mapped.max_value(), 980i64);
-    // assert_eq!(mapped.num_vals(), 100);
-    // let val_i64s: Vec<i64> = mapped.iter().collect();
-    // assert_eq!(val_i64s.len(), 100);
-    // for i in 0..100 {
-    // assert_eq!(val_i64s[i as usize], mapped.get_val(i));
-    // assert_eq!(val_i64s[i as usize], i64::from_u64(vals[i as usize]) * 10);
-    //}
-    // let mut buf = [0i64; 20];
-    // mapped.get_range(7, &mut buf[..]);
-    // assert_eq!(&val_i64s[7..][..20], &buf);
-    //}
+    #[test]
+    fn test_monotonic_mapping_get_range() {
+        let vals: Vec<u64> = (0..100u64).map(|el| el * 10).collect();
+        let col = VecColumn::from(&vals);
+        let mapped = monotonic_map_column(
+            col,
+            StrictlyMonotonicMappingInverter::from(
+                StrictlyMonotonicMappingToInternalGCDBaseval::new(10, 0),
+            ),
+        );
+
+        assert_eq!(mapped.min_value(), 0u64);
+        assert_eq!(mapped.max_value(), 9900u64);
+        assert_eq!(mapped.num_vals(), 100);
+        let val_u64s: Vec<u64> = mapped.iter().collect();
+        assert_eq!(val_u64s.len(), 100);
+        for i in 0..100 {
+            assert_eq!(val_u64s[i as usize], mapped.get_val(i));
+            assert_eq!(val_u64s[i as usize], vals[i as usize] * 10);
+        }
+        let mut buf = [0u64; 20];
+        mapped.get_range(7, &mut buf[..]);
+        assert_eq!(&val_u64s[7..][..20], &buf);
+    }
 }
--- a/fastfield_codecs/src/column.rs
+++ b/fastfield_codecs/src/column.rs
@@ -124,8 +124,7 @@ impl<'a, T: Copy + PartialOrd + Send + Sync> Column<T> for VecColumn<'a, T> {
 }
 
 impl<'a, T: Copy + Ord + Default, V> From<&'a V> for VecColumn<'a, T>
-where
-    V: AsRef<[T]> + ?Sized,
+where V: AsRef<[T]> + ?Sized
 {
     fn from(values: &'a V) -> Self {
         let values = values.as_ref();
@@ -217,8 +216,7 @@ where
 pub struct IterColumn<T>(T);
 
 impl<T> From<T> for IterColumn<T>
-where
-    T: Iterator + Clone + ExactSizeIterator,
+where T: Iterator + Clone + ExactSizeIterator
 {
     fn from(iter: T) -> Self {
         IterColumn(iter)
--- a/fastfield_codecs/src/column.rs
+++ b/fastfield_codecs/src/column.rs
@@ -126,8 +126,7 @@ impl<'a, T: Copy + PartialOrd + Send + Sync> Column<T> for VecColumn<'a, T> {
 }
 
 impl<'a, T: Copy + Ord + Default, V> From<&'a V> for VecColumn<'a, T>
-where
-    V: AsRef<[T]> + ?Sized,
+where V: AsRef<[T]> + ?Sized
 {
     fn from(values: &'a V) -> Self {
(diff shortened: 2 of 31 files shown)


src/fastfield/writer.rs
```diff
@@ -230,6 +269,161 @@ impl FastFieldsWriter {
...
+/// We cannot serialize earlier as the values are
+/// compressed to a compact number space and the number of
+/// bits required for bitpacking can only been known once
+/// we have seen all of the values.
+pub struct U128FastFieldWriter {
+    field: Field,
+    vals: Vec<u128>,
+    val_count: u32,
+
+    null_values: RoaringBitmap,
```
fulmicoton: We don't allow null_values yet.

Adding without a thought for other fields adds up with extra complexity.

src/fastfield/writer.rs
```diff
@@ -230,6 +269,161 @@ impl FastFieldsWriter {
...
+
+        let mut val_idx = 0;
+        for idx in 0..self.val_count {
+            if !self.null_values.contains(idx as u32) {
+                idx_to_val_idx[idx as usize] = val_idx as u32;
+                val_idx += 1;
+            }
+        }
+
+        struct RemappedFFWriter<'a> {
```
fulmicoton: No inline structs like this.

src/fastfield/writer.rs
```diff
@@ -230,6 +269,161 @@ impl FastFieldsWriter {
...
+
+            fn max_value(&self) -> u128 {
+                // unused by codec
+                unreachable!()
+            }
+
+            fn num_vals(&self) -> u64 {
+                self.val_count as u64
+            }
+            fn iter<'b>(&'b self) -> Box<dyn Iterator<Item = u128> + 'b> {
```
fulmicoton:
```suggestion
            fn iter(&self) -> Box<dyn Iterator<Item = u128> + '_> {
```

src/fastfield/writer.rs
PSeitz: yes, but I don't want to rollback all the null handling. The implementation is already broken for null because of the removed null handling. Edit: We allow null values, we just don't handle it right

src/query/query_parser/query_parser.rs
```diff
@@ -506,6 +507,13 @@ impl QueryParser {
                 let bytes_term = Term::from_field_bytes(field, &bytes);
                 Ok(vec![LogicalLiteral::Term(bytes_term)])
             }
+            FieldType::Ip(ref option) => {
+                if !option.is_indexed() {
+                    return Err(QueryParserError::FieldNotIndexed(field_name.to_string()));
+                }
+                let text_term = Term::from_field_text(field, phrase);
```
fulmicoton: I'm ok with using text for IP if you do it right, but this here is not good at all.
A given IP have way more than a single representation.

e.g.
```
- 129.144.50.56
- 129.144.050.056
- 129.144.050.56
- 129.144.050.056
- ::ffff:8190:3238
- 0:0:0:0:0:ffff:8190:3238
- 0000:0000:0000:0000:0000:ffff:8190:3238
```

are actually the same ip address.

src/fastfield/writer.rs
PSeitz: I think it increases readability, because of the reduced scope

src/query/query_parser/query_parser.rs
PSeitz: I'll remove it from this PR, we can add it later

src/fastfield/writer.rs
fulmicoton: I disagree.

PSeitz: I switched to a iter api, it's much less code and also faster

src/fastfield/writer.rs
fulmicoton: Remove it.

src/schema/document.rs
```diff
@@ -97,6 +98,11 @@ impl Document {
         self.add_field_value(field, value);
     }
 
+    /// Add a u64 field
```
fulmicoton:
```suggestion
    /// Add a IP address field
```

src/schema/document.rs
```diff
@@ -97,6 +98,11 @@ impl Document {
...
+    pub fn add_ip(&mut self, field: Field, value: IpAddr) {
```
fulmicoton:
```suggestion
    pub fn add_ip_addr(&mut self, field: Field, value: IpAddr) {
```

src/schema/field_entry.rs
```diff
@@ -60,6 +61,11 @@ impl FieldEntry {
         Self::new(field_name, FieldType::Date(date_options))
     }
 
+    /// Creates a new ip field entry.
+    pub fn new_ip(field_name: String, ip_options: IpOptions) -> FieldEntry {
```
fulmicoton:
```suggestion
    pub fn new_ip_addr(field_name: String, ip_options: IpOptions) -> FieldEntry {
```

src/postings/per_field_postings_writer.rs
```diff
@@ -50,6 +50,7 @@ fn posting_writer_from_field_entry(field_entry: &FieldEntry) -> Box<dyn Postings
         | FieldType::Bool(_)
         | FieldType::Date(_)
         | FieldType::Bytes(_)
+        | FieldType::Ip(_)
```
fulmicoton:
```suggestion
        | FieldType::IpAddr(_)
```

src/indexer/merger.rs
```diff
@@ -303,6 +322,82 @@ impl IndexMerger {
...
     }
 
+    // used to merge `u128` single fast fields.
+    fn write_u128_multi_fast_field(
+        &self,
+        field: Field,
+        fast_field_serializer: &mut CompositeFastFieldSerializer,
+        doc_id_mapping: &SegmentDocIdMapping,
+    ) -> crate::Result<()> {
+        let segment_and_ff_readers = self
```
fulmicoton: The type is non trivial.
```
        let segment_and_ff_readers: Vec<(&SegmentReader, MultivaluedU128FastFieldReader<u128>)> = self
```

src/indexer/merger.rs
```diff
@@ -303,6 +322,82 @@ impl IndexMerger {
...
+
+        let fast_field_readers = segment_and_ff_readers
+            .into_iter()
+            .map(|(_, ff_reader)| ff_reader)
+            .collect::<Vec<_>>();
+
+        let iter = || {
+            doc_id_mapping.iter_old_doc_addrs().flat_map(|doc_addr| {
+                let fast_field_reader = &fast_field_readers[doc_addr.segment_ord as usize];
+                let mut out = vec![];
```
fulmicoton: Please remove the allocation in each iteration of the Iterator.

fastfield_codecs/src/serialize.rs
```diff
@@ -129,28 +134,28 @@ pub fn estimate<T: MonotonicallyMappableToU64>(
...
         FastFieldCodecType::Bitpacked => BitpackedCodec::estimate(&normalized_column),
         FastFieldCodecType::Linear => LinearCodec::estimate(&normalized_column),
         FastFieldCodecType::BlockwiseLinear => BlockwiseLinearCodec::estimate(&normalized_column),
     }
 }
 
-pub fn serialize_u128(
-    typed_column: impl Column<u128>,
+pub fn serialize_u128<F: Fn() -> I, I: Iterator<Item = u128>>(
+    iter_gen: F,
```
fulmicoton, line 152: why not Column?

src/fastfield/multivalued/writer.rs
```diff
@@ -312,6 +314,137 @@ impl<'a, C: Column> Iterator for MultivalueStartIndexIter<'a, C> {
...
+                    0,
+                )?;
+            } else {
+                serializer.create_auto_detect_u64_fast_field_with_idx(self.field, col, 0)?;
+            }
+        }
+        {
+            let field_write = serializer.get_field_writer(self.field, 1);
+
+            let iter = || self.get_ordered_values(doc_id_map).flatten().cloned();
```
fulmicoton:
```suggestion
            let iter_gen = || self.get_ordered_values(doc_id_map).flatten().cloned();
```

src/fastfield/multivalued/writer.rs
```diff
@@ -312,6 +314,137 @@ impl<'a, C: Column> Iterator for MultivalueStartIndexIter<'a, C> {
...
+            serialize_u128(iter, self.vals.len() as u64, field_write)?;
```
fulmicoton:
```suggestion
            serialize_u128(iter_gen, self.vals.len() as u64, field_write)?;
```

src/fastfield/multivalued/writer.rs
```diff
@@ -312,6 +314,137 @@ impl<'a, C: Column> Iterator for MultivalueStartIndexIter<'a, C> {
...
+
+/// Returns an iterator over values per doc_id in ascending doc_id order.
+///
+/// Normally the order is simply iterating self.doc_id_index.
+/// With doc_id_map it accounts for the new mapping, returning values in the order of the
+/// new doc_ids.
+fn get_ordered_values<'a: 'b, 'b, T>(
+    vals: &'a [T],
+    doc_index: &'a [u64],
+    doc_id_map: Option<&'b DocIdMapping>,
```
fulmicoton, line 385: Can you open a ticket to make `DocIdMapping` and enum? The option makes the code ugly in many places.

```
enum DocIdMapping {
     Shuffled(DocIdPermutation),
     Trivial { max_doc: DocId }
}

File: src/fastfield/multivalued/writer.rs [hunk 15]
```diff
@@ -312,6 +314,137 @@ impl<'a, C: Column> Iterator for MultivalueStartIndexIter<'a, C> {
...
+    }
+
+    /// Shift to the next document and adds
+    /// all of the matching field values present in the document.
+    pub fn add_document(&mut self, doc: &Document) {
+        self.next_doc();
+        for field_value in doc.field_values() {
+            if field_value.field == self.field {
+                let value = field_value.value();
+                let ip_addr = value.as_ip().unwrap();
```
fulmicoton: `unwrap` is really bad here. This is stuff that the user is giving us. We need at the very least to panic with a decent error message.

src/fastfield/multivalued/writer.rs
```diff
@@ -312,6 +314,137 @@ impl<'a, C: Column> Iterator for MultivalueStartIndexIter<'a, C> {
...
+            field,
+            vals: Vec::new(),
+            doc_index: Vec::new(),
+        }
+    }
+
+    /// The memory used (inclusive childs)
+    pub fn mem_usage(&self) -> usize {
+        self.vals.capacity() * std::mem::size_of::<UnorderedTermId>()
+            + self.doc_index.capacity() * std::mem::size_of::<u64>()
```
fulmicoton, line 300: If we have this capacity * mem size item in many places of the code, it might be useful to have a helper funciton somewhere.
```
fn size_of_vec<T>(vec: &Vec<T>) -> usize
```
The point being that it would make sure more resilient to refactoring. Right now if sometimes change the content of the Vec, they might easily forget to update this funciton.

src/fastfield/multivalued/reader.rs
```diff
@@ -99,6 +99,153 @@ impl<Item: FastValue> MultiValueLength for MultiValuedFastFieldReader<Item> {
...
+                if Some(cur_doc) == last_doc {
+                    break;
+                }
+                docs.push(cur_doc);
+                last_doc = Some(cur_doc);
+                break;
+            }
+            cur_doc += 1;
+        }
+    }
```
fulmicoton:
```suggestion
```

src/fastfield/multivalued/reader.rs
```diff
@@ -99,6 +99,153 @@ impl<Item: FastValue> MultiValueLength for MultiValuedFastFieldReader<Item> {
...
+    pub fn get_vals(&self, doc: DocId, vals: &mut Vec<T>) {
+        let range = self.range(doc);
+        self.get_vals_for_range(range, vals);
+    }
+
+    /// Returns all docids which are in the provided value range
+    pub fn get_between_vals(&self, range: RangeInclusive<T>) -> Vec<DocId> {
+        let positions = self.vals_reader.get_between_vals(range);
+
+        positions_to_docids(&positions, self)
```
fulmicoton:
```suggestion
        self.vals_reader.get_between_vals(range)
```

fulmicoton: No handling of null for the moment.

src/indexer/merger.rs
PSeitz: It's not so easy, we tried that in the other case unsuccessfully. It would be easier with an iterator instead of get_vals.

src/fastfield/multivalued/writer.rs
PSeitz: https://github.com/quickwit-oss/tantivy/issues/1571

src/fastfield/multivalued/writer.rs
PSeitz: good catch, the behavior in tantivy is inconsistent here. The posting writer part returns an error and the fastfields panic. I think we should return errors here everywhere

fulmicoton: Error would be better yes.

src/indexer/merger.rs
fulmicoton: > It's not so easy, we tried that in the other case unsuccessfully. It would be easier with an iterator instead of get_vals.

I sent you a solution before. You need an iterator implementation.

src/fastfield/multivalued/writer.rs
PSeitz: We could add a trait, and implement it on Vec and our types
``` 
trait MemUsage{
    fn mem_usage(&self) -> usize;
}
```

fulmicoton: ok with that solution

src/fastfield/multivalued/reader.rs
PSeitz: That's not null handling, but mulitivalue handling

src/indexer/merger.rs
PSeitz: I would rather keep the code simple and slightly slower here.

src/fastfield/multivalued/writer.rs
PSeitz: https://github.com/quickwit-oss/tantivy/issues/1572

src/fastfield/multivalued/writer.rs
PSeitz: https://github.com/quickwit-oss/tantivy/issues/1573

fastfield_codecs/benches/bench.rs
```diff
@@ -100,9 +100,10 @@ mod tests {
 
     fn get_u128_column_from_data(data: &[u128]) -> Arc<dyn Column<u128>> {
         let mut out = vec![];
-        serialize_u128(VecColumn::from(&data), &mut out).unwrap();
+        let iter = || data.iter().cloned();
```
fulmicoton:
```suggestion
        let iter_fn = || data.iter().cloned();
```

fastfield_codecs/src/column.rs
```diff
@@ -242,7 +258,7 @@ mod tests {
     fn test_monotonic_mapping() {
         let vals = &[1u64, 3u64][..];
         let col = VecColumn::from(vals);
-        let mapped = monotonic_map_column(col, |el| el + 4);
+        let mapped = monotonic_map_column(col, |el| el + 4, |el| el);
```
fulmicoton: I don't understand. Isn't the second mapping supposed to be the inverse mapping?

fastfield_codecs/src/column.rs
```diff
@@ -262,7 +278,7 @@ mod tests {
     fn test_monotonic_mapping_iter() {
         let vals: Vec<u64> = (-1..99).map(i64::to_u64).collect();
         let col = VecColumn::from(&vals);
-        let mapped = monotonic_map_column(col, |el| i64::from_u64(el) * 10i64);
+        let mapped = monotonic_map_column(col, |el| i64::from_u64(el) * 10i64, i64::to_u64);
```
fulmicoton: same thing here

fastfield_codecs/src/column.rs
```diff
@@ -273,7 +289,7 @@ mod tests {
     fn test_monotonic_mapping_get_range() {
         let vals: Vec<u64> = (-1..99).map(i64::to_u64).collect();
         let col = VecColumn::from(&vals);
-        let mapped = monotonic_map_column(col, |el| i64::from_u64(el) * 10i64);
+        let mapped = monotonic_map_column(col, |el| i64::from_u64(el) * 10i64, i64::to_u64);
```
fulmicoton: same

fastfield_codecs/src/column.rs
```diff
@@ -137,60 +137,76 @@ where V: AsRef<[T]> + ?Sized
     }
 }
 
-struct MonotonicMappingColumn<C, T, Input> {
+struct MonotonicMappingColumn<C, T, U, Input> {
     from_column: C,
-    monotonic_mapping: T,
+    monotonic_mapping_to_output: T,
+    monotonic_mapping_to_input: U,
```
fulmicoton: because of the naming I thought you wanted to build something like 

```
g(col(f(i)))
```
Can you call it maybe
`monotonic_mapping` and
`monotonic_mapping_inv`

Also... note there is something tricky here. By introducing the reciprocal function, you forbid the use of non-strictly monotonic function.

fastfield_codecs/src/column.rs
```diff
@@ -137,60 +137,76 @@ where V: AsRef<[T]> + ?Sized
...
     _phantom: PhantomData<Input>,
 }
 
 /// Creates a view of a column transformed by a monotonic mapping.
```
fulmicoton: update the comment.

fastfield_codecs/src/lib.rs
```diff
@@ -100,10 +110,18 @@ fn open_specific_codec<C: FastFieldCodec, Item: MonotonicallyMappableToU64>(
...
+            monotonic_mapping,
+            Item::to_u64,
+        )))
     } else {
         let monotonic_mapping = move |val: u64| Item::from_u64(min_value + val);
-        Ok(Arc::new(monotonic_map_column(reader, monotonic_mapping)))
+        Ok(Arc::new(monotonic_map_column(
+            reader,
+            monotonic_mapping,
+            Item::to_u64,
```
fulmicoton: again...

src/directory/mmap_directory.rs
```diff
@@ -610,6 +610,7 @@ mod tests {
             assert!(num_segments <= 4);
             let num_components_except_deletes_and_tempstore =
                 crate::core::SegmentComponent::iterator().len() - 2;
+            std::thread::sleep(Duration::from_millis(100));
```
fulmicoton: what is this?

src/indexer/segment_writer.rs
```diff
@@ -296,6 +296,13 @@ impl SegmentWriter {
                         ctx,
                     )?;
                 }
+                FieldType::IpAddr(_) => {
+                    for value in values {
+                        let ip_val = value.as_ip().ok_or_else(make_schema_error)?;
+                        term_buffer.set_text(&ip_val.to_string());
+                        postings_writer.subscribe(doc_id, 0u32, term_buffer, ctx);
```
fulmicoton: this hasn't been addressed.

src/query/query_parser/query_parser.rs
```diff
@@ -506,6 +507,7 @@ impl QueryParser {
                 let bytes_term = Term::from_field_bytes(field, &bytes);
                 Ok(vec![LogicalLiteral::Term(bytes_term)])
             }
+            FieldType::IpAddr(_) => Err(QueryParserError::FieldNotIndexed(field_name.to_string())),
```
fulmicoton, line 512: !?

fastfield_codecs/src/column.rs
PSeitz: We need that for
`fn get_between_vals(&self, range: RangeInclusive<T>) -> Vec<u64> `

The user provides the original values and we need to monotonic map them in the same way the serialization does before calling the underlying column.

> By introducing the reciprocal function, you forbid the use of non-strictly monotonic function.

We had this limitation before, but only implicit, via serialization and reading. I'll move the code together to make that more explicit.

The API is a little strange for serialization, since semantically it doesn't make sense to have an inverse. In that case it's a one way and the inverse is unimplemented.


yes, `monotonic_mapping` and `monotonic_mapping_inv` are better names. Input and output flip depending if we are serializing or reading, and that's gets confusing with the mapping.

fastfield_codecs/src/column.rs
PSeitz: it's unused I replaced it with `unimplemented()`

fastfield_codecs/src/lib.rs
PSeitz: That's a bug

src/directory/mmap_directory.rs
PSeitz: the mmap_released test is still flaky, that was a test to fix it.

fastfield_codecs/src/compact_space/mod.rs
```diff
@@ -171,10 +171,10 @@ pub struct IPCodecParams {
 
 impl CompactSpaceCompressor {
     /// Taking the vals as Vec may cost a lot of memory. It is used to sort the vals.
-    pub fn train_from(column: &impl Column<u128>) -> Self {
+    pub fn train_from(iter: impl Iterator<Item = u128>, num_vals: u64) -> Self {
```
fulmicoton, line 174: Would ExactSizeIterator be usable here?

fastfield_codecs/src/column.rs
```diff
@@ -137,37 +137,53 @@ where V: AsRef<[T]> + ?Sized
...
     from_column: C,
     monotonic_mapping: T,
+    monotonic_mapping_inv: U,
     _phantom: PhantomData<Input>,
 }
 
 /// Creates a view of a column transformed by a monotonic mapping.
-pub fn monotonic_map_column<C, T, Input: PartialOrd, Output: PartialOrd>(
+/// E.g. apply a gcd monotonic_mapping([100, 200, 300]) == [1, 2, 3]
+/// The provided mappings need to be the inverse of each other.
```
fulmicoton: That's a tad inaccurate.

`monotonic_mapping` is expected to be injective, and we should always have 
`monotonic_mapping_inv(monotonic_mapping(el)) == el`.

fulmicoton: (the opposite is not true... Take the GCD mapping for instance.)

fastfield_codecs/src/column.rs
```diff
@@ -137,37 +137,53 @@ where V: AsRef<[T]> + ?Sized
...
+/// The user provides the original value range and we need to monotonic map them in the same way the
+/// serialization does before calling the underlying column.
+///
+/// Note that when opening a codec, the monotonic_mapping should be the inverse of the mapping
+/// during serialization. And therefore the monotonic_mapping_inv when opening is the same as
+/// monotonic_mapping during serialization.
+pub fn monotonic_map_column<C, T, U, Input: PartialOrd, Output: PartialOrd + Clone>(
     from_column: C,
     monotonic_mapping: T,
+    monotonic_mapping_inv: U,
```
fulmicoton: I think this is a great idea to introduce the notion of inverse mapping.

Can we go a step further and introduce a `StrictlyMonotonicMapping` trait.

```rust
pub trait StrictlyMonotonicFn<Input, Ouput> {
    fn mapping(&self, inp: Input) -> Output;
    fn inverse(&self, inp: Output) -> Input;
}
```

Two benefits:
- The code will be a tad clearer.
- The contract will be clearer.
- The GCD stuff will be implemented once and for all and reuse instead of being implemented twice with a useless unimplemented!().

src/fastfield/multivalued/reader.rs
```diff
@@ -99,6 +99,153 @@ impl<Item: FastValue> MultiValueLength for MultiValuedFastFieldReader<Item> {
...
+/// TODO: Instead of a linear scan we can employ a binary search to match a docid to its value
+/// position.
+fn positions_to_docids<T: MultiValueLength>(positions: &[u64], multival_idx: &T) -> Vec<DocId> {
+    let mut docs = vec![];
+    let mut cur_doc = 0u32;
+    let mut last_doc = None;
+
+    for pos in positions {
+        loop {
+            let range = multival_idx.get_range(cur_doc);
```
fulmicoton: We don't need the range. The end index was sufficient. that's two gets we do here when one was sufficient.

src/fastfield/multivalued/reader.rs
```diff
@@ -99,6 +99,153 @@ impl<Item: FastValue> MultiValueLength for MultiValuedFastFieldReader<Item> {
...
+
+/// Converts a list of positions of values in a 1:n index to the corresponding list of DocIds.
+///
+/// Since there is no index for value pos -> docid, but docid -> value pos range, we scan the index.
+///
+/// Correctness: positions needs to be sorted.
+///
+/// TODO: Instead of a linear scan we can employ a binary search to match a docid to its value
+/// position.
+fn positions_to_docids<T: MultiValueLength>(positions: &[u64], multival_idx: &T) -> Vec<DocId> {
```
fulmicoton: This function deserves unit testing.

src/fastfield/multivalued/reader.rs
```diff
@@ -99,6 +99,153 @@ impl<Item: FastValue> MultiValueLength for MultiValuedFastFieldReader<Item> {
...
+    }
+}
+
+/// Converts a list of positions of values in a 1:n index to the corresponding list of DocIds.
+///
+/// Since there is no index for value pos -> docid, but docid -> value pos range, we scan the index.
+///
+/// Correctness: positions needs to be sorted.
+///
+/// TODO: Instead of a linear scan we can employ a binary search to match a docid to its value
```
fulmicoton: i would suggest exponential search into binary search

src/fastfield/multivalued/writer.rs
```diff
@@ -312,6 +314,139 @@ impl<'a, C: Column> Iterator for MultivalueStartIndexIter<'a, C> {
...
+    /// all of the matching field values present in the document.
+    pub fn add_document(&mut self, doc: &Document) {
+        self.next_doc();
+        for field_value in doc.field_values() {
+            if field_value.field == self.field {
+                let value = field_value.value();
+                let ip_addr = value
+                    .as_ip()
+                    .unwrap_or_else(|| panic!("expected and ip, but got {:?}", value));
+                let value = ip_addr.to_u128();
```
fulmicoton:
```suggestion
                let ip_addr_u128 = ip_addr.to_u128();
```

src/fastfield/multivalued/writer.rs
```diff
@@ -312,6 +314,139 @@ impl<'a, C: Column> Iterator for MultivalueStartIndexIter<'a, C> {
...
+                self.add_val(value);
```
fulmicoton:
```suggestion
                self.add_val(ip_addr_u128);
```

src/fastfield/writer.rs
```diff
@@ -230,6 +268,101 @@ impl FastFieldsWriter {
...
+
+        if let Some(doc_id_map) = doc_id_map {
+            let iter_gen = || {
+                doc_id_map
+                    .iter_old_doc_ids()
+                    .map(|idx| self.vals[idx as usize])
+            };
+            serialize_u128(iter_gen, self.val_count as u64, field_write)?;
+        } else {
+            let iter_gen = || (0..self.val_count).map(|idx| self.vals[idx as usize]);
```
fulmicoton: isn't it the same as?
```suggestion
            let iter_gen = || { self.vals[0..self.val_count].iter().copied() };
```

fastfield_codecs/src/compact_space/mod.rs
PSeitz: in principle yes, but the api would be much harder to use

src/fastfield/serializer/mod.rs
```diff
@@ -79,6 +79,11 @@ impl CompositeFastFieldSerializer {
         self.composite_write.for_field_with_idx(field, 1)
     }
 
+    /// Gets the underlying writer
+    pub fn get_field_writer(&mut self, field: Field, idx: usize) -> &mut impl Write {
+        self.composite_write.for_field_with_idx(field, idx)
```
fulmicoton: I think the other fast field writer leaves the serialization logic in the serializer.

THe fast field writer accumulates values in memory, and then is in charge of passing a view over the data to the serializer which in turn serializes. 

If this is correct, why did you change the logic?

src/query/query_parser/query_parser.rs
```diff
@@ -400,6 +400,7 @@ impl QueryParser {
                 let bytes = base64::decode(phrase).map_err(QueryParserError::ExpectedBase64)?;
                 Ok(Term::from_field_bytes(field, &bytes))
             }
+            FieldType::IpAddr(_) => Ok(Term::from_field_text(field, phrase)),
```
fulmicoton: What is this?

src/schema/field_type.rs
```diff
@@ -210,6 +228,7 @@ impl FieldType {
             | FieldType::F64(ref int_options)
             | FieldType::Bool(ref int_options) => int_options.is_fast(),
             FieldType::Date(ref date_options) => date_options.is_fast(),
+            FieldType::IpAddr(ref options) => options.is_fast(),
```
fulmicoton:
```suggestion
            FieldType::IpAddr(ref ip_addr_options) => ip_addr_options.is_fast(),
```

src/schema/schema.rs
```diff
@@ -144,6 +145,28 @@ impl SchemaBuilder {
         self.add_field(field_entry)
     }
 
+    /// Adds a ip field.
+    /// Returns the associated field handle
+    /// Internally, Tantivy simply stores ips as u64,
+    /// while the user supplies IpAddr values for convenience.
```
fulmicoton:
```suggestion
```

fulmicoton: This is not useful and not really correct .

src/schema/schema.rs
```diff
@@ -144,6 +145,28 @@ impl SchemaBuilder {
...
+    ///
+    /// # Caution
+    ///
+    /// Appending two fields with the same name
+    /// will result in the shadowing of the first
+    /// by the second one.
+    /// The first field will get a field id
+    /// but only the second one will be indexed
```
fulmicoton, line 157: This is an excellent point, but it is very strange to put it here only (!?)

Can you create an issue to fix this.
I think we should panic instead, and document that this is forbidden.

src/schema/term.rs
```diff
@@ -415,6 +415,14 @@ fn debug_value_bytes(typ: Type, bytes: &[u8], f: &mut fmt::Formatter) -> fmt::Re
                 debug_value_bytes(typ, bytes, f)?;
             }
         }
+        Type::Ip => {
+            let s = as_str(bytes); // TODO: change when serialization changes
+            write_opt(f, s)?;
+        }
+        Type::U128 => {
```
fulmicoton: I don't think we want a u128.

src/schema/term.rs
```diff
@@ -415,6 +415,14 @@ fn debug_value_bytes(typ: Type, bytes: &[u8], f: &mut fmt::Formatter) -> fmt::Re
                 debug_value_bytes(typ, bytes, f)?;
             }
         }
+        Type::Ip => {
+            let s = as_str(bytes); // TODO: change when serialization changes
+            write_opt(f, s)?;
```
fulmicoton: write!(f, "<ipaddr>");  // TODO change once we actually have IP address terms.

src/schema/value.rs
```diff
@@ -418,7 +445,7 @@ mod binary_serialize {
                         _ => Err(io::Error::new(
                             io::ErrorKind::InvalidData,
                             format!(
-                                "No extended field type is associated with code {:?}",
+                                "No extened field type is associated with code {:?}",
```
fulmicoton: ?

src/indexer/index_writer.rs
```diff
@@ -1744,6 +1771,59 @@ mod tests {
...
+            .flat_map(|segment_reader| {
+                let ff_reader = segment_reader.fast_fields().ip_addr(ip_field).unwrap();
+                segment_reader.doc_ids_alive().flat_map(move |doc| {
+                    let val = ff_reader.get_val(doc as u64);
+                    if val == IpAddr::from_u128(0) {
+                        None
+                    } else {
+                        Some(val)
+                    }
+                })
```
fulmicoton: why?
The sentinel logic was something that is meant to live only in the compact code,
It has NOT been adopted.
IpAddr 0 is a real thing. It is not "None".

```suggestion
                segment_reader.doc_ids_alive().map(move |doc| {
                    ff_reader.get_val(doc as u64)
                })
```

src/indexer/index_writer.rs
```diff
@@ -1744,6 +1771,59 @@ mod tests {
...
+            .collect::<HashSet<_>>();
+        let ips: HashSet<IpAddr> = searcher
+            .segment_readers()
+            .iter()
+            .flat_map(|segment_reader| {
+                let ff_reader = segment_reader.fast_fields().ip_addrs(ips_field).unwrap();
+                segment_reader.doc_ids_alive().flat_map(move |doc| {
+                    let mut vals = vec![];
+                    ff_reader.get_vals(doc, &mut vals);
+                    vals.into_iter().filter(|val| val.to_u128() != 0)
```
fulmicoton: why again

src/query/query_parser/query_parser.rs
fulmicoton: Return an UnsupportedQuery

src/schema/field_type.rs
```diff
@@ -176,6 +193,7 @@ impl FieldType {
             FieldType::Facet(ref _facet_options) => true,
             FieldType::Bytes(ref bytes_options) => bytes_options.is_indexed(),
             FieldType::JsonObject(ref json_object_options) => json_object_options.is_indexed(),
+            FieldType::IpAddr(_) => false,
```
fulmicoton, line 191: Finish the renaming`Ip -> IpAddr` everywhere...

IpOptions --> IpAddrOptions
add_ip_field -> add_ip_addr_field

etc.

If you think this is a bad renaming, then switch back teh `FieldType::IpAddr -> FieldType::Ip` but don't half bake this.

src/schema/value.rs
```diff
@@ -32,6 +33,8 @@ pub enum Value {
     Bytes(Vec<u8>),
     /// Json object value.
     JsonObject(serde_json::Map<String, serde_json::Value>),
+    /// Ip
+    Ip(IpAddr),
```
fulmicoton: We should enforce (and document) normalization whenever we build an IpAddr address.

Here that means 
```suggestion
    Ip(Ipv6Addr),
```

fulmicoton: and mapping Ip v4 to Ip V6 using mapped IpV4 address.

(I don't think there are other normalziation to do, but you can dig further)

src/schema/value.rs
```diff
@@ -366,6 +389,10 @@ mod binary_serialize {
                     serde_json::to_writer(writer, &map)?;
                     Ok(())
                 }
+                Value::Ip(ref ip) => {
+                    IP_CODE.serialize(writer)?;
+                    ip.to_string().serialize(writer) // TODO Check best format
```
fulmicoton: I'd rather use IPv6 all the time and encode stuff over 16 bytes.

src/schema/value.rs
fulmicoton: We lose some info here (was this IPv4 address supplied in ipv4 format or ipv6 mapped) but this is a good thing. We will have to day that for search anyway.

fastfield_codecs/src/column.rs
PSeitz: Yes, I also thought about that, it's probably clearer.

Input and Output are flipped for serialization and opening, so I think we still have two implementations. Maybe we can have an inverse over the trait for that ... but I'm not sure that's very helpful for clarity

src/indexer/merger.rs
```diff
@@ -303,6 +322,82 @@ impl IndexMerger {
...
+
+        let fast_field_readers = segment_and_ff_readers
+            .into_iter()
+            .map(|(_, ff_reader)| ff_reader)
+            .collect::<Vec<_>>();
+
+        let iter_gen = || {
+            doc_id_mapping.iter_old_doc_addrs().flat_map(|doc_addr| {
+                let fast_field_reader = &fast_field_readers[doc_addr.segment_ord as usize];
+                let mut out = vec![];
```
fulmicoton: You can fix it if you rebase now, with the flat_map_with_buffer thingy

src/fastfield/serializer/mod.rs
PSeitz: It's not true for the bytes fast field.

I think `CompositeFastFieldSerializer` has evolved over time to something where we can consider removing it. I can pass everything through it, but there's no logic inside it.

src/indexer/index_writer.rs
PSeitz: This is only temporary. As I said, null is broken after the removed null handling. Currently "0" is pushed when no value exists in the document. I'll add a TODO to it.

src/indexer/merger.rs
PSeitz: nice

src/fastfield/writer.rs
PSeitz: yep

src/schema/schema.rs
PSeitz: https://github.com/quickwit-oss/tantivy/issues/1601

src/schema/term.rs
PSeitz: removed

src/fastfield/serializer/mod.rs
fulmicoton: The bytes fast field is a bit special in that it does not do anything with the bytes.

The idea is that the serializer contains code about the format of the fast field.
That code is common to both the merge and to segment writing.

Please fix

src/schema/field_entry.rs
fulmicoton: new_ip_addr...

src/schema/field_type.rs
```diff
@@ -62,9 +66,13 @@ pub enum Type {
     Bytes = b'b',
     /// Leaf in a Json object.
     Json = b'j',
+    /// IpAddr
+    Ip = b'p',
```
fulmicoton: IpAddr... well or swithc back to ip. I don't care.

src/schema/value.rs
PSeitz: Currently we do the normalization internally when converting to u128. 
I'm not sure about the advantage of putting it into the API?

src/schema/value.rs
PSeitz: I did some tests and u128 is probably better most of the time

src/fastfield/multivalued/reader.rs
PSeitz: There are some other things I'd like to improve with the multivalue index, but I don't want to put it in this PR (e.g. the translation should be part of the index). 

https://github.com/quickwit-oss/tantivy/issues/1602

src/fastfield/multivalued/reader.rs
PSeitz: I added a test

fastfield_codecs/src/column.rs
PSeitz: Okay done, it's a little more verbose than before. I chose `<External, Internal>`, since it's less ambigious. Since we have nameable types now we could evaluate to move it to `Column` and drop the monotonic column wrapping, which does cost some performance.

src/schema/field_type.rs
PSeitz: I think I got all now

fastfield_codecs/src/monotonic_mapping.rs
```diff
@@ -11,6 +17,136 @@ pub trait MonotonicallyMappableToU64: 'static + PartialOrd + Copy + Send + Sync
...
+/// used in fast field codecs.
+///
+/// The monotonic mapping is required so that `PartialOrd` can be used on `Internal` without
+/// converting to `External`.
+///
+/// All strictly monotonic functions are invertible because they are guaranteed to have a one-to-one
+/// mapping from their range to their domain. The `inverse` method is required when opening a codec,
+/// so a value can be converted back to its original domain (e.g. ip address or f64) from its
+/// internal representation.
+pub trait StrictlyMonotonicFn<External, Internal> {
```
fulmicoton, line 30: unit tests

fulmicoton, line 30: nitpick:
I think associated types would have been nicer here.

fastfield_codecs/src/monotonic_mapping.rs
```diff
@@ -11,6 +17,136 @@ pub trait MonotonicallyMappableToU64: 'static + PartialOrd + Copy + Send + Sync
...
+    /// Strictly monotonically maps the value from External to Internal.
+    fn mapping(&self, inp: External) -> Internal;
+    /// Inverse of `mapping`. Maps the value from Internal to External.
+    fn inverse(&self, out: Internal) -> External;
+}
+
+/// Inverts a strictly monotonic mapping from `StrictlyMonotonicFn<A, B>` to
+/// `StrictlyMonotonicFn<B, A>`.
```
fulmicoton, line 38:
```suggestion
/// `StrictlyMonotonicFn<B, A>`.
///
/// # Warning
///
/// This type comes with a footgun. A type being strictly monotonic does not impose that the inverse mapping is strictly monotonic over the entire space External. e.g. a -> a * 2.
/// Use at your own risks.
```

fastfield_codecs/src/monotonic_mapping.rs
```diff
@@ -11,6 +17,136 @@ pub trait MonotonicallyMappableToU64: 'static + PartialOrd + Copy + Send + Sync
...
+    fn mapping(&self, inp: External) -> u64 {
+        External::to_u64(inp)
+    }
+
+    fn inverse(&self, out: u64) -> External {
+        External::from_u64(out)
+    }
+}
+
+/// Strictly monotonic mapping with a gcd and a base value.
```
fulmicoton:
```suggestion
/// Mapping dividing by  gcd and a base value.
/// 
/// The function is assumed to be only called on values divided by passed
/// gcd value. (It is necessary for the function to be monotonic.)
```

fastfield_codecs/src/serialize.rs
```diff
@@ -94,6 +96,17 @@ impl Header {
     }
 }
 
+pub fn normalize_column<C: Column>(
+    from_column: C,
+    min_value: u64,
+    gcd: Option<NonZeroU64>,
+) -> impl Column {
+    let gcd = gcd.map(|gcd| gcd.get()).unwrap_or(1);
+
```
fulmicoton:
```suggestion
```

src/fastfield/multivalued/reader.rs
```diff
@@ -99,12 +99,193 @@ impl<Item: FastValue> MultiValueLength for MultiValuedFastFieldReader<Item> {
...
+/// The `vals_reader` will access the concatenated list of all
+/// values for all reader.
+/// The `idx_reader` associated, for each document, the index of its first value.
+#[derive(Clone)]
+pub struct MultiValuedU128FastFieldReader<T: MonotonicallyMappableToU128> {
+    idx_reader: Arc<dyn Column<u64>>,
+    vals_reader: Arc<dyn Column<T>>,
+}
+
+impl<T: MonotonicallyMappableToU128> MultiValuedU128FastFieldReader<T> {
```
fulmicoton, line 116: Can you open a ticket to investigate if we can refactor the monotonicallymappableto{u64,u128} trait and the StrictlyMonotonicMapping trait into one thing?

src/fastfield/multivalued/reader.rs
```diff
@@ -99,12 +99,193 @@ impl<Item: FastValue> MultiValueLength for MultiValuedFastFieldReader<Item> {
...
+        Self {
+            idx_reader,
+            vals_reader,
+        }
+    }
+
+    /// Returns `[start, end)`, such that the values associated
+    /// to the given document are `start..end`.
+    #[inline]
+    fn range(&self, doc: DocId) -> Range<u64> {
```
fulmicoton, line 130: can you open a ticket to rename this into something else than range?

src/fastfield/multivalued/reader.rs
```diff
@@ -99,12 +99,193 @@ impl<Item: FastValue> MultiValueLength for MultiValuedFastFieldReader<Item> {
...
+
+    /// Returns the array of values associated to the given `doc`.
+    #[inline]
+    pub fn get_vals(&self, doc: DocId, vals: &mut Vec<T>) {
+        let range = self.range(doc);
+        self.get_vals_for_range(range, vals);
+    }
+
+    /// Returns all docids which are in the provided value range
+    pub fn get_between_vals(&self, range: RangeInclusive<T>) -> Vec<DocId> {
```
fulmicoton, line 162: can you open a ticket to rename this too?

It does not suggest that DocIds are returned.

src/fastfield/multivalued/reader.rs
fulmicoton: Yes, can you at least change 2 lines to avoid doing twice the work.

`.start` in the range is not important.

```
let end = multival_idx.get_val(cur_doc + 1);
if end > pos {
```

src/fastfield/multivalued/reader.rs
```diff
@@ -99,12 +99,193 @@ impl<Item: FastValue> MultiValueLength for MultiValuedFastFieldReader<Item> {
...
+    use crate::fastfield::MultiValueLength;
     use crate::schema::{Cardinality, Facet, FacetOptions, NumericOptions, Schema};
 
+    #[test]
+    fn test_positions_to_docid() {
+        let positions = vec![10u64, 11, 15, 20, 21, 22];
+
+        let offsets = vec![0, 10, 12, 15, 22, 23];
+
+        struct MultiValueLenghtIdx {
```
fulmicoton:
```suggestion
        struct MultiValueLengthIdx {
```

src/fastfield/multivalued/reader.rs
```diff
@@ -99,12 +99,193 @@ impl<Item: FastValue> MultiValueLength for MultiValuedFastFieldReader<Item> {
...
+
+/// Converts a list of positions of values in a 1:n index to the corresponding list of DocIds.
+///
+/// Since there is no index for value pos -> docid, but docid -> value pos range, we scan the index.
+///
+/// Correctness: positions needs to be sorted.
+///
+/// TODO: Instead of a linear scan we can employ a expotential search into binary search to match a
+/// docid to its value position.
+fn positions_to_docids<T: MultiValueLength>(positions: &[u64], multival_idx: &T) -> Vec<DocId> {
```
fulmicoton: There is a hidden contract in this function.

It is only correct if the `MultiValueLength` store the different range of vals in order.
I don't see this property documented on the `MultiValueLength` and I assume it is not true for some objects.

Can you at the very least docuemnt?

fulmicoton: Or better, change T for `MultiValuedU128FastFieldReader<T>`. It is only used for this type.

fastfield_codecs/src/monotonic_mapping.rs
```diff
@@ -11,6 +17,136 @@ pub trait MonotonicallyMappableToU64: 'static + PartialOrd + Copy + Send + Sync
...
+pub(crate) struct StrictlyMonotonicMappingToInternalGCDBaseval {
```
fulmicoton, line 109: unit tests here

src/fastfield/readers.rs
```diff
@@ -143,6 +149,56 @@ impl FastFieldReaders {
...
+        Ok(MultiValuedU128FastFieldReader::open(
+            idx_reader,
+            vals_reader,
+        ))
+    }
+
+    /// Returns the `u128` fast field reader reader associated to `field`.
+    ///
+    /// If `field` is not a u128 fast field, this method returns an Error.
+    pub fn u128(&self, field: Field) -> crate::Result<Arc<dyn Column<u128>>> {
```
fulmicoton: The existance of u128 is internal.

```suggestion
    pub(crate) fn u128(&self, field: Field) -> crate::Result<Arc<dyn Column<u128>>> {
```

src/fastfield/readers.rs
```diff
@@ -143,6 +149,56 @@ impl FastFieldReaders {
         self.typed_fast_field_reader(field)
     }
 
+    /// Returns the `ip` fast field reader reader associated to `field`.
+    ///
+    /// If `field` is not a u128 fast field, this method returns an Error.
+    pub fn ip_addr(&self, field: Field) -> crate::Result<Arc<dyn Column<IpAddr>>> {
```
fulmicoton:
```suggestion
    pub fn ip_addr(&self, field: Field) -> crate::Result<Arc<dyn Column<Ipv6Addr>>> {
```

src/schema/document.rs
```diff
@@ -97,6 +98,11 @@ impl Document {
         self.add_field_value(field, value);
     }
 
+    /// Add a IP address field
+    pub fn add_ip_addr(&mut self, field: Field, value: IpAddr) {
```
fulmicoton:
```suggestion
    pub fn add_ip_addr(&mut self, field: Field, value: IpAddrV6) {
```

fulmicoton: That's a tricky one. We probably want to drop the notion of IpAddr to only work with IpAddrV6, and document everywhere if we get an ipv4 when say, we parse json, it gets mapped ot IpAddrV6.

In other word, "code" gets and returns  `std::net::IpAddrV6`.
"data" when interpreted, can get IPv4 representations, but converts them to IPv6. Does that make sense?

src/schema/value.rs
```diff
@@ -32,6 +33,8 @@ pub enum Value {
     Bytes(Vec<u8>),
     /// Json object value.
     JsonObject(serde_json::Map<String, serde_json::Value>),
+    /// Ip Address value
+    IpAddr(IpAddr),
```
fulmicoton:
```suggestion
    IpAddr(Ipv6Addr),
```

fulmicoton: This was in a previous comment by the way

src/schema/field_type.rs
```diff
@@ -304,37 +320,43 @@ impl FieldType {
...
+                }),
+                FieldType::Facet(_) => Ok(Value::Facet(Facet::from(&field_text))),
+                FieldType::Bytes(_) => base64::decode(&field_text)
+                    .map(Value::Bytes)
+                    .map_err(|_| ValueParsingError::InvalidBase64 { base64: field_text }),
+                FieldType::JsonObject(_) => Err(ValueParsingError::TypeError {
+                    expected: "a json object",
+                    json: JsonValue::String(field_text),
+                }),
+                FieldType::IpAddr(_) => Ok(Value::IpAddr(IpAddr::from_str(&field_text).map_err(
```
fulmicoton: This is the tricky bit.
Here the logic we want is something like:

```
let ip_addr: IpAddr = IpAddr::from_str(&field_text).map_err(
                    |err| ValueParsingError::ParseError {
                        error: err.to_string(),
                        json: JsonValue::String(field_text),
                    },
                )?;
let ip_addr_v6: Ipv6Addr = convert_into_ip_addr_v6(ip_addr);
...
```

so we do accept ip addr v4 in documents.

fulmicoton (changes requested):
I still have one major comment.

Internally we convert stuff to Ipv6. We want that to be clear for the user.
We accept ipv4 as an input but we convert it internally to IpV6.

fastfield_codecs/src/monotonic_mapping.rs
PSeitz: currently its generic over MonotonicallyMappableToU64. Maybe we can convert it, when they can be merged

fulmicoton (approved):

PR #1554, opened 2022-09-26
prepare for ip field

PSeitz:


src/fastfield/bytes/reader.rs
```diff
@@ -31,36 +31,39 @@ impl BytesFastFieldReader {
         Ok(BytesFastFieldReader { idx_reader, values })
     }
 
-    fn range(&self, doc: DocId) -> (usize, usize) {
+    fn range(&self, doc: DocId) -> std::ops::Range<u64> {
```
fulmicoton:
```suggestion
    fn range(&self, doc: DocId) -> Range<usize> {
```

src/fastfield/bytes/reader.rs
```diff
@@ -31,36 +31,39 @@ impl BytesFastFieldReader {
...
+        let stop = self.idx_reader.get_val(idx + 1);
+        start..stop
     }
 
     /// Returns the bytes associated to the given `doc`
     pub fn get_bytes(&self, doc: DocId) -> &[u8] {
-        let (start, stop) = self.range(doc);
-        &self.values.as_slice()[start..stop]
+        let range = self.range(doc);
+        &self.values.as_slice()[range.start as usize..range.end as usize]
```
fulmicoton, line 45:
```suggestion
        &self.values.as_slice()[range]
```

src/fastfield/bytes/reader.rs
```diff
@@ -31,36 +31,39 @@ impl BytesFastFieldReader {
         Ok(BytesFastFieldReader { idx_reader, values })
     }
 
-    fn range(&self, doc: DocId) -> (usize, usize) {
+    fn range(&self, doc: DocId) -> std::ops::Range<u64> {
         let idx = doc as u64;
-        let start = self.idx_reader.get_val(idx) as usize;
-        let stop = self.idx_reader.get_val(idx + 1) as usize;
-        (start, stop)
+        let start = self.idx_reader.get_val(idx);
```
fulmicoton, line 37:
```suggestion
        let start = self.idx_reader.get_val(idx) as usize;
```

src/fastfield/bytes/reader.rs
```diff
@@ -31,36 +31,39 @@ impl BytesFastFieldReader {
...
+        let stop = self.idx_reader.get_val(idx + 1);
```
fulmicoton:
```suggestion
        let end = self.idx_reader.get_val(idx + 1) as usize;
```

src/fastfield/bytes/reader.rs
```diff
@@ -31,36 +31,39 @@ impl BytesFastFieldReader {
...
+        start..stop
```
fulmicoton:
```suggestion
        start..end
```

src/fastfield/bytes/reader.rs
```diff
@@ -31,36 +31,39 @@ impl BytesFastFieldReader {
...
-        &self.values.as_slice()[start..stop]
+        let range = self.range(doc);
+        &self.values.as_slice()[range.start as usize..range.end as usize]
     }
 
     /// Returns the length of the bytes associated to the given `doc`
-    pub fn num_bytes(&self, doc: DocId) -> usize {
-        let (start, stop) = self.range(doc);
-        stop - start
+    pub fn num_bytes(&self, doc: DocId) -> u64 {
```
fulmicoton, line 48:
```suggestion
    pub fn num_bytes(&self, doc: DocId) -> usize {
```

src/fastfield/bytes/reader.rs
```diff
@@ -31,36 +31,39 @@ impl BytesFastFieldReader {
...
+        let range = self.range(doc);
+        range.end - range.start
```
fulmicoton, line 51:
```suggestion
        range.len() as u64
```

fulmicoton: You might need to use std::iter::ExactSizeIterator for that one.

fulmicoton (approved):

src/fastfield/bytes/reader.rs
PSeitz: Why usize? Multivalue ff values can be u64

fulmicoton: The index you mean? They are consumed as usize and it is a private method here.

(Actually I think we should make all column indexes `u32` but that's another discussion.)

fulmicoton: for this PR you don't have to take the suggestions if you judge better as it is.

PSeitz: The method is also used in the trait, which is u64. I also think we can change that, 4GB should be enough for a segment. For the time being I would be explicit with the type.

PR #1555, opened 2022-09-26
Move ArcBytes,WeakArcBytes to mmap_directory.

waywardmonkeys:
When building without default features (so without mmap, etc), there are some warnings about unused things. This fixes the ones related to `ArcBytes` and `WeakArcBytes`, which are only used with the `mmap_directory` code.

### DIFF
--- a/src/directory/file_slice.rs
+++ b/src/directory/file_slice.rs
@@ -1,5 +1,5 @@
 use std::ops::{Deref, Range};
-use std::sync::{Arc, Weak};
+use std::sync::Arc;
 use std::{fmt, io};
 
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
I left these as `pub` within the crate, but I did remove the re-`use` of them from the `crate::directory` module. They should either be re-`use` still or they shouldn't be `pub`.

They aren't used in any public API, so keeping them internal to the mmap code seems like the right option?

fulmicoton:
Yes that's much better. Thank you @waywardmonkeys

fulmicoton (approved):

PR #1556, opened 2022-09-26
POLLING_INTERVAL needn't be pub.

waywardmonkeys:
This is only used within the file watcher and is const, so it can't be configured.


waywardmonkeys:
This was another thing I noticed when building with `--no-default-features` since I was building for a WebAssembly target.

fulmicoton (approved):

PR #1559, opened 2022-09-26
Avoid slicing a string in DocParsingError.

### LINKED ISSUE #1339: Avoid slicing a string in DocParsingError
Done here: https://github.com/quickwit-oss/tantivy/blob/main/src/schema/schema.rs#L402-L412

Better with `invalid_json.chars().take(20)` :)

waywardmonkeys:
Fixes #1339.

### DIFF
--- a/src/schema/schema.rs
+++ b/src/schema/schema.rs
@@ -422,12 +422,8 @@ pub enum DocParsingError {
 impl DocParsingError {
     /// Builds a NotJson DocParsingError
     fn invalid_json(invalid_json: &str) -> Self {
-        let sample_json: String = if invalid_json.len() < 20 {
-            invalid_json.to_string()
-        } else {
-            format!("{:?}...", &invalid_json[0..20])
-        };
-        DocParsingError::InvalidJson(sample_json)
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
I'm not convinced that this is better since there are still allocations, etc ... but the code is shorter and it is only an error path thing.

src/schema/schema.rs
```diff
@@ -422,12 +422,8 @@ pub enum DocParsingError {
 impl DocParsingError {
     /// Builds a NotJson DocParsingError
     fn invalid_json(invalid_json: &str) -> Self {
-        let sample_json: String = if invalid_json.len() < 20 {
-            invalid_json.to_string()
-        } else {
-            format!("{:?}...", &invalid_json[0..20])
-        };
-        DocParsingError::InvalidJson(sample_json)
+        let sample = invalid_json.chars().take(20).collect();
```
fulmicoton, line 425: the best would be `std::floor_char_boundary` but it is not available in stable.

fulmicoton (approved):

fmassot:
nice :)
thanks @waywardmonkeys

PR #1561, opened 2022-09-26
Read only index.

### LINKED ISSUE #557: Read only mode
My problem is that I am attempting to deploy `tantivy` to an environement which has a read-only file system.

Creating the index as follows
```rust
let index = Index::open_in_dir("my-index-path").unwrap();
```
results in this error
```
"Read-only file system"
```
It seems that if `.tantivy-meta.lock` is read only, creating the `index` gives this error. All other files in the `index` directory can be read-only, however.

One solution might be an option to open the index in read only mode - you cannot commit or merge any new documents at runtime.

Perhaps this could avoid the need for acquiring locks and so on?

I am happy to build the index locally/on CI first and then deploy the `index` as a static file. 

(The data I want to search through will change very infrequently).

waywardmonkeys:
An index can be opened in read-only mode, which will prevent an IndexWriter from being created, as well as mark any IndexReaders as read-only as well (through a new reload policy) so that the lock file need not be created.

This is useful for when you have index data stored on read-only storage.

Fixes #557.

### DIFF
--- a/src/core/index.rs
+++ b/src/core/index.rs
@@ -228,6 +228,21 @@ impl IndexBuilder {
             ))
         }
     }
+
+    /// Opens a new index in the provided directory in read-only mode.
+    pub fn open_read_only<T: Into<Box<dyn Directory>>>(self, dir: T) -> crate::Result<Index> {
+        let mut index = Index::open(dir.into())?;
+        index.set_read_only(true);
+        index.set_tokenizers(self.tokenizer_manager.clone());
+        if index.schema() == self.get_expect_schema()? {
+            Ok(index)
+        } else {
+            Err(TantivyError::SchemaError(
+                "An index exists but the schema does not match.".to_string(),
+            ))
+        }
+    }
+
     /// Creates a new index given an implementation of the trait `Directory`.
     ///
     /// If a directory previously existed, it will be erased.
@@ -256,6 +271,7 @@ pub struct Index {
     executor: Arc<Executor>,
     tokenizers: TokenizerManager,
     inventory: SegmentMetaInventory,
+    read_only: bool,
 }
 
 impl Index {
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
This doesn't yet have tests.

I also have gone back and forth over whether or not a new `ReloadPolicy::ReadOnly` is needed or just use `Index::is_read_only()`.

adamreichold:
This sounds nice to have for auxiliary indexes based on vocabulary data generated at build time.

fulmicoton:
The PR makes sense. However I'm not fond of adding features and complexity out of thin air. Do we have actual users who need this? @shikhar maybe?

waywardmonkeys:
@fulmicoton This originally came from something that was needed in a pilot project I worked on earlier this year using Tantivy and DataFusion where some datasets were on read-only mounts.

There was an issue filed about it, #557, from a couple of years ago and it included a suggestion for the implementation as well as an example of using a `ReadOnlyDirectory` bit of additional code.

I think that my patch might be simpler yet if it were only having the index be read-only rather than the additional flexibility of the reload policy, although I like the suggestion by @petr-tik in another commit to call it `ReloadPolicy::Never` rather than `ReloadPolicy::ReadOnly`.

adamreichold:
> Do we have actual users who need this?

As written above, I would use this if included and released for indexes over static vocabulary that we create at build time and ship during deployment.

What I am not sure is whether I should, i.e. the original intention seems to be able to work with read-only filesystems which would not apply to our environment. Would there be any additional benefits like access speed or resource consumption?

shikhar:
I can definitely see this being useful with the included `Directory` implementations over a shared FS as per @waywardmonkeys stated motivation.

It doesn't benefit our use-case directly because we have a custom `GcsDirectory` implementation that makes lock acquisition a no-op (unique writer is enforced via a `StatefulSet` for the indexing pods in Kubernetes), and has a read-only mode for serving replicas that denies writes.

PR #1564, opened 2022-09-28
log improvements

PSeitz:


src/indexer/segment_writer.rs
```diff
@@ -374,9 +374,9 @@ fn remap_and_write(
         doc_id_map,
     )?;
 
-    debug!("resort-docstore");
     // finalize temp docstore and create version, which reflects the doc_id_map
     if let Some(doc_id_map) = doc_id_map {
+        debug!("resort-docstore");
```
fulmicoton, line 379: yes that was annoyign me too.

fulmicoton (approved):

PR #1566, opened 2022-09-28
fix docstore settings for temp docstore

### LINKED ISSUE #1565: Use optimized docstore settings for sorting
The docstore should have optimized settings when sorting is enabled.
The blocks should be tiny to allow random access on the documents. The compression should be LZ4 or None.

PSeitz:
fixes #1565

### DIFF
--- a/src/indexer/segment_serializer.rs
+++ b/src/indexer/segment_serializer.rs
@@ -24,12 +24,25 @@ impl SegmentSerializer {
         // In the merge case this is not necessary because we can kmerge the already sorted
         // segments
         let remapping_required = segment.index().settings().sort_by_field.is_some() && !is_in_merge;
-        let store_component = if remapping_required {
-            SegmentComponent::TempStore
+        let settings = segment.index().settings().clone();
+        let store_writer = if remapping_required {
+            let store_write = segment.open_write(SegmentComponent::TempStore)?;
+            StoreWriter::new(
+                store_write,
+                crate::store::Compressor::None,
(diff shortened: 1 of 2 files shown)


src/indexer/segment_serializer.rs
```diff
@@ -24,12 +24,25 @@ impl SegmentSerializer {
...
-        let store_component = if remapping_required {
-            SegmentComponent::TempStore
+        let settings = segment.index().settings().clone();
+        let store_writer = if remapping_required {
+            let store_write = segment.open_write(SegmentComponent::TempStore)?;
+            StoreWriter::new(
+                store_write,
+                crate::store::Compressor::None,
+                0, // we want random access on the docs, so we choose a minimal block size. Every
+                // doc will be flushed
```
fulmicoton:
```suggestion
                // doc will get its own block.
```

src/indexer/segment_writer.rs
```diff
@@ -393,7 +393,8 @@ fn remap_and_write(
             serializer
                 .segment()
                 .open_read(SegmentComponent::TempStore)?,
-            50,
+            1, /* The docstore is configured for tiny docs for fast random access, we don't need
+                * the the cache */
```
fulmicoton, line 397:
```suggestion
                * the cache */
```

fulmicoton:
```suggestion
            1, /* The docstore is configured to have one doc per block, and each doc is accessed only 
            once: we don't need caching. */
```

fulmicoton (approved):

PR #1574, opened 2022-09-30
Removing default features for zstd

fulmicoton:


ppodolsky:
Wow, thank you! It was blocker for compiling tantivy on latest macOS to WASM due to a chain of weird compilation bugs

fulmicoton:
I picked that from your PR actually.

PR #1580, opened 2022-10-03
Use u64::from boolean more.

waywardmonkeys:
This case is inverted from the previous cases fixed.

This is from nightly clippy.


fulmicoton (approved):

fulmicoton:
I'm not fond of this one to be honest. Is it really more readable?

waywardmonkeys:
The previous commit that did the initial changes was: e9a384bb157347bd77c11a9ef06ad10d07a79bee

In that one, it is `==` leading to `1` or `0`. In this one, it was `==` leading to `0` or `1`, which is why I had to invert the comparison to `!=`.

I think once you know these exist, this makes things a bit more concise and readable, but I could probably argue the other way as well. I landed on the side of "If clippy does this, then it will become the norm, so might as well."

PR #1582, opened 2022-10-03
use groupby instead of vec allocation

PSeitz:
`get_sorted_field_values` is quite expensive and usually visible in traces. The allocations are unnecessary though.

```
Before (13 mulitivalue fast fields)
-    9,72%     1,21%  thrd-tantivy-in  tantivy               [.] tantivy::indexer::segment_writer::SegmentWriter::add_document                                                         
   - 8,51% tantivy::indexer::segment_writer::SegmentWriter::add_document                                                                                                               
      + 3,03% tantivy::store::store_compressor::BlockCompressor::compress_block_and_write                                                                                              
      + 2,42% tantivy::schema::document::Document::get_sorted_field_values                                                                                                             
      + 1,89% cfree                                                                                                                                                                    
        0,95% tantivy::fastfield::multivalued::writer::MultiValuedFastFieldWriter::add_document                                                                                        


After
-    8,10%     1,65%  thrd-tantivy-in  tantivy               [.] tantivy::indexer::segment_writer::SegmentWriter::add_document                                                        
   - 6,45% tantivy::indexer::segment_writer::SegmentWriter::add_document                                                                                                               
      + 3,52% tantivy::store::store_compressor::BlockCompressor::compress_block_and_write                                                                                              
        1,28% tantivy::fastfield::multivalued::writer::MultiValuedFastFieldWriter::add_document                                                                                        
        1,21% cfree                                                                                                                                                                    
   + 1,65% __clone                                                                                                                                                                     

```

### DIFF
--- a/src/indexer/segment_writer.rs
+++ b/src/indexer/segment_writer.rs
@@ -1,4 +1,5 @@
 use fastfield_codecs::MonotonicallyMappableToU64;
+use itertools::Itertools;
 
 use super::doc_id_mapping::{get_doc_id_mapping_from_field, DocIdMapping};
 use super::operation::AddOperation;
@@ -157,7 +158,13 @@ impl SegmentWriter {
 
     fn index_document(&mut self, doc: &Document) -> crate::Result<()> {
         let doc_id = self.max_doc;
-        for (field, values) in doc.get_sorted_field_values() {
+        let vals_grouped_by_field = doc
+            .field_values()
+            .iter()
+            .sorted_by_key(|el| el.field())
+            .group_by(|el| el.field());
+        for (field, field_values) in &vals_grouped_by_field {
+            let values = field_values.map(|field_value| field_value.value());
             let field_entry = self.schema.get_field_entry(field);
             let make_schema_error = || {
                 crate::TantivyError::SchemaError(format!(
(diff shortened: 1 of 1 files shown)


src/indexer/segment_writer.rs
```diff
@@ -157,7 +158,11 @@ impl SegmentWriter {
 
     fn index_document(&mut self, doc: &Document) -> crate::Result<()> {
         let doc_id = self.max_doc;
-        for (field, values) in doc.get_sorted_field_values() {
+        let mut field_vals: Vec<_> = doc.field_values().iter().collect();
+        field_vals.sort_by_key(|el| el.field());
+        let iter = field_vals.into_iter().group_by(|el| el.field());
```
shikhar: you can use https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.sorted_by_key

since the `Vec` is just intermediary anyway

src/indexer/segment_writer.rs
```diff
@@ -157,7 +158,11 @@ impl SegmentWriter {
...
+        for (field, field_values) in &iter {
```
fulmicoton: can it be 

```
for (field, field_values) in iter {
}
```

src/indexer/segment_writer.rs
fulmicoton: `iter` is a sloppy name.

Can you call it `vals_grouped_by_field_iter` for instance?

fulmicoton (approved):

src/indexer/segment_writer.rs
PSeitz: no, that's some weirdness of the group_by api

src/indexer/segment_writer.rs
PSeitz: yep, but I called it vals_grouped_by_field, since it's not an iter but IntoIter

PR #1587, opened 2022-10-04
remove get_val in serialization

PSeitz:
remove get_val in serialization and mark as unimplemented!()
replace get_val with iter in linear codec
remove MultivalueStartIndexRandomSeeker
replace MultivalueStartIndexIter with closure
Sample 100 values in linear codec
remove Mutexes

### DIFF
--- a/fastfield_codecs/src/linear.rs
+++ b/fastfield_codecs/src/linear.rs
@@ -126,18 +126,21 @@ impl FastFieldCodec for LinearCodec {
             return None; // disable compressor for this case
         }
 
-        // let's sample at 0%, 5%, 10% .. 95%, 100%
-        let num_vals = column.num_vals() as f32 / 100.0;
-        let sample_positions = (0..20)
-            .map(|pos| (num_vals * pos as f32 * 5.0) as u64)
-            .collect::<Vec<_>>();
+        let limit_num_vals = column.num_vals().min(100_000);
 
-        let line = Line::estimate(column, &sample_positions);
+        let num_samples = 100;
+        let step_size = (limit_num_vals / num_samples).max(1); // 20 samples
+        let mut sample_positions_and_values: Vec<_> = Vec::new();
+        for (idx, val) in column.iter().step_by(step_size as usize).enumerate() {
+            let pos = idx as u64 * step_size;
+            sample_positions_and_values.push((pos, val));
+        }
 
-        let estimated_bit_width = sample_positions
+        let line = Line::estimate(&sample_positions_and_values);
+
+        let estimated_bit_width = sample_positions_and_values
             .into_iter()
-            .map(|pos| {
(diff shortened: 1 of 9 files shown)


fastfield_codecs/src/linear.rs
```diff
@@ -126,18 +126,21 @@ impl FastFieldCodec for LinearCodec {
...
         }
 
-        // let's sample at 0%, 5%, 10% .. 95%, 100%
-        let num_vals = column.num_vals() as f32 / 100.0;
-        let sample_positions = (0..20)
-            .map(|pos| (num_vals * pos as f32 * 5.0) as u64)
-            .collect::<Vec<_>>();
+        let limit_num_vals = column.num_vals().min(100_000);
+
+        let num_samples = 100;
```
fulmicoton, line 131: I like the step_by solution. 

Can you change the column.num_vals() limit above to 100 too, to make proofreading even easier.

fastfield_codecs/src/linear.rs
```diff
@@ -126,18 +126,21 @@ impl FastFieldCodec for LinearCodec {
...
+        let step_size = (limit_num_vals / num_samples).max(1); // 20 samples
+        let mut sample_positions_and_values: Vec<_> = Vec::new();
+        for (idx, val) in column.iter().step_by(step_size as usize).enumerate() {
```
fulmicoton: nitpick
I think you could have swapoed `step_by` and `enumerate`, and avoided l.135.

src/fastfield/multivalued/writer.rs
```diff
@@ -204,112 +203,68 @@ impl MultiValuedFastFieldWriter {
...
+) -> impl Iterator<Item = u64> + 'a {
+    let mut offset = 0;
+    doc_id_map
+        .iter_old_doc_ids()
+        .chain(std::iter::once(u32::MAX))
+        .map(move |old_doc| {
+            if old_doc == u32::MAX {
+                // sentinel value for last offset
+                return offset;
+            }
```
fulmicoton, line 261: wouldn't doing the chain below do the same thing?

```
 doc_id_map
        .iter_old_doc_ids()
        .map(move |old_doc| { ... })
        .chain(iter::once(offset))
```

fulmicoton (approved):
see comments inline

fastfield_codecs/src/linear.rs
PSeitz: I don't understand, 100 instead of 100_000?

src/fastfield/multivalued/writer.rs
PSeitz: It would, but offset is already owned by the closure

fastfield_codecs/src/linear.rs
PSeitz: yes, that's better

PR #1594, opened 2022-10-05
Removing alloc on all .next() in MultiValueColumn

fulmicoton:


PSeitz (approved):
Awesome solution. I'll push a change to the branch to make it as ergonomic as `flat_map`

src/indexer/flat_map_with_buffer.rs
```diff
@@ -0,0 +1,74 @@
+struct FlatMapWithgBuffer<T, F, Iter> {
```
ppodolsky: Did you really mean **g** in FlatMapWith**g**Buffer?

PSeitz: Updated the PR

PR #1599, opened 2022-10-05
WIP: add conversion for f32 and fixed precision number

### LINKED ISSUE #1530: Add f32 support
Right now we only offer f64 support. 
f64 take a large amount of space and are often overkill.
(Interestingly the gcd trick are possibly helping today.)

f32 without any compression would be a nice improvement

trinity-1686a:
will fix #1530

Having some form of fixed precision has been discussed with @fulmicoton, but it turned out to be way harder than expected due to rounding errors and f64 not being infinite-precision numbers.

After that, I wonder if it wouldn't be easier, and possibly more helpful, to allow custom floating points, and encode them in the same scheme as we do for f64, with leading zeroes for unused bits. Such a scheme would allow guaranteed no Generation loss (which is generally true for the encoding I made, but sometime isn't especially when using 50+ bits of precision).

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -34,7 +34,8 @@ impl<T: Deref<Target = [u8]>> HasLen for T {
     }
 }
 
-const HIGHEST_BIT: u64 = 1 << 63;
+const HIGHEST_BIT_64: u64 = 1 << 63;
+const HIGHEST_BIT_32: u32 = 1 << 31;
 
 /// Maps a `i64` to `u64`
 ///
@@ -58,13 +59,13 @@ const HIGHEST_BIT: u64 = 1 << 63;
 /// The reverse mapping is [`u64_to_i64()`].
 #[inline]
 pub fn i64_to_u64(val: i64) -> u64 {
-    (val as u64) ^ HIGHEST_BIT
+    (val as u64) ^ HIGHEST_BIT_64
 }
 
 /// Reverse the mapping given by [`i64_to_u64()`].
 #[inline]
 pub fn u64_to_i64(val: u64) -> i64 {
-    (val ^ HIGHEST_BIT) as i64
+    (val ^ HIGHEST_BIT_64) as i64
 }
 
 /// Maps a `f64` to `u64`
@@ -88,7 +89,7 @@ pub fn u64_to_i64(val: u64) -> i64 {
 pub fn f64_to_u64(val: f64) -> u64 {
     let bits = val.to_bits();
     if val.is_sign_positive() {
-        bits ^ HIGHEST_BIT
+        bits ^ HIGHEST_BIT_64
     } else {
         !bits
     }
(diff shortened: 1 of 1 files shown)


common/src/lib.rs
```diff
@@ -97,26 +98,148 @@ pub fn f64_to_u64(val: f64) -> u64 {
...
+    })
+}
+
+/// Maps a `f64` to a fixed point representation.
+/// Lower bound is inclusive, upper bound is exclusive.
+/// `precision` is the number of bits used to represent the number.
+///
+/// This is a lossy, affine transformation. All provided values must be finite and non-NaN.
+/// Care should be taken to not provide values which would cause loss of precision such as values
+/// low enough to get sub-normal numbers, value high enough rounding would cause ±Inf to appear, or
```
fulmicoton, line 141: really? Do you have an example causing +Inf?

common/src/lib.rs
```diff
@@ -97,26 +98,148 @@ pub fn f64_to_u64(val: f64) -> u64 {
...
+/// a precision larger than 50b.
+///
+/// # See also
+/// The reverse mapping is [`fixed_point_to_f64()`].
+#[inline]
+pub fn f64_to_fixed_point(val: f64, min: f64, max: f64, precision: u8) -> u64 {
+    debug_assert!((1..53).contains(&precision));
+    debug_assert!(min < max);
+
```
fulmicoton, line 150: precision must be greater than 0 I assume?

fulmicoton:
@trinity-1686a this seems difficult indeed.

I am not exactly sure what caused the extra gymnastics.
I suspect it might have been simpler by aligning ourselves with f32 scale.

Could we avoid all the complexity, by sticking to power of 2 arithmetics?

In pseudo-code it would look like this...
I did not consider signs.

```rust
struct Float {
    significant: u64,
    exponent: i8,
}

fn f64_to_u64(val: f64, precision: u8, exponent: i8) -> u64 {
    assert!(val >= 0); //< that's not great
    assert!(precision < 52); //< above we cannot do the f64 copmutation accurately.
    assert!(val < max_expressable_val(precision, exponent));
    let float: Float = deconstruct(val);
    let bitshift = exponent - float.exponent + 53 - precision;
    float.significant >> bitshift  //< can bitshift be negative?
}
```

The idea is, let's imagine we had a new standard called f19 with 
7 bits of exponent and 12 bits of significant, how would we write the conversion code from f64 to this weird f19?

With this approach (provided we have guarantees on the val being within a given range), the conversion boils down to a bitshift and a mask. Inverse conversion is a bitshift and or.

common/src/lib.rs
trinity-1686a: yes, it makes no sens to encode something using 0 bits

common/src/lib.rs
trinity-1686a: rounding to infinity wasn't the right wording, overflowing would be.
min = -1e+308;
max = +1e+308;
delta = max - min;
delta == +Inf

PR #1607, opened 2022-10-07
Store bytes as b64

nigel-andrews:
### Description
Instead of serialising an array of bytes, encode that array to a base64 string and serialise it as a string instead.

### Testing
Added tests to check the output of the serialisation.

### DIFF
--- a/src/schema/value.rs
+++ b/src/schema/value.rs
@@ -462,7 +462,7 @@ mod tests {
         let mut doc = Document::default();
         doc.add_bytes(bytes_field, "this is a test".as_bytes());
         let json_string = schema.to_json(&doc);
-        assert_eq!(json_string, "{\"my_bytes\":[\"dGhpcyBpcyBhIHRlc3Q=\"]}");
+        assert_eq!(json_string, r#"{"my_bytes":["dGhpcyBpcyBhIHRlc3Q="]}"#);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


src/schema/value.rs
```diff
@@ -448,9 +448,52 @@ mod binary_serialize {
...
+    #[test]
+    fn test_parse_bytes_doc() {
+        let mut schema_builder = Schema::builder();
+        let bytes_options = BytesOptions::default();
+        let bytes_field = schema_builder.add_bytes_field("my_bytes", bytes_options);
+        let schema = schema_builder.build();
+        let mut doc = Document::default();
+        doc.add_bytes(bytes_field, "this is a test".as_bytes());
+        let json_string = schema.to_json(&doc);
+        assert_eq!(json_string, "{\"my_bytes\":[\"dGhpcyBpcyBhIHRlc3Q=\"]}");
```
fulmicoton: nitpick for next time.

you can use raw string literals to avoid struggling with escaping . 
```suggestion
        assert_eq!(json_string, r#"{"my_bytes":["dGhpcyBpcyBhIHRlc3Q="]}"#);
```

fulmicoton (approved):

src/schema/value.rs
nigel-andrews: Noted, I have changed the strings in the tests, thank you.

PR #1608, opened 2022-10-07
Serialise bytes as base64 strings instead of arrays.

nigel-andrews:
### Description
Instead of serialising an array of bytes, encode that array to a base64 string and serialise it as a string instead.

### Testing
Added tests to check the output of the serialisation.


fmassot (approved):
LGTM. (PR is the same as https://github.com/quickwit-oss/tantivy/pull/1607 which was already approved).

Issue #1433, opened 2022-08-03
`NOT` and `-` operators not always consistent


fmassot:
Here is a test that is failing on the second assertion

```rust
#[test]
fn test_parse_test_not() {
    test_parse_query_to_ast_helper("tata -toto", "(*\"tata\" -\"toto\")");
    test_parse_query_to_ast_helper("tata NOT toto", "(*\"tata\" -\"toto\")");
}
```

with output

```bash
thread 'query_grammar::test::test_parse_test_query_other' panicked at 'assertion failed: `(left == right)`
  left: `"(*\"tata\" *(-\"toto\"))"`,
 right: `"(*\"tata\" -\"toto\")"`', query-grammar/src/query_grammar.rs:445:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I would have expected the same behavior from the operator "-" and "NOT".

bazhenov:
This happens basically because two provided queries parse results are different:

1. _tata -toto_ parsed down to 2 simple clauses
    ```
    Clause([
      (None,          Leaf("tata")),
      (Some(MustNot), Leaf("toto"))
    ])
    ```
2. while _tata NOT toto_ producing nested second clause:
    ```
    Clause([
      (None,          Leaf("tata")),
      (None,          Clause([
        (Some(MustNot), Leaf("toto"))
      ]))
    ])
    ```

I think this issue can be fixed by optimizing `UserInputAst` after it has been parsed from string. The optimization rule is quite simple. Each element in `UserInputAst::Clause` vector can be replaced by it's inner single clause if `Occur` is missing.

`(None, UserInputAst::Clause(v))` → `(v[0].0, v[0].1)` if and only if `v.len() == 1`.

I can provide PR.

bazhenov:
#1609

trinity-1686a:
this has been fixed, and there is a [test to detect regression](https://github.com/quickwit-oss/tantivy/blob/1a9fc10be981b681f891359757926684ada2405c/query-grammar/src/query_grammar.rs#L1593-L1596), closing

PR #1609, opened 2022-10-09
Enforcing "NOT" and "-" queries consistency in UserInputAst

bazhenov:
One additional phase added to `parse_to_ast()` (`rewrite_ast()`) which removes unnecessary nesting of clauses in `UserInputAst::Clause`

Motivated by #1433

### DIFF
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -374,6 +374,25 @@ pub fn parse_to_ast<'a>() -> impl Parser<&'a str, Output = UserInputAst> {
     spaces()
         .with(optional(ast()).skip(eof()))
         .map(|opt_ast| opt_ast.unwrap_or_else(UserInputAst::empty_query))
+        .map(|ast| rewrite_ast(ast))
+}
+
+/// Removes unnecessary children clauses in AST
+///
+/// Motivated by [issue #1433](https://github.com/quickwit-oss/tantivy/issues/1433)
+fn rewrite_ast(input: UserInputAst) -> UserInputAst {
+    if let UserInputAst::Clause(terms) = input {
+        UserInputAst::Clause(terms.into_iter().map(rewrite_ast_clause).collect())
+    } else {
+        input
+    }
+}
+
+fn rewrite_ast_clause(input: (Option<Occur>, UserInputAst)) -> (Option<Occur>, UserInputAst) {
+    match input {
+        (None, UserInputAst::Clause(mut clauses)) if clauses.len() == 1 => clauses.remove(0),
+        other => other,
+    }
 }
 
 #[cfg(test)]
@@ -748,5 +767,11 @@ mod test {
         test_parse_query_to_ast_helper("\"a b\"~3", "\"a b\"~3");
         test_parse_query_to_ast_helper("foo:\"a b\"~300", "\"foo\":\"a b\"~300");
         test_parse_query_to_ast_helper("\"a b\"~300^2", "(\"a b\"~300)^2");
+    }
+
+    #[test]
+    fn test_not_queries_are_consistent() {
+        test_parse_query_to_ast_helper("tata -toto", "(*\"tata\" -\"toto\")");
+        test_parse_query_to_ast_helper("tata NOT toto", "(*\"tata\" -\"toto\")");
     }
 }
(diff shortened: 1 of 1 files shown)


query-grammar/src/query_grammar.rs
```diff
@@ -374,6 +374,25 @@ pub fn parse_to_ast<'a>() -> impl Parser<&'a str, Output = UserInputAst> {
...
         .map(|opt_ast| opt_ast.unwrap_or_else(UserInputAst::empty_query))
+        .map(|ast| rewrite_ast(ast))
+}
+
+/// Removes unnecessary children clauses in AST
+///
+/// Motivated by [issue #1433](https://github.com/quickwit-oss/tantivy/issues/1433)
+fn rewrite_ast(input: UserInputAst) -> UserInputAst {
+    if let UserInputAst::Clause(terms) = input {
+        UserInputAst::Clause(terms.into_iter().map(rewrite_ast_clause).collect())
```
adamreichold: I understand that `Vec::into_iter().collect::<Vec<_>>()` can be optimized to act in-place instead of re-allocating. Can we rely on this here and do we want to? What downsides do you see to something like

```rust
if let UserInputAst::Clause(terms) = &mut input {
  for term in terms {
    rewrite_ast_clause(&mut term);
  }
}
```

query-grammar/src/query_grammar.rs
```diff
@@ -374,6 +374,25 @@ pub fn parse_to_ast<'a>() -> impl Parser<&'a str, Output = UserInputAst> {
...
+    } else {
+        input
+    }
+}
+
+fn rewrite_ast_clause(input: (Option<Occur>, UserInputAst)) -> (Option<Occur>, UserInputAst) {
+    match input {
+        (None, UserInputAst::Clause(mut clauses)) if clauses.len() == 1 => clauses.remove(0),
```
adamreichold: I fear that the compiler might not be able to remove that part of `Vec::remove`'s code which does not apply for a vector of length one. Maybe this could use `clauses.pop().unwrap()` to guide it towards a simpler method to begin with?

bazhenov: Yes, indeed. Forgot about `pop()`. Thank you.

query-grammar/src/query_grammar.rs
bazhenov: > Can we rely on this here and do we want to?

It's hard yo say. If this code is on a hot path, it's better to be safe I guess.

> Which downsides do you see for something like

Apart from it's maybe kinda not so idiomatic in Rust I can't think of any.

Rewrote `rewrite_ast_clause()` as `fn(&mut (Option<Occur>, UserInputAst))`

query-grammar/src/query_grammar.rs
```diff
@@ -374,6 +374,30 @@ pub fn parse_to_ast<'a>() -> impl Parser<&'a str, Output = UserInputAst> {
...
+        }
+    }
+    input
+}
+
+fn rewrite_ast_clause(input: &mut (Option<Occur>, UserInputAst)) {
+    match input {
+        (None, UserInputAst::Clause(ref mut clauses)) if clauses.len() == 1 => {
+            let new_term = clauses.pop().unwrap(); // safe because clasues.len() == 1
+            input.0 = new_term.0;
```
adamreichold:
```suggestion
            *input = new_term;
```

or does that not compile?

adamreichold: (Theoretically, even `*input = clauses.pop().unwrap();` but I guess this does not pass the borrow checker before Polonius is merged?)

bazhenov: +1

bazhenov: > (Theoretically, even `*input = clauses.pop().unwrap();` but I guess this does not pass the borrow checker before Polonius is merged?)

In fact it does. I was pretty sure it won't compile! 🤦

adamreichold (approved):
Not that it means anything coming from a bystander, but I like the change as presented now.

fulmicoton (approved):

fulmicoton:
@bazhenov @adamreichold Apologies for not merging this earlier. Thank you for the ocntirbution.

PR #1610, opened 2022-10-09
avoid prepare_doc allocation

PSeitz:
avoid prepare_doc allocation, ~10% more thoughput best case

### DIFF
--- a/src/schema/document.rs
+++ b/src/schema/document.rs
@@ -191,6 +191,34 @@ impl Document {
     pub fn get_first(&self, field: Field) -> Option<&Value> {
         self.get_all(field).next()
     }
+
+    /// Serializes stored field values.
+    pub fn serialize_stored<W: Write>(&self, schema: &Schema, writer: &mut W) -> io::Result<()> {
+        let stored_field_values = || {
+            self.field_values()
+                .iter()
+                .filter(|field_value| schema.get_field_entry(field_value.field()).is_stored())
+        };
+        let num_field_values = stored_field_values().count();
+
+        VInt(num_field_values as u64).serialize(writer)?;
+        for field_value in stored_field_values() {
+            match field_value {
+                FieldValue {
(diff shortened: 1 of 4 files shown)


adamreichold:
This is nice! Did you find this via allocation profiling? If so, which tool did you use?

PSeitz:
> This is nice! Did you find this via allocation profiling? If so, which tool did you use?

@adamreichold I used perf `perf record --call-graph dwarf`. There was a relative large chunk of `free` in the calling function, so I was checking for allocations (`free` is still relative large though). It's also visible with an allocation profiler `valgrind --tool=dhat`.

src/schema/document.rs
```diff
@@ -191,6 +191,34 @@ impl Document {
     pub fn get_first(&self, field: Field) -> Option<&Value> {
         self.get_all(field).next()
     }
+
+    /// Serializes stored field values.
+    pub fn serialize_stored<W: Write>(&self, schema: &Schema, writer: &mut W) -> io::Result<()> {
```
fulmicoton, line 196: can we make this a function in store/writer.rs?

fulmicoton (commented):
Nice!

I have one blocking comment.

fulmicoton:
@PSeitz Awesome. I love everything that removes allocation without makign the code too complex!
If you have creative ideas on how to remove the Doc related allocation that would be nice too.

I think we still have some in the tokenizer (T_T) and the document forces some too.
I think we can remove all of them.

src/schema/document.rs
PSeitz: I think the abstraction is better like this, so the store doesn't need to know about `Document` internals like FieldValues and de/serialization is in the same file

fulmicoton: Ah yes you are right!

fulmicoton (approved):

PR #1614, opened 2022-10-11
refactor Term

### LINKED ISSUE #1585: Remove assert_eq or add a comment
I found this assert a bit cryptic. I understood the 5 after reading the `set_field` method. I would remove it and would expect that a test on `indext_text` will properly avoid any error. 

https://github.com/quickwit-oss/tantivy/blob/main/src/indexer/segment_writer.rs#L220

```rust
                  for mut token_stream in token_streams {
                      assert_eq!(term_buffer.as_slice().len(), 5);
                      postings_writer.index_text(
                          doc_id,
                          &mut *token_stream,
                          term_buffer,
                          ctx,
                          &mut indexing_position,
                          self.fast_field_writers.get_term_id_writer_mut(field),
                      );
                  }
```

PSeitz:
fixes some issues with Term
Remove duplicate calls to truncate or resize
Replace Magic Number 5 with constant
Enforce minimum size of 5 for metadata
Fix broken truncate docs
use constructor instead new + set calls
normalize constructor stack
remove internal term exposure for json term writer
replace assert on internal behavior fixes #1585

### DIFF
--- a/src/indexer/json_term_writer.rs
+++ b/src/indexer/json_term_writer.rs
@@ -242,10 +242,12 @@ pub(crate) fn set_string_and_get_terms(
 ) -> Vec<(usize, Term)> {
     let mut positions_and_terms = Vec::<(usize, Term)>::new();
     json_term_writer.close_path_and_set_type(Type::Str);
-    let term_num_bytes = json_term_writer.term_buffer.as_slice().len();
+    let term_num_bytes = json_term_writer.term_buffer.len_bytes();
     let mut token_stream = text_analyzer.token_stream(value);
     token_stream.process(&mut |token| {
-        json_term_writer.term_buffer.truncate(term_num_bytes);
+        json_term_writer
+            .term_buffer
+            .truncate_value_bytes(term_num_bytes);
         json_term_writer
             .term_buffer
             .append_bytes(token.text.as_bytes());
@@ -260,7 +262,11 @@ pub struct JsonTermWriter<'a> {
 }
 
 impl<'a> JsonTermWriter<'a> {
-    pub fn from_json_path(json_path: &str, field: Field, term_buffer: &'a mut Term) -> Self {
+    pub fn from_field_and_json_path(
+        field: Field,
+        json_path: &str,
+        term_buffer: &'a mut Term,
+    ) -> Self {
         term_buffer.set_field_and_type(field, Type::Json);
         let mut json_term_writer = Self::wrap(term_buffer);
         for segment in json_path.split('.') {
@@ -272,7 +278,7 @@ impl<'a> JsonTermWriter<'a> {
     pub fn wrap(term_buffer: &'a mut Term) -> Self {
         term_buffer.clear_with_type(Type::Json);
         let mut path_stack = Vec::with_capacity(10);
-        path_stack.push(5);
+        path_stack.push(0);
         Self {
             term_buffer,
             path_stack,
@@ -281,28 +287,28 @@ impl<'a> JsonTermWriter<'a> {
 
     fn trim_to_end_of_path(&mut self) {
         let end_of_path = *self.path_stack.last().unwrap();
-        self.term_buffer.truncate(end_of_path);
+        self.term_buffer.truncate_value_bytes(end_of_path);
     }
 
     pub fn close_path_and_set_type(&mut self, typ: Type) {
         self.trim_to_end_of_path();
-        let buffer = self.term_buffer.as_mut();
+        let buffer = self.term_buffer.value_bytes_mut();
         let buffer_len = buffer.len();
         buffer[buffer_len - 1] = JSON_END_OF_PATH;
-        buffer.push(typ.to_code());
+        self.term_buffer.append_bytes(&[typ.to_code()]);
     }
 
     pub fn push_path_segment(&mut self, segment: &str) {
         // the path stack should never be empty.
         self.trim_to_end_of_path();
-        let buffer = self.term_buffer.as_mut();
+        let buffer = self.term_buffer.value_bytes_mut();
         let buffer_len = buffer.len();
         if self.path_stack.len() > 1 {
             buffer[buffer_len - 1] = JSON_PATH_SEGMENT_SEP;
         }
-        buffer.extend(segment.as_bytes());
-        buffer.push(JSON_PATH_SEGMENT_SEP);
-        self.path_stack.push(buffer.len());
+        self.term_buffer.append_bytes(segment.as_bytes());
+        self.term_buffer.append_bytes(&[JSON_PATH_SEGMENT_SEP]);
+        self.path_stack.push(self.term_buffer.len_bytes());
     }
 
     pub fn pop_path_segment(&mut self) {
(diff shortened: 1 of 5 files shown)


src/schema/term.rs
```diff
@@ -42,18 +30,50 @@ impl AsMut<Vec<u8>> for Term {
...
-    pub(crate) fn new() -> Term {
-        Term(Vec::with_capacity(100))
+    pub(crate) fn with_capacity(capacity: usize) -> Term {
+        let mut data = Vec::with_capacity(TERM_METADATA_LENGTH + capacity);
+        data.resize(TERM_METADATA_LENGTH, 0u8);
+        Term(data)
+    }
+
+    pub(crate) fn with_type_and_field(typ: Type, field: Field) -> Term {
+        let mut term = Self::with_capacity(8);
```
adamreichold, line 38: This 8 could be considered somewhat magical as well? It is the bytes used to encode the length of the value, right? So maybe `size_of::<u64>()` or a named constant?

This method also seems to place undue pressure on the optimizer as it will zero out the first 5 bytes of the underlying buffer only to rewrite them in `Term::set_field_and_type`. Maybe

```rust
let mut term = Self(Vec::with_capacity(TERM_METADATA_LENGTH + VALUE_LENGTH_LENGTH));
term.set_field_and_type(typ, field);
term
```

would also work even though it adds a bit of redundancy between the constructors?

PSeitz: It's a magic number, but it doesn't affect correctness. It's on the lower end on how much payload we expect.

We can `resize` and set or `extend`, but `reserve` and will not change the vec's length.

adamreichold: > We can `resize` and set or `extend`, but `reserve` and will not change the vec's length.

Indeed the suggestion is incorrect. My question would then be whether a special case that pushes the field and type instead of setting would be reasonable based on how often this constructor is called?

PSeitz: I don't know if that makes a relevant performance difference

src/indexer/json_term_writer.rs
```diff
@@ -260,12 +260,8 @@ pub struct JsonTermWriter<'a> {
 }
 
 impl<'a> JsonTermWriter<'a> {
-    pub fn from_field_and_json_path(
```
fulmicoton: was it unused? it seems more safe than the alternative

PSeitz: I changed it to the more safe alternative

src/indexer/json_term_writer.rs
```diff
@@ -260,12 +260,8 @@ pub struct JsonTermWriter<'a> {
...
-        field: Field,
-        json_path: &str,
-        term_buffer: &'a mut Term,
-    ) -> Self {
-        term_buffer.set_field(Type::Json, field);
+    pub fn from_json_path(json_path: &str, field: Field, term_buffer: &'a mut Term) -> Self {
```
fulmicoton: why change the order?

field is coming first logically and in the representation

src/schema/term.rs
```diff
@@ -42,18 +30,50 @@ impl AsMut<Vec<u8>> for Term {
...
+        term.set_field_and_type(field, typ);
+        term.0.extend_from_slice(bytes);
+        term
     }
 
     fn from_fast_value<T: FastValue>(field: Field, val: &T) -> Term {
-        let mut term = Term(vec![0u8; FAST_VALUE_TERM_LEN]);
-        term.set_field(T::to_type(), field);
+        let mut term = Self::with_type_and_field(T::to_type(), field);
+        term.set_field_and_type(field, T::to_type());
```
fulmicoton: is this line needed?

src/schema/term.rs
```diff
@@ -42,18 +30,50 @@ impl AsMut<Vec<u8>> for Term {
...
         term.set_u64(val.to_u64());
         term
     }
 
+    /// Panics when there are byte values.
```
fulmicoton, line 63: What is the panics doc about?

src/schema/term.rs
```diff
@@ -42,18 +30,50 @@ impl AsMut<Vec<u8>> for Term {
...
+    ///
+    /// Sets field and the type.
+    pub(crate) fn set_field_and_type(&mut self, field: Field, typ: Type) {
+        assert_eq!(self.0.len(), TERM_METADATA_LENGTH);
```
fulmicoton: assert!(self.is_empty()));

src/indexer/json_term_writer.rs
PSeitz: is it? `from_json_path` and json_path was the argument in the middle. field is only used for the term_buffer

src/schema/term.rs
fulmicoton:
```suggestion
    /// Panics when the term is not empty... ie: some value is set.
```

src/schema/term.rs
```diff
@@ -156,18 +172,13 @@ impl Term {
...
 
-    /// Removes the value_bytes and set the type code.
-    pub fn clear_with_type(&mut self, typ: Type) {
-        self.truncate(5);
-        self.0[4] = typ.to_code();
-    }
-
-    /// Truncate the term right after the field and the type code.
+    /// Truncates the term. The new length needs to be at least 5, which is reserved for metadata.
     pub fn truncate(&mut self, len: usize) {
```
fulmicoton: You changed the logic to make sure that the metadata being encoded in same buffer is an implementation detail, and change the semantics of length for `is_empty` , `with_capacity`... But not this one.

src/schema/term.rs
```diff
@@ -293,7 +304,7 @@ where B: AsRef<[u8]>
     /// Returns `None` if the field is not of string type
     /// or if the bytes are not valid utf-8.
     pub fn as_str(&self) -> Option<&str> {
-        if self.as_slice().len() < 5 {
+        if self.as_slice().len() < TERM_METADATA_LENGTH {
```
fulmicoton: Is it still possible? After you fix truncate?
If the API enforces this invariant, maybe we can make this if-statement  an assert instead... or remove it entirely.

src/schema/term.rs
```diff
@@ -321,7 +332,7 @@ where B: AsRef<[u8]>
     ///
     /// Returns `None` if the field is not of bytes type.
     pub fn as_bytes(&self) -> Option<&[u8]> {
-        if self.as_slice().len() < 5 {
+        if self.as_slice().len() < TERM_METADATA_LENGTH {
```
fulmicoton: same as above.

fulmicoton (changes requested):
See comments inline.

src/indexer/json_term_writer.rs
fulmicoton: Yes.

Representation is:
Field / Type / json_path / value

Logic is 
Field > json_path > value

src/schema/term.rs
PSeitz: no

src/schema/term.rs
PSeitz: I saw this relative late and there were some weird dependencies on the internal state, so I simply added an assert.
I fixed it now

src/schema/term.rs
PSeitz: yes, I think we can remove it

src/schema/term.rs
```diff
@@ -156,18 +166,28 @@ impl Term {
...
     }
 
-    /// Removes the value_bytes and set the type code.
-    pub fn clear_with_type(&mut self, typ: Type) {
-        self.truncate(5);
-        self.0[4] = typ.to_code();
+    /// Truncates the term. The new length needs to be at least 5, which is reserved for metadata.
+    fn truncate(&mut self, len: usize) {
+        assert!(len >= TERM_METADATA_LENGTH);
+        self.0.truncate(len);
```
fulmicoton: this is not fixed. `len` means something different than in other place of the code right now.

fulmicoton: Is this method necessary, or can we leave with just truncate_value_bytes

PSeitz: yes I replaced it with `truncate_value_bytes(0)`

fulmicoton (approved):

PR #1625, opened 2022-10-17
index ip field

PSeitz:
- refactor Term
- fix truncate, remove mutable access from term
- remove truncate
- add indexing for ip field

### DIFF
--- a/src/query/query_parser/query_parser.rs
+++ b/src/query/query_parser/query_parser.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::net::{AddrParseError, IpAddr};
 use std::num::{ParseFloatError, ParseIntError};
 use std::ops::Bound;
 use std::str::{FromStr, ParseBoolError};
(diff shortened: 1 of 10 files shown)


src/query/query_parser/query_parser.rs
```diff
@@ -401,7 +405,7 @@ impl QueryParser {
                 Ok(Term::from_field_bytes(field, &bytes))
             }
             FieldType::IpAddr(_) => Err(QueryParserError::UnsupportedQuery(
-                "Range query are not supported on IpAddr field.".to_string(),
+                "Range query are not supported on ip field.".to_string(),
```
fulmicoton, line 408: Is there an issue open for this?

PSeitz: https://github.com/quickwit-oss/tantivy/issues/1626

src/schema/ip_options.rs
```diff
@@ -79,6 +134,8 @@ impl From<StoredFlag> for IpAddrOptions {
 impl From<IndexedFlag> for IpAddrOptions {
     fn from(_: IndexedFlag) -> Self {
         IpAddrOptions {
+            fieldnorms: true,
```
fulmicoton, line 137: maybe we can drop fieldnorms for IP addresses?
what do you think?

fulmicoton (approved):

src/schema/ip_options.rs
PSeitz: I would keep it to be able to use BM25

PR #1627, opened 2022-10-18
add range query handling for ip via term dictionary

PSeitz:
since IPs are mapped monotonically we can use the term dictionary for range queries


fulmicoton:
@PSeitz we can merge this but the point was to use the fast field whenever available.

fulmicoton (approved):

PR #1631, opened 2022-10-19
add test for phrase search on multi text field

PSeitz:


fulmicoton (approved):
i don't understand the text but surely it cannot hurt :D

PR #1632, opened 2022-10-19
Bugfix position broken.

fulmicoton:
needs a unit test

### DIFF
--- a/src/indexer/segment_writer.rs
+++ b/src/indexer/segment_writer.rs
@@ -751,4 +751,38 @@ mod tests {
         let phrase_query = PhraseQuery::new(vec![nothello_term, happy_term]);
         assert_eq!(searcher.search(&phrase_query, &Count).unwrap(), 0);
     }
+
+    #[test]
+    fn test_bug_regression_1629_position_when_array_with_a_field_value_that_does_not_contain_any_token(
+    ) {
+        // We experienced a bug where we would have a position underflow when computing position
+        // delta in an horrible corner case.
+        //
+        // See the commit with this unit test if you want the details.
+        let mut schema_builder = Schema::builder();
+        let text = schema_builder.add_text_field("text", TEXT);
+        let schema = schema_builder.build();
+        let doc = schema
+            .parse_document(r#"{"text": [ "bbb", "aaa", "", "aaa"]}"#)
+            .unwrap();
+        let index = Index::create_in_ram(schema);
+        let mut index_writer = index.writer_for_tests().unwrap();
+        index_writer.add_document(doc).unwrap();
+        // On debug this did panic on the underflow
+        index_writer.commit().unwrap();
+        let reader = index.reader().unwrap();
+        let searcher = reader.searcher();
+        let seg_reader = searcher.segment_reader(0);
+        let inv_index = seg_reader.inverted_index(text).unwrap();
+        let term = Term::from_field_text(text, "aaa");
+        let mut postings = inv_index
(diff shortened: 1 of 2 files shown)


src/indexer/segment_writer.rs
```diff
@@ -751,4 +751,37 @@ mod tests {
...
+        let mut schema_builder = Schema::builder();
+        let text = schema_builder.add_text_field("text", TEXT);
+        let schema = schema_builder.build();
+        let doc = schema
+            .parse_document(r#"{"text": [ "bbb", "aaa", "", "aaa"]}"#)
+            .unwrap();
+        let index = Index::create_in_ram(schema);
+        let mut index_writer = index.writer_for_tests().unwrap();
+        index_writer.add_document(doc).unwrap();
+        // On debug this would panic on the underflow
```
PSeitz:
```suggestion
        // On debug this did panic on the underflow
```

src/indexer/segment_writer.rs
```diff
@@ -751,4 +751,37 @@ mod tests {
         let phrase_query = PhraseQuery::new(vec![nothello_term, happy_term]);
         assert_eq!(searcher.search(&phrase_query, &Count).unwrap(), 0);
     }
+
+    #[test]
+    fn test_reproduce_bug_position_when_array_with_a_field_value_that_does_not_contain_any_token() {
```
PSeitz:
```suggestion
    fn test_bug_regression_1629_position_when_array_with_a_field_value_that_does_not_contain_any_token() {
```

src/indexer/segment_writer.rs
```diff
@@ -751,4 +751,37 @@ mod tests {
...
+        let inv_index = seg_reader.inverted_index(text).unwrap();
+        let term = Term::from_field_text(text, "aaa");
+        let mut postings = inv_index
+            .read_postings(&term, IndexRecordOption::WithFreqsAndPositions)
+            .unwrap()
+            .unwrap();
+        assert_eq!(postings.doc(), 0u32);
+        let mut positions = Vec::new();
+        postings.positions(&mut positions);
+        // On release this would be [2, 1]. (< note the decreasing values)
```
PSeitz:
```suggestion
        // On release this was [2, 1]. (< note the decreasing values)
```

src/postings/postings_writer.rs
```diff
@@ -155,7 +155,7 @@ pub(crate) trait PostingsWriter: Send + Sync {
     ) {
         let end_of_path_idx = term_buffer.len_bytes();
         let mut num_tokens = 0;
-        let mut end_position = 0;
+        let mut end_position = indexing_position.end_position;
```
PSeitz, line 158: As I understand it, after `token_stream.process`, end_position is equivalent to

`
indexing_position.end_position + last_token.position +  last_token.position_length as u32;
`

Why do we add position_length (original token length)? The code is tested only for `position_length == 1` currently. It could be 0 or N

fulmicoton: let me see.

fulmicoton: Passing the position is done, taking in acount the token length a few lines below.

`end_position = start_position + token.position_length as u32;`

The bug we experienced was only happening in the case where there was a field value that contained no token.
(e.g empty string, just punctuation etc.)

In that case we were reinitializing the end position to 0 stupidly.
After the fix we just make the same value as the previous end position `+ POSITION_GAP`

PR #1637, opened 2022-10-20
add range query via ip fast field

PSeitz:


fastfield_codecs/src/column.rs
```diff
@@ -31,13 +31,21 @@ pub trait Column<T: PartialOrd = u64>: Send + Sync {
         }
     }
 
-    /// Return the positions of values which are in the provided range.
+    /// Get the positions of values which are in the provided value range.
+    ///
+    /// Note that position == docid for single value fast fields
     #[inline]
```
fulmicoton, line 37:
```suggestion
```

src/query/range_query_ip_fastfield.rs
```diff
@@ -0,0 +1,197 @@
...
+        let field_type = reader.schema().get_field_entry(self.field).field_type();
+        match field_type.fastfield_cardinality().unwrap() {
+            Cardinality::SingleValue => {
+                let ip_addr_fast_field = reader.fast_fields().ip_addr(self.field)?;
+                let value_range =
+                    bound_to_value_range(&self.left_bound, &self.right_bound, &ip_addr_fast_field);
+                let docset = IpRangeDocSet::new(value_range, ip_addr_fast_field);
+                Ok(Box::new(ConstScorer::new(docset, boost)))
+            }
+            Cardinality::MultiValues => unimplemented!(),
```
fulmicoton, line 55: Do we plan to implement taht rapidly or is will it be released as is?
If the latter, I'd rather return an error on the construction of the Weight.

src/query/range_query_ip_fastfield.rs
```diff
@@ -0,0 +1,197 @@
...
+        let explanation = Explanation::new("Const", scorer.score());
+
+        Ok(explanation)
+    }
+}
+
+fn bound_to_value_range(
+    left_bound: &Bound<Ipv6Addr>,
+    right_bound: &Bound<Ipv6Addr>,
+    column: &Arc<dyn Column<Ipv6Addr>>,
```
fulmicoton:
```suggestion
    column: &dyn Column<Ipv6Addr>,
```

src/query/range_query_ip_fastfield.rs
```diff
@@ -0,0 +1,197 @@
...
+        Bound::Included(ip_addr) => *ip_addr,
+        Bound::Excluded(ip_addr) => Ipv6Addr::from(ip_addr.to_u128() - 1),
+        Bound::Unbounded => column.max_value(),
+    };
+    start_value..=end_value
+}
+
+/// Helper to have a cursor over a vec
+struct VecCursor {
+    data: Vec<u64>,
```
fulmicoton: data is a terrible name.

src/query/range_query_ip_fastfield.rs
```diff
@@ -0,0 +1,197 @@
...
+    pos_in_data: usize,
```
fulmicoton: pos_in_data too

src/query/range_query_ip_fastfield.rs
```diff
@@ -0,0 +1,197 @@
...
+        Self {
+            data: Vec::with_capacity(32),
+            pos_in_data: 0,
+        }
+    }
+    fn next(&mut self) -> Option<u32> {
+        self.pos_in_data += 1;
+        self.current()
+    }
+    fn current(&self) -> Option<u32> {
```
fulmicoton, line 109:
```suggestion
    #[inline]
    fn current(&self) -> Option<u32> {
```

fastfield_codecs/src/column.rs
```diff
@@ -31,13 +31,21 @@ pub trait Column<T: PartialOrd = u64>: Send + Sync {
...
-    fn get_between_vals(&self, range: RangeInclusive<T>) -> Vec<u64> {
+    fn get_positions_for_value_range(
+        &self,
+        value_range: RangeInclusive<T>,
+        doc_id_range: Range<u64>,
```
fulmicoton: We should pass a `&mut Vec<u64>` here. (And evenutally a `&mut Vec<u32>`)

src/query/range_query_ip_fastfield.rs
```diff
@@ -0,0 +1,197 @@
...
+            value_range,
+            ip_addr_fast_field,
+            loaded_docs: VecCursor::new(),
+            fetched_until_doc: 0,
+        };
+        ip_range_docset.fetch_block();
+        ip_range_docset
+    }
+
+    /// Returns true if more data could be fetched
```
fulmicoton, line 146:
```suggestion
    /// Returns true if more data could be fetched
```
the comment is wrong

src/query/range_query_ip_fastfield.rs
```diff
@@ -0,0 +1,197 @@
...
+            end = limit;
+            finished_to_end = true;
+        }
+
+        let data = self.ip_addr_fast_field.get_positions_for_value_range(
+            self.value_range.clone(),
+            self.fetched_until_doc as u64..end,
+        );
+        self.loaded_docs.set_data(data);
+        self.fetched_until_doc += horizon as u32;
```
fulmicoton: should it be 
```suggestion
        self.fetched_until_doc = end;
```

src/query/range_query_ip_fastfield.rs
PSeitz: add rapidly, as kimbro mentioned they have multivalues for IP

src/query/range_query_ip_fastfield.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+    fn is_empty(&self) -> bool {
+        self.pos_in_data >= self.data.len()
+    }
+}
+
+struct IpRangeDocSet {
+    value_range: RangeInclusive<Ipv6Addr>,
+    ip_addr_fast_field: Arc<dyn Column<Ipv6Addr>>,
+    fetched_until_doc: u32,
```
fulmicoton, line 127: add a comment to define what this is... does this include this value or not?

fastfield_codecs/src/column.rs
PSeitz: I agree, I wanted to keep the PR small

src/query/range_query_ip_fastfield.rs
```diff
@@ -0,0 +1,197 @@
...
+            self.fetched_until_doc as u64..end,
+        );
+        self.loaded_docs.set_data(data);
+        self.fetched_until_doc += horizon as u32;
+        finished_to_end
+    }
+}
+
+impl DocSet for IpRangeDocSet {
+    fn advance(&mut self) -> DocId {
```
fulmicoton, line 181: we need an implementation of seek.

fulmicoton: is it for another PR?

src/query/range_query_ip_fastfield.rs
```diff
@@ -0,0 +1,197 @@
...
+        self.loaded_docs
+            .current()
+            .map(|el| el as u32)
+            .unwrap_or(TERMINATED)
+    }
+
+    fn size_hint(&self) -> u32 {
+        0 // heuristic possible by checking number of hits when fetching a block
+    }
+}
```
fulmicoton, line 203: There are a lot of corner case. We need a lot of unit tests. Proptest could really help.

fastfield_codecs/src/column.rs
fulmicoton: I don't think it makes the code longer... but ok... Can you keep that in your todo list or open a ticket?

src/query/range_query_ip_fastfield.rs
PSeitz: Yes, to keep the PR small.

src/query/range_query_ip_fastfield.rs
fulmicoton: ok can you open a super short ticket

fulmicoton (changes requested):
Just add unit tests.

Make sure there is a list of remaining stuff somewhere. 
Checklist on a ticket is super compact fr that kind of stuff.

I am not sure the horizon model you used is actually addressing the problem.

it might be worth adding a comment explaining which case it tries to 
adapts too. I suspect you miscalculated what it does.
For instance, isn't the current approach is extremely slow for a simple IP range query matching a lot of docs?

fulmicoton (approved):

PR #1643, opened 2022-10-21
allow more characters in range query

### LINKED ISSUE #1642: allow more characters in query_parser range queries
`ip:[::0.0.0.52 TO ::0.0.0.52]` is not allowed by the parser currently, because ":" is a reserved char.

In that context, we can accept more chars, since the grammar doesn't e.g. allow fields here anyway.

PSeitz:
closes #1642

### DIFF
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -62,6 +62,20 @@ fn word<'a>() -> impl Parser<&'a str, Output = String> {
         })
 }
 
+// word variant that allows more characters, e.g. for range queries that don't allow field
+// specifier
+fn relaxed_word<'a>() -> impl Parser<&'a str, Output = String> {
+    (
+        satisfy(|c: char| {
+            !c.is_whitespace() && !['`', '{', '}', '"', '[', ']', '(', ')'].contains(&c)
+        }),
+        many(satisfy(|c: char| {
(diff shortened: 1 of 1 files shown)


fulmicoton (approved):

query-grammar/src/query_grammar.rs
```diff
@@ -62,6 +62,20 @@ fn word<'a>() -> impl Parser<&'a str, Output = String> {
         })
 }
 
+// word variant that allows more characters, e.g. for range queries that don't allow field
+// specifier
+fn relaxed_word<'a>() -> impl Parser<&'a str, Output = String> {
+    (
+        satisfy(|c: char| {
+            !c.is_whitespace() && !['`', '{', '}', '"', '[', ']', '(', ')'].contains(&c)
```
fulmicoton, line 70: what is special about the backquote?

PSeitz: I wasn't sure, I think we can remove it?

fulmicoton: I guess you don't know this was already in `word` :-S
I can't recall either

fulmicoton (approved):

query-grammar/src/query_grammar.rs
PSeitz: I copied it from `word`, and kept it since I don't know why it's there. I would keep it for now, we can always remove it later

PR #1646, opened 2022-10-26
No score calls if score is not requested

PSeitz:
Those score calls can add add up in cost

### DIFF
--- a/src/query/boolean_query/boolean_weight.rs
+++ b/src/query/boolean_query/boolean_weight.rs
@@ -224,7 +224,7 @@ impl<TScoreCombiner: ScoreCombiner + Sync> Weight for BooleanWeight<TScoreCombin
         reader: &SegmentReader,
         callback: &mut dyn FnMut(DocId),
     ) -> crate::Result<()> {
-        let scorer = self.complex_scorer(reader, 1.0, &self.score_combiner_fn)?;
+        let scorer = self.complex_scorer(reader, 1.0, || DoNothingCombiner)?;
         match scorer {
             SpecializedScorer::TermUnion(term_scorers) => {
                 let mut union_scorer = Union::build(term_scorers, &self.score_combiner_fn);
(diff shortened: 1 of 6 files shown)


src/query/weight.rs
```diff
@@ -7,12 +7,21 @@ use crate::{DocId, Score, TERMINATED};
 /// `DocSet` and push the scored documents to the collector.
 pub(crate) fn for_each_scorer<TScorer: Scorer + ?Sized>(
     scorer: &mut TScorer,
+    requires_scoring: bool,
```
fulmicoton: Can we not add this parameter to this function and instead create a separate function:

```rust
pub(crate) fn for_each_docset<TDocSet: DocSet + ?Sized>(
    docset: &mut TDocset,
    callback: &mut dyn FnMut(DocId)
) -> { ... }
```

PSeitz: I updated the PR, but it still adds a fake score when `collect()`ing

src/query/weight.rs
```diff
@@ -16,6 +16,16 @@ pub(crate) fn for_each_scorer<TScorer: Scorer + ?Sized>(
     }
 }
 
+/// Iterates through all of the documents matched by the DocSet
+/// `DocSet`.
+pub(crate) fn for_each_docset<T: DocSet + ?Sized>(scorer: &mut T, callback: &mut dyn FnMut(DocId)) {
```
shikhar: probably calling the param `docset` makes more sense than `scorer`

src/query/boolean_query/boolean_weight.rs
```diff
@@ -219,6 +219,24 @@ impl<TScoreCombiner: ScoreCombiner + Sync> Weight for BooleanWeight<TScoreCombin
         Ok(())
     }
 
+    fn for_each_no_score(
+        &self,
+        reader: &SegmentReader,
+        callback: &mut dyn FnMut(DocId),
+    ) -> crate::Result<()> {
+        let scorer = self.complex_scorer(reader, 1.0, &self.score_combiner_fn)?;
```
shikhar: Does `&DoNothingCombiner` make sense here?

PSeitz: I think so, but it's not called anyway

fulmicoton: Can we still use the DoNothingCombiner?

```suggestion
        let scorer = self.complex_scorer(reader, 1.0, &DoNothingCombiner)?;
```

fulmicoton (approved):

PR #1648, opened 2022-10-26
Avoid unconditional allocation in StemmerTokenStream.

adamreichold:
This fixes the TODO in two ways: If the stemmer already yields an owned string, it is used directly as the new text of the token. Otherwise, a temporary buffer is used to copy the stemmed text (just as before) and then swapping it into the token to reuse its existing buffer.


PSeitz (approved):

PSeitz:
Nice, thanks

PR #1649, opened 2022-10-26
RFC: Add dictionary-based SplitCompoundWords token filter.

adamreichold:
We use a variant of this code to split compound nouns in German texts into their constituents. After removing the integration with a specific dictionary, the token filter itself seemed reasonably generic so that it could be included here.


src/tokenizer/split_compound_words.rs
```diff
@@ -0,0 +1,195 @@
...
+
+        {
+            let mut stream = tokenizer.token_stream("foobar");
+            assert_eq!(stream.next().unwrap().text, "foo");
+            assert_eq!(stream.next().unwrap().text, "bar");
+            assert_eq!(stream.next(), None);
+        }
+
+        {
+            let mut stream = tokenizer.token_stream("foobarbaz");
```
PSeitz, line 191: can you add a tests for 
"foobar foo bar foobar"
"foobar foobar"
"foobar abc foobar"
"foobazbar bar foo foobar"
"barfoo"

PSeitz: tests with actual compounds words like Brotbackautomat may give a better grasp of its capabilities

adamreichold: > can you add a tests for

Done

> tests with actual compounds words like Brotbackautomat may give a better grasp of its capabilities

I thought about that as well as but then, the capabilities are mainly defined by the dictionary. For example, with "Brotbackautomat", this will only work if not just "backen" but also its stem "back" is in the dictionary.

I updated the example with both "Dampfschifffahrt" and "Brotbackautomat" given the dictionary "Dampf, Schiff, Fahrt, Brot, backen, Automat" to illustrate why one works and the other fails.

PSeitz: perfect, that should give a good idea how to use it

PSeitz (approved):

PSeitz:
Thanks!

PR #1650, opened 2022-10-26
Replace FNV by rustc-hash

adamreichold:
Both construction have similar goals but rustc-hash ist better suited for contemporary CPU as it works one word at a time instead of byte per byte.


PSeitz:
Nice thanks. I'll update the benchmarks to see if there's an difference https://pseitz.github.io/tantivy_bench_graph/index.html

PSeitz (approved):

PSeitz:
Nice improvement on strings, regression on numbers
results_agg

PR #1651, opened 2022-10-26
RFC: Include MIT-licensed data from Stopwords ISO

adamreichold:
Adds pre-defined stop word filters for three common European languages based on the MIT-licensed data from [Stopwords ISO](https://github.com/stopwords-iso). This is put behind a non-default feature to avoid bloating the binary for users who are not interested in this while still providing batteries-included usability for those who are.


fulmicoton:
This list of stopwords seems way too long for search.

adamreichold:
> This list of stopwords seems way too long for search.

Alright. Do you know a more reasonable source with compatible licensing terms?

Otherwise, I'd propose that I just break out the first two general improvements to `StopWordFilter` into a new PR and drop this one?

fulmicoton:
We can probably steal something from the Lucene world?

PR #1652, opened 2022-10-27
add docstore version

### LINKED ISSUE #1589: add format version to doc store
add doc store format version to doc store In order to enable allow format changes in the future

PSeitz:
closes #1589

### DIFF
--- a/src/store/footer.rs
+++ b/src/store/footer.rs
@@ -25,7 +25,13 @@ impl BinarySerializable for DocStoreFooter {
     }
 
     fn deserialize<R: io::Read>(reader: &mut R) -> io::Result<Self> {
-        let _doc_store_version = u32::deserialize(reader)?;
+        let doc_store_version = u32::deserialize(reader)?;
+        if doc_store_version != DOC_STORE_VERSION {
+            panic!(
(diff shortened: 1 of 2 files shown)


src/store/footer.rs
```diff
@@ -17,13 +17,15 @@ pub struct DocStoreFooter {
...
     fn serialize<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
+        BinarySerializable::serialize(&DOC_STORE_VERSION, writer)?;
         BinarySerializable::serialize(&self.offset, writer)?;
         BinarySerializable::serialize(&self.decompressor.get_id(), writer)?;
         writer.write_all(&[0; 15])?;
         Ok(())
     }
 
     fn deserialize<R: io::Read>(reader: &mut R) -> io::Result<Self> {
+        let _doc_store_version = u32::deserialize(reader)?;
```
fulmicoton: Can we have a check on the version?
(so that we have an error if we try to open a different version)

PSeitz: yep, I added one

fulmicoton (approved):

PR #1653, opened 2022-10-27
switch to fx hashmap

PSeitz:


adamreichold:
Isn't it possible that the aggregation keys are user-controlled (say if Tantivy is used to power a web service with a generic interface) and hence this would enable DoS attacks?

PSeitz:
I think there are a ton of attack vectors to DoS a search engine, when you have control over the query

adamreichold:
> I think there are a ton of attack vectors to DoS a search engine, when you have control over the query

Agreed. For me, it is enough to know that you are aware of this.

fulmicoton (approved):
Is it actually helpful?

PR #1656, opened 2022-10-31
move multivalue index to own file

PSeitz:


fastfield_codecs/src/column.rs
```diff
@@ -240,6 +240,7 @@ where
     // and we do not have any specialized implementation anyway.
 }
 
+/// Wraps anything that returns an iterator into a `Column`.
```
fulmicoton:
```suggestion
/// Wraps an iterator into a `Column`.
```

fulmicoton: !? What is the deal with min_value and max_value?

That's a bug isn't it?

fulmicoton (approved):

fastfield_codecs/src/column.rs
PSeitz: Indeed that's a bug, it's only used in tests though

PR #1659, opened 2022-11-02
add support for ip range query on multivalue fastfields

PSeitz:


fastfield_codecs/benches/bench.rs
```diff
@@ -145,11 +145,7 @@ mod tests {
 
         b.iter(|| {
             let mut positions = Vec::new();
-            column.get_positions_for_value_range(
-                0..=u128::MAX,
-                0..data.len() as u32,
-                &mut positions,
-            );
+            column.get_docids_for_value_range(0..=u128::MAX, 0..data.len() as u32, &mut positions);
```
fulmicoton, line 148: Can we have a less edge-case  benchmark?

Range containing everything is not uninteresting, I would keep it, but it really makes things too easy for the branch prediction.

src/fastfield/multivalued/index.rs
```diff
@@ -26,6 +26,21 @@ impl MultiValueIndex {
         start..end
     }
 
+    /// Returns `[start, end)`, such that the values associated with
+    /// the given documents are `start..end`.
+    ///
+    /// The passed end range is allowed to be out of bounds.
+    #[inline]
+    pub(crate) fn docid_range_to_position_range(&self, range: Range<DocId>) -> Range<u32> {
+        let end_docid = range.end.min(self.num_docs() - 1);
```
fulmicoton:
```suggestion
        let end_docid = range.end.min(self.num_docs());
```

fulmicoton: I think this is a bug.

src/fastfield/multivalued/index.rs
```diff
@@ -54,40 +70,52 @@ impl MultiValueIndex {
...
-        let mut docs = vec![];
-        let mut cur_doc = docid_start;
+    pub(crate) fn positions_to_docids(&self, doc_id_range: Range<u32>, positions: &mut Vec<u32>) {
+        let mut cur_doc = doc_id_range.start;
         let mut last_doc = None;
 
-        for pos in positions {
+        let mut write_doc_pos = 0;
+        for i in 0..positions.len() {
+            let pos = positions[i];
```
fulmicoton, line 94: I think there is a hidden contract that I do not see document nor asserted.


I believe correctness requires that we have
```
assert!(
   self.idx.get_val(doc_id_range.start()) <= positions[0]
);
```

src/fastfield/multivalued/index.rs
```diff
@@ -54,40 +70,52 @@ impl MultiValueIndex {
...
                         break;
                     }
-                    docs.push(cur_doc);
+                    positions[write_doc_pos] = cur_doc;
+                    write_doc_pos += 1;
+
                     last_doc = Some(cur_doc);
                     break;
                 }
                 cur_doc += 1;
```
fulmicoton: I think rustc will not be able to remove the branch here. (not 100% sure, but the boundary check is probably preventing it)

```rust
                let end = self.idx.get_val(cur_doc + 1) as u32;
                if end > pos {
                    write_doc_pos += 
                         if last_doc == Some(cur_doc) {
                            0
                         } else {
                            1
                         };
                    positions[write_doc_pos] = cur_doc;
                    last_doc = Some(cur_doc);
                    break;
                }
```

You could also remove the other branch prediction doing the same trick on `i` but I don't know if this is beneficial.
as that's a lot of extra store.

src/fastfield/multivalued/reader.rs
```diff
@@ -187,6 +175,44 @@ impl<T: MonotonicallyMappableToU128> MultiValuedU128FastFieldReader<T> {
     }
 }
 
+impl<T: MonotonicallyMappableToU128> Column<T> for MultiValuedU128FastFieldReader<T> {
```
fulmicoton, line 184: Do we need it to implement `Column<T>`? It seems like a `MultiColumn<T>` trait would make more sense.

src/query/range_query_ip_fastfield.rs
```diff
@@ -139,29 +152,33 @@ struct IpRangeDocSet {
...
+    /// If fast field is multivalue.
+    is_multivalue: bool,
 }
 
-const DEFALT_FETCH_HORIZON: u32 = 128;
+const DEFAULT_FETCH_HORIZON: u32 = 128;
 impl IpRangeDocSet {
     fn new(
         value_range: RangeInclusive<Ipv6Addr>,
         ip_addr_fast_field: Arc<dyn Column<Ipv6Addr>>,
```
fulmicoton, line 163: ok so that's why you wanted the Column trait?

Should we have 

4 trait?

```rust
trait ColumnMeta<T: Default> {
   // len, min_val, max_val
   // If there are actually no values, min_value, max_value return T::default()
}


trait Column<T>: ColumnMeta<T> {};
trait ColumnOption<T: Copy>: ColumnMeta<T> {
  fn get() -> Option<T>;
  fn unwrap_or(&self, default_val: T) -> Arc<dyn Column<T>>
  // ...
}
trait ColumnMulti<T>: ColumnMeta<T> {
   fn get_vals(&self, output: &mut Vec<T>);
   fn first_value(&self,) -> Arc<dyn ColumnOption<T>>
   // ...
}
```

src/query/range_query_ip_fastfield.rs
```diff
@@ -589,7 +649,139 @@ mod bench {
...
+
+    #[bench]
+    fn bench_ip_range_hit_90_percent_intersect_with_10_percent_multi(bench: &mut Bencher) {
+        let index = get_index_0_to_100();
+
+        bench.iter(|| {
+            let start = Ipv6Addr::from_u128(0);
+            let end = Ipv6Addr::from_u128(90 * 1000);
+
+            excute_query("ips", start, end, "AND id:few", &index)
```
fulmicoton:
```suggestion
            execute_query("ips", start, end, "AND id:few", &index)
```

src/query/range_query_ip_fastfield.rs
```diff
@@ -435,14 +493,16 @@ mod bench {
         index
     }
     fn excute_query(
+        field: &str,
         start_inclusive: Ipv6Addr,
```
fulmicoton: `ip_range: RangeInclusive<IpV6Addr>` would be more expressive.

src/query/range_query_ip_fastfield.rs
```diff
@@ -321,14 +355,20 @@ mod tests {
...
-    fn ip_range_regression2_test() {
-        let ops = vec![doc_from_id_1(0)];
+    fn ip_range_regression3_test() {
+        let ops = vec![doc_from_id_1(1), doc_from_id_1(2), doc_from_id_1(3)];
         assert!(test_ip_range_for_docs(ops).is_ok());
     }
 
     pub fn create_index_from_docs(docs: &[Doc]) -> Index {
         let mut schema_builder = Schema::builder();
         let ip_field = schema_builder.add_ip_addr_field("ip", INDEXED | STORED | FAST);
```
fulmicoton: Is INDEXED and STORED useful here?

src/query/range_query_ip_fastfield.rs
```diff
@@ -589,7 +649,139 @@ mod bench {
...
             let end = Ipv6Addr::from_u128(90 * 1000);
 
-            excute_query(start, end, "AND id:veryfew", &index)
+            excute_query("ip", start, end, "AND id:veryfew", &index)
+        });
+    }
+
+    #[bench]
+    fn bench_ip_range_hit_90_percent_multi(bench: &mut Bencher) {
+        let index = get_index_0_to_100();
```
fulmicoton, line 626: can we seed those, and add an assert on the number of docs returned?

The bench is strongly relying on the non documented quirky spec of `get_index_0_to_100`.
I'd like to make sure we don't break some bench by modifying the spec of that function.

fulmicoton (changes requested):
One bug and a bunch of important comments.

src/fastfield/multivalued/index.rs
PSeitz: yep

src/query/range_query_ip_fastfield.rs
PSeitz: For the range query we only need 2 methods:
`get_docids_for_value_range` and `num_docs`
The effective horizon is different for multivalue fastfields after translating, but that's a small tradeoff for the same code. 

I'm not sure about splitting the trait, it could be a good idea, but could also be a hassle to work with different traits.
I wouldn't do it in this PR though.

src/query/range_query_ip_fastfield.rs
PSeitz: `STORED` is not used in the tests currently

fastfield_codecs/benches/bench.rs
PSeitz: There's already another benchmark for the 50% hit case

src/fastfield/multivalued/index.rs
PSeitz: I added the assert

src/query/range_query_ip_fastfield.rs
PSeitz: I moved the selectors to functions and added a comment

src/fastfield/multivalued/index.rs
```diff
@@ -26,6 +26,22 @@ impl MultiValueIndex {
         start..end
     }
 
+    /// Returns `[start, end)`, such that the values associated with
+    /// the given documents are `start..end`.
```
fulmicoton:
```suggestion
    /// Given a range of documents, returns the Range of value offsets fo
    /// these documents.
    ///
    /// For instance, `given start_doc..end_doc`,
    /// if we assume Document #start_doc end #end_doc both
    /// have values, this function returns `start..end`
    /// such that `value_column.get(start_doc)` is the first value of
    /// `start_doc` (well, if there is one), and `value_column.get(end_doc - 1)`
    /// is the last value of `end_doc`. 
```

src/fastfield/multivalued/index.rs
```diff
@@ -26,6 +26,22 @@ impl MultiValueIndex {
...
+    ///
+    /// The passed end range is allowed to be out of bounds.
```
fulmicoton:
```suggestion
    /// The passed end range is allowed to be out of bounds, in which case 
    /// it will be clipped to make it valid.
```

src/fastfield/multivalued/reader.rs
```diff
@@ -191,6 +179,44 @@ impl<T: MonotonicallyMappableToU128> MultiValuedU128FastFieldReader<T> {
...
```
fulmicoton, line 184:
```suggestion
// TODO having something that looks like MultiColumn trait. 
// See discussion in #1679
impl<T: MonotonicallyMappableToU128> Column<T> for MultiValuedU128FastFieldReader<T> {
```

fulmicoton: I added a comment. I'd really feel better if we can remove this... Is there a path where we just use the concrete type for the moment?

fulmicoton (approved):

PR #1663, opened 2022-11-07
fix clippy

PSeitz:


fulmicoton (approved):

src/termdict/tests.rs
```diff
@@ -229,10 +229,10 @@ fn test_empty_string() -> crate::Result<()> {
     let buffer: Vec<u8> = {
         let mut term_dictionary_builder = TermDictionaryBuilder::create(vec![]).unwrap();
         term_dictionary_builder
-            .insert(&[], &make_term_info(1_u64))
```
fulmicoton, line 232: Clippy asked for that one? 
:-S

Silly question, but does that generate twice as much bytecode?

PSeitz: > Clippy asked for that one?

yep

> Silly question, but does that generate twice as much bytecode?

These are different types now, so I think yes

PR #1666, opened 2022-11-07
Include stop word lists from the Snowball project

adamreichold:
Following the suggestion in https://github.com/quickwit-oss/tantivy/pull/1651#issuecomment-1298194705, this follows Apache Lucene by incorporating the stop word lists from the BSD-licensed [Snowball](https://snowballstem.org/) project.


src/tokenizer/stop_word_filter.rs
```diff
@@ -23,6 +23,26 @@ pub struct StopWordFilter {
     words: Arc<FxHashSet<String>>,
 }
 
+macro_rules! language {
+    ($lang:ident) => {
+        #[cfg(feature = "stopwords")]
+        /// Create a `StopWorldFilter` for the named language
+        pub fn $lang() -> Self {
+            Self::remove(
+                include_str!(concat!("stopwords/", stringify!($lang), ".txt"))
```
fulmicoton: It would be nicer to have it directly in a const &str array.
I assume you did not do that to make the source as clear as possible?

I'd rather use a simple array.
That's:
- less code for me to review
- less code to test: we can eyeball the results
- less code to compile
- less code to maintain
- less bytes in all binaries containing tantivy (not that much of course, but still)
- less macro

We don't need to go as far as doing build.rs magic.
I think it is ok to just have those in a feature gated module as `const &'static [&'str; N]`, 
and put the origin of the data in a comment.

src/tokenizer/stopwords/english.txt
```diff
@@ -0,0 +1,312 @@
...
+ |            | MUST
+ |must
+ |            | OUGHT
+
+ought
+
+           | COMPOUND FORMS, increasingly encountered nowadays in 'formal' writing
+              | pronoun + verb
+
+i'm
```
fulmicoton: I don't think those make sense with our tokenizer.

src/tokenizer/stopwords/english.txt
```diff
@@ -0,0 +1,312 @@
...
+themselves     | reflexive
+             | other forms (demonstratives, interrogatives)
+what
+which
+who
+whom
+this
+that
+these
+those
```
fulmicoton: I'm ok with this list, although it is still a tad long to my liking I think.

For reference, the one in Lucene 
https://cs.github.com/apache/lucene/blob/682e5c94e83320fee929a16f375fd0ebe4ddc99c/lucene/analysis/common/src/java/org/apache/lucene/analysis/en/EnglishAnalyzer.java?q=EnglishAnalyzer+org%3Aapache+#L46-L51

fulmicoton:
The list in Lucene is much shorter I believe @adamreichold .
Also, can we inline the words as literals?

fulmicoton (changes requested):
See comments inline

src/tokenizer/stopwords/english.txt
adamreichold: So pass the words through `SimpleTokenizer` to add the tokens to the `HashSet`? Or just skip those which are not a single token?

src/tokenizer/stopwords/english.txt
adamreichold: Lucene does not seem to use Snowball for English, only for other languages like German.

I can drop the English one and replace it by the open-coded list from Lucene? (Apache Licensed should be compatible I think?)

src/tokenizer/stopwords/english.txt
adamreichold: (Is this the only example? If so, I'd say this is moot if we go for Lucene's list in https://github.com/quickwit-oss/tantivy/pull/1666#discussion_r1016060491)

src/tokenizer/stop_word_filter.rs
adamreichold: > I assume you did not do that to make the source as clear as possible?

The main reason was not having any extra build-time processing step.

So I propose, I'll add a small Python script which will download and process the Snowball files and produce a generated module that is checked into source control? (The data files themselves would not end up in source control then.)

src/tokenizer/stopwords/english.txt
fulmicoton: No. I'd rather handcraft the list of tokens outside of the project and hardcodign them in the code.

I'd rather use the lucene, but if you want a larger list Im happy to debate it. Let's just make sure it only contains actually tokens.

src/tokenizer/stopwords/english.txt
fulmicoton: ok for me.

adamreichold: I noticed that the current stop word list for English already is the one from Lucene and hence made the change purely additive. (I did opt to make the `StopWordFilter::english` method public though for consistency. I am also not really convinced the `Default` impl but did not opt to change it.)

src/tokenizer/stopwords/english.txt
adamreichold: As discussed below, I dropped the stop word list from the Snowball project for English for this PR and the others do not seem to have this issue.

src/tokenizer/stop_word_filter.rs
adamreichold: Did that and checked in both the script and its result.

src/tokenizer/stop_word_filter/stopwords.rs
```diff
@@ -0,0 +1,2115 @@
+/*
+Copyright (c) 2001, Dr Martin Porter
```
fulmicoton, line 5: Can you add a comment/url to explain the origin of this data?

src/tokenizer/stop_word_filter/stopwords.rs
```diff
@@ -0,0 +1,2115 @@
...
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+#![allow(non_upper_case_globals)]
+
+pub const danish: &[&str] = &[
+    "og",
```
fulmicoton, line 37: Thank you!

src/tokenizer/stop_word_filter/mod.rs
```diff
@@ -31,15 +45,28 @@ impl StopWordFilter {
...
             "a", "an", "and", "are", "as", "at", "be", "but", "by", "for", "if", "in", "into",
             "is", "it", "no", "not", "of", "on", "or", "such", "that", "the", "their", "then",
             "there", "these", "they", "this", "to", "was", "will", "with",
         ];
 
-        StopWordFilter::remove(words.iter().map(|&s| s.to_string()))
+        Self::remove(WORDS.iter().map(|&word| word.to_owned()))
     }
+
+    language!(swedish);
```
fulmicoton: The macro is no useful here.

```rust
fn from_words_slice(words: &[&str]) -> Self {
       // ...
}

fn dutch(&self) -> Self {
     Self::from_words_slice(&stopwords::DUTCH)
}

fn italian(&self) -> Self {
     Self::from_words_slice(&stopwords::ITALIAN)
}
```


I think in a separate ticket, we should introduce a Language enum, and rely on that for the stop words filter and the stemmer filter.

src/tokenizer/stop_word_filter/stopwords.rs
```diff
@@ -0,0 +1,2115 @@
...
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+#![allow(non_upper_case_globals)]
```
fulmicoton:
```suggestion
```

Let's ditch the macro and use normal uppercase here.

src/tokenizer/stop_word_filter/mod.rs
```diff
@@ -31,15 +45,28 @@ impl StopWordFilter {
         }
     }
 
-    fn english() -> StopWordFilter {
-        let words: [&'static str; 33] = [
+    /// Create a `StopWorldFilter` for the named language
+    pub fn english() -> Self {
+        const WORDS: &[&str] = &[
```
fulmicoton, line 58: Add one line to explain the origin of this list. (Copied from Lucene).

fulmicoton (changes requested):
See comments inline

fulmicoton:
@adamreichold I still have a bunch of comments.  Sorry for the multiple iterations. I hope it does not seem too arbitrary. Your energy and attention to detail is actually well appreciated! ❤️

src/tokenizer/stop_word_filter/mod.rs
adamreichold: Done.

src/tokenizer/stop_word_filter/stopwords.rs
adamreichold: Done.

src/tokenizer/stop_word_filter/mod.rs
adamreichold: I think there already is [`tantivy::tokenizer::Language`](https://docs.rs/tantivy/latest/tantivy/tokenizer/enum.Language.html) used by the Stemmer which we could just apply here.

I think the main usability issue here would be that we do not have stop word lists for all of these languages. Should this fail to produce a filter then? Or produce a filter using an empty list?

src/tokenizer/stop_word_filter/stopwords.rs
adamreichold: Done.

adamreichold:
> @adamreichold I still have a bunch of comments. Sorry for the multiple iterations. I hope it does not seem too arbitrary. Your energy and attention to detail is actually well appreciated! heart

Better to take care and try to do it right than to slowly turn a FOSS project into a dumping ground for half-assed code drops. I don't mind iterating on this at all.

src/tokenizer/stop_word_filter/mod.rs
fulmicoton: You could return an option yes.

adamreichold: Will add that as a separate PR.

fulmicoton (approved):

fulmicoton:
Thank you!

PR #1667, opened 2022-11-08
Refactoring `Query::Weight` to not need a Searcher when scoring is disabled.

fulmicoton:
This refactoring makes it possible to remove the IndexReader from the IndexWriter.
Delete queries do not require scoring, and should not require a searcher.

### DIFF
--- a/src/core/searcher.rs
+++ b/src/core/searcher.rs
@@ -4,7 +4,7 @@ use std::{fmt, io};
 
 use crate::collector::Collector;
 use crate::core::{Executor, SegmentReader};
-use crate::query::Query;
+use crate::query::{EnableScoring, Query};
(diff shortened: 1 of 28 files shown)


trinity-1686a (commented):
I believe not reloading the `IndexReader` prevent deletion of old segments.
We should clarify if it is indeed the case, or not. And if it is, find a place to reload the `IndexReader` often enough it's not keeping too many files alive

fulmicoton:
> I believe not reloading the `IndexReader` prevent deletion of old segments. We should clarify if it is indeed the case, or not. And if it is, find a place to reload the `IndexReader` often enough it's not keeping too many files alive

Great great point! I'll find a workaround.

trinity-1686a (approved):
LGTM

PR #1671, opened 2022-11-09
Make the built-in stop word lists selectable via the Language enum already used by the Stemmer filter.

adamreichold:


src/tokenizer/stop_word_filter/mod.rs
```diff
@@ -28,95 +27,54 @@ pub struct StopWordFilter {
 }
 
 impl StopWordFilter {
+    /// Creates a new [`StopWordFilter`] for the given [`Language`]
+    ///
+    /// Returns `Some` if a list of stop words is available and `None` otherwise.
+    pub fn new(language: Language) -> Option<Self> {
```
fulmicoton, line 34: Can we put the feature gate on the new function directly?

src/tokenizer/stop_word_filter/mod.rs
```diff
@@ -160,7 +118,7 @@ impl<'a> TokenStream for StopWordFilterStream<'a> {
 
 impl Default for StopWordFilter {
     fn default() -> StopWordFilter {
-        StopWordFilter::english()
+        StopWordFilter::new(Language::English).unwrap()
```
fulmicoton: Can we remove the entire `StopWordFilter`?

adamreichold: You mean feature-gate the entire `stop_word_filter` module and hence `StopWorldFilter` type behind `#[cfg(feature = "stopwords")]` (or maybe `#[cfg(feature = "stop-word-filter")]` then)?

adamreichold: Personally, I'd say `StopWordFilter` can be useful without the built-in lists (For example, it can remove any list of words, not just stop words.) so I would suggest removing the `Default` impl (which I find questionable being tied to specific language in any case) and feature-gate `StopWorldFilter::new` so `StopWorldFilter::remove` could still be used with custom word lists.

fulmicoton: Sorry I mistyped! 
 I meant can we remove the default implementation for StopWordFilter.

I think we are on the same page here.
No more Default implementation.
Feature gate StopWordFilter::new.

src/tokenizer/stop_word_filter/mod.rs
adamreichold: Done.

src/tokenizer/stop_word_filter/mod.rs
adamreichold: Indeed. Changed as suggested.

fulmicoton (approved):

PR #1672, opened 2022-11-10
Allow range query on fastfield without INDEXED

PSeitz:


fulmicoton (approved):

fulmicoton:
do we allow range query on ip fields if they are not fastfield?

If so, can you add a simple unit test for both case to cover that possibillity (can be in a separate PR)?

PSeitz:
> do we allow range query on ip fields if they are not fastfield?
> 
> If so, can you add a simple unit test for both case to cover that possibillity (can be in a separate PR)?

Yes, and there's a test already: https://github.com/quickwit-oss/tantivy/pull/1627

PR #1674, opened 2022-11-10
add header with codec type for u128

PSeitz:


fastfield_codecs/src/serialize.rs
```diff
@@ -167,7 +190,11 @@ pub fn serialize_u128<F: Fn() -> I, I: Iterator<Item = u128>>(
     num_vals: u32,
     output: &mut impl io::Write,
 ) -> io::Result<()> {
-    // TODO write header, to later support more codecs
+    let header = U128Header {
+        num_vals,
+        codec_type: U128FastFieldCodecType::CompactSpace,
+    };
+    header.serialize(output)?;
```
fulmicoton, line 197: Can you add a dumb unit test testing `deserialize(serialize(T)) == T` on a simple sample?

fulmicoton (approved):

fastfield_codecs/src/serialize.rs
```diff
@@ -98,6 +98,29 @@ impl Header {
     }
 }
 
+#[derive(Debug, Copy, Clone)]
+pub(crate) struct U128Header {
+    pub num_vals: u32,
+    pub codec_type: U128FastFieldCodecType,
```
fulmicoton, line 104: I'm a  bit confused... Why do we need `num_vals` and are you planning to handle sparsity in a non-generic way?

PSeitz: sparsity would be a general solution with a separate index

PR #1680, opened 2022-11-14
[Discussion] Removing the need for column on Multivalued blabla

fulmicoton:


src/query/range_query_ip_fastfield.rs
```diff
@@ -202,47 +200,72 @@ impl IpRangeDocSet {
             true
         }
     }
+}
 
-    /// Fetches a block for docid range [next_fetch_start .. next_fetch_start + HORIZON]
+trait SingleOrMultivalued {
```
fulmicoton, line 205: A better name would be nicer...

PR #1687, opened 2022-11-19
Added a `expand_dots` JsonObjectOptions.

fulmicoton:


src/schema/json_object_options.rs
```diff
@@ -26,6 +28,16 @@ impl JsonObjectOptions {
         self.indexing.is_some()
     }
 
+    /// Returns `true` iff the json object should be indexed.
+    pub fn is_expand_dots_enabled(&self) -> bool {
```
trinity-1686a: doc comment is copy-pasted from above, not actually describing *this* function

src/schema/json_object_options.rs
```diff
@@ -26,6 +28,16 @@ impl JsonObjectOptions {
...
+        self.expand_dots_enabled
+    }
+
+    pub fn set_expand_dots_enabled(mut self) -> Self {
```
trinity-1686a, line 49: missing doc comment

trinity-1686a (changes requested):
See comments.
Also I think having a `const EXPAND_DOT` or something along the line would be a nice to have

trinity-1686a (approved):

PR #1691, opened 2022-11-22
prepare fastfield format for null index

PSeitz:
With the upcoming null handling/sparse data, the format will change for fast fields. 
Serialize a null index footer that reflects a `Full` column, to avoid compatibility issues for the next tantivy version

### DIFF
--- a/fastfield_codecs/src/compact_space/mod.rs
+++ b/fastfield_codecs/src/compact_space/mod.rs
@@ -373,7 +373,7 @@ impl CompactSpaceDecompressor {
 
         let compact_from = compact_from.unwrap_or_else(|pos| {
             // Correctness: Out of bounds, if this value is Err(last_index + 1), we early exit,
-            // since the to_value also mapps into the same non-mapped spacemich
+            // since the to_value also mapps into the same non-mapped space
             let range_mapping = self.params.compact_space.get_range_mapping(pos);
             range_mapping.compact_start
         });
--- a/fastfield_codecs/src/compact_space/mod.rs
+++ b/fastfield_codecs/src/compact_space/mod.rs
@@ -457,7 +457,7 @@ mod tests {
 
     use super::*;
     use crate::format_version::read_format_version;
-    use crate::null_index_footer::NullIndexFooter;
+    use crate::null_index_footer::read_null_index_footer;
     use crate::serialize::U128Header;
     use crate::{open_u128, serialize_u128};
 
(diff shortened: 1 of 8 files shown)


fulmicoton:
I don't understand what this PR is about. Can you edit the description and the commit message? @PSeitz

PSeitz:
> I don't understand what this PR is about. Can you edit the description and the commit message? @PSeitz

I modified the commit message

fulmicoton (changes requested):
I don't think this is putting us on the right track for backward compatibility.

Here are the issue I have:
- The footer is constant length.
- It implements stuff for one future that is not there.
- The footer is 11 bytes long and replicated for every column. once we have sparse columns we may end up with a lot of columns. 11 bytes is a bit wasteful (not so terrible but well)

I think we could...
End the fastfield file by a codec version and a magic number.

A codec object would have  a `fn load(&self, data: OwnedBytes) -> Result<_, _>`  instead of the load method in the fastfield_codecs crate.

PSeitz:
> It implements stuff for one future that is not there.

In just describes the current state, which will be compatible with the future.

> The footer is 11 bytes long and replicated for every column. once we have sparse columns we may end up with a lot of columns. 11 bytes is a bit wasteful (not so terrible but well)

It's 18bytes, that's ~2kb for 100columns, I don't think this will ever be relevant. No problem to change it though.

> I think we could...
> End the fastfield file by a codec version and a magic number.

I think that's a good idea in general and should be added. 

The downside to do that _instead_,  is to introduce a new version and with it all the annoying version handling when one will be fine.



> A codec object would have a fn load(&self, data: OwnedBytes) -> Result<_, _> instead of the load method in the fastfield_codecs crate.

We already have that 
```
fn open_from_bytes(mut data: OwnedBytes, header: NormalizedHeader) -> io::Result<Self::Reader>
```

fastfield_codecs/src/compact_space/mod.rs
```diff
@@ -373,7 +373,7 @@ impl CompactSpaceDecompressor {
 
         let compact_from = compact_from.unwrap_or_else(|pos| {
             // Correctness: Out of bounds, if this value is Err(last_index + 1), we early exit,
-            // since the to_value also mapps into the same non-mapped space
+            // since the to_value also mapps into the same non-mapped spacemich
```
fulmicoton:
```suggestion
            // since the to_value also mapps into the same non-mapped space
```

fastfield_codecs/src/null_index_footer.rs
```diff
@@ -0,0 +1,121 @@
...
+    // Unused for NullIndexCodec::Full
+    pub(crate) null_index_byte_range: Range<u64>,
+}
+
+impl BinarySerializable for NullIndexFooter {
+    fn serialize<W: Write>(&self, writer: &mut W) -> io::Result<()> {
+        self.cardinality.serialize(writer)?;
+        self.null_index_codec.serialize(writer)?;
+        VInt(self.null_index_byte_range.start).serialize(writer)?;
+        VInt(self.null_index_byte_range.end).serialize(writer)?;
```
fulmicoton: Can you switch to encoding `end-start` instead of `end`

fulmicoton (approved):

PR #1692, opened 2022-11-23
Move `split_full_path` to `Schema`

boraarslan:
Having this method on the `Schema` level will help simplify the code on Quickwit for the issue https://github.com/quickwit-oss/quickwit/pull/2260.

Also, this comes with a bonus of removing the `field_names` hashmap from `QueryParser`.

### DIFF
--- a/src/schema/schema.rs
+++ b/src/schema/schema.rs
@@ -252,6 +252,31 @@ impl Eq for InnerSchema {}
 #[derive(Clone, Eq, PartialEq, Debug)]
 pub struct Schema(Arc<InnerSchema>);
 
+// Returns the position (in byte offsets) of the unescaped '.' in the `field_path`.
+//
+// This function operates directly on bytes (as opposed to codepoint), relying
+// on a encoding property of utf-8 for its correctness.
+fn locate_splitting_dots(field_path: &str) -> Vec<usize> {
+    let mut splitting_dots_pos = Vec::new();
+    let mut escape_state = false;
+    for (pos, b) in field_path.bytes().enumerate() {
+        if escape_state {
+            escape_state = false;
+            continue;
+        }
+        match b {
+            b'\\' => {
+                escape_state = true;
+            }
+            b'.' => {
+                splitting_dots_pos.push(pos);
+            }
+            _ => {}
+        }
+    }
+    splitting_dots_pos
+}
+
 impl Schema {
     /// Return the `FieldEntry` associated with a `Field`.
     pub fn get_field_entry(&self, field: Field) -> &FieldEntry {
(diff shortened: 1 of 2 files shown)


src/schema/schema.rs
```diff
@@ -358,6 +383,21 @@ impl Schema {
         }
         Ok(doc)
     }
+
+    /// Splits a full_path, into a field name and a json path.
+    pub fn split_full_path<'a>(&self, full_path: &'a str) -> Option<(Field, &'a str)> {
```
guilload: `find_field`?

src/schema/schema.rs
```diff
@@ -358,6 +383,21 @@ impl Schema {
         }
         Ok(doc)
     }
+
+    /// Searches for a full_path, returning the field name and a json path.
```
fulmicoton: This is a public method in a library.

Can you elaborate the comment here?
What happens if there is a non-empty json path, but the field is not a json field for instance?

src/schema/schema.rs
```diff
@@ -358,6 +383,21 @@ impl Schema {
...
+    pub fn find_field<'a>(&self, full_path: &'a str) -> Option<(Field, &'a str)> {
+        if let Some(field) = self.0.fields_map.get(full_path) {
```
fulmicoton, line 396: No unit tests for find_field!? Can we add some?

boraarslan: I mostly copied the same unit test on `QueryParser`. I can remove the test from `QueryParser` or add more tests to `Schema` if that's what we want.

fulmicoton: yes please that would be awesome!

src/schema/schema.rs
```diff
@@ -358,6 +383,28 @@ impl Schema {
...
         Ok(doc)
     }
+
+    /// Searches for a full_path in the schema, returning the field name and a JSON path.
+    ///
+    /// This function works by checking if the field exists for the exact given full_path.
+    /// If it's not, it splits the full_path at non-escaped '.' chars and tries to match the
+    /// prefix with the field names, favoring the longest field names.
+    ///
+    /// This does not check if field is a JSON field. It is possible for this functions to
```
fulmicoton, line 393: Would it make sense to improve the spec and avoid this?

boraarslan: It might be. I changed the function to continue when the prefix is not a JSON field, and all of the tests passed (probably does not mean it is completely ok to change), but I think it might be better to open a separate PR for it.

fulmicoton: Can you open a ticket?

boraarslan: #1699

fulmicoton (approved):

Issue #1702, opened 2022-11-30
Max bucket limit in aggregation


PSeitz:
A max bucket limit was introduced here: https://github.com/quickwit-oss/tantivy/issues/1331.

It does not take into account a lot of final buckets when merging intermediate buckets.

PSeitz:
Superseeded by https://github.com/quickwit-oss/tantivy/issues/1822

PR #1703, opened 2022-11-30
Fix max bucket limit in histogram

PSeitz:
The max bucket limit in histogram was broken, since some code introduced temporary filtering of buckets, which then resulted in an incorrect increment on the bucket count.
The provided solution handles more use cases correctly, but there are still some scenarios unhandled (See #1702).

https://github.com/quickwit-oss/quickwit/issues/2503

### DIFF
--- a/src/aggregation/bucket/histogram/histogram.rs
+++ b/src/aggregation/bucket/histogram/histogram.rs
@@ -217,25 +217,22 @@ impl SegmentHistogramCollector {
         agg_with_accessor: &BucketAggregationWithAccessor,
     ) -> crate::Result<IntermediateBucketResult> {
         // Compute the number of buckets to validate against max num buckets
-        // Note: We use min_doc_count here, but it's only an estimate here, since were are on the
-        // intermediate level and after merging min_doc_count could be different.
+        // Note: We use min_doc_count here, but it's only an lowerbound here, since were are on the
+        // intermediate level and after merging the number of documents of a bucket could exceed
+        // `min_doc_count`.
         {
             let cut_off_buckets_front = self
                 .buckets
                 .iter()
                 .take_while(|bucket| bucket.doc_count <= self.min_doc_count)
                 .count();
-            let cut_off_buckets_end = self
-                .buckets
+            let cut_off_buckets_back = self.buckets[cut_off_buckets_front..]
                 .iter()
                 .rev()
                 .take_while(|bucket| bucket.doc_count <= self.min_doc_count)
                 .count();
-            let estimate_num_buckets = self
-                .buckets
-                .len()
-                .saturating_sub(cut_off_buckets_front)
-                .saturating_sub(cut_off_buckets_end);
+            let estimate_num_buckets =
+                self.buckets.len() - cut_off_buckets_front - cut_off_buckets_back;
 
             agg_with_accessor
                 .bucket_count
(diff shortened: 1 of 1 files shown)


fulmicoton (approved):

fulmicoton (changes requested):
...

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -215,6 +216,33 @@ impl SegmentHistogramCollector {
...
+        // Note: We use min_doc_count here, but it's only an estimate here, since were are on the
+        // intermediate level and after merging min_doc_count could be different.
+        {
+            let cut_off_buckets_front = self
+                .buckets
+                .iter()
+                .take_while(|bucket| bucket.doc_count <= self.min_doc_count)
+                .count();
+            let cut_off_buckets_end = self
+                .buckets
```
```suggestion
                .buckets[cut_off_buckets_front..]
```

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -215,6 +216,33 @@ impl SegmentHistogramCollector {
...
+                .iter()
+                .rev()
+                .take_while(|bucket| bucket.doc_count <= self.min_doc_count)
+                .count();
+            let estimate_num_buckets = self
+                .buckets
+                .len()
+                .saturating_sub(cut_off_buckets_front)
+                .saturating_sub(cut_off_buckets_end);
```
```suggestion
                - cut_off_buckets_front - cut_off_buckets_back
```

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -215,6 +216,33 @@ impl SegmentHistogramCollector {
...
+        // Compute the number of buckets to validate against max num buckets
+        // Note: We use min_doc_count here, but it's only an estimate here, since were are on the
+        // intermediate level and after merging min_doc_count could be different.
+        {
+            let cut_off_buckets_front = self
+                .buckets
+                .iter()
+                .take_while(|bucket| bucket.doc_count <= self.min_doc_count)
+                .count();
+            let cut_off_buckets_end = self
```
fulmicoton: nitpick... The opposite of front is back.

```suggestion
            let cut_off_buckets_back = self
```

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -215,6 +216,33 @@ impl SegmentHistogramCollector {
         self,
         agg_with_accessor: &BucketAggregationWithAccessor,
     ) -> crate::Result<IntermediateBucketResult> {
+        // Compute the number of buckets to validate against max num buckets
+        // Note: We use min_doc_count here, but it's only an estimate here, since were are on the
```
```suggestion
        // Note: We use min_doc_count here, but it's only an lowerbound here, since were are on the
```

It's an lowerbound, correct?

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -215,6 +216,33 @@ impl SegmentHistogramCollector {
...
+        // intermediate level and after merging min_doc_count could be different.
```
```suggestion
        // intermediate level and after merging the number of documents of a bucket could exceed `min_doc_count`.
```

src/aggregation/bucket/histogram/histogram.rs
PSeitz: yes

fulmicoton (approved):

PR #1718, opened 2022-12-12
Isolating sstable and stacker in independant crates.

fulmicoton:
Both crate will be used in the new (optional + dynamic) fastfield work.


stacker/src/arena_hashmap.rs
```diff
@@ -41,15 +40,16 @@ impl KeyValue {
...
-/// stored in a user defined memory arena.
+/// Its main particularity is that rather than storing its
+/// keys in the heap, keys are stored in a memory arena
+/// inline with the values.
 ///
 /// The quirky API has the benefit of avoiding
 /// the computation of the hash of the key twice,
 /// or copying the key as long as there is no insert.
-pub struct TermHashMap {
+pub struct ArenaHashMap {
```
PSeitz, line 52: I think the new name is too generic, the implementation is very specific, mapping &[u8] keys to ids

stacker/src/arena_hashmap.rs
```diff
@@ -153,22 +158,30 @@ impl TermHashMap {
         }
     }
 
-    fn set_bucket(&mut self, hash: u32, key_value_addr: Addr, bucket: usize) -> UnorderedTermId {
+    #[inline]
+    fn set_bucket(&mut self, hash: u32, key_value_addr: Addr, bucket: usize) -> UnorderedId {
         self.occupied.push(bucket);
-        let unordered_term_id = self.len as UnorderedTermId;
+        let unordered_term_id = self.len as UnorderedId;
```
PSeitz:
```suggestion
        let unordered_id = self.len as UnorderedId;
```

stacker/src/arena_hashmap.rs
```diff
@@ -153,22 +158,30 @@ impl TermHashMap {
...
         self.len += 1;
         self.table[bucket] = KeyValue {
             key_value_addr,
             hash,
-            unordered_term_id,
+            unordered_id: unordered_term_id,
```
PSeitz:
```suggestion
            unordered_id,
```

stacker/src/arena_hashmap.rs
fulmicoton: Yeah... So `Term` will soon be incorrect, as I use the stacker with field names as keys in dynamic fast fields.

I could have something like `ArenaBytesHashMap` or something like this, but I think at this point `ArenaHashMap` is ok.

PR #1726, opened 2022-12-21
Refactoring to prepare for the addition of dynamic fast field

fulmicoton:


common/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tantivy-common"
-version = "0.4.0"
+version = "0.5.0"
```
PSeitz, line 3: I think it's better to let the versioning handle by cargo release, to keep everything in sync

PR #1736, opened 2022-12-22
use group_by in translate_codec_idx_to_original_idx

PSeitz:
The `group_by` has some overhead, and for the strictly sparse case it's just worse. But it will address a pathological case when there are cluster of dense blocks in a sparse field with a many search results in it. I think that's a valid scenario, when users start to use the JSON type to ingest and search different kinds of data into a single index

```
Index has 1mio docs

Now
running 5 tests
test null_index::sparse::bench::bench_translate_codec_to_orig_1percent_filled_0comma005percent_hit  ... bench:         101 ns/iter (+/- 1)
test null_index::sparse::bench::bench_translate_codec_to_orig_1percent_filled_10percent_hit         ... bench:      10,805 ns/iter (+/- 682)
test null_index::sparse::bench::bench_translate_codec_to_orig_1percent_filled_full_scan             ... bench:      47,879 ns/iter (+/- 3,955)
test null_index::sparse::bench::bench_translate_codec_to_orig_90percent_filled_0comma005percent_hit ... bench:       9,267 ns/iter (+/- 66)
test null_index::sparse::bench::bench_translate_codec_to_orig_90percent_filled_full_scan            ... bench:  16,630,831 ns/iter (+/- 315,576)


Before
test null_index::sparse::bench::bench_translate_codec_to_orig_1percent_filled_0comma005percent_hit  ... bench:          91 ns/iter (+/- 2)
test null_index::sparse::bench::bench_translate_codec_to_orig_1percent_filled_10percent_hit         ... bench:       6,505 ns/iter (+/- 256)
test null_index::sparse::bench::bench_translate_codec_to_orig_1percent_filled_full_scan             ... bench:      33,322 ns/iter (+/- 2,482)
test null_index::sparse::bench::bench_translate_codec_to_orig_90percent_filled_0comma005percent_hit ... bench:      14,599 ns/iter (+/- 225)
test null_index::sparse::bench::bench_translate_codec_to_orig_90percent_filled_full_scan            ... bench: 295,434,573 ns/iter (+/- 31,193,127)
```

### DIFF
--- a/common/src/group_by.rs
+++ b/common/src/group_by.rs
@@ -1,0 +1,166 @@
+use std::cell::RefCell;
+use std::iter::Peekable;
+use std::rc::Rc;
+
+pub trait GroupByIteratorExtended: Iterator {
+    /// Return an `Iterator` that groups iterator elements. Consecutive elements that map to the
+    /// same key are assigned to the same group.
+    ///
+    /// The returned Iterator item is `(K, impl Iterator)`, where Iterator are the items of the
+    /// group.
+    ///
+    /// ```
+    /// use tantivy_common::GroupByIteratorExtended;
+    ///
+    /// // group data into blocks of larger than zero or not.
+    /// let data: Vec<i32> = vec![1, 3, -2, -2, 1, 0, 1, 2];
+    /// // groups:               |---->|------>|--------->|
+    ///
+    /// let mut data_grouped = Vec::new();
+    /// // Note: group is an iterator
+    /// for (key, group) in data.into_iter().group_by(|val| *val >= 0) {
+    ///     data_grouped.push((key, group.collect()));
+    /// }
+    /// assert_eq!(data_grouped, vec![(true, vec![1, 3]), (false, vec![-2, -2]), (true, vec![1, 0, 1, 2])]);
+    /// ```
+    fn group_by<K, F>(self, key: F) -> GroupByIterator<Self, F, K>
+    where
+        Self: Sized,
+        F: FnMut(&Self::Item) -> K,
+        K: PartialEq + Copy,
+        Self::Item: Copy,
+    {
+        GroupByIterator::new(self, key)
+    }
+}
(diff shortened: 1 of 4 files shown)


common/src/group_by.rs
```diff
@@ -0,0 +1,166 @@
...
+    F: FnMut(&I::Item) -> K,
+{
+    // I really would like to avoid the Rc<RefCell>, but the Iterator is shared between
+    // `GroupByIterator` and `GroupIter`. In practice they are used consecutive and
+    // `GroupByIter` is finished before calling next on `GroupByIterator`. I'm not sure there
+    // is a solution with lifetimes for that, because we would need to enforce it in the usage
+    // somehow.
+    //
+    // One potential solution would be to replace the iterator approach with something similar.
+    inner: Rc<RefCell<GroupByShared<I, F, K>>>,
```
fulmicoton, line 50: This might be a use case for GAT + `LendingIterator` maybe? But that's not an `Iterator` unfortunately.

fulmicoton (approved):
Let's merge this for the moment but I suspect there is a problem in the API.

My understanding is that this function (the id conversion thing) is meant to be used in the range queries. The current Column trait get_range is not working as an iterator.

If we focus on catering to this usage, we do not need an `Iterator`.
The group by logic can be implemented as loops.

It will probably unlock `more` from your group_by optimisation:
the Iterator won't be boxed anymore , and within a given block, the
loop will be doing static dispatch.

PSeitz:
Yes this is for range queries, but for that we call:
```    
pub fn get_docids_for_value_range(
    &self,
    value_range: RangeInclusive<T>,
    doc_id_range: Range<u32>,
    positions: &mut Vec<u32>,
) {
```
So we need to translate the `positions` from the codex index to orig index. We could replace the iterator and pass the positions directly. That will remove the dispatch issue. We could also replace the group_by, since we work on a slice then.

fulmicoton:
@PSeitz yes. let's merge this for the moment.

PR #1749, opened 2022-12-29
refactor multivalue fastfield, refactor range query

PSeitz:
Introduce MakeZero trait, remove make_zero from FastValue
Merge two multivalue fastfield implementations into one
prepare range query on fastfield for different types

### DIFF
--- a/src/query/range_query/fast_field_range_query.rs
+++ b/src/query/range_query/fast_field_range_query.rs
@@ -1,0 +1,209 @@
+use std::ops::RangeInclusive;
+use std::sync::Arc;
+
+use fastfield_codecs::Column;
+
+use crate::fastfield::{MakeZero, MultiValuedFastFieldReader};
+use crate::{DocId, DocSet, TERMINATED};
+
+/// Helper to have a cursor over a vec of docids
+struct VecCursor {
+    docs: Vec<u32>,
+    current_pos: usize,
+}
+impl VecCursor {
+    fn new() -> Self {
+        Self {
+            docs: Vec::with_capacity(32),
+            current_pos: 0,
+        }
+    }
+    fn next(&mut self) -> Option<u32> {
+        self.current_pos += 1;
+        self.current()
+    }
+    #[inline]
+    fn current(&self) -> Option<u32> {
+        self.docs.get(self.current_pos).copied()
+    }
+    fn get_cleared_data(&mut self) -> &mut Vec<u32> {
+        self.docs.clear();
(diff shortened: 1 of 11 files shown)


src/query/range_query/fast_field_range_query.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+    fn next(&mut self) -> Option<u32> {
+        self.current_pos += 1;
+        self.current()
+    }
+    #[inline]
+    fn current(&self) -> Option<u32> {
+        self.docs.get(self.current_pos).copied()
+    }
+    fn get_cleared_data(&mut self) -> &mut Vec<u32> {
```
fulmicoton, line 29: That's a very strange API.
I'm not sure where you are going with this.

Could we use `Into<Vec<u32>>` instead?

src/query/range_query/fast_field_range_query.rs
```diff
@@ -0,0 +1,212 @@
...
+        self.docs.clear();
+        self.current_pos = 0;
+        &mut self.docs
+    }
+    fn last_value(&self) -> Option<u32> {
+        self.docs.iter().last().cloned()
+    }
+    fn is_empty(&self) -> bool {
+        self.current_pos >= self.docs.len()
```
fulmicoton: alternatively `self.current.is_none()`...
there are pros and cons to both.

src/query/range_query/fast_field_range_query.rs
```diff
@@ -0,0 +1,212 @@
...
+        while doc < target {
+            doc = self.advance();
+        }
+        self.last_seek_pos_opt = Some(target);
+        doc
+    }
+
+    fn size_hint(&self) -> u32 {
+        0 // heuristic possible by checking number of hits when fetching a block
+    }
```
fulmicoton, line 208: can you add unit tests on this module? I think there are a lot of tricky logic in this

fulmicoton (approved):

src/query/range_query/fast_field_range_query.rs
PSeitz: I like yours better

src/query/range_query/fast_field_range_query.rs
PSeitz: There are tests in `range_query_ip_fastfield.rs`. I will add another PR that enables range queries on fastfield for all types. I'll try to cover all branches with tests there

PR #1750, opened 2022-12-31
Enable usage of FuzzyTermQuery for specific fields via QueryParser

adamreichold:
Following the lead of `QueryParser::set_field_boost`, this adds `QueryParser::set_field_fuzzy` which sets whether literals will be matched against the given field using `FuzzyTermQuery` instead of `TermQuery` using the given parameters `prefix`, `distance` and `transposition_cost_one`.

My personal use case for this is a single category-like `STRING` field which I would like to include in default set but as I expect users to make spelling mistakes or use prefixes instead of the full categories.

### DIFF
--- a/src/query/query_parser/query_parser.rs
+++ b/src/query/query_parser/query_parser.rs
@@ -1,9 +1,9 @@
-use std::collections::HashMap;
 use std::net::{AddrParseError, IpAddr};
 use std::num::{ParseFloatError, ParseIntError};
 use std::ops::Bound;
 use std::str::{FromStr, ParseBoolError};
 
+use rustc_hash::FxHashMap;
 use tantivy_query_grammar::{UserInputAst, UserInputBound, UserInputLeaf, UserInputLiteral};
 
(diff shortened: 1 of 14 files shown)


src/query/query_parser/query_parser.rs
```diff
@@ -182,7 +189,8 @@ pub struct QueryParser {
     default_fields: Vec<Field>,
     conjunction_by_default: bool,
     tokenizer_manager: TokenizerManager,
-    boost: HashMap<Field, Score>,
+    boost: FxHashMap<Field, Score>,
+    fuzzy: FxHashMap<Field, (bool, u8, bool)>,
```
PSeitz: Can you create a struct for the (bool, u8, bool) tuple? With two bools it's easy to confuse them

adamreichold: Done.

fulmicoton (approved):

fulmicoton:
Thank you @adamreichold

PR #1755, opened 2023-01-03
Bump dependency on lru to from version 0.7.5 to version 0.9.0.

adamreichold:


fulmicoton:
Thank you!

fulmicoton:
Thank you! @adamreichold.
@evanxg852000 can you review?

evanxg852000 (dismissed):

src/store/reader.rs
```diff
@@ -33,23 +34,29 @@ pub struct StoreReader {
 
 /// The cache for decompressed blocks.
 struct BlockCache {
-    cache: Mutex<LruCache<usize, Block>>,
-    cache_hits: Arc<AtomicUsize>,
-    cache_misses: Arc<AtomicUsize>,
+    cache: Option<Mutex<LruCache<usize, Block>>>,
```
evanxg852000, line 37: I believe this option is just for being able to use `NonZeroUsize`. I think it's simpler to stick to `cache: Mutex<LruCache<usize, Block>>`. If we want to forbid cache size of zero, an assert can be appropriate.

src/store/reader.rs
```diff
@@ -33,23 +34,29 @@ pub struct StoreReader {
 
 /// The cache for decompressed blocks.
 struct BlockCache {
-    cache: Mutex<LruCache<usize, Block>>,
-    cache_hits: Arc<AtomicUsize>,
-    cache_misses: Arc<AtomicUsize>,
```
evanxg852000, line 38: Nice :+1:

evanxg852000 (commented):
Nice work :+1:
Please can you avoid wrapping the cache in Option, or let me know if there are other reasons apart from using `NonZeroUsize`

src/store/reader.rs
adamreichold: I actually started by propagating the use of `NonZeroUsize` through to the code base, but there are places where, indirectly, zero-size caches are created. To continue to support this, I chose the `Option<..>` layer which is not so much about using `NonZeroUsize` but more about supporting zero-sized caches which `lru` does just not support any more.

Personally, I see two viable options:
* Using `Option<..>` to continue supporting zero-sized caches.
* Propagate the `usize -> NonZeroUsize` type change throughout the API and change any callers to use at least cache size one.

Just asserting a non-zero cache might lead to hard to diagnose panics IMHO as there are multiple layers of indirection between constructing readers, writers, etc. and the usage of `LruCache`.

evanxg852000: Ok, to keep the change in scope let's stick with Option<...> then.

evanxg852000 (approved):

src/store/reader.rs
adamreichold: I have the second alternative ready in [a branch](https://github.com/adamreichold/tantivy/tree/disallow-zero-size-block-cache). So if you want that, I can propose that as a follow-up PR since I do agree that having the local change first is much easier to review.

PR #1756, opened 2023-01-03
Make construction of LevenshteinAutomatonBuilder for FuzzyTermQuery instances fully lazy.

adamreichold:


src/query/fuzzy_query.rs
```diff
@@ -129,10 +112,15 @@ impl FuzzyTermQuery {
...
 
     fn specialized_weight(&self) -> crate::Result<AutomatonWeight<DfaWrapper>> {
-        // LEV_BUILDER is a HashMap, whose `get` method returns an Option
-        match LEV_BUILDER.get(&(self.distance, self.transposition_cost_one)) {
-            // Unwrap the option and build the Ok(AutomatonWeight)
-            Some(automaton_builder) => {
+        static BUILDER: Lazy<RwLock<FxHashMap<(u8, bool), LevenshteinAutomatonBuilder>>> =
+            Lazy::new(Default::default);
+
+        loop {
```
ChillFish8: Do we really want a RwLock here? Although this makes the logic lazy, it adds additional overhead to every read we try to create the scorer, and I believe the standard library RwLocks still have some performance issues plaguing them when being accessed concurrently by several threads even when reading only, and this is potentially a very hot path.

Secondly, this removes the cap on the maximum edit distance per term, which although isn't bad in itself, does open the potential for people to pick a number *too* high, 3 tends to be the realistic maximum you'd use, although it depends on the word length.

adamreichold: > Do we really want a RwLock here? Although this makes the logic lazy, it adds additional overhead to every read we try to create the scorer, and I believe the standard library RwLocks still have some performance issues plaguing them when being accessed concurrently by several threads even when reading only, and this is potentially a very hot path.

Yes, I am somewhat unhappy with the `RwLock` too, but some form of synchronization will have to be necessary if laziness is desired.

Alternatives I see are:
* Using `Mutex` which should be faster but would prevent concurrent reads after the relevant builders have been initialized. Alternatively, `Arc<LevenshteinAutomatonBuilder>` could be cached and cloned under the mutex so that usage can proceed without it.
* But if we do go for the additional indirection of `Arc<LevenshteinAutomatonBuilder>`, we might also use `ArcSwap` instead of `RwLock` (already a dependency) and always publish a new (larger) hash table when another `LevenshteinAutomatonBuilder` had to be initialized containing copies of all the previous `Arc<LevenshteinAutomatonBuilder>` instances. (`LevenshteinAutomatonBuilder` itself is not clone, but I guess that could also be changed upstream to avoid the additional `Arc` indirection in this case.)

> Secondly, this removes the cap on the maximum edit distance per term, which although isn't bad in itself, does open the potential for people to pick a number too high, 3 tends to be the realistic maximum you'd use, although it depends on the word length.

The limitation seems artificial IMHO and related to the eager construction of the cached `LevenshteinAutomatonBuilder` instances which is why I removed it after making things lazy.

I have also read about 3 being a realistic limit, but I think this would better be served by a soft limit, e.g. advice given in the documentation of the `new` and `new_prefix` methods.

adamreichold: Opened https://github.com/quickwit-oss/levenshtein-automata/pull/13 to enable cloning of the builders.

fulmicoton, line 118: I understand what you are doing with the loop but this deserves a comment.

src/query/fuzzy_query.rs
```diff
@@ -34,22 +33,6 @@ impl Automaton for DfaWrapper {
     }
 }
 
-/// A range of Levenshtein distances that we will build DFAs for our terms
-/// The computation is exponential, so best keep it to low single digits
-const VALID_LEVENSHTEIN_DISTANCE_RANGE: Range<u8> = 0..3;
```
fulmicoton, line 39: I don't want to let user build arbitrary distance. Above 5, all computer will catch fire.
We still need a guard and return an error. I think <3 is a reasonable limit.

src/query/fuzzy_query.rs
fulmicoton: We could have a hashmap (or arrays) with once_cell as value? That would remove the RwLock , the loop, and the possible expensive multiple levenshtein buider building upon race condition.

adamreichold: > We could have a hashmap (or arrays) with once_cell as value? 

We could lazily initialize only the entries and eagerly construct the data structure itself, but if we keep the static limit on the distance, then the current approach - lazily constructing all builders at once - seems to most reasonable to me and I would just remove the TODO.

adamreichold: Pushed an implementation of that approach, i.e. using `[[OnceCell<LevenshteinAutomatonBuilder>; 2]; 3]`.

src/query/fuzzy_query.rs
```diff
@@ -129,30 +110,39 @@ impl FuzzyTermQuery {
...
                 ))
-            }
-            None => Err(InvalidArgument(format!(
-                "Levenshtein distance of {} is not allowed. Choose a value in the {:?} range",
-                self.distance, VALID_LEVENSHTEIN_DISTANCE_RANGE
-            ))),
-        }
+            })?
+            .get(self.transposition_cost_one as usize)
+            .unwrap()
```
fulmicoton, line 129:
```suggestion
            [self.transposition_cost_one as usize]
```

fulmicoton: would not that be the same?

fulmicoton (approved):

src/query/fuzzy_query.rs
adamreichold: It would certainly be the same, but it formats very unfavorably:

```rust
let automaton_builder = AUTOMATON_BUILDER
    .get(self.distance as usize)
    .ok_or_else(|| {
        InvalidArgument(format!(
            "Levenshtein distance of {} is not allowed. Choose a value less than {}",
            self.distance,
            AUTOMATON_BUILDER.len()
        ))
    })?[self.transposition_cost_one as usize]
    .get_or_init(|| {
        LevenshteinAutomatonBuilder::new(self.distance, self.transposition_cost_one)
    });
```

which is why I went for this somewhat verbose way of doing it. But I am not really invested in this.

Which one do you prefer after seeing the formatting, `[..]` or `get(..).unwrap()`?

PR #1758, opened 2023-01-04
Implement RegexTokenizer

### LINKED ISSUE #1670: Add a configurable RegexTokenizer
The regex pattern woudl be a parameter of the tokenizer.
The tokenizer would "search" the regex and emit one token per match.

Gearme:
Implementation of a RegexTokenizer for #1670.

To make this a decently flexible solution I implemented support for capture groups as well as simple find matches: The RegexTokenStream will check on construction if there are any capture groups (besides the entire match) and use either `captures_read_at` for capture support or `find_at` for simple matches. Tests and examples included.

Closes https://github.com/quickwit-oss/tantivy/issues/1670

### DIFF
--- a/src/tokenizer/regex_tokenizer.rs
+++ b/src/tokenizer/regex_tokenizer.rs
@@ -9,9 +9,9 @@ use super::{BoxTokenStream, Token, TokenStream, Tokenizer};
 /// the entire matched expression as well (as capture group 0). However, since there is no way to
 /// filter these out later, the `RegexTokenizer` discards them itself. If matching of the entire
 /// expression is desired, be sure to include a capture group around the entire expression.
-/// 
+///
 /// # Simple example (no capture groups)
-/// 
+///
 /// ```rust
 /// use tantivy::tokenizer::*;
 ///
@@ -30,9 +30,8 @@ use super::{BoxTokenStream, Token, TokenStream, Tokenizer};
 ///     assert_eq!(token.offset_to, 13);
 /// }
 /// assert!(stream.next().is_none());
-/// 
 /// ```
-/// 
+///
 /// # Capture groups example
 ///
 /// ```rust
--- a/src/tokenizer/regex_tokenizer.rs
+++ b/src/tokenizer/regex_tokenizer.rs
@@ -1,0 +1,294 @@
+use regex::{CaptureLocations, Error, Match, Regex};
+
+use super::{BoxTokenStream, Token, TokenStream, Tokenizer};
+
+/// Tokenize the text by splitting based on the captures of a regular expression:
+/// - if no capture groups are specified, all matches of the entire expression will be tokenized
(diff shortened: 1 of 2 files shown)


Gearme:
I found some edge-cases in regex patterns which need addressing, converting back to draft.

fulmicoton:
@Gearme @mkleen any consensus on which PR we should progress on to land?
I think groups can be useful.
 
@PSeitz can you take over and make sure we end up merging something?

src/tokenizer/regex_tokenizer.rs
```diff
@@ -0,0 +1,293 @@
...
+impl<'a> RegexTokenStream<'a> {
+    fn new(text: &'a str, regex: &Regex) -> RegexTokenStream<'a> {
+        let capture_locations = if regex.captures_len() > 1 {
+            Some(regex.capture_locations())
+        } else {
+            None
+        };
+        RegexTokenStream {
+            text,
+            regex: regex.clone(),
```
PSeitz, line 129: This is called for every text. Arc may be preferable to `clone`

src/tokenizer/regex_tokenizer.rs
```diff
@@ -0,0 +1,293 @@
...
+                true
+            })
+    }
+}
+
+impl<'a> TokenStream for RegexTokenStream<'a> {
+    fn advance(&mut self) -> bool {
+        let tail_index = self.regex_match.map_or(0, |regex_match| regex_match.end());
+        if self.capture_locations.is_some() {
+            while !self.next_capture_group() {
```
PSeitz, line 196: `if let Some()` would be preferable to `is_some` and `unwrap`

src/tokenizer/regex_tokenizer.rs
adamreichold: See also https://github.com/quickwit-oss/tantivy/pull/1759#discussion_r1061421507

adamreichold:
@PSeitz @fulmicoton What do you think about merging #1759 first for the basic implementation and then rebasing this PR on top of it to add support for capture groups? This way, everyone gets kudos and we should end up with two easy to review PR.

Gearme:
@adamreichold Sounds good to me!

fulmicoton:
@adamreichold done sorry for the delay.

PSeitz:
@Gearme The other PR is merged now

fulmicoton:
What do we do with this PR? @Gearme @PSeitz

Gearme:
I'm afraid I've been rather busy recently, I'll try and get to finishing and proofing this feature as soon as possible.

mkleen:
> I found some edge-cases in regex patterns which need addressing, converting back to draft.

@Gearme Could you maybe add some of the edge cases to the tests? Then I could take over if it's ok for you.

fulmicoton:
Closing this PR as it seems stalled.

PR #1759, opened 2023-01-04
Add regex tokenizer

### LINKED ISSUE #1670: Add a configurable RegexTokenizer

mkleen:
This adds a regex tokenizer which tokenizes the text by using a regex pattern to split. 

Closes https://github.com/quickwit-oss/tantivy/issues/1670

### DIFF
--- a/src/tokenizer/regex_tokenizer.rs
+++ b/src/tokenizer/regex_tokenizer.rs
@@ -1,0 +1,149 @@
+use regex::Regex;
+use crate::TantivyError;
+
+use super::{BoxTokenStream, Token, TokenStream, Tokenizer};
+
+/// Tokenize the text by using a regex pattern to split.
+/// Each match of the regex emits a distinct token, empty tokens will not be emitted. Anchors such
+/// as `\A` will match the text from the part where the last token was emitted or the beginning of
+/// the complete text if no token was emitted yet.
+///
+/// Example: `` 'aaa' bbb 'ccc' 'ddd' `` with the pattern `` '(?:\w*)' `` will be tokenized as followed:
+///
+/// | Term     | aaa  | ccc    | ddd   |
+/// |----------|------|--------|-------|
+/// | Position | 1    | 2      | 3     |
+/// | Offsets  |0,5   | 10,15  | 16,21 |
+///
+///
+/// # Example
+///
+/// ```rust
+/// use tantivy::tokenizer::*;
+///
+/// let tokenizer = RegexTokenizer::new(r"'(?:\w*)'");
+/// let mut stream = tokenizer.token_stream("'aaa' bbb 'ccc' 'ddd'");
+/// {
+///     let token = stream.next().unwrap();
+///     assert_eq!(token.text, "'aaa'");
+///     assert_eq!(token.offset_from, 0);
+///     assert_eq!(token.offset_to, 5);
+/// }
+/// {
+///   let token = stream.next().unwrap();
+///     assert_eq!(token.text, "'ccc'");
+///     assert_eq!(token.offset_from, 10);
+///     assert_eq!(token.offset_to, 15);
+/// }
+/// {
+///   let token = stream.next().unwrap();
+///     assert_eq!(token.text, "'ddd'");
+///     assert_eq!(token.offset_from, 16);
+///     assert_eq!(token.offset_to, 21);
+/// }
+/// assert!(stream.next().is_none());
+/// ```
+
+#[derive(Clone)]
+pub struct RegexTokenizer {
+    regex: Regex
+}
+
+impl RegexTokenizer {
+    pub fn new(regex_pattern: &str) -> crate::Result<RegexTokenizer> {
+        Regex::new(regex_pattern)
+            .map_err(|_| TantivyError::InvalidArgument(regex_pattern.to_owned()))
+            .map(|regex| Self { regex })
+    }
+}
+
+impl Tokenizer for RegexTokenizer {
+    fn token_stream<'a>(&self, text: &'a str) -> BoxTokenStream<'a> {
+        BoxTokenStream::from(RegexTokenStream {
+            regex: self.regex.clone(),
+            text,
+            token: Token::default(),
+        })
+    }
+}
+
+pub struct RegexTokenStream<'a> {
+    regex: Regex,
+    text: &'a str,
+    token: Token,
+}
+
+impl<'a> TokenStream for RegexTokenStream<'a> {
+    fn advance(&mut self) -> bool {
+        if let Some(m) = self.regex.find(self.text) {
+            if !m.as_str().is_empty() {
+                self.token.text.clear();
+                self.token.text.push_str(&self.text[m.start()..m.end()]);
+
+                self.token.offset_from = self.token.offset_to + m.start();
+                self.token.offset_to = self.token.offset_to + m.end();
+
+                self.token.position = self.token.position.wrapping_add(1);
+
+                self.text = &self.text[m.end()..];
+                return true
+            }
+        }
+        false
+    }
+
+    fn token(&self) -> &Token {
+        &self.token
+    }
+
+    fn token_mut(&mut self) -> &mut Token {
+        &mut self.token
+    }
+}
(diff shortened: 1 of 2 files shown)


src/tokenizer/regex_tokenizer.rs
```diff
@@ -0,0 +1,149 @@
...
+        return Regex::new(regex_pattern)
+            .map_err(|_| TantivyError::InvalidArgument(regex_pattern.to_string()))
+            .map(|p| RegexTokenizer { regex: p });
+    }
+}
+
+impl Tokenizer for RegexTokenizer {
+    fn token_stream<'a>(&self, text: &'a str) -> BoxTokenStream<'a> {
+        BoxTokenStream::from(RegexTokenStream {
+            regex: self.regex.clone(),
```
mkleen, line 63: The regex is currently cloned here, this is inefficient. However, the regex should be only compiled once per Regex Tokenizer, so I think it needs to be passed down. I think this would need another lifetime annotation for the regex, which would require to modify BoxTokenStream. I s there a better way to do this ?

Gearme:
Wow, snap! I just authored #1759 to do much the same, albeit with capture group support.
Didn't find a better solution for the cloned Regex either though ;)

src/tokenizer/regex_tokenizer.rs
```diff
@@ -0,0 +1,149 @@
...
+/// assert!(stream.next().is_none());
+/// ```
+
+#[derive(Clone)]
+pub struct RegexTokenizer {
+    regex: Regex
+}
+
+impl RegexTokenizer {
+    pub fn new(regex_pattern: &str) -> crate::Result<RegexTokenizer> {
```
mkleen, line 53: Having a peek at https://github.com/apache/lucene/blob/main/lucene/analysis/common/src/java/org/apache/lucene/analysis/pattern/PatternTokenizer.java i realized they also allow to configure a group in addition to the regex which seems handy. Maybe it would make sense to also have that.

mkleen:
> Wow, snap! I just authored #1759 to do much the same, albeit with capture group support. Didn't find a better solution for the cloned Regex either though ;)

Haha, great. If you have capture group support then you have probably the better version.

src/tokenizer/regex_tokenizer.rs
adamreichold: You could store `Arc<Regex>` everywhere and clone that. This is done by e.g. the `StopWordFilter` and `SplitCompoundWords` filters.

But I think this is a non-obvious trade-off: `Regex` is internally an `Arc<..>` of a read-only part itself and the second part is a pool of cache memory for the various search algorithms. So cloning the `Regex` itself might actually fit this use case pretty well, c.f. https://github.com/rust-lang/regex/blob/master/PERFORMANCE.md#using-a-regex-from-multiple-threads

src/tokenizer/regex_tokenizer.rs
```diff
@@ -0,0 +1,149 @@
...
+            if !m.as_str().is_empty() {
+                self.token.text.clear();
+                self.token.position = self.token.position.wrapping_add(1);
+                self.token.text.push_str(&self.text[m.start()..m.end()]);
+                self.token.offset_from = m.start();
+                self.token.offset_to = m.end();
+                return true
+            }
+        }
+        return false;
```
fulmicoton: nitpick...
```suggestion
        false
```

src/tokenizer/regex_tokenizer.rs
```diff
@@ -0,0 +1,144 @@
...
+/// | Position | 1    | 2      | 3     |
+/// | Offsets  |0,5   | 10,15  | 16,21 |
+///
+///
+/// # Example
+///
+/// ```rust
+/// use tantivy::tokenizer::*;
+///
+/// let tokenizer = RegexTokenizer::new(r"'(\w*)'");
```
adamreichold: I think it would be nice if this used a non-capturing group, e.g. `(?:\w*)`, as this can have better performance and the capture group is never used.

src/tokenizer/regex_tokenizer.rs
adamreichold: While I do agree, I also think it would be nicer to do this as a follow-up PR to keep this one simpler.

src/tokenizer/regex_tokenizer.rs
```diff
@@ -0,0 +1,144 @@
...
+/// ```
+
+#[derive(Clone)]
+pub struct RegexTokenizer {
+    regex: Regex
+}
+
+impl RegexTokenizer {
+    pub fn new(regex_pattern: &str) -> crate::Result<RegexTokenizer> {
+        return Regex::new(regex_pattern)
```
adamreichold: I don't think the `return` is necessary here. I would also suggest using `.to_owned()` instead of `.to_string()` (you are making copy, not formatting something) and a slight rename of the temporary binding:

```rust
pub fn new(regex_pattern: &str) -> Self {
    Regex::new(regex_pattern)
        .map_err(|_| TantivyError::InvalidArgument(regex_pattern.to_owned()))
        .map(|regex| Self { regex })
}
```

src/tokenizer/regex_tokenizer.rs
```diff
@@ -0,0 +1,144 @@
...
+
+pub struct RegexTokenStream<'a> {
+    regex: Regex,
+    text: &'a str,
+    token: Token,
+}
+
+impl<'a> TokenStream for RegexTokenStream<'a> {
+    fn advance(&mut self) -> bool {
+        if let Some(m) = self.regex.find_at(self.text, self.token.offset_to) {
```
adamreichold: Reading [its documentation](https://docs.rs/regex/latest/regex/struct.Regex.html#method.find_at), I am not sure whether we want `find_at`. It depends on how anchors like `\A` are supposed to work: Should they reference the start of the whole text or just the part of the text after the last token was matched?

For the latter, I would recommend just advancing `self.text`, e.g.

```rust
[..]
self.token.offset_to = m.end();

self.text = &self.text[m.end()..];
```

In any case, the meaning of the anchors should be documented in the doc comment of `RegexTokenizer`.

src/tokenizer/regex_tokenizer.rs
```diff
@@ -0,0 +1,144 @@
...
+            if !m.as_str().is_empty() {
+                self.token.text.clear();
```
adamreichold, line 80: For readability, I would suggest grouping this, e.g.

```rust
self.token.text.clear();
self.token.text.push_str(&self.text[m.start()..m.end()]);

self.token.offset_from = m.start();
self.token.offset_to = m.end();

self.token.position = self.token.position.wrapping_add(1);
```

adamreichold (commented):
I think this would work. Some nits on the formatting, one concern about the intended meaning of regular expression anchors.

I would suggest doing capture group support as a follow-up.

I think the performance question of whether to clone `Regex` or `Arc<Regex>` can only be solved via benchmarks. I would suggest keeping it as it is for now if only for simplicity until such benchmarks are written.

mkleen:
@adamreichold Thank you for your feedback and the review. I like your idea of first merging this simpler version and then going ahead and adding the group capture support afterwards. I can wrap this up pretty quickly. @Gearme would that be OK for you as well? We could make the follow-up based on your implementation. I would also like to try out a benchmark to address https://github.com/quickwit-oss/tantivy/pull/1759#pullrequestreview-1238542538, but also as a follow-up.

mkleen:
Ups, @PSeitz removing you from review was not intentionally.

mkleen:
Sorry for requesting multiple reviews. I did not realize I can only request one review at the time.

adamreichold (approved):

mkleen:
I just discovered there is a off-by-one on the offsets of the second token. I will address that.

mkleen:
> I just discovered there is a off-by-one on the offsets of the second token. I will address that.

I think this is fine, wrong alarm.

fulmicoton (approved):

PR #1760, opened 2023-01-05
handle user input on get_docid_for_value_range

### LINKED ISSUE #1757: `get_docids_for_value_range` is broken w/certain fast fields that use a GCD inverse
**Describe the bug**

When inserting fast fields in a certain order, segments will occasionally end up with a column that matches extra documents when calling `get_docids_for_value_range` using nonexistant values.

The testcase at the bottom of this report illustrates a failing case, though you need to run it repeatedly to cause it to fail.

Expected behaviour: passing fastfield values/ranges to `get_docids_for_value_range` that cannot possibly exist in a given segment reader should return no matching documents.

0.19.0

**To Reproduce**

This code will reproduce the bug, but it won't happen every time. Running it repeatedly will eventually give you a `SegmentReader` with the min/max values `[6341073085727221 3373930417471920086]`, and that `SegmentReader` will always match document zero when querying in the invalid value `7999999999999999`.

```

    #[test]
    fn test_gcd_bug() {
        let dir = RamDirectory::create();
        let url_norm_hash_field = schema_builder.add_i64_field("url_norm_hash", FAST | INDEXED);  
        let index = Index::create(dir, schema, IndexSettings::default()).unwrap();
        let mut writer = index.writer(50_000_000).unwrap();
        for i in [6341073085727221_i64, 6341073085727221_i64, 6341073085727221_i64, 6341073085727221_i64, 6341073085727221_i64, 6341073085727221_i64, 3373930417471920086_i64, 3373930417471920086_i64, 3373930417471920086_i64, 3373930417471920086_i64] {
            writer.add_document(doc! {
                url_norm_hash_field => i,
            }).unwrap();
        }
        writer.commit();
        for segment in searcher.segment_readers() {
            let field = segment.fast_fields().i64(url_norm_hash_field).unwrap();
            let (min, max) = (field.min_value(), field.max_value());
            println!("{} {}", min, max);
            if 7999999999999999_i64 >= min && 7999999999999999_i64 <= max {
                let mut vec = vec![];
                field.get_docids_for_value_range(7999999999999999_i64..=7999999999999999_i64, 0..u32::MAX, &mut vec);
                println!("{:?}", vec);
            }
        }
    }
```

PSeitz:
fixes #1757

### DIFF
--- a/fastfield_codecs/src/column.rs
+++ b/fastfield_codecs/src/column.rs
@@ -1,3 +1,4 @@
+use std::fmt::{self, Debug};
 use std::marker::PhantomData;
 use std::ops::{Range, RangeInclusive};
 
@@ -6,7 +7,7 @@ use tantivy_bitpacker::minmax;
 use crate::monotonic_mapping::StrictlyMonotonicFn;
 
 /// `Column` provides columnar access on a field.
-pub trait Column<T: PartialOrd = u64>: Send + Sync {
+pub trait Column<T: PartialOrd + Debug = u64>: Send + Sync {
     /// Return the value associated with the given idx.
     ///
     /// This accessor should return as fast as possible.
@@ -83,7 +84,7 @@ pub struct VecColumn<'a, T = u64> {
     max_value: T,
 }
 
-impl<'a, C: Column<T>, T: Copy + PartialOrd> Column<T> for &'a C {
+impl<'a, C: Column<T>, T: Copy + PartialOrd + fmt::Debug> Column<T> for &'a C {
     fn get_val(&self, idx: u32) -> T {
         (*self).get_val(idx)
     }
@@ -109,7 +110,7 @@ impl<'a, C: Column<T>, T: Copy + PartialOrd> Column<T> for &'a C {
     }
 }
 
-impl<'a, T: Copy + PartialOrd + Send + Sync> Column<T> for VecColumn<'a, T> {
+impl<'a, T: Copy + PartialOrd + Send + Sync + Debug> Column<T> for VecColumn<'a, T> {
     fn get_val(&self, position: u32) -> T {
         self.values[position as usize]
     }
@@ -177,8 +178,8 @@ pub fn monotonic_map_column<C, T, Input, Output>(
 where
     C: Column<Input>,
     T: StrictlyMonotonicFn<Input, Output> + Send + Sync,
-    Input: PartialOrd + Send + Sync + Clone,
-    Output: PartialOrd + Send + Sync + Clone,
+    Input: PartialOrd + Send + Sync + Clone + Debug,
+    Output: PartialOrd + Send + Sync + Clone + Debug,
 {
     MonotonicMappingColumn {
         from_column,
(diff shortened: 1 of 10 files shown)


fastfield_codecs/src/column.rs
```diff
@@ -228,12 +229,26 @@ where
         doc_id_range: Range<u32>,
         positions: &mut Vec<u32>,
     ) {
-        self.from_column.get_docids_for_value_range(
-            self.monotonic_mapping.inverse(range.start().clone())
-                ..=self.monotonic_mapping.inverse(range.end().clone()),
-            doc_id_range,
-            positions,
-        )
+        if range.start() > &self.max_value() || range.end() < &self.min_value() {
```
fulmicoton, line 232:
```suggestion
        if range.start > self.max_value() || range.end < self.min_value() {
```

fastfield_codecs/src/monotonic_mapping.rs
```diff
@@ -34,6 +37,24 @@ pub trait StrictlyMonotonicFn<External, Internal> {
...
+    /// It may be necessary to coerce the value if it is outside the value space.
+    /// In that case it tries to find the next greater value in the value space.
+    ///
+    /// Returns a bool to mark if a value was outside the value space and had to be coerced _up_.
+    /// With that information we can detect if two values in a range both map outside the same value
+    /// space.
+    ///
+    /// coerce_up means the next valid upper value in the value space will be chosen if the value
+    /// has to be coerced.
+    fn mapping_coerce(&self, inp: External, _coerce_up: bool) -> (bool, Internal) {
```
fulmicoton: Would something like `.map_lower_than(&self, bound: Bound<To>) -> Bound<From>`  do the job?

If it would, I feel like the contract is more obvious.

fulmicoton (changes requested):
See comment inline

fastfield_codecs/src/monotonic_mapping.rs
PSeitz: We could pass a Range in `mapping_coerce` instead (we don't have bounds here anymore). We still need the early exit for correctness if both map in front of the value space, and for that return a 1..0 range.

fastfield_codecs/src/monotonic_mapping.rs
```diff
@@ -142,6 +176,27 @@ impl<External: MonotonicallyMappableToU64> StrictlyMonotonicFn<External, u64>
     fn inverse(&self, out: u64) -> External {
         External::from_u64(self.min_value + out * self.gcd)
     }
+
+    #[inline]
+    #[allow(clippy::reversed_empty_ranges)]
+    fn mapping_coerce(&self, inp: RangeInclusive<External>) -> RangeInclusive<u64> {
+        let end = External::to_u64(*inp.end());
+        if end < self.min_value {
+            return 1..=0;
```
fulmicoton, line 185: Can we return a `Option<>` and not rely on this trick?

fulmicoton: actually not the trick is good to go

fulmicoton (approved):

PR #1768, opened 2023-01-09
Implement min aggregation

guilload:
Just needed the `min` for something I'm working on in Quickwit. Will add `max` if this change looks ok.

### DIFF
--- a/src/aggregation/metric/min.rs
+++ b/src/aggregation/metric/min.rs
@@ -1,0 +1,171 @@
+use std::fmt::Debug;
+
+use fastfield_codecs::Column;
+use serde::{Deserialize, Serialize};
+
+use crate::aggregation::f64_from_fastfield_u64;
+use crate::schema::Type;
+use crate::DocId;
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+/// A single-value metric aggregation that computes the minimum of numeric values that are
+/// extracted from the aggregated documents.
+/// Supported field types are u64, i64, and f64.
+/// See [super::SingleMetricResult] for return value.
+///
+/// # JSON Format
+/// ```json
+/// {
+///     "min": {
+///         "field": "price",
+///     }
+/// }
+/// ```
+pub struct MinAggregation {
+    /// The field name to compute the minimum on.
+    pub field: String,
+}
+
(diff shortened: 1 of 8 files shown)


adamreichold:
Is there a measurable performance benefit from using this versus `Stats` and ignoring everything but the `min` field? If there is, could we deduplicate the implementation by adding a single aggregation using any associative function `Fn(T, T) -> T`?

PSeitz:
> Is there a measurable performance benefit from using this versus `Stats` and ignoring everything but the `min` field? If there is, could we deduplicate the implementation by adding a single aggregation using any associative function `Fn(T, T) -> T`?

Yes, that was also my first thought. There's a stats benchmark already, it should be easy to compare with `bench_aggregation_stats_f64`

fulmicoton:
There is a min aggregation in elastic. (apart from stat) 
That's enough reason to have one. :)

We could reuse the implementation though, if computing just the min is not faster. 
Is this how average is done today?

src/aggregation/metric/min.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub fn field_name(&self) -> &str {
+        &self.field
+    }
+}
+
+/// Holds the intermediate state of the minimum computation.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct IntermediateMin {
+    pub(crate) min: f64,
+    pub(crate) is_some: bool,
```
fulmicoton, line 45: `Option<f64>` ?

PSeitz:
> There is a min aggregation in elastic. (apart from stat) That's enough reason to have one. :)
> 
> We could reuse the implementation though, if computing just the min is not faster. Is this how average is done today?


Yes, just code reuse, that's what I meant. Average has its own code, since it was there first. It could be refactored though.
The bench looks pretty much the same

```
test aggregation::tests::bench::bench_aggregation_average_f64                                                            ... bench:   8,574,992 ns/iter (+/- 283,628)
test aggregation::tests::bench::bench_aggregation_stats_f64                                                              ... bench:   8,235,065 ns/iter (+/- 319,069)
```

fulmicoton:
@PSeitz let's put the code in common then

PR #1770, opened 2023-01-10
RFC: Completely remove the option of a zero-sized block cache

adamreichold:
This was discussed in https://github.com/quickwit-oss/tantivy/pull/1755#discussion_r1063241523 and would simplify the cache implementation at the cost of some scattered changes around the code base and the requirement to at least use cache size one everywhere.

### DIFF
--- a/src/core/searcher.rs
+++ b/src/core/searcher.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeMap;
+use std::num::NonZeroUsize;
 use std::sync::Arc;
 use std::{fmt, io};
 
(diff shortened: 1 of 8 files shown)


PSeitz:
I think the `NonZero` API is worse than before. Disabling the cache is a valid use case (e.g. in very resource constrained environments or for tests). I created a ticket here: https://github.com/jeromefroe/lru-rs/issues/165

adamreichold:
> I think the `NonZero` API is worse than before. Disabling the cache is a valid use case (e.g. in very resource constrained environments or for tests). I created a ticket here: [jeromefroe/lru-rs#165](https://github.com/jeromefroe/lru-rs/issues/165)

Yes, I agree. Let's drop this and remove the `Option` layer when `lru` is bumped to something supporting zero-capacity caches again.

PR #1772, opened 2023-01-10
start migrate Field to &str

PSeitz:
start migrate Field to &str in preparation of columnar
return Result for get_field

### DIFF
--- a/examples/custom_collector.rs
+++ b/examples/custom_collector.rs
@@ -14,7 +14,7 @@ use fastfield_codecs::Column;
 // Importing tantivy...
 use tantivy::collector::{Collector, SegmentCollector};
 use tantivy::query::QueryParser;
-use tantivy::schema::{Field, Schema, FAST, INDEXED, TEXT};
+use tantivy::schema::{Schema, FAST, INDEXED, TEXT};
 use tantivy::{doc, Index, Score, SegmentReader};
 
 #[derive(Default)]
@@ -52,11 +52,11 @@ impl Stats {
 }
 
 struct StatsCollector {
-    field: Field,
+    field: String,
 }
 
 impl StatsCollector {
-    fn with_field(field: Field) -> StatsCollector {
+    fn with_field(field: String) -> StatsCollector {
         StatsCollector { field }
     }
 }
@@ -73,7 +73,7 @@ impl Collector for StatsCollector {
         _segment_local_id: u32,
         segment_reader: &SegmentReader,
     ) -> tantivy::Result<StatsSegmentCollector> {
-        let fast_field_reader = segment_reader.fast_fields().u64(self.field)?;
+        let fast_field_reader = segment_reader.fast_fields().u64(&self.field)?;
         Ok(StatsSegmentCollector {
             fast_field_reader,
             stats: Stats::default(),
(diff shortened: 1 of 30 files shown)


fulmicoton:
Switching as draft. This kind of change is a one way street, so this requires a lot of more reflexion.

adamreichold:
I would also be interested if this has any performance impact `&str` is more costly to search/compare/store than `u64` (which `Field` currently wraps).

I also fear that this makes API usage much more awkward as downstream code might be forced to manage `Cow<'static, str>` instead of `Field`. ATM, I take care to convert field names as early as possible into `Field` at my module boundaries so that I can easily pass/store/clone the `Field` values in its implementation.

This is doubly worrying as neither the commit message nor the PR cover letter give any rationale as to why this change is made beyond "in preparation of columnar" so that it is hard to understand whether this is a fundamental requirement or whether there are options/alternatives that could be discussed.

PSeitz:
> I would also be interested if this has any performance impact `&str` is more costly to search/compare/store than `u64` (which `Field` currently wraps).
> 
> I also fear that this makes API usage much more awkward as downstream code might be forced to manage `Cow<'static, str>` instead of `Field`. ATM, I take care to convert field names as early as possible into `Field` at my module boundaries so that I can easily pass/store/clone the `Field` values in its implementation.
> 
> This is doubly worrying as neither the commit message nor the PR cover letter give any rationale as to why this change is made beyond "in preparation of columnar" so that it is hard to understand whether this is a fundamental requirement or whether there are options/alternatives that could be discussed.

Thanks for the feedback. The preparation which was mentioned is adding fastfield support for JSON fields.
We can't use `Field` for that use case and we want to keep the API consistent between JSON and non-JSON fastfields. The name of the field is a better identifier that can cover both use cases.
The typical workflow as of involves `&str` anyway. `Schema::get_field("name")`   => load fastfield via `Field`

As of performance impact, a single lookup in a Hashmap won't show in any performance profile. Even if there are hundreds of fastfields, the lookups will likely be dwarfed by operations on the fastfield.

fulmicoton:
The change has a lot of other benefits. It will evenutally make it possible to modify tantivy's schema

adamreichold:
> The preparation which was mentioned is adding fastfield support for JSON fields.

> The change has a lot of other benefits. It will evenutally make it possible to modify tantivy's schema

If I understand you correctly, this is basically a prerequisite for schema-less operation? Otherwise, the schema could for example contain paths into JSON fields instead of just names, couldn't it?

> The typical workflow as of involves &str anyway. Schema::get_field("name") => load fastfield via Field

I think the main ergonomic downside (that is admittedly very Rust-specific) is that borrowing `&str` "infects" code with lifetimes (`Cow<'static, str>` often being the compromise used to avoid that) whereas with `Field(u64)`, passing by value is even cheaper than passing by reference.

Looking at this from an efficiency angle again, since the number of distinct field names is low in schema-bound operation, I would probably end up deduplicating those strings using something `string_cache` which is just a generic and hence less efficient variant of what `Field` enables today. (For example, during indexing I currently do the above `&str` to `Field` conversion exactly once during start up.)

> As of performance impact, a single lookup in a Hashmap won't show in any performance profile. Even if there are hundreds of fastfields, the lookups will likely be dwarfed by operations on the fastfield.

In principle, I agree. There is such a thing "distributed fat" though, which individually is hard to measure, yet it doesn't preclude that it has a measurable cost globally. For example, a hash table look-up might increase instruction cache pressure compared which can result in slowdowns elsewhere. But of course, without benchmarks this is just FUD.

And in the end, you are not accountable to me in any way and I also do not want to hold back any changes necessary for Tantivy to grow. I would be glad if big changes like this one would come with detailed rationales so that an outsider like me can understand that this is either necessary to move forward at all or will have additional benefits that are worth the costs.

PR #1762, opened 2023-01-06
enable range query on fast field for u64 compatible types

PSeitz:

PR #1777, opened 2023-01-10
Allow range queries via fast fields on non-indexed fields

guilload:
Generalize the existing IP address field's logic to the datetime and u64 fields.

### DIFF
--- a/src/query/query_parser/query_parser.rs
+++ b/src/query/query_parser/query_parser.rs
@@ -13,6 +13,7 @@ use crate::core::Index;
 use crate::indexer::{
     convert_to_fast_value_and_get_term, set_string_and_get_terms, JsonTermWriter,
 };
+use crate::query::range_query::is_type_valid_for_fastfield_range_query;
 use crate::query::{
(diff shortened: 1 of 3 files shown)


src/query/query_parser/query_parser.rs
```diff
@@ -847,6 +846,10 @@ fn convert_to_query(fuzzy: &FxHashMap<Field, Fuzzy>, logical_ast: LogicalAst) ->
     }
 }
 
+fn supports_ff_range_queries(field_type: &FieldType) -> bool {
+    matches!(field_type, FieldType::Date(_) | FieldType::IpAddr(_) | FieldType::U64(_) if field_type.is_fast())
```
PSeitz: IP or any type that maps to u64 is supported.

You can reuse that method:
https://github.com/quickwit-oss/tantivy/pull/1762/files#diff-7041df7b5853bd1b4eb09be3f79a4fbe0ae29301c7d40e192c6589c95fd2c957R299

guilload: Thanks for the pointer. Fixed.

fulmicoton (approved):

PR #1780, opened 2023-01-11
EnableScoring::Disabled - optional Searcher

shikhar:
We have a custom `Query` that requires the `Searcher` in `weight(..)` (for looking up cached data used in matching that is kept warmed and keyed on searcher generation), even when scoring is disabled (e.g. using `DocSetCollector`)

So I am proposing to make a further API change on top of https://github.com/quickwit-oss/tantivy/pull/1667 such that `EnableScoring::Disabled` can contain an optional `Searcher`, and nudging towards providing it when available. 

`Searcher::search(..)` has been updated accordingly.

### DIFF
--- a/src/query/query.rs
+++ b/src/query/query.rs
@@ -15,8 +15,12 @@ pub enum EnableScoring<'a> {
     Enabled(&'a Searcher),
     /// Pass this to disable scoring.
     /// This can improve performance.
-    /// Searcher should be provided if available.
-    Disabled(&'a Schema, Option<&'a Searcher>),
+    Disabled {
+        /// Schema is required.
+        schema: &'a Schema,
+        /// Searcher should be provided if available.
+        searcher_opt: Option<&'a Searcher>,
+    },
 }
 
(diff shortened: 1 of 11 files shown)


src/query/query.rs
```diff
@@ -15,24 +15,45 @@ pub enum EnableScoring<'a> {
     Enabled(&'a Searcher),
     /// Pass this to disable scoring.
     /// This can improve performance.
-    Disabled(&'a Schema),
+    /// Searcher should be provided if available.
+    Disabled(&'a Schema, Option<&'a Searcher>),
```
fulmicoton: nitpick

```suggestion
    Disabled {
            schema: &'a Schema, 
            searcher_opt: Option<&'a Searcher>),
      }
```

fulmicoton (approved):
approved but please consider addressing nitpick comment

fulmicoton:
@shikhar you can hit the squash and merge button I think you have the permissions!

shikhar:
yup was just waiting on CI :)

PR #1785, opened 2023-01-13
reuse stats for average

PSeitz:


src/aggregation/metric/average.rs
```diff
@@ -36,61 +33,19 @@ impl AverageAggregation {
...
-        }
-    }
-}
-
 /// Contains mergeable version of average data.
 #[derive(Default, Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct IntermediateAverage {
     pub(crate) sum: f64,
-    pub(crate) doc_count: u64,
+    pub(crate) doc_count: u32,
```
adamreichold: Is there any inherent limit so that this is limited to 2^32 documents?

PSeitz: Actually that change is incorrect, it's only valid on a segment level

fulmicoton (approved):

PR #1793, opened 2023-01-16
modify getters for sstable metadata

trinity-1686a:
this adds support for limit when ranging over an sstable, and makes a few function read less blocks by leveraging sstable metadata more.

This is ready for review, however I'd like to add tests before merging to make sure no of-by-one slipped in

### DIFF
--- a/sstable/src/dictionary.rs
+++ b/sstable/src/dictionary.rs
@@ -301,3 +301,96 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::ops::Range;
+    use std::sync::{Arc, Mutex};
+
+    use common::OwnedBytes;
+
+    use super::Dictionary;
+    use crate::MonotonicU64SSTable;
+
+    #[derive(Debug)]
+    struct PermissionedHandle {
+        bytes: OwnedBytes,
+        allowed_range: Mutex<Range<usize>>,
+    }
+
+    impl PermissionedHandle {
+        fn new(bytes: Vec<u8>) -> Self {
+            let bytes = OwnedBytes::new(bytes);
+            PermissionedHandle {
+                allowed_range: Mutex::new(0..bytes.len()),
+                bytes,
+            }
+        }
+
+        fn restrict(&self, range: Range<usize>) {
+            *self.allowed_range.lock().unwrap() = range;
+        }
+    }
+
+    impl common::HasLen for PermissionedHandle {
+        fn len(&self) -> usize {
+            self.bytes.len()
+        }
+    }
+
+    impl common::file_slice::FileHandle for PermissionedHandle {
+        fn read_bytes(&self, range: Range<usize>) -> std::io::Result<OwnedBytes> {
+            let allowed_range = self.allowed_range.lock().unwrap();
+            if !allowed_range.contains(&range.start) || !allowed_range.contains(&(range.end - 1)) {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    format!("invalid range, allowed {allowed_range:?}, requested {range:?}"),
+                ));
+            }
+
+            Ok(self.bytes.slice(range))
+        }
+    }
+
+    fn make_test_sstable() -> (Dictionary<MonotonicU64SSTable>, Arc<PermissionedHandle>) {
+        let mut builder = Dictionary::<MonotonicU64SSTable>::builder(Vec::new()).unwrap();
+
+        let alphabet = 'A'..='Z';
+        let mut count = 0;
+        // this makes 450k keys, enough to fill multiple blocks.
+        for a in alphabet.clone() {
+            for b in alphabet.clone() {
+                for c in alphabet.clone() {
+                    for d in alphabet.clone() {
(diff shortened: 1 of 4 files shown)


sstable/src/dictionary.rs
```diff
@@ -92,26 +94,71 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
...
+                        // this block is needed too
+                        last_required_block = block_id;
+                    }
                 }
+            } else {
+                last_block_id
             }
-            Bound::Unbounded => Bound::Unbounded,
+        } else {
+            last_block_id
```
fulmicoton: can you make that code simpler?

sstable/src/dictionary.rs
```diff
@@ -178,21 +225,35 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
...
-        for _ in 0..(ord + 1) {
+        // find block in which the term would be
+        let Some(block_addr) = self.sstable_index.search_term_ordinal(ord) else {
+            return Ok(false)
+        };
+        let first_ordinal = block_addr.first_ordinal;
+
+        // then search inside that block only
+        let mut sstable_reader = self.sstable_reader_block(block_addr)?;
+        for _ in first_ordinal..(ord + 1) {
```
fulmicoton: nitpick... (=feel free to ignore)
```suggestion
        for _ in first_ordinal..=ord {
```

sstable/src/dictionary.rs
```diff
@@ -178,21 +225,35 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
...
-        for _ in 0..(term_ord + 1) {
+        // find block in which the term would be
+        let Some(block_addr) = self.sstable_index.search_term_ordinal(term_ord) else {
+            return Ok(None)
+        };
+        let first_ordinal = block_addr.first_ordinal;
+
+        // then search inside that block only
+        let mut sstable_reader = self.sstable_reader_block(block_addr)?;
+        for _ in first_ordinal..(term_ord + 1) {
```
fulmicoton:
```suggestion
        for _ in first_ordinal..=term_ord {
```

sstable/src/dictionary.rs
```diff
@@ -178,21 +225,35 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
...
+        // find block in which the term would be
+        let Some(block_addr) = self.sstable_index.search_term_ordinal(ord) else {
+            return Ok(false)
+        };
+        let first_ordinal = block_addr.first_ordinal;
+
+        // then search inside that block only
+        let mut sstable_reader = self.sstable_reader_block(block_addr)?;
+        for _ in first_ordinal..(ord + 1) {
             if !sstable_reader.advance().unwrap_or(false) {
```
fulmicoton: I think we have a bug here. (not your doing though).

advance returning an io Error is different from not having the term.

sstable/src/dictionary.rs
```diff
@@ -178,21 +225,35 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
...
+        // find block in which the term would be
+        let Some(block_addr) = self.sstable_index.search_term_ordinal(term_ord) else {
+            return Ok(None)
+        };
+        let first_ordinal = block_addr.first_ordinal;
+
+        // then search inside that block only
+        let mut sstable_reader = self.sstable_reader_block(block_addr)?;
+        for _ in first_ordinal..(term_ord + 1) {
             if !sstable_reader.advance().unwrap_or(false) {
```
fulmicoton: same as above.

sstable/src/sstable_index.rs
```diff
@@ -16,10 +16,48 @@ impl SSTableIndex {
...
+            .map(|block_meta| block_meta.block_addr.clone())
+    }
+
+    pub(crate) fn search_block_id(&self, key: &[u8]) -> Option<usize> {
+        let pos = self
+            .blocks
+            .binary_search_by_key(&key, |block| &block.last_key_or_greater);
+        match pos {
+            Ok(mut pos) => {
+                // handle case where multiple blocks handle a single term (is that possible?)
```
fulmicoton: It is not possible, but I am glad you thought about that corner case. Keys are unique.

sstable/src/sstable_index.rs
```diff
@@ -16,10 +16,48 @@ impl SSTableIndex {
...
+    pub(crate) fn get_block_addr(&self, block_id: usize) -> Option<BlockAddr> {
         self.blocks
-            .iter()
-            .find(|block| &block.last_key_or_greater[..] >= key)
-            .map(|block| block.block_addr.clone())
+            .get(block_id)
+            .map(|block_meta| block_meta.block_addr.clone())
+    }
+
+    pub(crate) fn search_block_id(&self, key: &[u8]) -> Option<usize> {
```
fulmicoton: This function needs a doc... Especially if you were to handle multiple keys...

trinity-1686a: > Especially if you were to handle multiple keys...

not sure what you mean here

fulmicoton: In the implementaton you handled the case where there were keys appearing several time in the Sstable.
This is good... But in that case, that method has an ambiguous behavior (is it returning the first block, the last block, or any block that might contain the key?) that should have been documented.

fulmicoton (approved):

PR #1794, opened 2023-01-16
Add count, min, max, and sum aggregations

guilload:


fulmicoton:
I don't understand the point of this PR.

The goal is to catter for the need of subset of stats in quickwit API, correct?
I don't think it requires any PR in tantivy.

We can just call use the stats collector in quickwit and populate the reponse there.

fulmicoton:
Discussed outside of this PR...

Currently Quickwit directly relies on the deserialization of the aggregations object to convert a user request into a collector. This is not great, but this is what we have today.

I'll open a separate ticket to have tantivy act more bottom-up, and allow the DSL logic to be done on quickwit side...
Or at least to improve the logic on tantivy side.

fulmicoton:
Opened a ticket here https://github.com/quickwit-oss/tantivy/issues/1796

fulmicoton (approved):

PR #1805, opened 2023-01-18
add ip addr to columnar

PSeitz:


columnar/src/columnar/writer/column_operation.rs
```diff
@@ -25,12 +27,12 @@ struct ColumnOperationMetadata {
 
 impl ColumnOperationMetadata {
     fn to_code(self) -> u8 {
-        place_bits::<0, 4>(self.len) | place_bits::<4, 8>(self.op_type.to_code())
+        place_bits::<0, 6>(self.len) | place_bits::<6, 8>(self.op_type.to_code())
```
fulmicoton, line 30: why?

fulmicoton (approved):

columnar/src/columnar/writer/column_operation.rs
PSeitz: 4 bits are not enough for length 16

PR #1808, opened 2023-01-19
add columnar to workspace

PSeitz:


fulmicoton (approved):
the unit tests are too slow.... we need the attributes you added in the other PR. please merge this one after.

PR #1812, opened 2023-01-19
collect columns for merge

PSeitz:


columnar/src/column_index/mod.rs
```diff
@@ -44,9 +44,10 @@ impl<'a> ColumnIndex<'a> {
                 }
             }
             ColumnIndex::Multivalued(multivalued_index) => {
-                let start = multivalued_index.get_val(row_id);
-                let end = multivalued_index.get_val(row_id + 1);
-                start..end
+                unimplemented!()
+                // let start = multivalued_index.get_val(row_id);
```
fulmicoton: Why unimplemented!? Did you spot a bug?

columnar/src/columnar/merge.rs
```diff
@@ -31,3 +35,135 @@ pub fn merge_columnar(
...
+        }
+    }
+}
+
+/// Receives a list of columns of numerical types (u64, i64, f64)
+///
+/// Returns a list of `DynamicColumn` which are all of the same numerical type
+fn cast_to_common_numerical_column(columns: &[DynamicColumn]) -> Vec<DynamicColumn> {
+    assert!(columns.iter().all(|column| column.is_numerical()));
+    let coerce_to_i64: Vec<_> = columns
```
fulmicoton, line 86: we can probably do something a little bit nicer here, and be more coupled with the writer logic?

```rust
fn acceptable_numerical_type(numerical_types: &[NumericalType]) -> NumericalType {
     let mut compatible_numerical_types = CompatibleNumericalTypes::default();
     for &typ in numerical_types {
         compatible_numerical_types.accept(typ);
     }
     numerical_types.to_numeric_type()
}


struct DynamicColummn {
    fn column_type(&self) -> ColumnType { ... }
    fn coerce(&self, column_type: ColumnType) -> Option<DynamicColumn> { ... }
}
```

columnar/src/dynamic_column.rs
```diff
@@ -20,6 +22,114 @@ pub enum DynamicColumn {
...
+    }
+    pub fn is_f64(&self) -> bool {
+        matches!(self, DynamicColumn::F64(_))
+    }
+    pub fn is_u64(&self) -> bool {
+        matches!(self, DynamicColumn::U64(_))
+    }
+    pub fn is_i64(&self) -> bool {
+        matches!(self, DynamicColumn::I64(_))
+    }
```
fulmicoton: I'd replace those by a `.column_type()`...
That one already have a conversion to numerical_type()

columnar/src/dynamic_column.rs
```diff
@@ -20,6 +22,114 @@ pub enum DynamicColumn {
...
+                    values: Arc::new(monotonic_map_column(column.values, MapI64ToU64)),
+                }))
+            }
+            DynamicColumn::U64(_) => Some(self),
+            _ => None,
+        }
+    }
+}
+
+pub struct MapI64ToF64;
```
fulmicoton: Do you think we could replace those three thing by the following?
I don;'t know any trait boudnary for as.

If you want to stick to `as`, can you make those private to this module?
Technically these mapping are not monotonic.



```
pub struct Coerce<FromTyp, ToTyp>;

impl<FromTyp, ToTyp> StrictlyMonotonicFn<FromTyp, ToTyp> for MapI64ToF64 
where FromTyp: TryFrom<ToTyp>, ToTyp: TryFrom<FromTyp>
{
}
```

fulmicoton (changes requested):
see comments inline

columnar/src/columnar/merge.rs
PSeitz: the type is not enough, we need to know if the a i64 column could be a u64 column (!contains negative values)

columnar/src/dynamic_column.rs
PSeitz: maybe with num traits, but the use case is too small

fulmicoton (approved):

columnar/src/dynamic_column.rs
fulmicoton: ok

columnar/src/columnar/merge.rs
fulmicoton: ah good point

PR #1819, opened 2023-01-21
Implement extended stats aggregation

ritesh99rakesh:
Still WIP but hoping to get review about whether this is the right direction


guilload:
Hi @ritesh99rakesh, I messed up the link in the original issue. So we want a new aggregation named `extended_stats` matching this [API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-extendedstats-aggregation.html).

Issue #1820, opened 2023-01-22
v0.19.1 consumes much more memory than v0.18


Enter-tainer:
**Describe the bug**
- What did you do?
When building a chat history searcher using tantivy v0.19.1, I found that it takes quite a lot of memory (~800M) for only 80k entries. I use bytehound to ananlyze the memory usage but I cannot figure out why. I try v0.18 and it only takes ~80M.

**To Reproduce**

> If your bug is deterministic, can you give a minimal reproducing code?
> Some bugs are not deterministic. Can you describe with precision in which context it happened?
> If this is possible, can you share your code?

I cannot share my code because it contains sensitive data. But I will provide the flamegraph. Hope that helps.

![v0 19](https://user-images.githubusercontent.com/25521218/213909339-03e094f8-3693-4bfb-8ff3-008b716c1011.svg)

full backtrace for the most memory-consuming part:

<details>
<summary>Details</summary>

```
#0 [libc.so.6] 0x00007F80A0D21A5F
#1 [libc.so.6] 0x00007F80A0C9F8FC
#2 [tg-search] std::sys::unix::thread::Thread::new::thread_start [thread.rs:108]
#3 [tg-search] <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once [boxed.rs:1987]
#4 [tg-search] <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once [boxed.rs:1987]
#5 [tg-search] core::ops::function::FnOnce::call_once{{vtable.shim}} [function.rs:251]
#6 [tg-search] std::thread::Builder::spawn_unchecked_::{{closure}} [mod.rs:550]
#7 [tg-search] std::panic::catch_unwind [panic.rs:137]
#8 [tg-search] std::panicking::try [panicking.rs:447]
#9 [tg-search] std::panicking::try::do_call [panicking.rs:483]
#10 [tg-search] <core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once [unwind_safe.rs:271]
#11 [tg-search] std::thread::Builder::spawn_unchecked_::{{closure}}::{{closure}} [mod.rs:551]
#12 [tg-search] std::sys_common::backtrace::__rust_begin_short_backtrace [backtrace.rs:121]
#13 [tg-search] tantivy::directory::watch_event_router::WatchCallbackList::broadcast::{{closure}} [watch_event_router.rs:87]
[...]

PR #1821, opened 2023-01-23
fix doc store cache docs

PSeitz:
addresses an issue reported in #1820

### DIFF
--- a/src/reader/mod.rs
+++ b/src/reader/mod.rs
@@ -151,7 +151,7 @@ impl TryInto<IndexReader> for IndexReaderBuilder {
 }
 
 struct InnerIndexReader {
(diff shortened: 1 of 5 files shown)


src/reader/mod.rs
```diff
@@ -178,7 +178,7 @@ impl InnerIndexReader {
             &searcher_generation_inventory,
         )?;
         Ok(InnerIndexReader {
-            doc_store_cache_size,
+            doc_store_cache_num_blocks: doc_store_cache_size,
```
fulmicoton: we need to rename the argument too.

fulmicoton (approved):
thank you for following up on this. Can you rename the arguemtn as mentionned in my comment?

PR #1823, opened 2023-01-23
improve bucket size aggregation limit

### LINKED ISSUE #1822: Investigate elastic search max bucket behaviour
As reported here https://github.com/quickwit-oss/quickwit/issues/812, term queries run into the aggregation bucket limit.

Investigate behavior of elastic search for term aggregation queries. 

Opposed to other aggregations like `HistogramAggregation`, there is a half-way sensible upper limit due to the number of terms in the dictionary. 
This upper limit may be circumvented with nested aggregations.

PSeitz:
postpone bucket size aggregation limit for terms aggregation when converting to intermediate aggregation result
improve min_doc_count special use case 0, it loaded more texts from the
dict than segment_size limit

Note that we effectively check now that segment_size does not exceed the
bucket limit, since that's the number of bucket when converting into an intermediate bucket. 
segment_size is 10 * size. So requests would fail with size 6500 for a bucket limit of 65000.

That behavior is probably good for Quickwit, since we pass intermediate results between nodes, so we should have the circuit breaker on the intermediate result level.

It's not great for use cases where we have many segments in a single index, since they could easily add up to hit the bucket limit, since the formula is now: `#num_segments * segment_size < bucket limit` (if all segments have enough results)

closes https://github.com/quickwit-oss/tantivy/issues/1822

### DIFF
--- a/src/aggregation/bucket/term_agg.rs
+++ b/src/aggregation/bucket/term_agg.rs
@@ -11,7 +11,7 @@ use crate::aggregation::agg_req_with_accessor::{
 use crate::aggregation::intermediate_agg_result::{
     IntermediateBucketResult, IntermediateTermBucketEntry, IntermediateTermBucketResult,
 };
-use crate::aggregation::segment_agg_result::{BucketCount, SegmentAggregationResultsCollector};
+use crate::aggregation::segment_agg_result::SegmentAggregationResultsCollector;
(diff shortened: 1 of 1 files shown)


fulmicoton (approved):

PSeitz:
Superseeded by https://github.com/quickwit-oss/tantivy/pull/1942

PR #1834, opened 2023-01-28
#1833 Add ExistsQuery

### LINKED ISSUE #1833: Add ExistsQuery for finding all documents where a certain field exists
Sometimes it's necessary to find all documents that simply contain a field, regardless of its value

An ExistsQuery could be added that allows querying for documents containing a certain field.

Currently `RegexQuery::from_pattern(".*", field)?` can be used for this, and probably a RangeQuery for numeric fields.

Hodkinson:
This adds a simple ExistsQuery, that allows documents that simply contains a field to be found. 

Fixes #1833

### DIFF
--- a/src/query/exists_query.rs
+++ b/src/query/exists_query.rs
@@ -76,8 +76,8 @@ impl Weight for ExistsWeight {
         let mut doc_bitset = BitSet::with_max_value(max_doc);
 
         let inverted_index = reader.inverted_index(self.field)?;
-        let term_dict = inverted_index.terms();
-        let mut term_stream = term_dict.range().into_stream()?;
+        let mut term_stream = inverted_index.terms().stream()?;
+
         while term_stream.advance() {
             let term_info = term_stream.value();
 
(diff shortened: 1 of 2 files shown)


src/query/exists_query.rs
```diff
@@ -0,0 +1,183 @@
...
+}
+
+impl Weight for ExistsWeight {
+    fn scorer(&self, reader: &SegmentReader, boost: Score) -> Result<Box<dyn Scorer>> {
+        let max_doc = reader.max_doc();
+        let mut doc_bitset = BitSet::with_max_value(max_doc);
+
+        let inverted_index = reader.inverted_index(self.field)?;
+        let term_dict = inverted_index.terms();
+        let mut term_stream = term_dict.range().into_stream()?;
```
adamreichold: Isn't this equivalent to

```rust
let mut term_stream = term_dict.stream()?;
```

Hodkinson: yes, I didn't see that function, have simplified

Nickersoft:
Per #1836, what would be the lift for tweaking this to also support JSON fields?

rdettai:
This was implemented in https://github.com/quickwit-oss/tantivy/pull/2160 and https://github.com/quickwit-oss/tantivy/pull/2170

PR #1840, opened 2023-01-31
fix formatting

saroh:
`$ cargo +nightly fmt --all`


PSeitz:
Thanks, but already covered by https://github.com/quickwit-oss/tantivy/pull/1838

PR #1842, opened 2023-02-08
add PhrasePrefixQuery

trinity-1686a:
see https://github.com/quickwit-oss/quickwit/issues/2266


src/query/phrase_prefix_query/phrase_prefix_scorer.rs
```diff
@@ -0,0 +1,116 @@
...
+            if suffix.doc() > current_doc {
+                continue;
+            }
+            let doc = suffix.seek(current_doc);
+            if doc == current_doc {
+                suffix.positions_with_offset(self.suffix_offset, &mut positions);
+                if self.slop == 0 {
+                    count += intersection_count(pos_matching, &positions);
+                } else {
+                    count += intersection_count_with_slop(pos_matching, &positions, self.slop);
```
fulmicoton: can we ditch the slop logic?

src/query/phrase_prefix_query/phrase_prefix_scorer.rs
```diff
@@ -0,0 +1,187 @@
...
+            let (pos, posting) = term_postings
+                .pop()
+                .expect("PhrasePrefixScorer must have at least two terms");
+            let offset = suffixe_pos - pos;
+            PhraseKind::SinglePrefix(offset as u32, posting, Vec::with_capacity(100))
+        };
+        let mut res = PhrasePrefixScorer {
+            phrase_scorer,
+            suffixes,
+            suffix_offset: (max_offset - suffixe_pos) as u32,
```
fulmicoton:
```suggestion
            suffix_offset: (max_offset - suffix_pos) as u32,
```

src/query/phrase_prefix_query/phrase_prefix_scorer.rs
```diff
@@ -0,0 +1,187 @@
...
+
+pub struct PhrasePrefixScorer<TPostings: Postings> {
+    phrase_scorer: PhraseKind<TPostings>,
+    suffixes: Vec<TPostings>,
+    suffix_offset: u32,
+    phrase_count: u32,
+}
+
+impl<TPostings: Postings> PhrasePrefixScorer<TPostings> {
+    // If similarity_weight is None, then scoring is disabled.
```
fulmicoton, line 87:
```suggestion
    /// If similarity_weight is None, then scoring is disabled.
    /// # Panics
    /// If `term_postings.len() < 2`
```

src/query/phrase_prefix_query/phrase_prefix_scorer.rs
```diff
@@ -0,0 +1,187 @@
...
+                term_postings,
+                similarity_weight_opt,
+                fieldnorm_reader,
+                0,
+                1,
+            ))
+        } else {
+            let (pos, posting) = term_postings
+                .pop()
+                .expect("PhrasePrefixScorer must have at least two terms");
```
fulmicoton, line 114:
```suggestion
                .expect("PhrasePrefixScorer must have at least one term");
```

src/query/phrase_prefix_query/phrase_prefix_scorer.rs
```diff
@@ -0,0 +1,187 @@
...
+            let offset = suffixe_pos - pos;
+            PhraseKind::SinglePrefix(offset as u32, posting, Vec::with_capacity(100))
+        };
+        let mut res = PhrasePrefixScorer {
```
fulmicoton:
```suggestion
        let mut phrase_prefix_scorer = PhrasePrefixScorer {
```

src/query/phrase_prefix_query/phrase_prefix_scorer.rs
```diff
@@ -0,0 +1,187 @@
+use crate::docset::{DocSet, TERMINATED};
+use crate::fieldnorm::FieldNormReader;
+use crate::postings::Postings;
+use crate::query::bm25::Bm25Weight;
+use crate::query::phrase_query::{intersection_count, PhraseScorer};
+use crate::query::Scorer;
+use crate::{DocId, Score};
+
+enum PhraseKind<TPostings: Postings> {
+    SinglePrefix(u32, TPostings, Vec<u32>),
```
fulmicoton:
```suggestion
    SingleTermPrefix {
           position_offset: u32,
           postings: TPostings,
           positions: Vec<u32>
     },
```

src/query/phrase_prefix_query/phrase_prefix_query.rs
```diff
@@ -0,0 +1,160 @@
...
+        let field = terms[0].1.field();
+        assert!(
+            terms[1..].iter().all(|term| term.1.field() == field),
+            "All terms from a phrase query must belong to the same field"
+        );
+        PhrasePrefixQuery {
+            field,
+            prefix: terms.pop().unwrap(),
+            phrase_terms: terms,
+            max_expansions: 50,
```
fulmicoton: Can you add a  const DEFAULT_MAX_EXPANSIONS

src/query/phrase_prefix_query/phrase_prefix_query.rs
```diff
@@ -0,0 +1,160 @@
...
+            .map(|(_, term)| term.clone())
+            .collect::<Vec<Term>>()
+    }
+
+    /// Returns the [`PhrasePrefixWeight`] for the given phrase query given a specific `searcher`.
+    ///
+    /// This function is the same as [`Query::weight()`] except it returns
+    /// a specialized type [`PhraseQueryWeight`] instead of a Boxed trait.
+    /// If the query was only one term long, this returns `None` wherease [`Query::weight`]
+    /// returns a boxed [`RangeWeight`]
```
fulmicoton, line 90:
```suggestion
    /// returns a boxed [`RangeWeight`]
    ///
    /// Returns `None`, if phrase_terms is empty, which happens if the phrase prefix query was
    /// built with a single term.
```

src/query/phrase_prefix_query/phrase_prefix_query.rs
```diff
@@ -0,0 +1,160 @@
...
+
+impl Query for PhrasePrefixQuery {
+    /// Create the weight associated with a query.
+    ///
+    /// See [`Weight`].
+    fn weight(&self, enable_scoring: EnableScoring<'_>) -> crate::Result<Box<dyn Weight>> {
+        if let Some(phrase_weight) = self.phrase_query_weight(enable_scoring)? {
+            Ok(Box::new(phrase_weight))
+        } else {
+            let end_term = if let Some(end_value) = prefix_end(&self.prefix.1.value_bytes()) {
```
fulmicoton, line 141:
```suggestion
            /// There are no prefix. Let's just match the suffix.
            let end_term = if let Some(end_value) = prefix_end(&self.prefix.1.value_bytes()) {
```

src/query/phrase_prefix_query/phrase_prefix_query.rs
```diff
@@ -0,0 +1,160 @@
...
+                let mut end_term = Term::with_capacity(end_value.len());
+                end_term.set_field_and_type(self.field, self.prefix.1.typ());
+                end_term.append_bytes(&end_value);
+                Bound::Excluded(end_term)
+            } else {
+                Bound::Unbounded
+            };
+
+            RangeQuery::new_term_bounds(
```
fulmicoton, line 150: I think we lost the limit in the number of term to expand to which could be catastrophic.
(think about a use typing "t" and hitting enter)

Can you open an issue and auto-attribute it to you?

src/query/phrase_prefix_query/phrase_prefix_query.rs
```diff
@@ -0,0 +1,160 @@
...
+                    .to_owned(),
+                self.prefix.1.typ(),
+                &Bound::Included(self.prefix.1.clone()),
+                &end_term,
+            )
+            .weight(enable_scoring)
+        }
+    }
+
+    fn query_terms<'a>(&'a self, visitor: &mut dyn FnMut(&'a Term, bool)) {
```
fulmicoton, line 163: Is the named lifetime required here?

src/query/phrase_prefix_query/phrase_prefix_query.rs
```diff
@@ -0,0 +1,160 @@
...
+        for (_, term) in &self.phrase_terms {
+            visitor(term, true);
+        }
```
fulmicoton, line 166: you forgot the prefix (expanded prefix which is a suffix)

src/query/phrase_prefix_query/phrase_prefix_weight.rs
```diff
@@ -0,0 +1,257 @@
...
+                    term_postings_list.push((offset, postings));
+                } else {
+                    return Ok(None);
+                }
+            }
+        } else {
+            for &(offset, ref term) in &self.phrase_terms {
+                if let Some(postings) = reader
+                    .inverted_index(term.field())?
+                    .read_postings_no_deletes(term, IndexRecordOption::WithFreqsAndPositions)?
```
fulmicoton, line 71: I know you copied this from phrase_weight but the code is meaningless.. As far as I can tell read_postings_no_deletes does the same as read_postings.

Can you open an issue to cleanup this mess `read_postings_no_deletes`?

Correctness is (probably) not at stake here... We do the check at the collector level anyway.
There is a non-trivial optimization decision to take here.

Do we do 

check_phrase(intersection(remove_deletes(postings)))
or
remove_deletes(check_phrase(intersection(postings)))


The earlier probably seems trivially faster, but I suspect it depends on the filtering power of deletes.
Anyway for the moment we can probably stick to 

remove_deletes(check_phrase(intersection(postings)))

for simplificity, and clean up the existing code.

src/query/phrase_prefix_query/phrase_prefix_weight.rs
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+        let mut stream = stream.into_stream()?;
+
+        let mut suffixes = Vec::with_capacity(self.max_expansions as usize);
+        let mut new_term = self.prefix.1.clone();
+        while stream.advance() && (suffixes.len() as u32) < self.max_expansions {
+            new_term.clear_with_type(new_term.typ());
+            new_term.append_bytes(stream.key());
+            if reader.has_deletes() {
```
fulmicoton, line 103: same as above. we don't need the distinction

fulmicoton (approved):
LGTM but can you go throught the minor suggestions?

They are mostly about naming / lack of comments / one bug / and a WTF coming from the original phrase scorer.

src/query/phrase_prefix_query/phrase_prefix_query.rs
```diff
@@ -0,0 +1,160 @@
...
+            .collect::<Vec<Term>>()
+    }
+
+    /// Returns the [`PhrasePrefixWeight`] for the given phrase query given a specific `searcher`.
+    ///
+    /// This function is the same as [`Query::weight()`] except it returns
+    /// a specialized type [`PhraseQueryWeight`] instead of a Boxed trait.
+    /// If the query was only one term long, this returns `None` wherease [`Query::weight`]
+    /// returns a boxed [`RangeWeight`]
+    pub(crate) fn phrase_query_weight(
```
fulmicoton:
```suggestion
    pub(crate) fn phrase_prefix_query_weight(
```

src/query/phrase_prefix_query/phrase_prefix_weight.rs
```diff
@@ -0,0 +1,257 @@
...
+
+        let inv_index = reader.inverted_index(self.prefix.1.field())?;
+        let mut stream = inv_index.terms().range().ge(self.prefix.1.value_bytes());
+        if let Some(end) = prefix_end(self.prefix.1.value_bytes()) {
+            stream = stream.lt(&end);
+        }
+
+        #[cfg(feature = "quickwit")]
+        {
+            // TODO implement this on fst too so we don't need to check for feature flag.
```
fulmicoton:
```suggestion
            // We don't have this on the fst, hence  we end up needing a feature flag.
            //
            // This is not a problem however as we enforce the limit below too.
            // The point of `stream.limit` is to limit the number of term dictionary
            // blocks being downloaded.
```

fulmicoton: (I don't think we need to address this)

src/query/phrase_prefix_query/phrase_prefix_query.rs
trinity-1686a: it is
```
error[E0308]: method not compatible with trait
   --> src/query/phrase_prefix_query/phrase_prefix_query.rs:161:5
    |
161 |     fn query_terms(&self, visitor: &mut dyn FnMut(&Term, bool)) {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ one type is more general than the other
    |
    = note: expected fn pointer `fn(&'a PhrasePrefixQuery, &mut dyn FnMut(&'a Term, bool))`
               found fn pointer `fn(&PhrasePrefixQuery, &mut dyn for<'a> FnMut(&'a Term, bool))`
```

src/query/phrase_prefix_query/phrase_prefix_query.rs
trinity-1686a: I did not put it on purpose, we don't use that Term directly, only as a bound in a range (and at this instant, we can't expand it yet). `RangeQuery` isn't giving its bounds when calling that function, so it seemed more coherent to not do that here either. If you think I still should emit it, please reply back.

src/query/phrase_prefix_query/phrase_prefix_query.rs
adamreichold: The extra lifetime tying the self liftime to the term refereneces is required so that the closure can stash the term references elsewhere, e.g. into a separate `Vec`. Otherwise, the term references would be limited to the closure body.

src/query/phrase_prefix_query/phrase_prefix_query.rs
trinity-1686a: opened #1886

PR #1843, opened 2023-02-09
Make index compatible with virtual drives on Windows

### LINKED ISSUE #1724: Create index in virtual drives on Windows
On Windows, Tantivy will raise IO errors (`Os { code: 1, kind: Uncategorized, message: "Incorrect function." }`) if you put the index in some virtual drives (e.g., ramdisk, network drive, etc.). I find the errors occur in two places:

1. Calling `std::fs::canonicalize` on a path to virtual drive. Related Rust issues: <https://github.com/rust-lang/rust/issues/40282>, <https://github.com/rust-lang/rust/pull/86447>.

    This can be fixed by using [`normpath::PathExt::normalize`](https://docs.rs/normpath/latest/normpath/trait.PathExt.html#tymethod.normalize) instead.

2. Calling `sync_data` on a handle of directory on virtual drive. Under the hood it calls [`FlushFileBuffers`](https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-flushfilebuffers#remarks) Win32 function, but according to the documentation, `FlushFileBuffers` doesn't work with directories, and "If an application is performing multiple writes to disk and also needs to ensure critical data is written to persistent media, the application should use unbuffered I/O instead of frequently calling FlushFileBuffers.". Refs: [1](https://social.msdn.microsoft.com/Forums/Windowsdesktop/en-US/847a735b-f21a-4be2-880b-12660e5b98b4/flushfilebuffers-system-call-fails-on-network-mapped-drive), [2](https://social.msdn.microsoft.com/Forums/sqlserver/en-US/0e0b734b-2b73-414d-8833-8d2eed7043f6/sync-directories-on-windows?forum=windowsgeneraldevelopmentissues).

    So it seems better to make `sync_directory` an no-op on Windows?

gyk:
Fixes #1724.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,6 @@ regex = { version = "1.5.5", default-features = false, features = ["std", "unico
 aho-corasick = "0.7"
 tantivy-fst = "0.4.0"
 memmap2 = { version = "0.5.3", optional = true }
-normpath = "1.0.0"
 lz4_flex = { version = "0.10", default-features = false, features = ["checked-decode"], optional = true }
 brotli = { version = "3.3.4", optional = true }
 zstd = { version = "0.12", optional = true, default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ regex = { version = "1.5.5", default-features = false, features = ["std", "unico
 aho-corasick = "0.7"
 tantivy-fst = "0.4.0"
 memmap2 = { version = "0.5.3", optional = true }
+normpath = "1.0.0"
 lz4_flex = { version = "0.10", default-features = false, features = ["checked-decode"], optional = true }
 brotli = { version = "3.3.4", optional = true }
 zstd = { version = "0.12", optional = true, default-features = false }
--- a/src/directory/mmap_directory.rs
+++ b/src/directory/mmap_directory.rs
@@ -9,7 +9,6 @@ use std::{fmt, result};
 use common::StableDeref;
 use fs2::FileExt;
 use memmap2::Mmap;
-use normpath::PathExt;
 use serde::{Deserialize, Serialize};
 use tempfile::TempDir;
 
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -23,6 +23,7 @@ regex = { version = "1.5.5", default-features = false, features = ["std", "unico
 aho-corasick = "0.7"
 tantivy-fst = "0.4.0"
 memmap2 = { version = "0.5.3", optional = true }
+normpath = "1.0.0"
```
fulmicoton: I'd rather not introduce a dependency for this. Can you remove it and stick to std's canonicalize?

fulmicoton (changes requested):
@gyk can you remove the deps on normpath?

Cargo.toml
gyk: > Can you remove it and stick to std's canonicalize?

Unfortunately this is not possible currently as the [PR](https://github.com/rust-lang/rust/pull/86447) attempting to fix the problem was closed, but in the future we can switch to [`std::path::absolute`](https://doc.rust-lang.org/std/path/fn.absolute.html) once it is stabilized.

IMHO `normpath` is quite lightweight, and its biggest dependency is `windows-sys = { version = "0.45", features = ["Win32_Storage_FileSystem"] }` (not the macro-heavy `windows` crate), which only adds some reasonable amount of compilation time and little extra size to the release build. Also Tantivy has already included dependencies of `windows` 0.39 and `windows-sys` 0.42.

<details>
<summary>cargo-bloat output</summary>

```
    Analyzing target\release\bloat-normpath.exe

 File  .text     Size    Crate Name
 1.9%   2.6%   3.2KiB      std rustc_demangle::demangle
 1.7%   2.4%   3.0KiB      std <rustc_demangle::legacy::Demangle as core::fmt::Display>::fmt
 1.4%   1.9%   2.4KiB normpath normpath::normalize::normalize_virtually
 1.3%   1.7%   2.2KiB      std core::str::pattern::impl$28::is_contained_in
 1.1%   1.5%   1.9KiB      std <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
 1.0%   1.4%   1.8KiB      std rustc_demangle::v0::Printer::print_path
 1.0%   1.4%   1.7KiB      std std::sys_common::once::generic::Once::call
 1.0%   1.4%   1.7KiB      std std::sys::windows::path::maybe_verbatim
 1.0%   1.4%   1.7KiB      std rustc_demangle::v0::Printer::print_const
 1.0%   1.3%   1.7KiB      std core::str::count::do_count_chars
56.7%  77.7%  97.9KiB          And 443 smaller methods. Use -n N to show more.
73.0% 100.0% 126.0KiB          .text section size, the file size is 172.5KiB
```
</details>

fulmicoton: @gyk I won't merge this with the extra dependency. Isn't this PR useful without the canonicalization?

gyk: No, with `std::fs::canonicalize` the index still cannot be put on special drives. You might want to consider:

1. Make normpath a Windows-only dependency, or
2. Call `std::fs::canonicalize` like before, but in case it returns this kind of error but the path does exist, just use the original `directory_path` as a fallback.

fulmicoton: ah ok so canonicalize return an error for that kind of path. I see. I'm ok with solution 1 or solution 2. You can decide.

gyk: OK, I'll update the code later.

gyk: @fulmicoton Updated using approach 2 above. PTAL.

fulmicoton (approved):

PR #1855, opened 2023-02-09
Make BM25 scoring more flexible

alexcole:
This diff makes a few changes to expand the ways developers can use tantivy's BM25 scoring:
1. Make `Bm25Weight` public.
2. Remove the dependency of `Bm25Weight` on a `Searcher`. Instead it expects a `Bm25StatisticsProvider` (which is by default a `Searcher`).
3. Allow specifying a custom `Bm25StatisticsProvider` when running a query.

Some example use cases this unlocks:
1. Using `Bm25Weight` without the rest of tantivy.
    - Having a correct BM25 implementation is tricky! The `Bm25Weight` is also optimized to cache based on fieldnorm which is nice.
    - Now it has no dependency on actually storing documents in tantivy to use.
2. Adjusting the statistics to include updates not yet saved to tantivy.
    - If a developer is using tantivy as a part of a larger system, they may not store every document to tantivy immediately when the data is created. For example they may store recent documents in memory and only periodically sync them to tantivy. Or there may be documents that are created in a transaction that hasn't committed yet (so they shouldn't be saved to tantivy yet).
    - In these cases, the `Bm25StatisticsProvider` should be the `Searcher` with adjustments for the data not yet saved to tantivy.
3. Adjust the statistics to account for deletes
    - Currently the numbers of documents and tokens include deleted ones which is somewhat surprising. Being able to customize the `Bm25StatisticsProvider` opens the door to storing additional files to correctly track the number of undeleted documents and tokens.

My company (Convex) is using (1) and (2) and maybe some day we'll want to do (3) as well. 

Let me know if any of this doesn't make sense! Also happy to change the implementation if there is a better way to accomplish these goals.


fulmicoton:
@alexcole thanks for the contribution... espeically taking time to properly explain what your use case is. 
@shikhar can you review?

fulmicoton (approved):

shikhar (commented):
Good stuff, LGTM as well! Sorry about the delay.

src/query/bm25.rs
```diff
@@ -2,11 +2,53 @@ use serde::{Deserialize, Serialize};
...
 const B: Score = 0.75;
 
+/// An interface to compute the statistics needed in BM25 scoring.
+///
+/// The standard implementation is a [Searcher] but you can also
+/// create your own to adjust the statistics.
+pub trait Bm25StatisticsProvider {
+    /// The total number of tokens in a given field across all documents in
+    /// the index.
+    fn total_num_tokens(&self, field: Field) -> crate::Result<u64>;
```
adamreichold, line 18: Doesn't putting `total_num_tokens` and `total_num_docs` into separate methods mean that we access at least the metadata of the inverted indexes for all segments twice now instead of once as without the trait? Does this not have a measurable effect on performance? Would it suffice to provide these values in a single method, e.g.

```rust
fn total_numbers(&self, field: Field) -> crate::Result<(u64, u64)>;
```

src/query/query.rs
```diff
@@ -12,7 +13,16 @@ use crate::{DocAddress, Term};
...
-    Enabled(&'a Searcher),
+    Enabled {
+        /// The searcher to use during scoring.
+        searcher: &'a Searcher,
+
+        /// A [Bm25StatisticsProvider] used to compute BM25 scores.
+        ///
+        /// Normally this should be the [Searcher], but you can specify a custom
+        /// one to adjust the statistics.
+        statistics_provider: &'a dyn Bm25StatisticsProvider,
```
adamreichold, line 24: Are there any measurements of whether this additional dynamic dispatch affects performance of the scoring?

fulmicoton: (No need to measure.)

src/query/bm25.rs
fulmicoton: @adamreichold We all like performance, but this has no impact.

src/query/query.rs
shikhar: This is being accessed at most per term, I think static dispatch will only matter for per doc stuff like collection.

adamreichold: > This is being accessed at most per term, I think static dispatch will only matter for per doc stuff like collection.

Indeed that is a convincing argument that the additional overhead will not matter. Thank you for taking the time to write it down!

PR #1856, opened 2023-02-10
fix: doc store for files larger 4GB

PSeitz:
Fixes an issue in the skip list deserialization, which deserialized the byte start offset incorrectly as u32.
`get_doc` will fail for any docs that live in a block with start offset larger than u32::MAX (~4GB).
Causes index corruption, if a segment with a doc store file larger 4GB is merged.

tantivy version 0.19 is affected

### DIFF
--- a/src/store/index/block.rs
+++ b/src/store/index/block.rs
@@ -90,7 +90,7 @@ impl CheckpointBlock {
             return Ok(());
         }
         let mut doc = read_u32_vint(data);
-        let mut start_offset = read_u32_vint(data) as usize;
+        let mut start_offset = VInt::deserialize_u64(data)? as usize;
         for _ in 0..len {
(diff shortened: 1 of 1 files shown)


src/store/index/block.rs
```diff
@@ -90,7 +90,7 @@ impl CheckpointBlock {
             return Ok(());
         }
         let mut doc = read_u32_vint(data);
-        let mut start_offset = read_u32_vint(data) as usize;
+        let mut start_offset = VInt::deserialize_u64(data)? as usize;
```
fulmicoton, line 93: on the write side we use the u64 stuff?

fulmicoton (approved):

src/store/index/block.rs
PSeitz: yes

PR #1857, opened 2023-02-10
fix: validate index record option, instead of vint error

PSeitz:
Prev: 
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: IoError(Custom { kind: InvalidData, error: "Reach end of buffer while reading VInt" })', src/main.rs:46:14
```
Now:
 ```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: SchemaError("Received invalid index record option: WithFreqsAndPositions, Actual: Basic. Field \"source\"")', src/main.rs:46:14
```

### DIFF
--- a/src/query/term_query/term_query.rs
+++ b/src/query/term_query/term_query.rs
@@ -109,6 +109,7 @@ impl TermQuery {
         } else {
             IndexRecordOption::Basic
         };
+
         Ok(TermWeight::new(
             self.term.clone(),
             index_record_option,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/query/term_query/term_query.rs [hunk 1]
```diff
@@ -104,14 +98,26 @@ impl TermQuery {
...
+            let error_msg = format!("Field {:?} is not indexed.", field_entry.name());
+            return Err(crate::TantivyError::SchemaError(error_msg));
+        };
+        acutal_index_record_option
+            .validate_compatiblity(requested_index_record_option, field_entry.name())?;
+
         Ok(TermWeight::new(
             self.term.clone(),
-            index_record_option,
+            acutal_index_record_option,
```
fulmicoton:
```suggestion
            actual_index_record_option,
```

fulmicoton (approved):

PR #1864, opened 2023-02-13
improve range query performance

### LINKED ISSUE #1863: Perf Regression: Range queries on multivalued fast fields
https://pseitz.github.io/tantivy_bench_graph/index.html?start=&end=&filter=percent_multi

Root cause seems to be restart at 0, for every batch
```
ColumnIndex::Multivalued(multivalued_index) => {
    // TODO important: avoid using 0u32, and restart from the beginning all of the time.
    multivalued_index.select_batch_in_place(0u32, rank_ids)
}
```

PSeitz:
fix RowId vs DocId naming
fixes #1863

### DIFF
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -1870,7 +1870,7 @@ mod tests {
                         .column_opt::<Ipv6Addr>("ips")
                         .unwrap()
                         .unwrap();
-                    ff_reader.num_rows() as usize
+                    ff_reader.num_docs() as usize
                 })
                 .sum();
             assert_eq!(num_docs, num_docs_expected);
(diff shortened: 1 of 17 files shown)


src/indexer/index_writer.rs
```diff
@@ -2089,11 +2089,12 @@ mod tests {
 
             let do_search_ip_field = |term: &str| do_search(term, ip_field).len() as u64;
             // Range query on single value field
-            // let query = gen_query_inclusive("ip", ip, ip);
```
fulmicoton, line 2092: thanks for spotting this!

fulmicoton (changes requested):
Let's not rename RowId -> DocId yet.

I voluntarily called it `RowId` to keep columnar as independant a crate as possible.
I agree that it has some Cons:
a) It creates some minor naming mismatch when using it within tantivy.
b) We end up with RowId at the index level and at the value level. It is somewhat confusing. Right now, I call the latter "value row id".

I feel b) is more itchy than a) actually.
Anyway can we revisit that rename later and in a separate PR?

fulmicoton (approved):

fulmicoton (approved):

PR #1876, opened 2023-02-16
Issue/columnar for json

fulmicoton:


src/fastfield/writer.rs
```diff
@@ -1,34 +1,44 @@
...
+/// Only index JSON down to a depth of 20.
+/// This is mostly to guard us from a stack overflow triggered by malicious input.
+const JSON_DEPTH_LIMIT: usize = 20;
+
 /// The `FastFieldsWriter` groups all of the fast field writers.
 pub struct FastFieldsWriter {
     columnar_writer: ColumnarWriter,
     fast_field_names: Vec<Option<String>>, //< TODO see if we can cash the field name hash too.
     date_precisions: Vec<DatePrecision>,
+    expand_dots: Vec<bool>,
```
PSeitz, line 20: can you add a comment what this does?

src/fastfield/writer.rs
```diff
@@ -163,3 +189,243 @@ impl FastFieldsWriter {
...
+        json_path_buffer.push_str(key);
+        if expand_dots {
+            // This might include the separation byte, which is ok because it is not a dot.
+            let appended_segment = &mut json_path_buffer[len_path..];
+            // The unsafe below is safe as long as b'.' and JSON_PATH_SEGMENT_SEP are
+            // valid single byte ut8 strings.
+            // By utf-8 design, they cannot be part of another codepoint.
+            replace_in_place(b'.', JSON_PATH_SEGMENT_SEP, unsafe {
+                appended_segment.as_bytes_mut()
+            });
```
PSeitz, line 232: Do we replace JSON_PATH_SEGMENT_SEP with `.`, so we can query it?

Why don't we use `.` instead of JSON_PATH_SEGMENT_SEP

src/fastfield/writer.rs
```diff
@@ -163,3 +189,243 @@ impl FastFieldsWriter {
...
+        // popping our sub path.
+        json_path_buffer.truncate(len_path);
+    }
+}
+
+fn record_json_value_to_columnar_writer(
+    doc: DocId,
+    json_val: &serde_json::Value,
+    expand_dots: bool,
+    mut depth_limit: usize,
```
PSeitz:
```suggestion
    mut remaining_depth_limit: usize,
```

src/schema/json_object_options.rs
```diff
@@ -13,7 +13,8 @@ pub struct JsonObjectOptions {
     // If set to some, int, date, f64 and text will be indexed.
     // Text will use the TextFieldIndexing setting for indexing.
     indexing: Option<TextFieldIndexing>,
-
+    // Store all field as fast fields.
+    fast: bool,
```
PSeitz, line 17: I think a list of paths for which to create fast fields would be good

src/fastfield/writer.rs
fulmicoton: that's the other way around. we replace '.' by the separator

src/fastfield/writer.rs
fulmicoton: yes... let me document that in the JsonOption though.

src/schema/json_object_options.rs
fulmicoton: Not in tantivy... but I think at one point we might want some router indeed.

src/fastfield/mod.rs
```diff
@@ -1071,4 +1072,41 @@ mod tests {
...
+            .add_document(doc!(with_fast_field=>json!({"hello": "with1"})))
+            .unwrap();
+        writer.commit().unwrap();
+        let searcher = index.reader().unwrap().searcher();
+        let segment_reader = searcher.segment_reader(0u32);
+        let fast_fields = segment_reader.fast_fields();
+        let column_without_opt: Option<StrColumn> = fast_fields.str("without\u{1}hello").unwrap();
+        assert!(column_without_opt.is_none());
+        for (col_name, _) in fast_fields.columnar().list_columns().unwrap() {
+            dbg!(&col_name);
```
PSeitz: is this on purpose in the test?

PSeitz (approved):

src/fastfield/mod.rs
fulmicoton: i removed it

PR #1882, opened 2023-02-19
feat: Separate flag for fieldnorms

ppodolsky:
Current API does not allow to disable using fieldnorms without disabling using frequencies/positions. Just a suggestion how to fix it, feel free to use it or reimplement.
@PSeitz


PSeitz:
@ppodolsky generally a good idea, but we would need to document the implications

Issue #1689, opened 2022-11-22
Support term aggregation on non text fields


PSeitz:
Term aggregation only works on text fields currently. 
Extend support for (single and multi-value):

#### Bucket Key type 
u64 
- [x] u64
- [x] i64
- [x] f64
- [x] boolean
- [x] DateTime

&[u8]
- [ ] bytes

u128
- [x] ip address

fulmicoton:
ETA on bool / ipvaddr? (I don't think we care about bytes all that much)

PSeitz:
bool should be simple
for ip addresses we need to change the fast field API to get a u64 representation for the u128 values (which may limit possible codecs), to reuse the aggregation code

PR #1883, opened 2023-02-20
feat: add support for u64,i64,f64 fields in term aggregation

PSeitz:


src/aggregation/mod.rs
```diff
@@ -271,6 +271,22 @@ pub enum Key {
     /// `f64` key
     F64(f64),
 }
+impl Eq for Key {}
+impl std::hash::Hash for Key {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        core::mem::discriminant(self).hash(state);
```
PSeitz, line 277: This won't work well if someone indexes different values of `f64::NAN` and then does an aggregation on it. 
Do we care about this use case?

PSeitz:
https://github.com/quickwit-oss/tantivy/issues/1689

src/aggregation/mod.rs
fulmicoton: Let's forbid NaN entirely. If you really want to keep them we can work at least map them upon input to whatever `u64_to_f64(0u64)` is

fulmicoton: TIL about Discriminant

fulmicoton, line 277: You forgot to hash the value isn't it?

src/aggregation/mod.rs
```diff
@@ -271,6 +271,22 @@ pub enum Key {
...
+    }
+}
+
+impl PartialEq for Key {
```
fulmicoton, line 285: Could this be dervied?

src/aggregation/mod.rs
PSeitz: yep

src/aggregation/mod.rs
PSeitz: yep

PSeitz: ah no this is an clippy error

src/aggregation/bucket/term_agg.rs
```diff
@@ -423,41 +430,52 @@ impl SegmentTermCollector {
...
-                let key = std::str::from_utf8(key)
-                    .map_err(|utf8_err| DataCorruption::comment_only(utf8_err.to_string()))?;
-                if !dict.contains_key(key) {
-                    dict.insert(key.to_owned(), Default::default());
+                    let key = Key::Str(
+                        std::str::from_utf8(key)
+                            .map_err(|utf8_err| DataCorruption::comment_only(utf8_err.to_string()))?
+                            .to_string(),
+                    );
+                    dict.entry(key).or_insert_with(Default::default);
```
fulmicoton: silly nitpick

```suggestion
                    dict.entry(key).or_default();
```

fulmicoton (approved):

PR #1884, opened 2023-02-20
Adding a write schema to columnar's merge operations.

fulmicoton:
As far as I can tell the previous logic was not faulty.
As a reminder: the logic was: 
- enforce a write schema upon 1st generation segment write
- after merge, make sure we keep writing all columns in the union of the merged columnar table (regardless of deleted rows), and keep the same type if they are all of the same type.

The logic however was rather brittle and disconnected from its purpose... The situation was therefore error-prone, were we to make it possible to add fields for instance, or were we to take into account column deletion in the emission of columns.

This PR defensively makes it possible to enforce a write schema upon merge.


columnar/src/columnar/merge/mod.rs
```diff
@@ -54,14 +52,34 @@ impl From<ColumnType> for ColumnTypeCategory {
     }
 }
 
+/// Merge several columnar table together.
+///
+/// If several columns with the same name are conflicting with the numerical types in the
```
PSeitz, line 57: does that mean we can't have multiple types on a column name, e.g. ip and i64

fulmicoton: Yes.
Only i64, u64, f64 cannot coexist.

src/schema/mod.rs
```diff
@@ -158,6 +160,21 @@ pub fn is_valid_field_name(field_name: &str) -> bool {
     !field_name.is_empty() && !field_name.starts_with('-')
 }
 
+pub(crate) fn convert_to_column_type(typ: Type) -> Option<ColumnType> {
```
PSeitz:
```suggestion
pub(crate) fn value_type_to_column_type(typ: Type) -> Option<ColumnType> {
```

columnar/src/columnar/merge/mod.rs
```diff
@@ -174,97 +192,174 @@ fn merge_column(
...
+        // At the moment, only the numerical categorical column type has more than one possible
+        // column type.
+        assert_eq!(self.column_category, ColumnTypeCategory::Numerical);
+        merged_numerical_columns_type(self.columns.iter().flatten()).into()
+    }
+}
+
+/// Returns the type of the merged numerical column.
+///
+/// Otherwise, this function picks the first numerical type out of i64, u64, f64 (order matters
```
PSeitz:
```suggestion
/// This function picks the first numerical type out of i64, u64, f64 (order matters
```

columnar/src/columnar/merge/mod.rs
```diff
@@ -174,97 +192,174 @@ fn merge_column(
...
+                ),
+            ));
         }
+        Ok(column)
+    }
+}
+
+fn min_max_if_numerical(column: &DynamicColumn) -> Option<(NumericalValue, NumericalValue)> {
+    match column {
+        DynamicColumn::I64(column) => Some((column.min_value().into(), column.max_value().into())),
```
PSeitz, line 363: what happens for empty columns?

fulmicoton: This is not specific to this PR but I added a comment and a unit test.

PSeitz (approved):

PR #1888, opened 2023-02-21
remove schema in aggs

PSeitz:
- switch to ColumnType, move tests
- remove Schema dependency in agg

remove schema dependency in preparation to support aggregation on JSON fields


columnar/Cargo.toml
```diff
@@ -17,6 +17,7 @@ stacker = { path = "../stacker", package="tantivy-stacker"}
 sstable = { path = "../sstable", package = "tantivy-sstable" }
 common = { path = "../common", package = "tantivy-common" }
 tantivy-bitpacker = { version= "0.3", path = "../bitpacker/" }
+serde = "1.0.152"
```
fulmicoton, line 20: Can we do `serde="1"` ?

PSeitz: There's some recommendation to be explicit here: https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277
It's also the default behavior of `cargo add`

fulmicoton: yeah I was scared we could end up with two versions of serde in quickwit but as long as we don't use = on either side, it will take the most recent compatible version.

PSeitz: `=` is setting the minimum patch version, not the exact version. We should have one version in quickwit even with

```
serde = "1.0.152"
...
serde = "1"
..
serde = "1.0.133"
```

fulmicoton (approved):

PR #1892, opened 2023-02-21
Expose building  query from UserInputAst [W.I.P]

evanxg852000:
This is related  to  https://github.com/quickwit-oss/quickwit/pull/2863 therefore subject to discussion


adamreichold:
I would also find exposing this useful as I currently have to parse the query twice to feed some auxiliary system based on the raw/untokenized terms.

PR #1898, opened 2023-02-22
switch to sparse collection for histogram

### LINKED ISSUE #1370: Histogram: Smarter bounds for bucket creation
Currently the histogram pre-creates buckets in the segment collector for the whole data range in the fast field, adjusted by `hard_bounds` or `extended_bounds` if set.

In a use case where the histogram is a sub_aggregation from another aggregation on the same field which applies the some bounds, these bounds should be forwarded.

E.g.

Range on a timestamp, one bucket per month, and inside each bucket a histogram on the timestamp.

### LINKED ISSUE #1704: Evaluate sparse histogram collection
Currently histogram buckets are densely pre-created depending on the fastfields min max values and passed bounds. This allows fast computation of the bucket pos for incoming values.

In some scenarios this may cause issue with max bucket limit (defaults to 65000) and server memory consumption

#### Data example
```rust
{"value": 0}
{"value": 1_000_000_000}
```

A histogram query on `value` with `interval` 1 and `min_doc_count` > 0 would create 1 billion buckets, and potentially overload the server, even though only 2 buckets will be returned.

An alternative would be to have sparse histogram collection. It could also be a hybrid of lazy dense collection with automatic switching to sparse.

Sparse histogram collection may have a potential reuse capability with a a future date histogram.

Related Issues https://github.com/quickwit-oss/tantivy/pull/1703, https://github.com/quickwit-oss/quickwit/issues/2503

PSeitz:
Replaces histogram vec collection with a hashmap. This approach works much better for sparse data and enables use cases like drill downs (filter + small interval).
It is slower for dense cases (1.3x-2x slower). This can maybe be alleviated with a specialized hashmap in the future.
closes #1704
closes #1370

### DIFF
--- a/src/aggregation/bucket/histogram/histogram.rs
+++ b/src/aggregation/bucket/histogram/histogram.rs
@@ -203,8 +203,8 @@ impl SegmentHistogramBucketEntry {
 #[derive(Clone, Debug)]
 pub struct SegmentHistogramCollector {
     /// The buckets containing the aggregation data.
-    buckets: FxHashMap<u64, SegmentHistogramBucketEntry>,
-    sub_aggregations: FxHashMap<u64, Box<dyn SegmentAggregationCollector>>,
+    buckets: FxHashMap<i64, SegmentHistogramBucketEntry>,
+    sub_aggregations: FxHashMap<i64, Box<dyn SegmentAggregationCollector>>,
     sub_aggregation_blueprint: Option<Box<dyn SegmentAggregationCollector>>,
     column_type: ColumnType,
     interval: f64,
@@ -250,7 +250,7 @@ impl SegmentAggregationCollector for SegmentHistogramCollector {
         let bounds = self.bounds;
         let interval = self.interval;
         let offset = self.offset;
-        let get_bucket_pos = |val| (get_bucket_pos_f64(val, interval, offset) as i64) as u64;
+        let get_bucket_pos = |val| (get_bucket_pos_f64(val, interval, offset) as i64);
 
         for doc in docs {
             for val in accessor.values(*doc) {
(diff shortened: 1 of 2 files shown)


src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -289,65 +289,21 @@ impl SegmentHistogramCollector {
...
-            {
+        let mut buckets = Vec::with_capacity(self.buckets.len());
+
+        if self.sub_aggregation_blueprint.is_some() {
+            for bucket_res in self.buckets.into_iter().map(|(bucket_pos, bucket)| {
+                bucket.into_intermediate_bucket_entry(
+                    self.sub_aggregations.get(&bucket_pos).unwrap().clone(),
+                    &agg_with_accessor.sub_aggregation,
+                )
+            }) {
```
fulmicoton: The map seems useless here..
```rust
for (bucket_pos, bucket) in  self.buckets {
    let  sub_aggregation = self.sub_aggregations.get(&bucket_pos).unwrap().clone();
      let bucket_res =
    bucket.into_intermediate_bucket_entry(sub_aggregation,
                    &agg_with_accessor.sub_aggregation,
                )
}
```

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -359,104 +315,70 @@ impl SegmentHistogramCollector {
...
         &mut self,
-        bucket_pos: usize,
+        bucket_pos: u64,
         doc: DocId,
         bucket_with_accessor: &AggregationsWithAccessor,
+        interval: f64,
+        offset: f64,
     ) -> crate::Result<()> {
-        let bucket = &mut self.buckets[bucket_pos];
+        let bucket = &mut self.buckets.entry(bucket_pos).or_insert_with(|| {
```
fulmicoton: I don't think you need the &mut here do you?

fulmicoton (approved):

PR #1899, opened 2023-02-22
allow limiting the number of matched term in range query

### LINKED ISSUE #1886: PhrasePrefixQuery doesn't respect max_expansion when only one term is provided

trinity-1686a:
fix #1886 

I choose to not make `RangeQuery::limit` public because of its odd semantic. A final user should almost always use a collector to get the top docs instead.

I did not implement it for the fastfield range weights, as it seems non trivial due to the difference in encoding, and not useful for PhrasePrefixQueries

### DIFF
--- a/src/query/range_query/range_query.rs
+++ b/src/query/range_query/range_query.rs
@@ -70,6 +70,7 @@ pub struct RangeQuery {
     value_type: Type,
     left_bound: Bound<Vec<u8>>,
     right_bound: Bound<Vec<u8>>,
+    limit: Option<u64>,
 }
 
 impl RangeQuery {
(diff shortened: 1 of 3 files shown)


src/query/range_query/range_query.rs
```diff
@@ -364,7 +385,9 @@ impl Weight for RangeWeight {
         let inverted_index = reader.inverted_index(reader.schema().get_field(&self.field)?)?;
         let term_dict = inverted_index.terms();
         let mut term_range = self.term_range(term_dict)?;
-        while term_range.advance() {
+        let mut processed_count = 0;
+        while term_range.advance() && self.limit.map_or(true, |limit| processed_count < limit) {
```
fulmicoton: nitpick: (meaning no need to fix) `map_or` is not helping readability.

```rust
if let Some(limit) = self.limit {
    if processed_coutn < limit {
          break;
    }
}
```

within the body is my favorite solution on something like that.

fulmicoton (approved):

PR #1900, opened 2023-02-23
add date_histogram

PSeitz:
add date_histogram (DateHistogramReq) to aggregations

Limitations:
- only fixed_interval is supported, calendar-aware features like calendar_interval are not supported
- format is unsupported. Fixed to Rfc3339 currently.
- bounds are not yet supported to be provided via String, e.g. "2015-01-01T00:00:00Z"

### DIFF
--- a/src/aggregation/agg_req.rs
+++ b/src/aggregation/agg_req.rs
@@ -110,15 +110,13 @@ impl BucketAggregationInternal {
             _ => None,
         }
     }
-    pub(crate) fn as_histogram(&self) -> Option<HistogramAggregation> {
+    pub(crate) fn as_histogram(&self) -> crate::Result<Option<HistogramAggregation>> {
         match &self.bucket_agg {
-            BucketAggregationType::Histogram(histogram) => Some(histogram.clone()),
-            BucketAggregationType::DateHistogram(histogram) => Some(
-                histogram
-                    .to_histogram_req()
-                    .expect("could not convert date histogram req to histogram req"),
-            ),
-            _ => None,
+            BucketAggregationType::Histogram(histogram) => Ok(Some(histogram.clone())),
+            BucketAggregationType::DateHistogram(histogram) => {
+                Ok(Some(histogram.to_histogram_req()?))
+            }
+            _ => Ok(None),
         }
     }
     pub(crate) fn as_term(&self) -> Option<&TermsAggregation> {
--- a/src/aggregation/agg_req.rs
+++ b/src/aggregation/agg_req.rs
@@ -50,7 +50,7 @@ use std::collections::{HashMap, HashSet};
 use serde::{Deserialize, Serialize};
 
 pub use super::bucket::RangeAggregation;
-use super::bucket::{HistogramAggregation, TermsAggregation};
+use super::bucket::{DateHistogramAggregationReq, HistogramAggregation, TermsAggregation};
 use super::metric::{
     AverageAggregation, CountAggregation, MaxAggregation, MinAggregation, StatsAggregation,
     SumAggregation,
@@ -110,9 +110,14 @@ impl BucketAggregationInternal {
             _ => None,
         }
     }
-    pub(crate) fn as_histogram(&self) -> Option<&HistogramAggregation> {
+    pub(crate) fn as_histogram(&self) -> Option<HistogramAggregation> {
         match &self.bucket_agg {
-            BucketAggregationType::Histogram(histogram) => Some(histogram),
+            BucketAggregationType::Histogram(histogram) => Some(histogram.clone()),
+            BucketAggregationType::DateHistogram(histogram) => Some(
+                histogram
+                    .to_histogram_req()
+                    .expect("could not convert date histogram req to histogram req"),
+            ),
             _ => None,
         }
     }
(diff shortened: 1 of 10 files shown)


src/aggregation/agg_req.rs
```diff
@@ -110,9 +110,14 @@ impl BucketAggregationInternal {
...
     }
-    pub(crate) fn as_histogram(&self) -> Option<&HistogramAggregation> {
+    pub(crate) fn as_histogram(&self) -> Option<HistogramAggregation> {
         match &self.bucket_agg {
-            BucketAggregationType::Histogram(histogram) => Some(histogram),
+            BucketAggregationType::Histogram(histogram) => Some(histogram.clone()),
+            BucketAggregationType::DateHistogram(histogram) => Some(
+                histogram
+                    .to_histogram_req()
+                    .expect("could not convert date histogram req to histogram req"),
```
fulmicoton: I have no clue if this expect is justified or not... O_o

Can you remove it?

Expect is ok if it is reasonably obvious that it should never happen.

A good rule of thumb...
It is not ok if it depends on code running in a different file or a different computer to do "the right thing".

If it is something justified (it should never happen) trivially by the surrounding code, then expect is perfectly fine.

src/aggregation/bucket/histogram/date_histogram.rs
```diff
@@ -55,42 +66,144 @@ pub struct DateHistogramAggregationReq {
...
+    #[error("Unit not found in passed String {0:?}")]
     UnitMissing(String),
+    /// Offset invalid
+    #[error("passed offset is invalid {0:?}")]
+    InvalidOffset(String),
 }
 
-fn parse_into_milliseconds(input: &str) -> Result<u64, DateHistogramParseError> {
+fn parse_offset_into_microseconds(input: &str) -> Result<i64, AggregationError> {
+    let is_sign = |byte| &[byte] == b"-" || &[byte] == b"+";
```
fulmicoton, line 175: Why do we need to parse numbers?
`str::parse` does not work here?

fulmicoton: ah ok, so you wanted to reuse `parse_into_microseconds`, but offset can be negative or something like that?

src/aggregation/bucket/histogram/date_histogram.rs
```diff
@@ -55,42 +66,144 @@ pub struct DateHistogramAggregationReq {
...
     /// Unit not found in passed String
+    #[error("Unit not found in passed String {0:?}")]
     UnitMissing(String),
+    /// Offset invalid
+    #[error("passed offset is invalid {0:?}")]
+    InvalidOffset(String),
 }
 
-fn parse_into_milliseconds(input: &str) -> Result<u64, DateHistogramParseError> {
+fn parse_offset_into_microseconds(input: &str) -> Result<i64, AggregationError> {
```
fulmicoton, line 174: I did not find any unit tests.

src/aggregation/bucket/histogram/date_histogram.rs
fulmicoton: no really I don't get it... O_o

src/aggregation/bucket/histogram/date_histogram.rs
fulmicoton, line 174: I did not find unit test for this function, especially testing negative offset is important.

src/aggregation/bucket/histogram/date_histogram.rs
```diff
@@ -55,42 +66,144 @@ pub struct DateHistogramAggregationReq {
...
+    if input.is_empty() {
+        return Err(DateHistogramParseError::InvalidOffset(input.to_string()).into());
+    }
+
+    let has_sign = is_sign(input.as_bytes()[0]);
+    if has_sign {
+        let (sign, input) = input.split_at(1);
+        let mut val = parse_into_microseconds(input)?;
+        if sign == "-" {
+            val *= -1;
```
fulmicoton: This can overflow.

fulmicoton: nitpick,

```
if sign == "-" {
   -val
} else {
   val
}
```

would have been a tiny bit more readable... But overall I am missing the point of the whole fucntion.

fulmicoton (changes requested):
see comments inline

Can you help me understand the point of `parse_offset_into_microseconds` ?

src/aggregation/agg_req.rs
PSeitz: The method can fail in general, because of the parse into microseconds conversion, but not at this time, since the conversion was already done before.

I think I can change it to return `Result<Option>`

src/aggregation/bucket/histogram/date_histogram.rs
PSeitz: There's a test that tests
`"offset": "-4d"`

src/aggregation/bucket/histogram/date_histogram.rs
PSeitz: > ah ok, so you wanted to reuse parse_into_microseconds, but offset can be negative or something like that?

Yes, that's it. The sign is optional

e.g. offset: `5d` or `-5d`

src/aggregation/bucket/histogram/date_histogram.rs
PSeitz: Not a this point, that's an i64

src/aggregation/bucket/histogram/date_histogram.rs
PSeitz: I added a unit test

fulmicoton (approved):

PR #1904, opened 2023-02-24
add coerce option for text and numbers types

PSeitz:
allow to coerce the field type when indexing if the type does not match

### DIFF
--- a/src/schema/field_type.rs
+++ b/src/schema/field_type.rs
@@ -346,7 +346,7 @@ impl FieldType {
                     }
                     FieldType::I64(opt) => {
                         if opt.should_coerce() {
-                            Ok(Value::U64(field_text.parse().map_err(|_| {
+                            Ok(Value::I64(field_text.parse().map_err(|_| {
                                 ValueParsingError::TypeError {
                                     expected: "a i64 or a i64 as string",
                                     json: JsonValue::String(field_text),
@@ -361,7 +361,7 @@ impl FieldType {
                     }
                     FieldType::F64(opt) => {
                         if opt.should_coerce() {
-                            Ok(Value::U64(field_text.parse().map_err(|_| {
+                            Ok(Value::F64(field_text.parse().map_err(|_| {
                                 ValueParsingError::TypeError {
                                     expected: "a f64 or a f64 as string",
                                     json: JsonValue::String(field_text),
--- a/src/schema/field_type.rs
+++ b/src/schema/field_type.rs
@@ -374,10 +374,21 @@ impl FieldType {
                             })
                         }
                     }
-                    FieldType::Bool(_) => Err(ValueParsingError::TypeError {
-                        expected: "a boolean",
-                        json: JsonValue::String(field_text),
-                    }),
+                    FieldType::Bool(opt) => {
+                        if opt.should_coerce() {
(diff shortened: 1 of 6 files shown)


src/schema/field_type.rs
```diff
@@ -329,11 +329,50 @@ impl FieldType {
...
+                        } else {
+                            Err(ValueParsingError::TypeError {
+                                expected: "a u64",
+                                json: JsonValue::String(field_text),
+                            })
+                        }
+                    }
+                    FieldType::I64(opt) => {
+                        if opt.should_coerce() {
+                            Ok(Value::U64(field_text.parse().map_err(|_| {
```
fulmicoton:
```suggestion
                            Ok(Value::I64(field_text.parse().map_err(|_| {
```

src/schema/field_type.rs
```diff
@@ -329,11 +329,50 @@ impl FieldType {
...
+                        } else {
+                            Err(ValueParsingError::TypeError {
+                                expected: "a i64",
+                                json: JsonValue::String(field_text),
+                            })
+                        }
+                    }
+                    FieldType::F64(opt) => {
+                        if opt.should_coerce() {
+                            Ok(Value::U64(field_text.parse().map_err(|_| {
```
fulmicoton:
```suggestion
                            Ok(Value::F64(field_text.parse().map_err(|_| {
```

src/schema/field_type.rs
```diff
@@ -329,11 +329,50 @@ impl FieldType {
                         Ok(DateTime::from_utc(dt_with_fixed_tz).into())
                     }
                     FieldType::Str(_) => Ok(Value::Str(field_text)),
-                    FieldType::U64(_) | FieldType::I64(_) | FieldType::F64(_) => {
-                        Err(ValueParsingError::TypeError {
-                            expected: "an integer",
-                            json: JsonValue::String(field_text),
-                        })
+                    FieldType::U64(opt) => {
+                        if opt.should_coerce() {
```
fulmicoton, line 333: We need unit tests for this.

fulmicoton (changes requested):
Unit tests needed (that's a good change though)

PSeitz:
> Unit tests needed (that's a good change though)

Yes, I was just checking first if that's the right direction

src/schema/numeric_options.rs
```diff
@@ -17,6 +18,12 @@ pub struct NumericOptions {
     fieldnorms: bool, // This attribute only has an effect if indexed is true.
     fast: bool,
     stored: bool,
+    #[serde(skip_serializing_if = "is_false")]
```
fulmicoton, line 21: is there a reason to skip serialization? 
```suggestion
```

src/schema/text_options.rs
```diff
@@ -17,6 +17,14 @@ pub struct TextOptions {
     stored: bool,
     #[serde(default)]
     fast: bool,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "is_false")]
```
fulmicoton, line 21:
```suggestion
```

fulmicoton (approved):

src/schema/numeric_options.rs
PSeitz: just a little less noise when reading the meta.json

PR #1912, opened 2023-02-27
Adapting for quickwit2

fulmicoton:


src/core/json_utils.rs
```diff
@@ -296,6 +296,32 @@ fn split_json_path(json_path: &str) -> Vec<String> {
     json_path_segments
 }
 
+/// Takes a field name, a json path as supplied by a user, and whether we should expand dots, and
+/// return a column key, as expected by the columnar crate.
+///
+/// This function wil detect unescaped dots in the path, and split over them.
```
PSeitz:
```suggestion
/// This function will detect unescaped dots in the path, and split over them.
```

src/fastfield/mod.rs
```diff
@@ -1068,13 +1068,94 @@ mod tests {
...
+        let column_with_opt: Option<StrColumn> = fast_fields.str("with.hello").unwrap();
         let column_with: StrColumn = column_with_opt.unwrap();
         assert!(column_with.term_ords(0).next().is_none());
         assert!(column_with.term_ords(1).eq([0]));
         assert!(column_with.term_ords(2).eq([2]));
         assert!(column_with.term_ords(3).eq([1]));
     }
+
+    #[test]
+    fn test_fast_field_in_json_field_no_expand_dots_disabled() {
```
PSeitz:
```suggestion
    fn test_fast_field_in_json_field_expand_dots_disabled() {
```

src/fastfield/mod.rs
```diff
@@ -1068,13 +1068,94 @@ mod tests {
...
+        let mut index_writer = index.writer_for_tests().unwrap();
+        index_writer
+            .add_document(doc!(field_with_dot => 32i64))
+            .unwrap();
+        index_writer.commit().unwrap();
+        let reader = index.reader().unwrap();
+        let searcher = reader.searcher();
+        let reader = searcher.segment_reader(0u32);
+        let fast_field_reader = reader.fast_fields();
+        for test_column_name in &["field.with.dot"] {
```
PSeitz: clippy

src/fastfield/mod.rs
```diff
@@ -1068,13 +1068,94 @@ mod tests {
...
+                .column_opt::<i64>(test_column_name)
+                .unwrap()
+                .unwrap();
+            let vals: Vec<i64> = column.values_for_doc(0u32).collect();
+            assert_eq!(&vals, &[32]);
+        }
+    }
+
+    #[test]
+    fn test_fast_field_dot_in_schema_field_name() {
```
PSeitz, line 1134: can we add a test for json field with dot

src/fastfield/readers.rs
```diff
@@ -18,16 +19,22 @@ use crate::space_usage::{FieldUsage, PerFieldSpaceUsage};
...
         let columnar = Arc::new(ColumnarReader::open(fast_field_file)?);
-        Ok(FastFieldReaders { columnar })
+        Ok(FastFieldReaders { columnar, schema })
     }
 
-    pub(crate) fn columnar(&self) -> &ColumnarReader {
-        self.columnar.as_ref()
+    fn resolve_field(&self, column_name: &str) -> Option<String> {
+        let default_field_opt: Option<Field> = if cfg!(feature = "quickwit") {
+            self.schema.get_field("_dynamic").ok()
```
PSeitz, line 33: I think it would be better if this was a configurable parameter. Not sure if that would add too much complexity though

src/fastfield/readers.rs
```diff
@@ -46,18 +53,70 @@ impl FastFieldReaders {
...
+    /// `schema.find_field(..)`. This method will attempt to split the user splied fast field
+    /// name by non-escaped dots, and find the longest matching schema field name.
+    /// In our case, it would return the (attribute_field, "color").
+    ///
+    /// If no field is found, but a dynamic field is supplied, then we
+    /// will simply assuem the user is targetting the dynamic field. (This feature is used in
+    /// Quickwit.)
+    ///
+    /// We then encode the `(field, path)` into the right `columnar_key`.
+    fn resolve_column_name_given_default_field<'a>(
```
PSeitz, line 81: shouldn't this be part of `Schema` (maybe replace find_field)?

src/fastfield/mod.rs
fulmicoton: added a unit test

src/fastfield/readers.rs
fulmicoton: I agree...but let's do this in another ticket.
https://github.com/quickwit-oss/tantivy/issues/1918

src/fastfield/readers.rs
fulmicoton: Not really no: that logic is very specific to the quirky way we encode column names.

`schema` does not need to "know" about that.

src/fastfield/mod.rs
fulmicoton: addressed

PR #1923, opened 2023-03-02
remove dynamic dispatch in collect_segment

PSeitz:


src/collector/mod.rs
```diff
@@ -173,28 +173,32 @@ pub trait Collector: Sync + Send {
 
         match (reader.alive_bitset(), self.requires_scoring()) {
             (Some(alive_bitset), true) => {
-                weight.for_each(reader, &mut |doc, score| {
+                let mut scorer = weight.scorer(reader, 1.0)?;
```
fulmicoton, line 176: I think you just disconned blockwand if you do this?

fulmicoton (approved):

Issue #1654, opened 2022-10-29
Indexing an array of small pieces of text is much slower than indexing a big string 


fmassot:
I still don't know if there is room for optimization there but the performance drop is significant. I observed on a dataset with documents containing a big array of small text, a drop of 25% in indexing throughput.

I pushed on a branch a file `hdfs-with-array.json` (hdfs json with a body made of array of text) to reproduce the performance drop in the bench. Here are the result:

```
Benchmarking index-hdfs/index-hdfs-no-commit: Warming up for 3.0000 s
Warning: Unable to complete 20 samples in 5.0s. You may wish to increase target time to 11.4s, or reduce sample count to 10
index-hdfs/index-hdfs-no-commit
                        time:   [538.24 ms 541.50 ms 545.32 ms]
                        change: [-3.6931% -2.4507% -1.2364%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) high mild
Benchmarking index-hdfs/index-hdfs-with-array-no-commit: Warming up for 3.0000 s
Warning: Unable to complete 20 samples in 5.0s. You may wish to increase target time to 17.7s, or reduce sample count to 10.
index-hdfs/index-hdfs-with-array-no-commit
                        time:   [879.80 ms 883.73 ms 888.34 ms]
                        change: [-2.6182% -1.1315% +0.0697%] (p = 0.12 > 0.05)
                        No change in performance detected.
Found 2 outliers among 20 measurements (10.00%)
```

Here is the branch: https://github.com/quickwit-oss/tantivy/tree/fmassot--bench-hdfs-with-array

PSeitz:
here's a diff on the profile.

There's some noise, but we can see the alloc overhead for the array (array itself + multiple strings), slower deserialization, slower value_from_json.

`9.61%     +4.74%  libc.so.6                     [.] 0x000000000005a0d9` could be another alloc call, e.g. `calloc`.

```
Check IO/CPU overload!

# Event 'cycles'
#
# Baseline  Delta Abs  Shared Object                 Symbol                                                                                                                                                                                                                          >
# ........  .........  ............................  ................................................................................................................................................................................................................................>
#
    24.16%     -9.51%  index_bench-075aa16d259e3fd0  [.] crossbeam_channel::flavors::array::Channel<T>::send
     9.61%     +4.74%  libc.so.6                     [.] 0x000000000005a0d9
[...]

PR #1924, opened 2023-03-03
refactor tokenization pipeline to use GATs

trinity-1686a:
while working on #1654, I found a substantial part of the difference between "normal" strings and pre-split strings is the amount of allocation/deallocation done. It's one per string and per filter, as filter take a `BoxTokenStream` and returns a `BoxTokenStream`. This leverage GATs so only the the final `TokenStream` get boxed.

I've also found that pre-allocating a String for the LowerCaser is harmful to performance if some strings may be ascii only, which happens a lot more when strings are pre-split (and making matters worse, a lot more `LowerCaserTokenStream` are also instantiated in that case).

Measuring on `fmassot--bench-hdfs-with-array`, I get a 3% uplift in the general case, and a 19% with pre-split strings.

### DIFF
--- a/src/tokenizer/empty_tokenizer.rs
+++ b/src/tokenizer/empty_tokenizer.rs
@@ -5,7 +5,7 @@ pub(crate) struct EmptyTokenizer;
 
 impl Tokenizer for EmptyTokenizer {
     type TokenStream<'a> = EmptyTokenStream;
-    fn token_stream<'a>(&self, _text: &'a str) -> EmptyTokenStream {
+    fn token_stream(&self, _text: &str) -> EmptyTokenStream {
         EmptyTokenStream::default()
     }
 }
--- a/src/tokenizer/empty_tokenizer.rs
+++ b/src/tokenizer/empty_tokenizer.rs
@@ -1,16 +1,17 @@
-use crate::tokenizer::{BoxTokenStream, Token, TokenStream, Tokenizer};
+use crate::tokenizer::{Token, TokenStream, Tokenizer};
 
 #[derive(Clone)]
 pub(crate) struct EmptyTokenizer;
 
 impl Tokenizer for EmptyTokenizer {
-    fn token_stream<'a>(&self, _text: &'a str) -> BoxTokenStream<'a> {
-        EmptyTokenStream::default().into()
+    type TokenStream<'a> = EmptyTokenStream;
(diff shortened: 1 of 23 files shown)


src/tokenizer/empty_tokenizer.rs
```diff
@@ -1,16 +1,17 @@
...
 
 #[derive(Clone)]
 pub(crate) struct EmptyTokenizer;
 
 impl Tokenizer for EmptyTokenizer {
-    fn token_stream<'a>(&self, _text: &'a str) -> BoxTokenStream<'a> {
-        EmptyTokenStream::default().into()
+    type TokenStream<'a> = EmptyTokenStream;
+    fn token_stream<'a>(&self, _text: &'a str) -> EmptyTokenStream {
+        EmptyTokenStream::default()
```
fulmicoton, line 9: can you address clippy's comment?

fulmicoton, line 9:
```suggestion
        EmptyTokenStream
```

src/tokenizer/raw_tokenizer.rs
```diff
@@ -11,7 +10,8 @@ pub struct RawTokenStream {
 }
 
 impl Tokenizer for RawTokenizer {
-    fn token_stream<'a>(&self, text: &'a str) -> BoxTokenStream<'a> {
+    type TokenStream<'a> = RawTokenStream;
+    fn token_stream<'a>(&self, text: &'a str) -> RawTokenStream {
```
fulmicoton: clippy

tokenizer-api/src/lib.rs
```diff
@@ -139,39 +144,48 @@ pub trait TokenStream {
...
-        BoxTokenFilter(Box::new(tokenizer))
-    }
-}
-
-pub trait TokenFilterClone {
-    fn box_clone(&self) -> BoxTokenFilter;
-}
+// Simple wrapper of `Box<dyn TokenFilter + 'a>`.
+//
+// See [`TokenFilter`] for more information.
```
fulmicoton: let's remove this.

fulmicoton (approved):
Good job. 
See inline comments.

src/tokenizer/empty_tokenizer.rs
trinity-1686a: `EmptyTokenStream` isn't an empty struct, it contains a default `Token`

PR #1927, opened 2023-03-06
fix clippy

PSeitz:
fix clippy

### DIFF
--- a/src/aggregation/agg_tests.rs
+++ b/src/aggregation/agg_tests.rs
@@ -208,36 +208,36 @@ fn test_aggregation_level1() -> crate::Result<()> {
         ("average".to_string(), get_avg_req("score")),
         (
             "range".to_string(),
-            Aggregation::Bucket(BucketAggregation {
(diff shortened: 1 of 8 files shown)


fulmicoton:
if it is not performance critical i'd rather have boxing T_T

PSeitz:
In this case there are a lot of tests that would need to be adjusted

fulmicoton:
@PSeitz let's adjust the tests

src/aggregation/agg_tests.rs
```diff
@@ -590,7 +590,7 @@ fn test_aggregation_invalid_requests() -> crate::Result<()> {
     Ok(())
 }
 
-#[cfg(all(test, feature = "unstable"))]
+//#[cfg(all(test, feature = "unstable"))]
```
fulmicoton: this is a mistake i believe?

fulmicoton (approved):
approved to make the exchange smooth, but have a look at the comment in agg_test.
I think you commented out some feature fence by mistake.

PR #1930, opened 2023-03-06
fix panic in dict column merge

PSeitz:
seems not to fix the issue reported in discord #1929, but I think it does fix a possible issue

### DIFF
--- a/columnar/src/columnar/merge/merge_dict_column.rs
+++ b/columnar/src/columnar/merge/merge_dict_column.rs
@@ -52,20 +52,18 @@ impl<'a> Iterable for RemappedTermOrdinalsValues<'a> {
 
 impl<'a> RemappedTermOrdinalsValues<'a> {
     fn boxed_iter_stacked(&self) -> Box<dyn Iterator<Item = u64> + '_> {
-        let iter = self
-            .bytes_columns
-            .iter()
-            .enumerate()
-            .flat_map(|(segment_ord, byte_column)| {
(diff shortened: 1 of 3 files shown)


PSeitz:
https://github.com/quickwit-oss/tantivy/issues/1929

columnar/src/columnar/merge/merge_dict_column.rs
```diff
@@ -52,20 +52,18 @@ impl<'a> Iterable for RemappedTermOrdinalsValues<'a> {
...
-                        .ords()
-                        .values
-                        .iter()
-                        .map(move |term_ord| segment_ord[term_ord as usize])
-                })
-            });
+        let mut ord = 0;
+        let iter = self.bytes_columns.iter().flat_map(move |byte_column| {
+            let segment_ord = self.term_ord_mapping.get_segment(ord as u32);
+            ord += 1;
```
fulmicoton: I don't understand the difference !?

ChillFish8: If a column is `None`, it won't increase the ordinal. 

The original would still increase the ordinal regardless of if it was `None` or not IIRC

fulmicoton: Thanks! That is not how flat_map works however. It is called on None as well.
I suspect this is the bug you encounterred. I think I fixed it, can you try and reproduce with the last commit?

fulmicoton (approved):

PR #1933, opened 2023-03-07
Export vint codec from postings::compression

sujayakar:
I'd like to use Tantivy's `vint` compression in one of my projects, and this PR is a first pass at making it public. I haven't added docstrings to the newly public exports since I'm not sure this is the right approach.

Would it be better to extract these functions into a new `tantivy-vint` crate underneath the repo root?


fulmicoton:
No this is internal stuff. Feel free to copy that code and create a crate. Don't forget to keep the MIT license for that code however.

sujayakar:
sounds good!

PR #1936, opened 2023-03-13
fix validation in date histogram

PSeitz:
fix validation in date histogram for parameters interval and date_interval


src/aggregation/bucket/histogram/date_histogram.rs
```diff
@@ -112,7 +114,7 @@ impl DateHistogramAggregationReq {
         self.validate()?;
         Ok(HistogramAggregation {
             field: self.field.to_string(),
-            interval: parse_into_microseconds(&self.fixed_interval)? as f64,
+            interval: parse_into_microseconds(&self.fixed_interval.as_ref().unwrap())? as f64,
```
fulmicoton: clippy

fulmicoton (approved):

PR #1940, opened 2023-03-14
perf: faster term hash map

PSeitz:
perf: faster arena hashmap

add inlines
faster initialization by replacing iter with `vec!`
remove occupied array and use table_entry.is_empty instead (saves 4 bytes per entry)
reduce saturation threshold from 1/3 to 1/2 to reduce memory consumption
use u32 for UnorderedId (we have the 4billion limit anyways on the Columnar stuff)
fix naming LinearProbing
remove byteorder dependency

### DIFF
--- a/stacker/src/arena_hashmap.rs
+++ b/stacker/src/arena_hashmap.rs
@@ -1,7 +1,5 @@
-use std::{iter, mem, slice};
-
-use byteorder::{ByteOrder, NativeEndian};
-use murmurhash32::murmurhash2;
+use std::iter::{Cloned, Filter};
+use std::mem;
 
 use super::{Addr, MemoryArena};
 use crate::memory_arena::store;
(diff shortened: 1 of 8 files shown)


stacker/src/arena_hashmap.rs
```diff
@@ -50,43 +55,47 @@ impl KeyValue {
 /// the computation of the hash of the key twice,
 /// or copying the key as long as there is no insert.
 pub struct ArenaHashMap {
-    table: Box<[KeyValue]>,
+    table: Vec<KeyValue>,
```
fulmicoton, line 58: Is this helpful?

stacker/src/arena_hashmap.rs
```diff
@@ -205,15 +221,13 @@ impl ArenaHashMap {
         let new_len = (self.table.len() * 2).max(1 << 13);
         let mask = new_len - 1;
         self.mask = mask;
-        let new_table = vec![KeyValue::default(); new_len].into_boxed_slice();
+        let new_table = vec![KeyValue::default(); new_len];
         let old_table = mem::replace(&mut self.table, new_table);
-        for old_pos in self.occupied.iter_mut() {
-            let key_value: KeyValue = old_table[*old_pos];
-            let mut probe = QuadraticProbing::compute(key_value.hash as usize, mask);
+        for key_value in old_table.into_iter().filter(|kv| !kv.is_empty()) {
```
fulmicoton:
```suggestion
        for key_value in old_table.into_iter().filter(KeyValue::is_not_empty_ref) {
```

fulmicoton (approved):

fulmicoton (approved):

PR #1942, opened 2023-03-14
add memory limit for aggregations

PSeitz:
introduce AggregationLimits to set memory consumption limit and bucket limits
memory limit is checked during aggregation, bucket limit is checked before returning the aggregation request.

TODO: add check for term aggregation

### DIFF
--- a/src/aggregation/agg_limits.rs
+++ b/src/aggregation/agg_limits.rs
@@ -1,0 +1,91 @@
+use std::collections::HashMap;
+use std::sync::atomic::AtomicU64;
+use std::sync::Arc;
+
+use super::collector::DEFAULT_MEMORY_LIMIT;
+use super::{AggregationError, DEFAULT_BUCKET_LIMIT};
+use crate::TantivyError;
+
+/// An estimate for memory consumption
+pub trait MemoryConsumption {
+    fn memory_consumption(&self) -> usize;
+}
+
+impl<K, V, S> MemoryConsumption for HashMap<K, V, S> {
+    fn memory_consumption(&self) -> usize {
+        let num_items = self.len();
+        (std::mem::size_of::<K>() + std::mem::size_of::<V>()) * num_items
+    }
+}
+
+/// Aggregation memory limit after which the request fails. Defaults to DEFAULT_MEMORY_LIMIT
+/// (500MB). The limit is shared by all SegmentCollectors
+pub struct AggregationLimits {
+    /// The counter which is shared between the aggregations for one request.
+    memory_consumption: Arc<AtomicU64>,
+    /// The memory_limit in bytes
+    memory_limit: u64,
+    /// The maximum number of buckets _returned_
+    /// This is not counting intermediate buckets.
+    bucket_limit: u32,
+}
+impl Clone for AggregationLimits {
+    fn clone(&self) -> Self {
+        Self {
+            memory_consumption: Arc::clone(&self.memory_consumption),
+            memory_limit: self.memory_limit,
+            bucket_limit: self.bucket_limit,
+        }
+    }
+}
+
+impl Default for AggregationLimits {
+    fn default() -> Self {
+        Self {
+            memory_consumption: Default::default(),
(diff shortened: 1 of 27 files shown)


src/aggregation/agg_limits.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// An estimate for memory consumption
+pub trait MemoryConsumption {
+    fn memory_consumption(&self) -> usize;
+}
+
+impl<K, V, S> MemoryConsumption for HashMap<K, V, S> {
+    fn memory_consumption(&self) -> usize {
+        let num_items = self.len();
+        (std::mem::size_of::<K>() + std::mem::size_of::<V>()) * num_items
```
fulmicoton:
```suggestion
        let num_items = self.capacity();
        (std::mem::size_of::<K>() + std::mem::size_of::<V>()) * num_items
```

src/aggregation/agg_limits.rs
```diff
@@ -0,0 +1,91 @@
...
+use super::collector::DEFAULT_MEMORY_LIMIT;
+use super::{AggregationError, DEFAULT_BUCKET_LIMIT};
+use crate::TantivyError;
+
+/// An estimate for memory consumption
+pub trait MemoryConsumption {
+    fn memory_consumption(&self) -> usize;
+}
+
+impl<K, V, S> MemoryConsumption for HashMap<K, V, S> {
```
fulmicoton, line 16:
```suggestion
impl<K: Copy, V: Copy, S> MemoryConsumption for HashMap<K, V, S> {
```

src/aggregation/error.rs
```diff
@@ -3,7 +3,29 @@ use super::bucket::DateHistogramParseError;
 /// Error that may occur when opening a directory
 #[derive(Debug, Clone, PartialEq, Eq, Error)]
 pub enum AggregationError {
-    /// Failed to open the directory.
+    /// Date histogram parse error
     #[error("Date histogram parse error: {0:?}")]
     DateHistogramParseError(#[from] DateHistogramParseError),
+    /// Memory limit exceeded
+    #[error(
+        "Aborting aggregation because memory limit was exceeded. Limit: {limit:?}, Current: \
```
fulmicoton, line 13: that would be nice to display a value expressed in MB, with a unit.

src/error.rs
```diff
@@ -55,7 +55,7 @@ impl fmt::Debug for DataCorruption {
 #[derive(Debug, Clone, Error)]
 pub enum TantivyError {
     /// Error when handling aggregations.
-    #[error("AggregationError {0:?}")]
+    #[error(transparent)]
```
fulmicoton, line 58: ah cool ! I did not know about `transparent`

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -287,6 +296,12 @@ impl SegmentAggregationCollector for SegmentHistogramCollector {
 }
 
 impl SegmentHistogramCollector {
+    fn get_memory_consumption(&self) -> usize {
```
fulmicoton, line 299: we should really add byte-unit as a dependency of tantivy at one point... if you agree can you open a ticket?

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -287,6 +296,12 @@ impl SegmentAggregationCollector for SegmentHistogramCollector {
...
+        let self_mem = std::mem::size_of::<Self>();
+        let sub_aggs = self.sub_aggregations.memory_consumption();
+        let buckets = self.buckets.memory_consumption();
+        self_mem + sub_aggs + buckets
```
```suggestion
        let self_mem = std::mem::size_of::<Self>();
        let sub_aggs_mem = self.sub_aggregations.memory_consumption();
        let buckets_mem = self.buckets.memory_consumption();
        self_mem + sub_aggs_mem + buckets_mem
```

fulmicoton (approved):

src/aggregation/error.rs
PSeitz: I changed the existing `ByteCount` to cover that

Issue #1827, opened 2023-01-24
encode dictionary type (fst, sstable)


PSeitz:
the dictionary is currently controlled by a feature flag to be an fst or sstable.

To make the index forward compatible we should store the dictionary type in the metadata

trinity-1686a:
#1943 adds a footer containing the dictionary type and its version, and implement its usage for sstables. The fst part isn't implemented yet, nor is a test that read one kind of dictionary with the other reader

Issue #1738, opened 2022-12-23
Improve SSTable format... or find another dictionary format?


fulmicoton:
inverse vint implementation.
~~remove serde_cbor.~~ removed in #1943 
add multilevel indexing?

oronsh:
Hi @fulmicoton !
may I naively ask why the sstable couldn't actually be implemented as `Vec<Fst>` rather than `Vec<Block>`?
it seemed like fst was great for a local tantivy but a problem with quickwit since you need to download the entire dictionary.
so theoretically, could the sstable be implemented from multiple fsts each contains subset of the keys range?

fulmicoton:
Yes you are correct. Having a bunch of fst blocks would solve the IO problem too.

Another reason we picked  `sstable` is because iterating through them is much faster and we originally wanted to build analytics based on this.

oronsh:
I see, so fsts would have better `get` performance where the sstable blocks would have better iteration performance where the get op is used to for the usual search queries and iteration for analytics?
In that case, may I ask how worse the performance of `get` with sstable blocks compared to fst? would it really affect queries that much?

Issue #1727, opened 2022-12-21
Add version/magic number in SSTable format.


fulmicoton:
- ~~add version & magic number in SSTable format. At the dictionary scale, not in each block.~~ done in #1943 
- remove `sstable::Dictionary::new` and use `sstable::Dictionary::create`
- refactor the magic number/ version logic.

PR #1943, opened 2023-03-16
 new sstable format

### LINKED ISSUE #1731: Make SSTable block size  configurable
... set a very small block size in prop tests to cover move ground.

### LINKED ISSUE #1851: Revisit the sstable format
The sstable format currently relies on serde cbor to encode its format. It was a decent shortcut at the time, but it would be great to remove this:
- to remove the dependency.
- to have a well-spec'ed format. 
- to have proper versioning in the format 
- to get something more compact... Right now a tiny dictionary can often top 100 bytes.

Ideally, it would be nice for the candidate format to help with the lookup.
We want the sstable index to be as small as possible as it is shipped into the hotcache.
A format that is 2x compact, means that we can have 2x smaller blocks, which in turn means 2x faster looker.

Maybe a 1 level sstable, or a k-level sstable pyramid could be a good solution.

trinity-1686a:
this implements and document a new format for sstables. Most of the format is the same as before, except the index is now a small sstable too instead of a cbor blob.

fix #1851
- [x] remove the dependency.
- [x] have a well-spec'ed format.
- [x] have proper versioning in the format
- [x]  get something more compact... Right now a tiny dictionary can often top 100 bytes.

groundwork for #1827 (encode dictionary type, but only for sstable, fst kept unchanged for now)

fix some of #1738
- [ ] inverse vint implementation (not sure what this is)
- [x] remove serde_cbor.
- [ ] add multilevel indexing (in some way this is a 2 level index I guess?)

fix #1731

fix some of #1727
- [x] add version & magic number in SSTable format. At the dictionary scale, not in each block.
- [ ] remove `sstable::Dictionary::new` and use `sstable::Dictionary::create`
- [ ] refactor the magic number/ version logic.

Size comparison with previous format:
- tests in `src/fastfield/mod.rs` and `columnar/src/tests.rs` show the difference for small indexes
- test payload in `sstable/src/lib.rs` went from 115 to 53 bytes
- on a split of exactly one day of gh-archive, the size of the "footer" went from 1075365 (1MB) to 86399 (84kB), ~92% smaller

The improvement comes from three sources:
- cbor keep the name of the fields as it aims to encode the same kind of data as json, so an index for a single block might look something like this: `$fblocks$$slast_key_or_greater$!!jblock_addr$jbyte_range$estartcend$mfirst_ordinal`, where there is more field name than actual metadata (impact both small and large indexes)
- using sstable allows to merge key prefixes when there are blocks ending with similar enough keys (impact only large indexes)
- block range and first_ord are encoded as delta, start offset are implicit, derived from the length of previous blocks (impact both small and large indexes)

This could probably be made to support reading the old format if we want some amount of backward compatibility, at the cost of reintroducing cbor

### DIFF
--- a/sstable/README.md
+++ b/sstable/README.md
@@ -87,9 +87,10 @@ Note: there is no ambiguity between both representation as Add is always guarant
 
 ### SSTFooter
 ```
-+-------+-------------+---------+---------+------+
-| Block | IndexOffset | NumTerm | Version | Type |
-+-------+-------------+---------+---------+------+
++-------+-------+-----+-------------+---------+---------+------+
+| Block | Block | ... | IndexOffset | NumTerm | Version | Type |
++-------+-------+-----+-------------+---------+---------+------+
+|----( # of blocks)---|
 ```
 - Block(SSTBlock): uses IndexValue for its Values format
 - IndexOffset(u64): Offset to the start of the SSTFooter
--- a/sstable/README.md
+++ b/sstable/README.md
@@ -38,7 +38,7 @@ Overview of the SSTable format. Unless noted otherwise, numbers are little-endia
 +-------+-------+-----+--------+
 |----( # of blocks)---|
 ```
-- Block(`SSTBlock`): list of independant block, terminated by a single empty block.
+- Block(`SSTBlock`): list of independent block, terminated by a single empty block.
 - Footer(`SSTFooter`)
 
 ### SSTBlock
@@ -48,8 +48,8 @@ Overview of the SSTable format. Unless noted otherwise, numbers are little-endia
 +----------+--------+-------+-------+-----+
                     |----( # of deltas)---|
 ```
-- BlockLen(u32): lenght of the block
-- Values: an application defined format storing a sequence of value, capable of determining it own lenght
+- BlockLen(u32): length of the block
+- Values: an application defined format storing a sequence of value, capable of determining it own length
 - Delta
 
 ### Delta
@@ -115,5 +115,5 @@ Note: there is no ambiguity between both representation as Add is always guarant
 | BlockLen | FirstOrdinal |
 +----------+--------------+
 ```
-- BlockLen(VInt): lenght of the block
+- BlockLen(VInt): length of the block
 - FirstOrdinal(VInt): ordinal of the first element in the given block
--- a/sstable/README.md
+++ b/sstable/README.md
@@ -87,15 +87,15 @@ Note: there is no ambiguity between both representation as Add is always guarant
 
 ### SSTFooter
 ```
-+-------+---------+---------+------+-------------+
-| Block | NumTerm | Version | Type | IndexOffset |
-+-------+---------+---------+------+-------------+
++-------+-------------+---------+---------+------+
+| Block | IndexOffset | NumTerm | Version | Type |
++-------+-------------+---------+---------+------+
 ```
 - Block(SSTBlock): uses IndexValue for its Values format
+- IndexOffset(u64): Offset to the start of the SSTFooter
 - NumTerm(u64): number of terms in the sstable
 - Version(u32): Currently defined to 0x00\_00\_00\_01
 - Type(u32): Defined to 0x00\_00\_00\_02
-- IndexOffset(u64): Offset to the start of the SSTFooter
 
 ### IndexValue
 ```
--- a/sstable/README.md
+++ b/sstable/README.md
@@ -29,92 +29,90 @@ the intersection with an automaton
 
 # On disk format
 
-The on disk format is made to allow for partial loading from a sparse FileSlice. A single SSTable is composed of a sequence of independant blocks followed by an index.
+Overview of the SSTable format. Unless noted otherwise, numbers are little-endian.
 
-This definition uses a rust like syntax. Numbers are encoded in little endian unless noted otherwise.
-```rust
-struct SSTable {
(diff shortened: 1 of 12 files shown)


sstable/README.md
```diff
@@ -26,3 +26,94 @@ possible.
...
+- Footer(`SSTFooter`)
+
+### SSTBlock
+```
++----------+--------+-------+-------+-----+
+| BlockLen | Values | Delta | Delta | ... |
++----------+--------+-------+-------+-----+
+                    |----( # of deltas)---|
+```
+- BlockLen(u32): lenght of the block
```
fulmicoton:
```suggestion
- BlockLen(u32): length of the block
```

sstable/README.md
```diff
@@ -26,3 +26,94 @@ possible.
...
+- Values: an application defined format storing a sequence of value, capable of determining it own lenght
```
fulmicoton:
```suggestion
- Values: an application defined format storing a sequence of value, capable of determining it own length
```

sstable/README.md
```diff
@@ -26,3 +26,94 @@ possible.
...
+```
++------+------+-----+
+| 0x01 | Keep | Add |
++------+------+-----+
+```
+- Add(VInt): number of bytes to push
+- Keep(VInt): number of bytes to pop
+
+
+Note: there is no ambiguity between both representation as Add is always guarantee to be non-zero, except for the very first key of an SSTable, where Keep is guaranteed to be zero.
```
fulmicoton, line 86: Great comment. This also implies that we do not support redundant keys. Is it written somewhere in this document?

```suggestion
Note: there is no ambiguity between both representation as Add is always guarantee to be non-zero, except for the very first key of an SSTable, where Keep is guaranteed to be zero.
```

sstable/src/dictionary.rs
```diff
@@ -178,10 +178,26 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
...
     /// Opens a `TermDictionary`.
     pub fn open(term_dictionary_file: FileSlice) -> io::Result<Self> {
-        let (main_slice, footer_len_slice) = term_dictionary_file.split_from_end(16);
+        let (main_slice, footer_len_slice) = term_dictionary_file.split_from_end(24);
         let mut footer_len_bytes: OwnedBytes = footer_len_slice.read_bytes()?;
+
         let index_offset = u64::deserialize(&mut footer_len_bytes)?;
         let num_terms = u64::deserialize(&mut footer_len_bytes)?;
+        let version = u32::deserialize(&mut footer_len_bytes)?;
+        let type_ = u32::deserialize(&mut footer_len_bytes)?;
```
fulmicoton: Can we introduce a footer object with serialize/deserialize method?

sstable/src/lib.rs
```diff
@@ -201,6 +201,14 @@ where
         }
     }
 
+    /// Set the target block lenght.
```
fulmicoton:
```suggestion
    /// Set the target block length.
```

sstable/src/lib.rs
```diff
@@ -201,6 +201,14 @@ where
...
+    ///
+    /// The delta part of a block will generaly be slighly larger than the requested `block_len`,
```
fulmicoton:
```suggestion
    /// The delta part of a block will generally be slighly larger than the requested `block_len`,
```

sstable/src/lib.rs
```diff
@@ -371,19 +383,25 @@ mod test {
...
+                1,   // num blocks
+                11,  // len of 1st block
+                0,   // first ord of 1st block
+                32, 17, 20, // keep 0 push 2 | 17 20
+                // end of block
+                0, 0, 0, 0, // no more blocks
+                15, 0, 0, 0, 0, 0, 0, 0, // index start offset
+                3, 0, 0, 0, 0, 0, 0, 0, // num_term
+                1, 0, 0, 0, // version
+                2, 0, 0, 0, // dictionary kind. sstable = 2
```
fulmicoton, line 410: Can we have the same footer on the fst type? Either in this PR or open an issue so we do not forget.

For the moment no need to be dynamic, but we need a unit test that takes a valid payload with sstable that gets an explicit error on the fst module and vice versa.

fulmicoton (approved):
Really great job. Thank you. I had a couple of comments.

fulmicoton (changes requested):
~~Can we add a ord_to_term benchmark and make sure that we don't have a regression here?(Previously we were using a binary_search...)~~ 

Scratch that. That part is unchanged at this point.

fulmicoton (approved):

sstable/README.md
trinity-1686a: I don't think it is, this could be rephrased as 
```suggestion
Note: as the SSTable does not support redundant keys, there is no ambiguity between both representation. Add is always guaranteed to be non-zero, except for the very first key of an SSTable, where Keep is guaranteed to be zero.
```
to include that information

sstable/src/lib.rs
trinity-1686a: I plan on doing that in a future PR. Wrt fixing #1827 , I consider this PR to be groundwork only, and not a fix yet

sstable/README.md
trinity-1686a: oups, I forgot to add that before merging. I'll add it in the compression PR, or as an independent PR if it compression end up not making it

PR #1945, opened 2023-03-20
tokenizer option on text fastfield

### LINKED ISSUE #1901: Lowercase transformation on text fast field
Fast fields for text have an embedded text dictionary (after v20).
The embedded terms dictionary in the fast field stores the text as is (equal to raw tokenizer).

Term aggregations in project airmail use a lower case transformation.

Possible solutions
1. Allow to configure a tokenizer on the fastfield
2. Add optional to_lowercase option on the fast field
3. Apply the lower case when aggregating (lower quality results)

PSeitz:
allow to set tokenizer option on text fastfield (fixes #1901)
handle PreTokenized strings in fast field

### DIFF
--- a/src/schema/text_options.rs
+++ b/src/schema/text_options.rs
@@ -16,11 +16,72 @@ pub struct TextOptions {
     #[serde(default)]
     stored: bool,
     #[serde(default)]
-    fast: bool,
+    fast: FastFieldOptions,
     #[serde(default)]
     #[serde(skip_serializing_if = "is_false")]
-    /// coerce values if they are not of type string
+    /// coerce values into string if they are not of type string
     coerce: bool,
+}
+
+#[derive(Clone, Debug, Default, PartialEq)]
+enum FastFieldOptions {
+    Enabled,
+    #[default]
+    Disabled,
+    WithTokenizer(TokenizerName),
+}
+
+impl Serialize for FastFieldOptions {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where S: serde::Serializer {
+        use serde::ser::SerializeMap;
+        match self {
+            FastFieldOptions::Enabled => serializer.serialize_bool(true),
+            FastFieldOptions::Disabled => serializer.serialize_bool(false),
+            FastFieldOptions::WithTokenizer(name) => {
+                let mut map = serializer.serialize_map(Some(1))?;
+                map.serialize_entry("tokenizer", name)?;
+                map.end()
+            }
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for FastFieldOptions {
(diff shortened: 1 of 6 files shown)


src/schema/text_options.rs
```diff
@@ -92,7 +165,7 @@ impl TextOptions {
 }
 
 #[derive(Clone, PartialEq, Debug, Eq, Serialize, Deserialize)]
-struct TokenizerName(Cow<'static, str>);
+pub struct TokenizerName(Cow<'static, str>);
```
fulmicoton: Should it be `pub(crate)` ? 
I don't think we use it public API do we?

src/schema/text_options.rs
```diff
@@ -16,13 +16,74 @@ pub struct TextOptions {
...
+    Disabled,
+    WithTokenizer(TokenizerName),
+}
+
+impl Serialize for FastFieldOptions {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where S: serde::Serializer {
+        use serde::ser::SerializeMap;
+        match self {
+            FastFieldOptions::Enabled => serializer.serialize_bool(true),
```
fulmicoton: I appreciate the effort to have backward compatibliity, but we are breaking backward compat big time anyway in next release :D

We can probably avoid have something a tad more straightforward.

```rust


struct TextOption {
    fast: Option<TextFastFieldOption> //< disabled if None
}

impl TextOption {
   pub fn set_fast(&self, text_fast_field_option: TextFastFieldOption) { ... }
}

#[derive(Serialize, Deserialize)]
struct TextFastFieldOption {
   tokenizer: Option<String>
}
```

fulmicoton (changes requested):
See comments inline

src/schema/text_options.rs
PSeitz: The compatibility is on a different layer, since `meta.json` is also used as input to create an index (e.g. `tantivy-cli`). Being able to parse existing `meta.json` files is quite helpful.

src/schema/text_options.rs
PSeitz: it's not reexported. I changed it to `pub(crate)`

src/schema/text_options.rs
fulmicoton: This is adding unneeded complexity.
If it is not to interact with legacy stuff or with humans, let's avoid quirks like "this thing serializes as a bool or a string".

If you really want backward compatibilty, can you make it on the Deserializer side only?

PSeitz: I replaced it with a enum that can deserialize from existing `meta.json` without custom de/serialization

fulmicoton (approved):

Issue #1850, opened 2023-02-09
Enable ZSTD compressino for large str field dictionary in the columnar format


fulmicoton:
The columnar format is now embedding a dictionary within the columnar format.
For large dictionaries, it can really be worth applying ZSTD compression.

This ticket is about tweaking the format to indicate whether compression is enabled or not (at the column level)
and compress the dictionary (or the entire column) if it makes sense.

We can use a threshold (if the dict is larger than 10KB for instance) to decide when compression should kick in.

trinity-1686a:
fixed by #1946

PR #1946, opened 2023-03-20
sstable compression

trinity-1686a:
this builds upon #1943 and add block-level compression and reduce block size to 4k.

in my experimentation, with 1 day worth of gh-archive, it reduces the size of the term dictionary by ~40%, and the size of the split as a whole by 5.5% compared to 32k blocks with no compression. The footer is bigger than with larger blocks, though thanks to the new format, still 45% smaller than what it was before, while being substantially more dense.

This increase the size of small tables by about 18 bytes when they are too small for compression to kick in. This could be lowered to 10 bytes by removing zstd magic number

related to #1850


PSeitz:
Just as a info, I saw on real life term aggregation queries that `ord_to_term` costs around 10%, so reducing the Blocksize is probably a good idea.


#### Top 50 Term Agg Query
flamegraph

sstable/src/block_reader.rs
```diff
@@ -14,6 +16,29 @@ fn read_u32(read: &mut dyn io::Read) -> io::Result<u32> {
...
+    fn new(read: T, budget: usize) -> ReadLimiter<T> {
+        ReadLimiter {
+            inner: read,
+            budget_left: budget,
+        }
+    }
+}
+
+impl<T: Read> Read for ReadLimiter<T> {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
```
fulmicoton: Can we have an implem of `read_exact` calling the underlying read_exact here too?

fulmicoton: (this is mostly defensive)

sstable/src/block_reader.rs
```diff
@@ -36,19 +61,18 @@ impl<'a> BlockReader<'a> {
...
-            return Ok(false);
-        }
-        self.buffer.resize(block_len as usize, 0u8);
-        self.reader.read_exact(&mut self.buffer[..])?;
+        self.buffer.clear();
+
+        let block_len = match read_u32(self.reader.as_mut()) {
+            Ok(0) => return Ok(false),
+            Ok(n) => n,
+            Err(err) if err.kind() == io::ErrorKind::UnexpectedEof => return Ok(false),
```
fulmicoton: Should we differentiate the case where the EOF happens after reading 0 bytes vs 1|2|3 bytes

fulmicoton:
> it reduces the size of the term dictionary by ~40%, and the size of the split as a whole by 5.5% compared to 32k blocks with no compression.

FYI this change is mostly about making aggregation faster by reducing the time it takes to download a column with a large dictionary.

fulmicoton (changes requested):
Can you check how bad does it hurt indexing performance?

Also some of our dictionaries will tiny. 
(E.g. status on a log: ERROR|INFO|WARN|CRITICAL|DEBUG|TRACE)
I don't think it makes any sense to compress those (it won't compress). 
Can we make it so that ZSTD compression only kicks in if the block is above a given threshold?

We then want a flag to at the block level to tell us if it is compressed or not.

sstable/src/block_reader.rs
trinity-1686a: it wasn't verified before, but I guess it would make sens. The latter is probably a sign of either a bug or corruption

sstable/src/block_reader.rs
trinity-1686a: I don't think `read_exact` is ever called on this, the `read_exact` we do is on the `ZstdDecoder`, which leverage a `BufReader`, itself using only `Read::read_buf` (that's all implementation detail).

trinity-1686a:
>  We then want a flag to at the block level to tell us if it is compressed or not.

How about reserving the MSB of block_len, for that metadata?

trinity-1686a:
> Can you check how bad does it hurt indexing performance?

Preliminary results: it impacts severely a "building an sstable" microbenchmark, being 3.5 to 4.5 time slower (this bench assumes an already sorted list of entry, with everything already in memory, and no reallocation in the writer)
```
entry count: 11_176_233

url+zstd
elapsed=1_911_560µs, size=177_437kB

md5(hex)+zstd
elapsed=1_187_112µs, size=191_087kB

md5(raw)+zstd
elapsed=562_105µs, size=161_706kB

url
elapsed=514_202µs, size=365_091kB

md5(hex)
elapsed=266_635µs, size=337_391kB

md5(raw)
elapsed=161_493µs, size=161_328kB
```

trinity-1686a:
zstd1 seems too slow for what we try to do. Lowering the block size makes `ord_to_term` around 5 times faster (but opening the dictionary is ~5.5 time slower). Zstd1 bring `ord_to_term` back to its old speed.

I've experimented with zstd-10 (that's minus 10, not just 10), which reduce substantially the possible size gain (on that specific dataset 53% -> 17% gain) and is "only" 2.25 time slower than no compression at the same block size (so around twice as fast when going from 32k to 4k blocks). All that on a dictionary that should compress particularly well (11m urls). 

I also gave a try to lz4, which gives slightly worst ratio, same open/`ord_to_term` speed as zstd-10, but faster building.

All in all, I'm not sure compression is that good of an idea, or at least not without more tuning, and probably a different compression algorithm.

trinity-1686a:
turns out a big part of the slowdown I measured on `ord_to_term` is time spent copying things around when using the `Read`-based stream API. I also tried the `Write` based API and got the same issue. The bulk API is better but comes with a few constraints: it wants slices, and Vec that are pre-allocated to be big enough without reallocating. For the former, we are already kinda using slices, the Read trait going everywhere is almost always an OwnedByte (except in tests where it's just slices or Vec), so de-genericizing things and using slices methods works fine. Pre-allocating a big enough Vec is a bit of an issue as we don't know how much the decompressed block is going to be. My approach isn't great, and I wonder if allocating a u32 per compressed block for "size of the uncompressed data" wouldn't be better.

With this iteration, at zstd -1 (and with lowered block size), `ord_to_term` is around twice as fast as main, zstd 1 is only ~10% faster (but yield much smaller blocks, negative level disable entropy coding I think?).
Building the sstable is still 3.5 to 4.5 time slower, but it does not show in `ingest-hdfs` benchmarks, where creating sstables account for 0.4% of samples, with this patch.

ownedbytes/src/lib.rs
```diff
@@ -139,6 +139,16 @@ impl OwnedBytes {
         self.advance(8);
         u64::from_le_bytes(octlet)
     }
+
+    /// Reads an `u32` encoded as little-endian from the `OwnedBytes` and advance by 4 bytes.
+    #[inline]
+    pub fn read_u32(&mut self) -> u32 {
+        assert!(self.len() > 3);
```
adamreichold, line 146: Doesn't `as_slice()[..4]` basically assert the same condition?

sstable/Cargo.toml
```diff
@@ -7,6 +7,7 @@ license = "MIT"
 [dependencies]
 common = {path="../common", package="tantivy-common"}
 tantivy-fst = "0.4"
+zstd = "0.12"
```
adamreichold: I am not quite clear whether this fits your MSRV, but workspace dependencies might be a good thing here as this is most likely not the first dependency on a specific pre-1.0 version of `zstd`, c.f. https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

sstable/src/block_reader.rs
```diff
@@ -36,19 +34,63 @@ impl<'a> BlockReader<'a> {
...
+
+        let real_block_len = (block_len & !COMPRESSION_FLAG) as usize;
+        if self.reader.len() < real_block_len {
+            return Err(io::Error::new(
+                io::ErrorKind::UnexpectedEof,
+                "failed to read block content",
+            ));
+        }
+        if real_block_len as u32 != block_len {
+            // TODO this way of growing the Vec isn't very good, but using the friendlier APIs
```
adamreichold: I think Zstd has some support to handle this internally as well, at least when the `experimental` feature is enabled, i.e. https://docs.rs/zstd/latest/zstd/bulk/struct.Decompressor.html#method.upper_bound

sstable/src/delta.rs
```diff
@@ -45,13 +46,33 @@ where
...
+
         let block_len = buffer.len() + self.block.len();
-        self.write.write_all(&(block_len as u32).to_le_bytes())?;
-        self.write.write_all(&buffer[..])?;
-        self.write.write_all(&self.block[..])?;
+
+        if block_len > 2048 {
+            buffer.extend_from_slice(&self.block);
+            self.block.clear();
+            zstd::stream::copy_encode(&**buffer, &mut self.block, -1)?;
```
adamreichold: Maybe this would benefit from using `Compressor::compress_to_buffer` as well to avoid the `io::Read/Write` based interfaces? Resizing an amortized output buffer should be possible using [`zstd::zstd_safe::compress_bound`](https://docs.rs/zstd-safe/6.0.3+zstd.1.5.2/zstd_safe/fn.compress_bound.html) as `Compressor::compress` does internally.

ownedbytes/src/lib.rs
trinity-1686a: It does, I copied from `read_u64`, not sure why it does that check either

sstable/src/delta.rs
trinity-1686a: When looking at flamegraphs, this part seemed fine for some reason. I rewrote it with `compress_bound` and the bulk api. I can't tell if it's any faster, results are within margin of error. However it seems to yield a more compact representation so I'll do the change anyway

adamreichold: > However it seems to yield a more compact representation so I'll do the change anyway

This might be the difference between stream and bulk API where the bulk API knows all of its input including the size upfront.

trinity-1686a: yeah, and I guess this different encoding is probably helpful to `Decompressor::upper_bound` too

sstable/src/block_reader.rs
```diff
@@ -36,19 +34,42 @@ impl<'a> BlockReader<'a> {
...
+            return Err(io::Error::new(
+                io::ErrorKind::UnexpectedEof,
+                "failed to read block content",
+            ));
+        }
+        if real_block_len as u32 != block_len {
+            let required_capacity =
+                Decompressor::upper_bound(&self.reader[..real_block_len]).unwrap_or(1024 * 1024);
+            self.buffer.reserve(required_capacity);
+            Decompressor::new()?
```
adamreichold, line 63: Not sure whether the difference is measurable, but maybe you can save some initialization overhead by moving the `Decompressor` into the `BlockReader` and re-using it for each call to `read_block`? (And similarly for the `Compressor` but I guess this might even be less significant in the compression path.)

trinity-1686a: I already tried it and it made no difference whatsoever. I think it has more of an impact when using a dictionary

sstable/README.md
```diff
@@ -46,9 +46,10 @@ Overview of the SSTable format. Unless noted otherwise, numbers are little-endia
 +----------+--------+-------+-------+-----+
 | BlockLen | Values | Delta | Delta | ... |
 +----------+--------+-------+-------+-----+
-                    |----( # of deltas)---|
+           |        |----( # of deltas)---|
+           |------(maybe compressed)------|
 ```
-- BlockLen(u32): length of the block
+- BlockLen(u32): length of the block. If MSB is 1, block is compressed with zstd, otherwise it's not.
```
fulmicoton: I'm not fond of the single bit trick here.
Can we have a byte in the begining instead. The bit solution will make it painful if we want to change the compression format.

sstable/src/delta.rs
```diff
@@ -45,13 +47,36 @@ where
...
-        self.write.write_all(&buffer[..])?;
-        self.write.write_all(&self.block[..])?;
+
+        if block_len > 2048 {
+            buffer.extend_from_slice(&self.block);
+            self.block.clear();
+
+            let max_len = zstd::zstd_safe::compress_bound(buffer.len());
+            self.block.reserve(max_len);
+            Compressor::new(-1)?.compress_to_buffer(&**buffer, &mut self.block)?;
```
fulmicoton: clippy

sstable/README.md
```diff
@@ -43,13 +43,14 @@ Overview of the SSTable format. Unless noted otherwise, numbers are little-endia
...
-```
-- BlockLen(u32): length of the block. If MSB is 1, block is compressed with zstd, otherwise it's not.
++----------+----------+--------+-------+-------+-----+
+| Compress | BlockLen | Values | Delta | Delta | ... |
++----------+----------+--------+-------+-------+-----+
+                      |        |----( # of deltas)---|
+                      |------(maybe compressed)------|
+```
+- Compress(u8): indicate whether block is compressed. 0 if not compressed, 1 if compressed.
+- BlockLen(u32): length of the block.
```
fulmicoton: We should intervert `BlockLen` and `Compress`

trinity-1686a:
passed the huge difference between -1 and 1, decompression speed seems to be basically unaffected by the compression level. Considering compressing the sstable has a negligible impact on building an index, if we decide level 1 is fine, we can probably go a bit higher to get a better ratio with no after cost

fulmicoton (approved):

trinity-1686a:
After discussing with Pascal, it seems the size gain from using a positive level is worth it

Issue #1948, opened 2023-03-21
SSTableIndex could be a multi layer SSTable


fulmicoton:
Right now we use the SSTable format to store block "checkpoints" in a Vec and do binary search on those.

An alternative could be to store several layers of SSTable:
- Layer 1 one like current SSTableIndex. Checkpoint every N / B docs.
- Layer 2 Checkpoint in layer 1. N / B^2 
- Layer 3 Checkpoint in layer 1. N / B^3
- ...

Thanks to geometric series, the overhead of having that stack of layers is minor for B = 16 for instance.
We would have transformed the binary search into a coarse to scale linear search, with good strong locality.

The main benefit would be to make opening a term dictionary very cheap, regardless of number of blocks.

@trinity-1686a @PSeitz let me know if I make sense?  To know if this is useful, we need to know 
how expensive it is to open a sstable with 10 millions terms today. 
To accept such a change, we will need also a bench on `ord_to_term` (if it does not exist already)

fulmicoton:
Assigned to @trinity-1686a  for the moment, but we don't even know if we want this.

trinity-1686a:
In the flamegraph PSeitz linked here https://github.com/quickwit-oss/tantivy/pull/1946#issuecomment-1476377510, `Dictionary::open` is very fast (0.42% of samples), whereas `Dictionary::ord_to_term` is 9.94% of samples. A naive approach could cause `ord_to_term` to be N time slower (with N the number of layers).
On some other workload (a simple TermQuery), the `Dictionary::open` is probably most of the cost and it could then make sens to do that kind of change (or maybe it doesn't).
When querying the index, the block being opened should probably be cached to amortize that cost on workload which may query the index a lot (columnar case)

trinity-1686a:
unassigning myself for now, while there is possibly something to gain for large segments, I believe there are probably more impactful changes to explore

PR #1953, opened 2023-03-21
fix bug with new sstable index format

trinity-1686a:
in #1943 , I introduced a small bug where the index-sstable wouldn't honor block size, and make a single block that's as large as needed. Fixing that uncovered a different issue: contrary to what the spec says, we did not actually support a multiblock index-sstable, for all but the 1st block, the `BlockAddr::byte_range` would be plain wrong.

Discovered while trying to assess the theoretical effect implementing something like #1948 could have

### DIFF
--- a/sstable/README.md
+++ b/sstable/README.md
@@ -100,13 +100,14 @@ Note: there is no ambiguity between both representation as Add is always guarant
 
 ### IndexValue
 ```
-+------------+-------+-------+-----+
-| EntryCount | Entry | Entry | ... |
-+------------+-------+-------+-----+
-             |---( # of entries)---|
++------------+----------+-------+-------+-----+
+| EntryCount | StartPos | Entry | Entry | ... |
++------------+----------+-------+-------+-----+
+                        |---( # of entries)---|
 ```
 
 - EntryCount(VInt): number of entries
+- StartPos(VInt): the start pos of the first (data) block referenced by this (index) block
 - Entry (IndexEntry)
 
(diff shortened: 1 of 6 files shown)


sstable/README.md
```diff
@@ -100,13 +100,14 @@ Note: there is no ambiguity between both representation as Add is always guarant
...
-+------------+-------+-------+-----+
-             |---( # of entries)---|
++------------+----------+-------+-------+-----+
+| EntryCount | StartPos | Entry | Entry | ... |
++------------+----------+-------+-------+-----+
+                        |---( # of entries)---|
 ```
 
 - EntryCount(VInt): number of entries
+- StartPos(VInt): the start pos of the first (data) block referenced by this (index) block
```
fulmicoton, line 110: pos is ambiguous. does that mean byte offset? is it not 0?

sstable/src/sstable_index.rs
```diff
@@ -151,6 +151,7 @@ impl SSTableIndexBuilder {
 
             sstable_writer.write_suffix(keep_len, &block.last_key_or_greater[keep_len..]);
             sstable_writer.write_value(&block.block_addr);
+            sstable_writer.flush_block_if_required()?;
```
fulmicoton, line 154: Do we handle having more than one block in the sstable_index itself? I thought is was a single monoblock we deserialized entirely to run binary search

fulmicoton (approved):

PR #1957, opened 2023-03-23
add memory check for term agg

PSeitz:


src/aggregation/agg_limits.rs
```diff
@@ -8,7 +8,7 @@ use super::collector::DEFAULT_MEMORY_LIMIT;
 use super::{AggregationError, DEFAULT_BUCKET_LIMIT};
 use crate::TantivyError;
 
-/// An estimate for memory consumption
+/// An estimate for memory consumption. Non recursive
```
fulmicoton, line 11: what does non recursive mean here?

fulmicoton (approved):

src/aggregation/agg_limits.rs
PSeitz: That it does not cover sub elements, e.g.

`Vec<Box<dyn SubCollector>>`

1. We would need to add trait bounds on the sub elements
2. It would be much slower to traverse whole trees

PR #1959, opened 2023-03-24
Exposing empty column builder.

fulmicoton:


columnar/src/column_values/mod.rs
```diff
@@ -135,6 +141,27 @@ pub trait ColumnValues<T: PartialOrd = u64>: Send + Sync {
...
+/// Empty column of values.
+pub struct EmptyColumnValues;
+
+impl<T: PartialOrd + Default> ColumnValues<T> for EmptyColumnValues {
+    fn get_val(&self, _idx: u32) -> T {
+        panic!("Internal Error: Called get_val of empty column.")
+    }
+
+    fn min_value(&self) -> T {
+        T::default()
```
PSeitz, line 153: I think this should be a panic, so we don't use this accidentally to compute min max

fulmicoton: I disagree :) There are valid uses for wanting a min_value and a max_value...

min_value is misnamed though.

PSeitz: It will yield incorrect min/max when empty columns are mixed in the computation

PR #1968, opened 2023-03-28
encode dictionary type in fst footer

### LINKED ISSUE #1827: encode dictionary type (fst, sstable)

trinity-1686a:
fix #1827


PSeitz:
Shouldnʼt there be layer above the dictionary, that stores the type?

trinity-1686a:
It probably could. I made it that way because it's easier (no such new layer to add), and I think format and version should be together, but version should be "inside" the dictionary, so a future dictionary can support decoding both version x and x+1. However it does mean there isn't a good way to read either an fst or a sstable, whenever both are compiled in

PSeitz:
If dictionary type is inside the dictionary, then we canʼt use it in the future if we want to support multiple dictionaries types.
Version should be part of the dict itself, so we can change the fst, sstable format separately.

PSeitz (approved):

Issue #1913, opened 2023-02-27
Handle missing columns or multiple columns in aggregation


fulmicoton:
After PR#1912 has landed, tantivy will be able to run aggregation over JSON fields.

This means we need to handle the case where columns are missing. 
The result shoudl be the same as if the column was not missing but "empty."

We will also need to face the possibility that we have more than one column with the same name (but different types).

As a first step we need to define a proper behavior.. The behavior is probably aggregation specific.
For instance:
- metric: pick the single numerical column (there can only be one) and ignore others.
- term aggregation: ideal implem would to the aggregation over all of the column and merge the results.

PSeitz:
> 
> As a first step we need to define a proper behavior.. The behavior is probably aggregation specific. For instance:
> 
>     * metric: pick the single numerical column (there can only be one) and ignore others.
> 
>     * term aggregation: ideal implem would to the aggregation over all of the column and merge the results.

For the term agg and mixed types we probably would need a separate virtual column over all column indices (`ColumnIndex`) to collect the missing values. So we would aggregate over up to 3 columns (term + number + null) in a term aggregation.

PR #1971, opened 2023-03-28
Better mixed types support in aggs and fix serialization issue

### LINKED ISSUE #1967: IntermediateTermBucketResult on numeric types is not serializable to json
Number keys are not supported in JSON. 
Serialization fails for `FxHashMap<Key::F64(_), _>`

```rust 
/// Term aggregation including error counts
pub struct IntermediateTermBucketResult {
    pub(crate) entries: FxHashMap<Key, IntermediateTermBucketEntry>,
    pub(crate) sum_other_doc_count: u64,
    pub(crate) doc_count_error_upper_bound: u64,
}

#[derive(Clone, Debug, Serialize, Deserialize, PartialOrd)]
/// The key to identify a bucket.
#[serde(untagged)]
pub enum Key {
    /// String key
    Str(String),
    /// `f64` key
    F64(f64),
}
```

PSeitz:
- Improve support for mixed types in JSON field aggregations (pick the right field, #1913)
- Resolve the issue with JSON serialization for numeric keys (fixes #1967)
- Add JSON round-trip test for term buckets
- Remove `u64_lenient`, as this is a footgun without the type
- move aggregation benchmarks

### DIFF
--- a/src/aggregation/agg_req_with_accessor.rs
+++ b/src/aggregation/agg_req_with_accessor.rs
@@ -46,6 +46,15 @@ pub struct BucketAggregationWithAccessor {
     pub(crate) column_block_accessor: ColumnBlockAccessor<u64>,
 }
 
+fn get_numeric_or_date_column_types() -> &'static [ColumnType] {
+    &[
+        ColumnType::F64,
+        ColumnType::U64,
+        ColumnType::I64,
+        ColumnType::DateTime,
+    ]
+}
+
 impl BucketAggregationWithAccessor {
     fn try_from_bucket(
         bucket: &BucketAggregationType,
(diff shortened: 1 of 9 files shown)


src/fastfield/readers.rs
```diff
@@ -252,33 +252,43 @@ impl FastFieldReaders {
...
-    /// Returns the `u64` column used to represent any `u64`-mapped typed (i64, u64, f64, DateTime).
+    /// Returns the `u64` column used to represent any `u64`-mapped typed (String/Bytes term ids,
+    /// i64, u64, f64, DateTime).
     ///
     /// Returns Ok(None) for empty columns
     #[doc(hidden)]
-    pub fn u64_lenient_with_type(
+    pub fn u64_lenient_for_type(
         &self,
+        type_white_list: Option<&[ColumnType]>,
```
fulmicoton:
```suggestion
        type_white_list_opt: Option<&[ColumnType]>,
```

src/fastfield/readers.rs
```diff
@@ -252,33 +252,43 @@ impl FastFieldReaders {
...
         field_name: &str,
     ) -> crate::Result<Option<(Column<u64>, ColumnType)>> {
         let Some(resolved_field_name) = self.resolve_field(field_name)? else {
             return Ok(None);
         };
         for col in self.columnar.read_columns(&resolved_field_name)? {
+            if let Some(type_white_list) = type_white_list {
```
fulmicoton: pls avoid shadowing  (when it is needless)

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -110,7 +131,11 @@ impl MetricAggregationWithAccessor {
             | MetricAggregation::Min(MinAggregation { field: field_name })
             | MetricAggregation::Stats(StatsAggregation { field: field_name })
             | MetricAggregation::Sum(SumAggregation { field: field_name }) => {
-                let (accessor, field_type) = get_ff_reader_and_validate(reader, field_name)?;
+                let (accessor, field_type) = get_ff_reader_and_validate(
+                    reader,
+                    field_name,
+                    Some(get_numeric_or_date_column_types()),
```
fulmicoton, line 137: does it make sense to allow sum for dates?
(I guess it is does not matter all that much to return a nonsense answer to a nonsense request but...)

fulmicoton (approved):

src/aggregation/agg_req_with_accessor.rs
PSeitz: since the baseline with 1970 is arbitrary, no doesn't make sense

PR #1976, opened 2023-04-02
Proptest merge columnar

fulmicoton:


columnar/src/column_index/merge/mod.rs
```diff
@@ -1,19 +1,70 @@
...
+        if cardinality_so_far >= cardinality_before_deletes {
+            // There won't be any improvement in the cardinality.
+            // We can early exit.
+            return cardinality_before_deletes;
+        }
+    }
+    cardinality_so_far
+}
+
 // For simplification, we never have cardinality go down due to deletes.
```
PSeitz: outdated comment

columnar/src/column_index/merge/mod.rs
```diff
@@ -1,19 +1,70 @@
...
+        return column_index.get_cardinality();
+    };
+    let cardinality_before_deletes = column_index.get_cardinality();
+    if cardinality_before_deletes == Cardinality::Full {
+        // The columnar cardinality can only become stricter in the presence of deletes.
+        return Cardinality::Full;
+    }
+    let mut cardinality_so_far = Cardinality::Full;
+    for doc_id in alive_bitset.iter() {
+        let num_values = column_index.value_row_ids(doc_id).len();
```
PSeitz, line 29: Iterating the other way around should be much faster, especially in cases with many sparse columns

The iterator could return `(doc_id, Cardinality)`

Not sure how much that matters, since the early exit should be triggered quite fast 
except for sparse multivalues

columnar/src/column_index/merge/mod.rs
```diff
@@ -1,19 +1,70 @@
...
+fn detect_cardinality_single_column_index(
+    column_index: &ColumnIndex,
+    alive_bitset_opt: &Option<ReadOnlyBitSet>,
+) -> Cardinality {
+    let Some(alive_bitset) = alive_bitset_opt else {
+        return column_index.get_cardinality();
+    };
+    let cardinality_before_deletes = column_index.get_cardinality();
+    if cardinality_before_deletes == Cardinality::Full {
+        // The columnar cardinality can only become stricter in the presence of deletes.
```
PSeitz: the strictness terminology is not great here imo, since it's not clear what strict means.

columnar/src/columnar/merge/merge_dict_column.rs
```diff
@@ -52,18 +52,23 @@ impl<'a> Iterable for RemappedTermOrdinalsValues<'a> {
...
-                    .term_ord_mapping
-                    .get_segment(seg_ord_with_column as u32);
+        let iter = self
+            .bytes_columns
+            .iter()
+            .enumerate()
+            .flat_map(|(columnar_ord, bytes_column_opt)| {
+                let bytes_column = bytes_column_opt.as_ref()?;
+                Some((columnar_ord, bytes_column))
+            })
```
PSeitz:
```suggestion
            .flat_map(|(seg_ord, bytes_column_opt)| {
                let bytes_column = bytes_column_opt.as_ref()?;
                Some((seg_ord, bytes_column))
            })
```

PSeitz: so the naming is consistent between both functions

columnar/src/columnar/merge/mod.rs
```diff
@@ -287,10 +289,44 @@ fn merged_numerical_columns_type<'a>(
...
+            false
+        }
+        MergeRowOrder::Shuffled(shuffled) => {
+            if let Some(alive_bitset) = &shuffled.alive_bitsets[columnar_id] {
+                let column_index = column.column_index();
+                for doc in alive_bitset.iter() {
+                    if column_index.has_value(doc) {
+                        return false;
+                    }
+                }
```
PSeitz: Similar like above, iterating the other way around should be much faster, especially in cases with many sparse columns

but here, it may take much longer to hit the early exit for very sparse columns

Pathological case:  10mio docs, 100k Columns with each one value. Average has_value early exit after 5mio

100k * 5mio = `500_000_000_000` calls to `has_value` (that's 8minutes for 1ns per call)

PSeitz (commented):
LGTM, but we could hit some pathological cases for very sparse data, which could be avoided with a `column_index.iter_docids() -> (docid, Cardinality)` or similar

columnar/src/column_index/merge/mod.rs
fulmicoton: > Iterating the other way around should be much faster, especially in cases with many sparse columns

I don't think it would help here actually... We need to iterate the docs that do not have any value.
Not the docs that have a value.

columnar/src/column_index/merge/mod.rs
fulmicoton: nice catch

columnar/src/column_index/merge/mod.rs
fulmicoton: making the comment more explicit.

columnar/src/columnar/merge/mod.rs
fulmicoton: yeah in this case it might be better. I was lazy tbh

columnar/src/column_index/merge/mod.rs
fulmicoton: changed the iteration loop

columnar/src/columnar/merge/mod.rs
fulmicoton: I made the change.

PSeitz (approved):

PR #1992, opened 2023-04-13
chore!:drop JSON support on intermediate agg result

PSeitz:
JSON support is flaky anyway due it's lack on f64::INF etc. handling

add support for other formats by removing skip_serialize and untagged

In conjunction with: https://github.com/quickwit-oss/quickwit/pull/3170

### DIFF
--- a/src/aggregation/intermediate_agg_result.rs
+++ b/src/aggregation/intermediate_agg_result.rs
@@ -54,6 +54,14 @@ impl From<Key> for IntermediateKey {
         }
     }
 }
+impl From<IntermediateKey> for Key {
+    fn from(value: IntermediateKey) -> Self {
+        match value {
+            IntermediateKey::Str(s) => Self::Str(s),
+            IntermediateKey::F64(f) => Self::F64(f),
+        }
+    }
+}
 
 impl Eq for IntermediateKey {}
 
--- a/src/aggregation/intermediate_agg_result.rs
+++ b/src/aggregation/intermediate_agg_result.rs
@@ -3,12 +3,12 @@
 //! indices.
 
 use std::cmp::Ordering;
+use std::hash::Hash;
 
 use columnar::ColumnType;
 use itertools::Itertools;
 use rustc_hash::FxHashMap;
-use serde::ser::SerializeSeq;
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::{Deserialize, Serialize};
 
 use super::agg_req::{Aggregation, AggregationVariants, Aggregations};
 use super::agg_result::{AggregationResult, BucketResult, MetricResult, RangeBucketEntry};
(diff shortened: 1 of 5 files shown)


fulmicoton:
I am lacking background on this. What problem does it solve?

PSeitz:
For sending intermediate results between nodes JSON is unnecessary brittle. It's easy to run into cases that are not supported without noticing.

Currently it fails sometimes for ddsketch deserialization, when there are f64 values with `f64::INF`, which serializes to null.
Same for `f64::NAN` https://github.com/serde-rs/json/issues/202

I'd like to avoid any special de/serialization that can be handled by switching the format.
I tested serde round-trips on a similar data-structure here:

https://github.com/PSeitz/test_serde_formats
| Format       | Result                                                                                |
|--------------|---------------------------------------------------------------------------------------|
| Json         | Err(invalid type: null, expected f64 at line 1 column 81)                             |
| Postcard     | Err(Found an Option discriminant that wasn't 0 or 1)                                  |
| Ron          | Ok(())                                                                                |
| MessagePack  | Err(invalid length 0, expected struct IntermediateAggregationResults with 2 elements) |
| Bincode      | Err(invalid value: integer `10`, expected variant index 0 <= i < 2)                   |
| Ciborium     | Ok(())                                                                                |
| Bson         | Err(Invalid map key type: 10)                                                         |

fulmicoton:
Is the problem only NaN or is there something else?

PSeitz:
> Is the problem only NaN or is there something else?

For JSON it's `NaN`, `INF`,`NEG_INF` and non-string keys in Hashmaps. Maybe more we don't know yet

src/aggregation/mod.rs
```diff
@@ -239,6 +245,14 @@ pub enum Key {
     /// `f64` key
     F64(f64),
 }
+impl From<IntermediateKey> for Key {
```
fulmicoton: nitpick It woudl have been more inspired to put that in 
intermediate_agg_result.rs,
with the other conversion.

src/aggregation/intermediate_agg_result.rs
```diff
@@ -29,11 +29,42 @@ use crate::TantivyError;
...
 /// intermediate results.
+///
+/// Notice: This struct should not be de/serialized via JSON format.
 #[derive(Default, Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct IntermediateAggregationResults {
     pub(crate) aggs_res: VecWithNames<IntermediateAggregationResult>,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, PartialOrd, PartialEq)]
+/// The key to identify a bucket.
```
fulmicoton, line 40:
```suggestion
/// The key to identify a bucket.
/// This might seem redundant with `Key`, but the point is to have a different
/// Serialize implementation.
```

fulmicoton (approved):

PR #1994, opened 2023-04-13
Expose `SimpleTokenStream` to use it in quickwit for the multilanguage tokenizer

fmassot:
Ref https://github.com/quickwit-oss/quickwit/pull/3145


trinity-1686a (approved):
~~I'm surprised this wasn't already public~~ or I'm not, it used to be boxed I guess

PR #1999, opened 2023-04-17
fix BooleanQuery document

RTEnzyme:
I think labeling `match at least one of the sub queries` as `Should` makes more sense.

### DIFF
--- a/src/query/boolean_query/boolean_query.rs
+++ b/src/query/boolean_query/boolean_query.rs
@@ -11,8 +11,8 @@ use crate::schema::{IndexRecordOption, Term};
 /// `Must` occurrence
 /// * match none of the sub queries associated with the
 /// `MustNot` occurrence.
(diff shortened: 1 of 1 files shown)


src/query/boolean_query/boolean_query.rs
```diff
@@ -11,8 +11,8 @@ use crate::schema::{IndexRecordOption, Term};
 /// `Must` occurrence
 /// * match none of the sub queries associated with the
 /// `MustNot` occurrence.
-/// * match at least one of the subqueries that is not
-/// a `MustNot` occurrence.
+/// * match at least one of the sub queries associated
+/// with the `Should` occurrence.
```
fulmicoton:
```suggestion
/// with the `Must` or `Should` occurrence.
```

PSeitz (approved):

Issue #1934, opened 2023-03-08
Phrase wit Slop Bug


inzanez:
Hi

Am I correct that proximity search does not handle transpositions? So having indexed something like:
`"This is a big grey wolf"`

the query

`"wolf big"~3`

would not match. Right? I would have to reverse this like:

`"wolf big"~3 OR "big wolf"~3`

if I wanted something like "the word `big` occurring next to `wolf` with a word distance of **up to '3'**, no matter which way around"
...

Quick update: on small samples I do get results that work in both ways. On my larger index datasets the proximity search seems not to be exact / consistent,...not really sure why.

inzanez:
Ok, quick update:

I built an index following the basic sample. I used a large document corpus (emails in text form). Now I also stored the file name of the email in the index, so I have the email-text and the filename itself indexed.

The email text I am looking at is as follows:
```
From    : Peter Pan [peterpan@test.com]
Sent    : 1/14/2115 2:03:52 PM
To      : Captain Hook [hook@somethingelse.com]
Subject : Re: This is for Wendy another test email some more information xxx long subject

Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam

e et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no

Sent from my iPhone

> On 13 Jan 2114, at 6:11 pm, Captain Hook <chook@sdfj.com> wrote:
>
> e et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no
```
I am running the following search:
`query_parser.parse_query("\"captain wendy\"~10")?;`

And I do get **zero** results. I just searched that email in the index by its filename and retrieved the "text" portion to check if the stored text really matches the email above:
[...]

PR #2000, opened 2023-04-17
fix proximity search #1934

sftse:
This is one possible way to fix proximity search.

It might be worthwhile to note that after we made this fix, we discovered that the tantivy [docs](https://docs.rs/tantivy/0.19.2/tantivy/query/struct.PhraseQuery.html#method.set_slop) are inconsistent with [Lucene](https://lucene.apache.org/core/8_0_0/core/org/apache/lucene/search/PhraseQuery.html#getSlop--). Our fix targets the definition of slop that is the maximum allowed offset distance between terms.

We've split the changes into two commits to highlight the race condition. One can test that one commit still leads to nondeterminism in the originally stated case in #1934 while the second commit eliminates this non-determinism.

### DIFF
--- a/columnar/src/column_values/mod.rs
+++ b/columnar/src/column_values/mod.rs
@@ -72,7 +72,7 @@ pub trait ColumnValues<T: PartialOrd = u64>: Send + Sync {
         let cutoff = indexes.len() - indexes.len() % step_size;
 
         for idx in cutoff..indexes.len() {
-            output[idx] = self.get_val(indexes[idx] as u32);
+            output[idx] = self.get_val(indexes[idx]);
         }
     }
 
--- a/src/query/intersection.rs
+++ b/src/query/intersection.rs
@@ -69,7 +69,6 @@ impl<TDocSet: DocSet> Intersection<TDocSet, TDocSet> {
     pub(crate) fn new(mut docsets: Vec<TDocSet>) -> Intersection<TDocSet, TDocSet> {
         let num_docsets = docsets.len();
         assert!(num_docsets >= 2);
-        docsets.sort_by_key(|docset| docset.size_hint());
         go_to_first_doc(&mut docsets);
         let left = docsets.remove(0);
         let right = docsets.remove(0);
--- a/src/query/phrase_query/phrase_scorer.rs
+++ b/src/query/phrase_query/phrase_scorer.rs
@@ -45,7 +45,7 @@ impl<TPostings: Postings> DocSet for PostingsWithOffset<TPostings> {
 
 pub struct PhraseScorer<TPostings: Postings> {
     intersection_docset: Intersection<PostingsWithOffset<TPostings>, PostingsWithOffset<TPostings>>,
-    num_terms: usize,
+    slops: Vec<u32>,
     left: Vec<u32>,
     right: Vec<u32>,
     phrase_count: u32,
@@ -161,6 +161,8 @@ fn intersection_with_slop(left: &mut [u32], right: &[u32], slop: u32) -> usize {
         if left_val < right_slop {
             left_index += 1;
         } else if right_slop <= left_val && left_val <= right_val {
+            left[count] = left_val;
+            count += 1;
             while left_index + 1 < left_len {
                 // there could be a better match
                 let next_left_val = left[left_index + 1];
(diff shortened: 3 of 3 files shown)


sftse (commented):
We've added a test case to demonstrate why we think skipping the saving of intermediate matches within the slop window in `fn intersection_with_slop` is incorrect.

fulmicoton:
@PSeitz can you have a look?

PSeitz:
Thanks for the PR, but as I already have shown in the minimal example in the original issue, this is not a race condition. In one commit it's mentioned that the segments are sorted before they are passed. These are not segments, but docsets on a single segment.

The root cause is the algorithm expects the order of the docsets to match the order of the terms. That is not uphold when the the docsets are sorted by size. https://github.com/quickwit-oss/tantivy/pull/2020 fixes that, by allowing slop in both directions.

sftse:
The minimal example is independent to what we're claiming. We split our changes into two commits to show that 1) our first commit fixes your minimal example, but 2) continues to be non-deterministic in the multithreaded case. We agree the root cause is the sorting of docsets, which should have been clear from our PR. It is possible we're mistaken, but we don't see how the minimal example provides the evidence that this is not a race condition. We don't disagree with your statement, but you seem to object to ours.

#2020 is incorrect. A slop that works in both directions interacts poorly with `PostingsWithOffset`: "wendy subject subject captain" matches the query "wendy subject captain"~1 which should match only "wendy subject captain" or "captain subject wendy". Another counterexample is "captain subject wendy" that does not match the query "wendy captain"~1.

We provided a test demonstrating that (updated to match new functions)

```
let mut left = vec![3, 5];
intersection_count_with_slop(&mut left, &[1, 2], 2, true);
intersection_count_with_slop(&mut left, &[4], 2, true);
```
is not equivalent with
```
let mut left = vec![3, 5];
intersection_count_with_slop(&mut left, &[4], 2, true);
intersection_count_with_slop(&mut left, &[1, 2], 2, true);
```
which seems like an obvious bug, which #2020 does not address, and our PR suggests a solution. We were looking for clarification in the subsequent comment to OP, can we ask you directly to clarify if this is intended behavior? We have not attempted to construct a minimal example with the public API to trigger these arguments, but it would be odd if intersection doesn't need to be commutative.

Edit: Our PR also does not make intersection commutative, but makes (A intersection B) intersection C = (A intersection C) intersection B hold which is the weakest property under which we could reason our solution to be correct. This is because we sort all docsets except the first. If arbitrary sorting is allowed, we don't know how to argue about the correctness without assuming A intersection B = B intersection A.

PSeitz:
> It is possible we're mistaken, but we don't see how the minimal example provides the evidence that this is not a race condition.

It shows that it's an error in the algorithm. 100% deterministic without any threading.

If you create an index, what data gets into which segment is not deterministic. A lot of algorithms in tantivy depend on the actual data in a segment. Just because they depend on what data is in a segment does not mean it's a race condition. After segment creation queries are deterministic.

> A slop that works in both directions interacts poorly with PostingsWithOffset: "wendy subject subject captain" matches the query "wendy subject captain"~1

Yes, as it should be. The distance is 1.

> Another counterexample is "captain subject wendy" that does not match the query "wendy captain"~1

Yes, this should not match, the distance is bigger than 1.

> We provided a test demonstrating that (updated to match new functions)
> ```
> let mut left = vec![3, 5];
> intersection_count_with_slop(&mut left, &[1, 2], 2, true);
> intersection_count_with_slop(&mut left, &[4], 2, true);
> ```
> is not equivalent with
> ```
> let mut left = vec![3, 5];
> intersection_count_with_slop(&mut left, &[4], 2, true);
> intersection_count_with_slop(&mut left, &[1, 2], 2, true);
> ```

Thanks for bringing that up, I was wondering how this algorithm could work correctly by only storing one value in left.
It doesn't, it's buggy.
This test fails for both PRs. On your PR even for slop = 2.

```rust
    #[test]
    pub fn test_phrase_score_with_slop_bug_2() -> crate::Result<()> {
        let index = create_index(&["a x b x c", "a a c"])?;
        let scores = test_query(1, &index, vec!["a", "b", "c"]);
        assert_eq!(scores.len(), 1);
        Ok(())
    }
````

I think we need to store multiple hits similar like in your PR.

I also checked how slop is handled in elastic, and there the slop is more like budget for between all terms. This would required a different approach, probably where we carry the remaining slop distance with each hit.

sftse:
I see, we were erroneously under the impression that the slop indicates the size of a window within which all the terms must be found. Our counterexamples and the PR that additionally assumes the terms must appear in order don't make much sense with any other definition.

fulmicoton:
@sftse @PSeitz  Is the problem entirely solved by #2020 and should we close this PR?

PSeitz:
I created a PR, that fixes open issues with terms > 2 and adds missing clarifications
https://github.com/quickwit-oss/tantivy/pull/2031

PR #2002, opened 2023-04-18
Bump aho-corasick dependency to version 1.0 and adjust to API changes

adamreichold:
* Drop additional `Arc`-layer as the automaton itself is now cheap-to-clone.
* Drop state ID type parameter as it is not exposed by the library any more.

Supersedes automatically generated #2001.


PSeitz:
Thanks!

PSeitz (approved):

PR #2008, opened 2023-04-21
Update schema.rs

tusharv01:


fulmicoton:
@tusharv01 ???

Closing as it looks like an error.

PR #2019, opened 2023-05-01
add seperate tokenizer manager for fast fields

PSeitz:


src/schema/text_options.rs
```diff
@@ -25,8 +25,12 @@ pub struct TextOptions {
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 #[serde(untagged)]
+/// Enum to control how the fast field setting of a text field.
 enum FastFieldOptions {
+    /// Flag to enable/disable
     IsEnabled(bool),
+    /// Enable with tokenizer. The tokenizer must be available on the fast field tokenizer manager.
+    /// `Index::tokenizer_fast_field`.
     EnabledWithTokenizer { with_tokenizer: TokenizerName },
```
fulmicoton, line 34: Do we call it normalizer or something like that?

PSeitz: I don't think we should restrict it to normalizer on tantivy side

src/core/index.rs
```diff
@@ -409,6 +411,16 @@ impl Index {
         &self.tokenizers
     }
 
+    /// Setter for the fast field tokenizer manager.
+    pub fn set_fast_field_tokenizers(&mut self, tokenizers: TokenizerManager) {
+        self.fast_field_tokenizers = tokenizers;
+    }
+
+    /// Accessor for the fast field tokenizer manager.
+    pub fn tokenizer_fast_field(&self) -> &TokenizerManager {
```
```suggestion
    pub fn fast_field_tokenizer(&self) -> &TokenizerManager {
```

.. or the other way around, but we should be more consistent

src/schema/text_options.rs
fulmicoton: you are right

fulmicoton (approved):
approved but can you align naming one way or another before merging?

PR #2020, opened 2023-05-01
allow slop in both directions

### LINKED ISSUE #1934: Phrase wit Slop Bug

PSeitz:
allow slop in both directions
so "big wolf"~3 can also match "wolf big"

This also fixes #1934, when the docsets were reordered by size and didn't
match the terms.

### DIFF
--- a/src/query/phrase_query/phrase_scorer.rs
+++ b/src/query/phrase_query/phrase_scorer.rs
@@ -46,8 +46,8 @@ impl<TPostings: Postings> DocSet for PostingsWithOffset<TPostings> {
 pub struct PhraseScorer<TPostings: Postings> {
     intersection_docset: Intersection<PostingsWithOffset<TPostings>, PostingsWithOffset<TPostings>>,
     num_terms: usize,
-    left: Vec<u32>,
-    right: Vec<u32>,
+    left_positions: Vec<u32>,
+    right_positions: Vec<u32>,
     phrase_count: u32,
     fieldnorm_reader: FieldNormReader,
     similarity_weight_opt: Option<Bm25Weight>,
@@ -104,7 +104,8 @@ pub(crate) fn intersection_count(left: &[u32], right: &[u32]) -> usize {
 /// resulting array in left.
 ///
 /// Returns the length of the intersection
-fn intersection(left: &mut [u32], right: &[u32]) -> usize {
+#[inline]
+fn intersection(left: &mut Vec<u32>, right: &[u32]) -> usize {
     let mut left_index = 0;
     let mut right_index = 0;
     let mut count = 0;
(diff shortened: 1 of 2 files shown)


src/query/phrase_query/phrase_scorer.rs
```diff
@@ -104,7 +104,8 @@ pub(crate) fn intersection_count(left: &[u32], right: &[u32]) -> usize {
 /// resulting array in left.
 ///
 /// Returns the length of the intersection
-fn intersection(left: &mut [u32], right: &[u32]) -> usize {
+#[inline]
+fn intersection(left: &mut Vec<u32>, right: &[u32]) -> usize {
```
adamreichold: Does this need to return `count` if `left` is truncated anyway so that the information is "returned" via `left.len()`?

PSeitz: No, but it's clear for the function caller

fulmicoton: What is the motivation for this change? If it is to simplify the code, as @adamreichold noted, the new API is really bad.

I'd rather keep the previous logic. If you want to simplify this part, do this in a separate PR, as we will need to make sure that it does not hurt performance.

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -128,115 +129,83 @@ fn intersection(left: &mut [u32], right: &[u32]) -> usize {
...
-        let right_slop = if right_val >= slop {
-            right_val - slop
-        } else {
-            0
-        };
-
-        if left_val < right_slop {
-            left_index += 1;
-        } else if right_slop <= left_val && left_val <= right_val {
+        let distance = left_val.abs_diff(right_val);
```
fulmicoton, line 157: neat

src/query/phrase_query/phrase_scorer.rs
```diff
@@ -128,115 +129,83 @@ fn intersection(left: &mut [u32], right: &[u32]) -> usize {
...
+/// Condition for match is that the distance between left and right is less than or equal to `slop`.
 ///
 /// Returns the length of the intersection
-fn intersection_with_slop(left: &mut [u32], right: &[u32], slop: u32) -> usize {
+#[inline]
+fn intersection_count_with_slop(
+    left_positions: &mut Vec<u32>,
+    right_positions: &[u32],
+    slop: u32,
+    update_left: bool,
```
fulmicoton, line 147: Should that be a const generics maybe?

src/query/phrase_query/phrase_scorer.rs
PSeitz: It's simpler to use, you get the results in left, instead of `left[..returned size]`

fulmicoton: agreed. But currently the change is halfway, can we remove the returned value?

Also can we check it improves or does not hurt perf on phrase queries?

PSeitz: Yes, I'll create a benchmark for it

fulmicoton: the search benchmark phrase tag + count is ok too.

PSeitz: Performance looks good
Type of Query: Phrase
tantivy-0.18 | tantivy-0.19 | tantivy-phrase
-- | -- | --
1,114 μs | 1,114 μs | 1,115 μs


TOP10_COUNT seems to have regressed a little though on current main (TOP10 and COUNT are fine)
Type of Query: All
Query | tantivy-0.18 | tantivy-0.19 | tantivy-phrase
-- | -- | -- | --
AVERAGE | 2,561 μs | 2,599 μs | 2,822 μs
the | 11,003  μs4,168,066 docs | 16,523  μs+50.2 %4,168,066 docs | 19,449  μs+76.8 %4,168,066 docs

fulmicoton: can you share results.json?

PSeitz: [results.zip](https://github.com/quickwit-oss/tantivy/files/11369965/results.zip)

PSeitz: > agreed. But currently the change is halfway, can we remove the returned value?

We need to return the count, in case left is not updated

fulmicoton: > [results.zip](https://github.com/quickwit-oss/tantivy/files/11369965/results.zip)

thx

fulmicoton: the results looks good!

fulmicoton: > We need to return the count, in case left is not updated

I don't understand. 
`left.len()` is not always the same as the result of the function?

PSeitz: sorry I thought was about `intersection_count_with_slop`
returned count is never used for `intersection` now, I removed it

fulmicoton (approved):

fulmicoton (approved):

fulmicoton (approved):

Issue #2021, opened 2023-05-02
Improve AggregationLimits


PSeitz:
- [X] Aggregations should release memory in `AggregationLimits` when they are finished
- [ ] Term Aggregations could prune regularly results while collecting when reaching a threshold

PR #2022, opened 2023-05-02
clear memory consumption in AggregationLimits

PSeitz:
clear memory consumption in AggregationLimits at the end of segment collection

### DIFF
--- a/src/aggregation/agg_limits.rs
+++ b/src/aggregation/agg_limits.rs
@@ -25,13 +25,6 @@ impl<K, V, S> MemoryConsumption for HashMap<K, V, S> {
 pub struct AggregationLimits {
     /// The counter which is shared between the aggregations for one request.
     memory_consumption: Arc<AtomicU64>,
-    /// The counter that is specific to one segment.
-    /// It's used mainly to release memory on the global counter.
-    /// Since this is copied to every aggregation in a aggregation tree, this is also
-    /// the memory consumption of the aggregation per segment.
-    ///
-    /// This could be Cell, but it's part of the SegmentCollector, which is Send + Sync.
-    segment_memory_consumption: AtomicU64,
     /// The memory_limit in bytes
     memory_limit: ByteCount,
     /// The maximum number of buckets _returned_
@@ -42,7 +35,6 @@ impl Clone for AggregationLimits {
     fn clone(&self) -> Self {
         Self {
             memory_consumption: Arc::clone(&self.memory_consumption),
-            segment_memory_consumption: AtomicU64::new(0),
             memory_limit: self.memory_limit,
             bucket_limit: self.bucket_limit,
         }
@@ -53,7 +45,6 @@ impl Default for AggregationLimits {
     fn default() -> Self {
         Self {
             memory_consumption: Default::default(),
-            segment_memory_consumption: Default::default(),
             memory_limit: DEFAULT_MEMORY_LIMIT.into(),
             bucket_limit: DEFAULT_BUCKET_LIMIT,
         }
@@ -73,9 +64,20 @@ impl AggregationLimits {
     pub fn new(memory_limit: Option<u64>, bucket_limit: Option<u32>) -> Self {
         Self {
             memory_consumption: Default::default(),
-            segment_memory_consumption: Default::default(),
             memory_limit: memory_limit.unwrap_or(DEFAULT_MEMORY_LIMIT).into(),
             bucket_limit: bucket_limit.unwrap_or(DEFAULT_BUCKET_LIMIT),
+        }
+    }
+
+    /// Crate a new ResourceLimitGuard, that will release the memory when dropped.
+    pub fn new_guard(&self) -> ResourceLimitGuard {
+        ResourceLimitGuard {
+            /// The counter which is shared between the aggregations for one request.
+            memory_consumption: Arc::clone(&self.memory_consumption),
+            /// The memory_limit in bytes
+            memory_limit: self.memory_limit,
+
+            allocated_with_the_guard: 0,
         }
     }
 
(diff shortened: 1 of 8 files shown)


PSeitz:
https://github.com/quickwit-oss/tantivy/issues/2021

src/aggregation/agg_limits.rs
```diff
@@ -80,15 +91,35 @@ impl AggregationLimits {
...
     }
+
+    /// Removes the memory consumed tracked by this _instance_ of AggregationLimits.
+    /// This is used to clear the segment specific memory consumption all at once.
+    pub fn clear_local_memory_consumption(&self) {
+        self.memory_consumption.fetch_sub(
+            self.segment_memory_consumption.load(Ordering::Relaxed),
+            Ordering::Relaxed,
+        );
+        self.segment_memory_consumption.store(0, Ordering::Relaxed);
```
adamreichold:
```suggestion
        self.memory_consumption.fetch_sub(
            self.segment_memory_consumption.swap(0, Ordering::Relaxed),
            Ordering::Relaxed,
        );
```

fulmicoton (commented):
The second Atomic does not seem to make sense.
Can we use a RAII / release on drop pattern here?

```
#{[derive(Debug)]
pub struct ResourceLimit {
    available: Arc<AtomicUsize>,
}

impl ResourceLimit {
    fn new_guard() -> ResourceLimitGuard {
        todo!()
    }
}

/// Does not impl Clone
pub struct ResourceLimitGuard {
    available: Arc<AtomicUsize>,
    allocated_with_the_guard: usize,
}

impl ResourceLimitGuard {
    pub fn resize_allocation(&mut self, new_allocation: usize) -> Result<(), ()> {
        /// fetch_sub/fetch_add is tempting here, but we probably need a compare and swap anyway.
        todo!()
    }
}

impl Drop for ResourceLimitGuard {
    fn drop(&self) {
        todo!();
    }

}
```

src/aggregation/agg_limits.rs
```diff
@@ -91,8 +93,6 @@ impl AggregationLimits {
         Ok(())
     }
     pub(crate) fn add_memory_consumed(&self, num_bytes: u64) {
```
fulmicoton: Can we merge `add_memory_consumed` and `validate_memory_consumption` and  internalize the computation of the delta instead of having the logic everywhere?

```fn set_memory_consumed(&mut self, num_bytes: u64) -> Result<(), AggregationError>```

Comment (PSeitz) [reply src/aggregation/agg_limits.rs]: I thought about internalizing the delta logic, as I don't like the replication either, but it can't be part of the guard, since the cardinality is different

The idea behind splitting `add_memory_consumed` and `validate_memory_consumption` is to have multiple `add_memory_consumed` calls and a more expensive `validate_memory_consumption` call, but it's never used like this currently.

Comment (fulmicoton) [reply src/aggregation/agg_limits.rs]: > I thought about internalizing the delta logic, as I don't like the replication either, but it can't be part of the guard, since the cardinality is different

Ah yes indeed.

Comment (fulmicoton) [reply src/aggregation/agg_limits.rs]: > The idea behind splitting add_memory_consumed and validate_memory_consumption is to have multiple add_memory_consumed calls and a more expensive validate_memory_consumption call, but it's never used like this currently.

If it is still not used can we shoot for factorized API?
`.add_memory_consumed(add_num_bytes: u64) -> Result<(), AggregationError>`

Comment (PSeitz) [reply src/aggregation/agg_limits.rs]: yep, updated

File: src/aggregation/agg_limits.rs [hunk 3]
```diff
@@ -68,28 +68,70 @@ impl AggregationLimits {
...
+        Ok(())
     }
+
     pub(crate) fn get_bucket_limit(&self) -> u32 {
         self.bucket_limit
     }
 }
+
+pub(crate) fn validate_memory_consumption(
+    memory_consumption: &Arc<AtomicU64>,
```
fulmicoton: nitpick
```suggestion
    memory_consumption: &AtomicU64,
```

src/aggregation/agg_limits.rs
```diff
@@ -68,28 +68,70 @@ impl AggregationLimits {
...
+        validate_memory_consumption(&self.memory_consumption, self.memory_limit)?;
+        Ok(())
     }
+
     pub(crate) fn get_bucket_limit(&self) -> u32 {
         self.bucket_limit
     }
 }
+
+pub(crate) fn validate_memory_consumption(
```
fulmicoton:
```suggestion
fn validate_memory_consumption(
```

src/aggregation/agg_limits.rs
```diff
@@ -68,28 +68,70 @@ impl AggregationLimits {
...
-    pub(crate) fn validate_memory_consumption(&self) -> crate::Result<()> {
-        if self.get_memory_consumed() > self.memory_limit {
-            return Err(TantivyError::AggregationError(
-                AggregationError::MemoryExceeded {
-                    limit: self.memory_limit,
-                    current: self.get_memory_consumed(),
-                },
-            ));
+
+    /// Crate a new ResourceLimitGuard, that will release the memory when dropped.
```
fulmicoton:
```suggestion
    /// Create a new ResourceLimitGuard, that will release the memory when dropped.
```

src/aggregation/agg_limits.rs
```diff
@@ -68,28 +68,70 @@ impl AggregationLimits {
...
+    // Load the estimated memory consumed by the aggregations
+    let memory_consumed: ByteCount = memory_consumption.load(Ordering::Relaxed).into();
+    if memory_consumed > memory_limit {
+        return Err(TantivyError::AggregationError(
+            AggregationError::MemoryExceeded {
+                limit: memory_limit,
+                current: memory_consumed,
+            },
+        ));
+    }
```
fulmicoton:
```suggestion
) -> Result<(), AggregationError> {
    // Load the estimated memory consumed by the aggregations
    let memory_consumed: ByteCount = memory_consumption.load(Ordering::Relaxed).into();
    if memory_consumed > memory_limit {
        return Err(
            AggregationError::MemoryExceeded {
                limit: memory_limit,
                current: memory_consumed,
            },
        );
    }
```

fulmicoton (approved):

fulmicoton:
Approved to make things smooth, but please have a look at the comments. Most of them are nitpicks but I think they are straightforward.

PR #2030, opened 2023-05-08
fix and extend benchmark

PSeitz:
- add benchmark, add missing inlines
- fix stacker bench
- add wiki benchmark


benches/index-bench.rs
```diff
@@ -28,28 +29,25 @@ pub fn hdfs_index_benchmark(c: &mut Criterion) {
...
     group.bench_function("index-hdfs-with-commit", |b| {
         b.iter(|| {
             let index = Index::create_in_ram(schema.clone());
             let mut index_writer = index.writer_with_num_threads(1, 100_000_000).unwrap();
-            for _ in 0..NUM_REPEATS {
-                for doc_json in HDFS_LOGS.trim().split('\n') {
-                    let doc = schema.parse_document(doc_json).unwrap();
-                    index_writer.add_document(doc).unwrap();
-                }
+            for doc_json in HDFS_LOGS.trim().split('\n') {
```
fulmicoton: this is not gigantic, but can we create a Vec<&str> outside of the loop with the trimmed pre-split line?

fulmicoton (approved):

PR #2031, opened 2023-05-09
fix slop

PSeitz:
Fix slop by carrying slop so far for terms > 2.
Define slop contract in the API

* The query will match if its terms are separated by `slop` terms at most.
The slop can be considered a budget between all terms.
E.g. "A B C" with slop 1 allows "A X B C", "A B X C", but not "A X B X C".
* Transposition costs 2, e.g. "A B" with slop 1 will not match "B A" but it would with slop 2
Transposition is not a special case, in the example above A is moved 1 position and B is moved 1 position, so the slop is 2.

* As a result slop works in both directions, so the order of the terms may changed as long as they respect the slop.

The slop carrying algorithm is only active for terms > 2 as it is slightly slower
```
running 6 tests
test query::phrase_query::phrase_scorer::bench::bench_intersection_count_short                                           ... bench:           2 ns/iter (+/- 0)
test query::phrase_query::phrase_scorer::bench::bench_intersection_medium                                                ... bench:         106 ns/iter (+/- 1)
test query::phrase_query::phrase_scorer::bench::bench_intersection_medium_slop                                           ... bench:         104 ns/iter (+/- 7)
test query::phrase_query::phrase_scorer::bench::bench_intersection_medium_slop_carrying                                  ... bench:         194 ns/iter (+/- 3)
test query::phrase_query::phrase_scorer::bench::bench_intersection_short                                                 ... bench:           3 ns/iter (+/- 0)
test query::phrase_query::phrase_scorer::bench::bench_intersection_short_slop                                            ... bench:          16 ns/iter (+/- 0)
```

### DIFF
--- a/src/query/phrase_query/phrase_query.rs
+++ b/src/query/phrase_query/phrase_query.rs
@@ -66,6 +66,16 @@ impl PhraseQuery {
     /// Slop allowed for the phrase.
     ///
     /// The query will match if its terms are separated by `slop` terms at most.
+    /// The slop can be considered a budget between all terms.
+    /// E.g. "A B C" with slop 1 allows "A X B C", "A B X C", but not "A X B X C".
+    ///
+    /// Transposition costs 2, e.g. "A B" with slop 1 will not match "B A" but it would with slop 2
+    /// Transposition is not a special case, in the example above A is moved 1 position and B is
+    /// moved 1 position, so the slop is 2.
+    ///
+    /// As a result slop works in both directions, so the order of the terms may changed as long as
+    /// they respect the slop.
+    ///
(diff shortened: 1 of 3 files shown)


fulmicoton (approved):

src/query/phrase_query/phrase_query.rs
```diff
@@ -66,6 +66,16 @@ impl PhraseQuery {
     /// Slop allowed for the phrase.
     ///
     /// The query will match if its terms are separated by `slop` terms at most.
+    /// The slop can be considered a budget between all terms.
+    /// E.g. "A B C" with slop 1 allows "A X B C", "A B X C", but not "A X B X C".
```
fulmicoton, line 70: awesome explanation

PR #2035, opened 2023-05-10
handle multiple agg results

PSeitz:
handle multiple intermediate aggregation results with the same name.

### DIFF
--- a/src/aggregation/intermediate_agg_result.rs
+++ b/src/aggregation/intermediate_agg_result.rs
@@ -3,6 +3,7 @@
 //! indices.
 
 use std::cmp::Ordering;
+use std::collections::hash_map::Entry;
 use std::hash::Hash;
 
 use columnar::ColumnType;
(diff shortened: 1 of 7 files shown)


src/aggregation/intermediate_agg_result.rs
```diff
@@ -77,8 +78,17 @@ impl std::hash::Hash for IntermediateKey {
 
 impl IntermediateAggregationResults {
     /// Add a result
-    pub fn push(&mut self, key: String, value: IntermediateAggregationResult) {
-        self.aggs_res.push(key, value);
+    pub fn push(&mut self, key: String, value: IntermediateAggregationResult) -> crate::Result<()> {
+        let entry = self.aggs_res.entry(key);
+        match entry {
+            Entry::Occupied(mut e) => {
```
fulmicoton, line 84: Can you add a comment explaining when this happens? (I assume this has to do with having an aggregation running on several columns of different types?)

fulmicoton (approved):

src/aggregation/intermediate_agg_result.rs
PSeitz: This is just an preparation PR for term aggs over different types, it doesn't happen currently

PR #2036, opened 2023-05-10
Added support for madvise

fulmicoton:


Cargo.toml
```diff
@@ -19,6 +19,7 @@ base64 = "0.21.0"
 byteorder = "1.4.3"
 crc32fast = "1.3.2"
 once_cell = "1.10.0"
+madvise = { version="0.1", optional= true }
```
adamreichold: Note that `memmap2` has integrated support for `madvise(2)`: https://docs.rs/memmap2/latest/memmap2/struct.Mmap.html#method.advise

PSeitz: Yes good point, let's use that and remove the feature flag

fulmicoton: Ah thank you!

fulmicoton:
@PSeitz can you have a look?

PSeitz (approved):

PR #2042, opened 2023-05-12
add support for warming up range of terms

trinity-1686a:
required for quickwit-oss/quickwit#2266

not yet tested. It should be final unless I find issues while implementing \#2266, so I'm keeping this as a draft for now


trinity-1686a:
I did not find any issues. This should be good to go

src/core/inverted_index_reader.rs
```diff
@@ -206,10 +206,50 @@ impl InvertedIndexReader {
...
+        };
+
+        let mut stream = range_builder.into_stream_async().await?;
+
+        let iter = std::iter::from_fn(move || stream.next().map(|(_k, v)| v.clone()));
+
+        // limit on stream is only an optimization to load less data, the stream may still return
+        // more than limit elements.
+        let mut limit = limit;
+        let iter = iter.take_while(move |_e| match limit {
```
trinity-1686a: maybe this should just be `.take(limit.unwrap_or(usize::MAX))`, even on 32b, limiting a range to "just" 4 billion terms shouldn't be an issue (or rather, there would probably be other problems before)

fulmicoton: I'd rather have that yes.

fulmicoton: yeah. let's go with the simpler code

fulmicoton: Can you make the change to use Iterator::take

fulmicoton (approved):

src/core/inverted_index_reader.rs
trinity-1686a: done

PR #2044, opened 2023-05-15
Expose phrase-prefix queries via the built-in query parser

adamreichold:
This proposes the less-than-imaginative syntax `field:"phrase ter"*` to perform a phrase prefix query against `field` using `phrase` and `ter` as the terms. The aim of this is to make this type of query more discoverable and simplify manual testing.

I did consider exposing the `max_expansions` parameter similar to how slop is handled, but I think that this is rather something that should be configured via the querser parser (similar to `set_field_boost` and `set_field_fuzzy`) as choosing it requires rather intimiate knowledge of the backing index.

### DIFF
--- a/src/query/query_parser/query_parser.rs
+++ b/src/query/query_parser/query_parser.rs
@@ -20,7 +20,7 @@ use crate::query::{
 };
 use crate::schema::{
     Facet, FacetParseError, Field, FieldType, IndexRecordOption, IntoIpv6Addr, JsonObjectOptions,
-    Schema, Term, Type,
+    Schema, Term, TextFieldIndexing, Type,
 };
 use crate::time::format_description::well_known::Rfc3339;
 use crate::time::OffsetDateTime;
@@ -70,6 +70,17 @@ pub enum QueryParserError {
     /// have any positions indexed.
     #[error("The field '{0}' does not have positions indexed")]
     FieldDoesNotHavePositionsIndexed(String),
+    /// A phrase-prefix query requires at least two terms
+    #[error(
+        "The phrase '{phrase:?}' does not produce at least two terms using the tokenizer \
+         '{tokenizer:?}'"
+    )]
+    PhrasePrefixRequiresAtLeastTwoTerms {
+        /// The phrase which triggered the issue
+        phrase: String,
+        /// The tokenizer configured for the field
+        tokenizer: String,
+    },
     /// The tokenizer for the given field is unknown
     /// The two argument strings are the name of the field, the name of the tokenizer
     #[error("The tokenizer '{tokenizer:?}' for the field '{field:?}' is unknown")]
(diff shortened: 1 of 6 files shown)


fulmicoton:
@adamreichold is this something that you need?
(this does seem useful...)

adamreichold:
> @adamreichold is this something that you need? (this does seem useful...)

I would say that it is something that I want: I have an internal UI for testing purposes which just uses the built-in query parser. Hence, if we want to try out some functionality before we commit to integrating it, it is easiest to expose it in the built-in query parser so the data people can play with it.

I also thought that it is a rather straight-forward extension of the existing support for phrase queries and slop that does not really introduce anything new conceptually. So the maintenance effort should be limited making this a reasonable trade-off?

fulmicoton:
@adamreichold that makes sense... Let's see if we can slip that into the next tantivy release.

fulmicoton:
@PSeitz can you take this review?

PSeitz:
@adamreichold  Can you add a test with data?

What happens for `"wo"*`

adamreichold:
> Can you add a test with data?

I believe those would not be added to the query parser tests but rather to the implementation of phrase-prefix weight (as with the other query types). But that weight already has tests using an index with data. Is there anything specific missing from those tests that should be added?
 
> What happens for `"wo"*`

Nice catch! I added an explicit error for that case as phrase-prefix queries only really make sense with at least two terms (in contrast to phrase queries which sensibly "downgrade" to term queries).

adamreichold:
> I believe those would not be added to the query parser tests but rather to the implementation of phrase-prefix weight (as with the other query types). But that weight already has tests using an index with data. Is there anything specific missing from those tests that should be added?

@PSeitz Friendly ping on the above question.

PSeitz:
> > Can you add a test with data?
> 
> I believe those would not be added to the query parser tests but rather to the implementation of phrase-prefix weight (as with the other query types). But that weight already has tests using an index with data. Is there anything specific missing from those tests that should be added?
> 

I was thinking of an end to end test, that makes sure everything is correctly connected. It's also a form of documentation and helps discovering features

adamreichold:
> I was thinking of an end to end test, that makes sure everything is correctly connected. It's also a form of documentation and helps discovering features

With this motivation in mind, I added a dedicated example as the discoverability of unit tests is limited and adding a doc test to `QueryParser`, but only for phrase-prefix queries would most likely confuse people reading that comment for the first time.

src/query/query_parser/query_parser.rs
```diff
@@ -774,8 +793,9 @@ fn generate_literals_for_str(
     field: Field,
     phrase: &str,
     slop: u32,
+    prefix: bool,
+    option: &TextFieldIndexing,
```
PSeitz: option is not a good variable name

src/query/query_parser/query_parser.rs
```diff
@@ -497,14 +504,14 @@ impl QueryParser {
...
                             tokenizer: option.tokenizer().to_string(),
                         })?;
-                let index_record_option = option.index_option();
                 Ok(generate_literals_for_str(
                     field_name,
                     field,
                     phrase,
                     slop,
+                    prefix,
+                    option,
```
PSeitz:
```suggestion
                    indexing_options,
```

adamreichold:
I replaced the variable name `option` by `indexing_options` as suggested in both cases.

PSeitz (approved):
LGTM Thanks!

Issue #2026, opened 2023-05-04
Date histogram extended bounds and hard bounds default unit are not aligned with ES API


fmassot:
ES API is not very clear on this but it does not expect nanoseconds.

It probably accepts milliseconds by default? 

Not sure of this one.... if a `format` is specified, it parses the bounds according to the defined format.

PSeitz:
It's milliseconds now. Parsing from Rfc3339 is not yet supported

PR #2045, opened 2023-05-16
switch to ms in histogram for date type

### LINKED ISSUE #2028: Date histogram should return keys in milliseconds to be compatible with ES
Currently, keys are returned in nanoseconds.

I suggest returning by default milliseconds and having a parameter to be able to get keys in another unit.

Note that `key_as_string` can be customized by the `format` parameter which we don't support. By default, `key_as_string` returns the first format specified in the field. The `format` support can be tracked in another issue.

```json
"aggregations" : {
    "metrics" : {
      "buckets" : [
        {
          "key_as_string" : "1674285600000",
          "key" : 1674285600000,
          "doc_count" : 0
        }
      ]
   }
}
```

PSeitz:
switch to ms in histogram, by adding a normalization step that converts
to nanoseconds precision when creating the collector.

closes #2028
related to #2026

### DIFF
--- a/src/aggregation/bucket/histogram/histogram.rs
+++ b/src/aggregation/bucket/histogram/histogram.rs
@@ -13,8 +13,6 @@ use crate::aggregation::agg_req_with_accessor::{
     AggregationWithAccessor, AggregationsWithAccessor,
 };
 use crate::aggregation::agg_result::BucketEntry;
-use crate::aggregation::date::format_date_ms;
-use crate::aggregation::f64_from_fastfield_u64;
 use crate::aggregation::intermediate_agg_result::{
     IntermediateAggregationResult, IntermediateAggregationResults, IntermediateBucketResult,
     IntermediateHistogramBucketEntry,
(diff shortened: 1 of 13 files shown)


src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -312,19 +330,15 @@ impl SegmentHistogramCollector {
     ) -> crate::Result<IntermediateBucketResult> {
         let mut buckets = Vec::with_capacity(self.buckets.len());
 
-        if self.sub_aggregation_blueprint.is_some() {
-            for (bucket_pos, bucket) in self.buckets.into_iter() {
-                let bucket_res = bucket.into_intermediate_bucket_entry(
-                    self.sub_aggregations.get(&bucket_pos).unwrap().clone(),
-                    &agg_with_accessor.sub_aggregation,
-                );
+        for (bucket_pos, bucket) in self.buckets.into_iter() {
```
fulmicoton: nitpick

```suggestion
        for (bucket_pos, bucket) in self.buckets {
```

fulmicoton (approved):

PR #2051, opened 2023-05-18
Rename `DatePrecision` to `DateTimePrecision`

guilload:
@PSeitz, as discussed on Discord.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 test:
-	echo "Run test only... No examples."
+	@echo "Run test only... No examples."
 	cargo test --tests --lib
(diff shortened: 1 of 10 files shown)


PSeitz (approved):
LGTM, we don't need the backwards compatibility since it's a breaking change anyway

guilload:
Just wanted to be gentle on the tantivy users. Do you want me to remove the deprecation stuff? Otherwise, I'll submit a clean-up after PR in one release or two.

PSeitz:
Okay then let's keep it to getting used to it

PR #2058, opened 2023-05-26
Improve Termmap Indexing Performance +~30%

PSeitz:
This contains many small changes to improve Termmap performance.
Most notably:
* Specialized byte compare and equality versions, instead of glibc calls.
* ExpUnrolledLinkedList to not contain inline items.

Allow compare hash only via a feature flag compare_hash_only:
64bits should be enough with a good hash function to compare strings by
their hashes instead of comparing the strings. Disabled by default

Benchmark without `compare_hash_only`
```bash
CreateHashMap/alice/174693
                        time:   [642.23 µs 643.80 µs 645.24 µs]
                        thrpt:  [258.20 MiB/s 258.78 MiB/s 259.41 MiB/s]
                 change:
                        time:   [-14.429% -13.303% -12.348%] (p = 0.00 < 0.05)
                        thrpt:  [+14.088% +15.344% +16.862%]
                        Performance has improved.
CreateHashMap/alice_expull/174693
                        time:   [877.03 µs 880.44 µs 884.67 µs]
                        thrpt:  [188.32 MiB/s 189.22 MiB/s 189.96 MiB/s]
                 change:
                        time:   [-26.460% -26.274% -26.091%] (p = 0.00 < 0.05)
                        thrpt:  [+35.301% +35.637% +35.981%]
                        Performance has improved.
CreateHashMap/numbers_zipf/8000000
                        time:   [9.1198 ms 9.1573 ms 9.1961 ms]
                        thrpt:  [829.64 MiB/s 833.15 MiB/s 836.57 MiB/s]
                 change:
                        time:   [-35.229% -34.828% -34.384%] (p = 0.00 < 0.05)
                        thrpt:  [+52.403% +53.440% +54.390%]
                        Performance has improved.
```

Benchmark with `compare_hash_only`
```bash
CreateHashMap/alice/174693
                        time:   [503.08 µs 505.07 µs 507.22 µs]
                        thrpt:  [328.46 MiB/s 329.86 MiB/s 331.16 MiB/s]
                 change:
                        time:   [-37.334% -36.881% -36.399%] (p = 0.00 < 0.05)
                        thrpt:  [+57.231% +58.432% +59.576%]
                        Performance has improved.
CreateHashMap/alice_expull/174693
                        time:   [863.67 µs 873.21 µs 883.15 µs]
                        thrpt:  [188.64 MiB/s 190.79 MiB/s 192.90 MiB/s]
                 change:
                        time:   [-35.224% -34.352% -33.461%] (p = 0.00 < 0.05)
                        thrpt:  [+50.288% +52.328% +54.379%]
                        Performance has improved.
```

### DIFF
--- a/stacker/Performance.md
+++ b/stacker/Performance.md
@@ -1,0 +1,14 @@
+
+# Notes
+
+- `extend_from_slice(&key)` calls memcpy, which is relatively slow, since most keys are relatively short. For now there's a specialized version toavoid memcpy calls.
+    Wild copy 16 bytes in a loop is faster, but would require a guard against overflow from the caller side. (We probably can do that). 
+- Comparing two slices of unknown length calls memcmp. Same as above, we can do a specialized version.
+
+fastcmp and fastcpy both employ the same trick, to compare slices of odd length, e.g. 2 operations unconditional on 4 bytes, instead 3 operations with conditionals (1 4byte, 1 2byte, 1 1byte).
+[1, 2, 3, 4, 5, 6, 7]
+[1, 2, 3, 4]
+         [4, 5, 6, 7]
+
+- Since the hashmap writes the values on every key insert/update, the values like expull should be small. Therefore inlining of the values has been removed.
+- Currently the first call to Expull will get a capacity of 0. It would be beneficial if it could be initialized with some memory, so that the first call doesn't have to allocate. But that would mean we don't have `Default` impls.
--- a/stacker/src/arena_hashmap.rs
+++ b/stacker/src/arena_hashmap.rs
@@ -2,6 +2,7 @@ use std::iter::{Cloned, Filter};
 use std::mem;
 
 use super::{Addr, MemoryArena};
+use crate::fastcpy::fast_short_slice_copy;
 use crate::memory_arena::store;
 use crate::UnorderedId;
 
@@ -12,7 +13,11 @@ pub fn compute_table_memory_size(capacity: usize) -> usize {
     capacity * mem::size_of::<KeyValue>()
 }
 
+#[cfg(not(feature = "compare_hash_only"))]
 type HashType = u32;
+
+#[cfg(feature = "compare_hash_only")]
+type HashType = u64;
 
 /// `KeyValue` is the item stored in the hash table.
 /// The key is actually a `BytesRef` object stored in an external memory arena.
@@ -132,8 +137,19 @@ impl ArenaHashMap {
     }
 
     #[inline]
+    #[cfg(not(feature = "compare_hash_only"))]
     fn get_hash(&self, key: &[u8]) -> HashType {
         murmurhash32::murmurhash2(key)
+    }
+
+    #[inline]
+    #[cfg(feature = "compare_hash_only")]
+    fn get_hash(&self, key: &[u8]) -> HashType {
+        /// Since we compare only the hash we need a high quality hash.
+        use std::hash::Hasher;
+        let mut hasher = ahash::AHasher::default();
+        hasher.write(key);
+        hasher.finish() as HashType
     }
 
     #[inline]
@@ -159,17 +175,19 @@ impl ArenaHashMap {
     #[inline]
     fn get_key_value(&self, addr: Addr) -> (&[u8], Addr) {
         let data = self.memory_arena.slice_from(addr);
-        let (key_bytes_len_bytes, data) = data.split_at(2);
+        let key_bytes_len_bytes = unsafe { data.get_unchecked(..2) };
         let key_bytes_len = u16::from_le_bytes(key_bytes_len_bytes.try_into().unwrap());
-        let key_bytes: &[u8] = &data[..key_bytes_len as usize];
+        let key_bytes: &[u8] = unsafe { data.get_unchecked(2..2 + key_bytes_len as usize) };
         (key_bytes, addr.offset(2 + key_bytes_len as u32))
     }
 
     #[inline]
     #[cfg(not(feature = "compare_hash_only"))]
     fn get_value_addr_if_key_match(&self, target_key: &[u8], addr: Addr) -> Option<Addr> {
+        use crate::fastcmp::fast_short_slice_compare;
+
         let (stored_key, value_addr) = self.get_key_value(addr);
-        if stored_key == target_key {
+        if fast_short_slice_compare(stored_key, target_key) {
             Some(value_addr)
         } else {
             None
@@ -178,6 +196,8 @@ impl ArenaHashMap {
     #[inline]
     #[cfg(feature = "compare_hash_only")]
     fn get_value_addr_if_key_match(&self, _target_key: &[u8], addr: Addr) -> Option<Addr> {
+        // For the compare_hash_only feature, it would make sense to store the keys at a different
+        // memory location. Here they will just pollute the cache.
         let data = self.memory_arena.slice_from(addr);
         let key_bytes_len_bytes = &data[..2];
         let key_bytes_len = u16::from_le_bytes(key_bytes_len_bytes.try_into().unwrap());
@@ -283,9 +303,9 @@ impl ArenaHashMap {
         }
         let hash = self.get_hash(key);
         let mut probe = self.probe(hash);
+        let mut bucket = probe.next_probe();
+        let mut kv: KeyValue = self.table[bucket];
         loop {
-            let bucket = probe.next_probe();
-            let kv: KeyValue = self.table[bucket];
             if kv.is_empty() {
                 // The key does not exist yet.
                 let val = updater(None);
@@ -293,14 +313,16 @@ impl ArenaHashMap {
                 let key_addr = self.memory_arena.allocate_space(num_bytes);
                 {
                     let data = self.memory_arena.slice_mut(key_addr, num_bytes);
-                    data[..2].copy_from_slice(&(key.len() as u16).to_le_bytes());
+                    let key_len_bytes: [u8; 2] = (key.len() as u16).to_le_bytes();
+                    data[..2].copy_from_slice(&key_len_bytes);
                     let stop = 2 + key.len();
-                    data[2..stop].copy_from_slice(key);
+                    fast_short_slice_copy(key, &mut data[2..stop]);
                     store(&mut data[stop..], val);
                 }
 
                 return self.set_bucket(hash, key_addr, bucket);
-            } else if kv.hash == hash {
+            }
+            if kv.hash == hash {
                 if let Some(val_addr) = self.get_value_addr_if_key_match(key, kv.key_value_addr) {
                     let v = self.memory_arena.read(val_addr);
                     let new_v = updater(Some(v));
@@ -308,6 +330,9 @@ impl ArenaHashMap {
                     return kv.unordered_id;
                 }
             }
+            // This allows fetching the next bucket before the loop jmp
+            bucket = probe.next_probe();
+            kv = self.table[bucket];
         }
     }
 }
@@ -354,5 +379,24 @@ mod tests {
         assert_eq!(compute_previous_power_of_two(9), 8);
         assert_eq!(compute_previous_power_of_two(7), 4);
         assert_eq!(compute_previous_power_of_two(u64::MAX as usize), 1 << 63);
+    }
+
+    #[test]
+    fn test_many_terms() {
+        let mut terms: Vec<String> = (0..20_000).map(|val| val.to_string()).collect();
+        let mut hash_map: ArenaHashMap = ArenaHashMap::default();
+        for term in terms.iter() {
+            hash_map.mutate_or_create(term.as_bytes(), |_opt_val: Option<u32>| 5u32);
+        }
+        let mut terms_back: Vec<String> = hash_map
+            .iter()
+            .map(|(bytes, _, _)| String::from_utf8(bytes.to_vec()).unwrap())
+            .collect();
+        terms_back.sort();
+        terms.sort();
+
+        for pos in 0..terms.len() {
+            assert_eq!(terms[pos], terms_back[pos]);
+        }
     }
 }
--- a/stacker/src/expull.rs
+++ b/stacker/src/expull.rs
@@ -2,40 +2,10 @@ use std::mem;
 
 use common::serialize_vint_u32;
 
-use crate::memory_arena::{load, store};
+use crate::fastcpy::fast_short_slice_copy;
 use crate::{Addr, MemoryArena};
 
-const MAX_BLOCK_LEN: u32 = 1u32 << 15;
-const FIRST_BLOCK: usize = 16;
-const INLINED_BLOCK_LEN: usize = FIRST_BLOCK + mem::size_of::<Addr>();
-
-enum CapacityResult {
-    Available(u32),
-    NeedAlloc(u32),
-}
-
-fn len_to_capacity(len: u32) -> CapacityResult {
-    match len {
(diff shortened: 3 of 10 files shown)


stacker/src/fastcpy.rs
```diff
@@ -0,0 +1,113 @@
...
+            src_len, dst_len,
+        );
+    }
+
+    if src.len() != dst.len() {
+        len_mismatch_fail(src.len(), dst.len());
+    }
+    let len = src.len();
+
+    if src.is_empty() {
```
adamreichold, line 23:
```suggestion
    if len == 0 {
```

stacker/src/fastcpy.rs
```diff
@@ -0,0 +1,113 @@
...
+        return;
+    }
+
+    if len <= 32 {
+        double_copy_trick::<16>(src, dst);
+        return;
+    }
+
+    /// The code will use the vmovdqu instruction to copy 32 bytes at a time.
+    #[cfg(target_feature = "avx")]
```
fulmicoton, line 48: !?

stacker/src/fastcmp.rs
```diff
@@ -0,0 +1,159 @@
...
+// Note: The straigthforward left.chunks_exact(SIZE).zip(right.chunks_exact(SIZE)) produces slower
+// assembly
+fn fast_nbyte_slice_compare<const SIZE: usize>(mut left: &[u8], mut right: &[u8]) -> bool {
+    let left_orig = left;
+    let right_orig = right;
+
+    loop {
+        if left.len() < SIZE || right.len() < SIZE {
+            break;
+        }
```
fulmicoton:
```suggestion
    assert_eq!(left.len(), right.len()); //< I assume this will go away after inlining?
    while left.len() >= SIZE { //< I assume this one cannot go away after inlining? (but maybe LLVM is stronger than I think)
```

stacker/src/fastcmp.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    double_check_trick::<8>(left, right)
+}
+
+#[inline]
+// Note: The straigthforward left.chunks_exact(SIZE).zip(right.chunks_exact(SIZE)) produces slower
+// assembly
+fn fast_nbyte_slice_compare<const SIZE: usize>(mut left: &[u8], mut right: &[u8]) -> bool {
+    let left_orig = left;
```
fulmicoton: nitpick: it would be a tad nicer to keep the argument mut, and
create cursor local variables.

stacker/Performance.md
```diff
@@ -0,0 +1,14 @@
+
+# Notes
+
+- `extend_from_slice(&key)` calls memcpy, which is relatively slow, since most keys are relatively short. For now there's a specialized version toavoid memcpy calls.
```
fulmicoton, line 4: great explanation!

stacker/src/fastcmp.rs
```diff
@@ -0,0 +1,159 @@
...
+    let right_orig = right;
+
+    loop {
+        if left.len() < SIZE || right.len() < SIZE {
+            break;
+        }
+        if &left[..SIZE] != &right[..SIZE] {
```
fulmicoton: the boundary is not removed here unfortunately. (not sure why)

stacker/src/fastcmp.rs
fulmicoton: I confirmed with godbolt this does two checks with the current code.

stacker/src/fastcpy.rs
```diff
@@ -0,0 +1,113 @@
...
+    // length 1-3
+    dst[0] = src[0];
+    if len >= 2 {
+        double_copy_trick::<2>(src, dst);
+    }
+}
+
+#[inline]
+fn double_copy_trick<const SIZE: usize>(src: &[u8], dst: &mut [u8]) {
+    dst[0..SIZE].copy_from_slice(&src[0..SIZE]);
```
fulmicoton, line 82: can we add an assert or at least a debug assert on the size of src?

stacker/src/fastcpy.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+#[inline]
+fn short_copy(src: &[u8], dst: &mut [u8]) {
+    let len = src.len();
+
+    // length 1-3
+    dst[0] = src[0];
+    if len >= 2 {
+        double_copy_trick::<2>(src, dst);
```
fulmicoton, line 74: I don't understand... Is it faster than? 

```
if len == 1 {
        dst[0] = src[0];
} else {
        // len >= 2
        double_copy_trick::<2>(src, dst);
}
```

stacker/src/arena_hashmap.rs
```diff
@@ -284,31 +304,36 @@ impl ArenaHashMap {
         }
         let hash = self.get_hash(key);
         let mut probe = self.probe(hash);
+        let mut bucket = probe.next_probe();
```
fulmicoton, line 306: what is the goal of putting these outside of the loop?

stacker/src/expull.rs
```diff
@@ -52,50 +22,56 @@ fn len_to_capacity(len: u32) -> CapacityResult {
...
 pub struct ExpUnrolledLinkedList {
-    len: u32,
+    // u16, since the max size of each block is (1<<next_cap_pow_2)
+    // Limited to 15, so we don't overflow remaining_cap.
+    remaining_cap: u16,
+    // To get the current number of blocks: block_num - FIRST_BLOCK_NUM
+    block_num: u16,
+    head: Addr,
     tail: Addr,
-    inlined_data: [u8; INLINED_BLOCK_LEN],
```
fulmicoton, line 65: before we remove this... do we have  a bench testing what happens with a unique id field?

fulmicoton:
Cool stuff!

stacker/src/arena_hashmap.rs
PSeitz: To fetch the next bucket before the loop jmp. The comment for it is at the end of the loop.
The idea behind it is to allow unconditionally prefetching data (without the branch predictor)

stacker/src/memory_arena.rs
```diff
@@ -134,44 +128,66 @@ impl MemoryArena {
...
+    data: Box<[u8; PAGE_SIZE]>,
 }
 
 impl Page {
     fn new(page_id: usize) -> Page {
         Page {
             page_id,
             len: 0,
-            data: vec![0u8; PAGE_SIZE].into_boxed_slice(),
+            data: Box::new([0u8; PAGE_SIZE]),
```
fulmicoton: Did it help? If it does not I'd rather avoid the stack allocation,

PSeitz: Both are heap allocations. It removes the bounds check in combination with the mask, since the size is known to the compiler

stacker/src/fastcpy.rs
PSeitz: I think both are equal performance wise, but I didn't test much on this case

stacker/src/fastcmp.rs
PSeitz: In my tests it is, but it needs the checks from the caller function. (e.g. empty checks)

These functions should probably be marked as inline always to either inline the whole function or not

stacker/src/memory_arena.rs
adamreichold: They both end up on the heap, but Rust will often allocate the array on the stack before moving it to the heap, c.f. https://github.com/rust-lang/rust/issues/53827

stacker/src/fastcmp.rs
adamreichold: Not sure about the produced assembly, but maybe to make it more concise and give it fewer boundary checks to remove in the first place:

```rust
while left.len() > SIZE {
  let (left_head, left_tail) = left.split_at(SIZE);
  let (right_head, right_tail) = right.split_at(SIZE);

  if left_head != right_head { return false; }

  left = left_tail;
  right = right_tail;
}
```

stacker/src/fastcmp.rs
PSeitz: This does not remove the bounds checks.
It seems that the compiler can't deduce from same length + same increments on left and right to remove bounds checks

stacker/src/fastcmp.rs
PSeitz: This version doesn't remove the bounds checks

stacker/src/memory_arena.rs
PSeitz: I don't see that issue looking at the assembly, but `vec![0, PAGE_SIZE]` is more efficient since it calls alloc zeroed vs. alloc + memset

I"ll switch to `vec![0u8; PAGE_SIZE].into_boxed_slice().try_into().unwrap()`

stacker/src/expull.rs
PSeitz: I added a benchmark for this case, the performance seems to be the same
```
CreateHashMap/ids_expull/8000000
                        time:   [72.704 ms 73.010 ms 73.357 ms]
                        thrpt:  [104.00 MiB/s 104.50 MiB/s 104.94 MiB/s]
                 change:
                        time:   [-0.9446% -0.3605% +0.2415%] (p = 0.24 > 0.05)
                        thrpt:  [-0.2409% +0.3618% +0.9536%]
                        No change in performance detected.
```

stacker/src/fastcpy.rs
PSeitz: I added a debug_assert, but it will fail very fast with an out of bounds if that is incorrect anyways

stacker/src/fastcpy.rs
PSeitz: I prefer is_empty

stacker/src/fastcmp.rs
fulmicoton: hmm I checked with godbolt and that's not true...

stacker/src/arena_hashmap.rs
fulmicoton: What do you mean by unconditionally:  this is not a conditional jump here.
also... 
I would not be surprised if for LLVM, the two are exactly the same...

stacker/src/arena_hashmap.rs
```diff
@@ -284,31 +304,36 @@ impl ArenaHashMap {
...
+        let mut kv: KeyValue = self.table[bucket];
         loop {
-            let bucket = probe.next_probe();
-            let kv: KeyValue = self.table[bucket];
             if kv.is_empty() {
```
fulmicoton, line 309: in both case the condition jump is here, and happens after fetching KeyValue

stacker/src/fastcpy.rs
```diff
@@ -0,0 +1,115 @@
...
+    //
+    // The theory should be that the checks above don't cost much relative to the copy call for
+    // larger copies.
+    // The bounds checks in `copy_from_slice` are elided.
+    dst.copy_from_slice(src);
+}
+
+#[inline]
+fn short_copy(src: &[u8], dst: &mut [u8]) {
+    let len = src.len();
```
fulmicoton, line 69:
```suggestion
    debug_assert_ne!(src.len(), 0);
    debug_assert_eq!(src.len(), dst.len());
    let len = src.len();
```

stacker/src/arena_hashmap.rs
PSeitz: Sorry, that part with the branch predictor didn't make sense. It helps the instruction prefetcher a little bit I think.
The two version are indeed the same, but only with unsafe access. With bounds checks llvm can't optimize the code as well.

https://godbolt.org/z/x7xc8sTcY

stacker/src/expull.rs
```diff
@@ -108,59 +84,68 @@ impl<'a> ExpUnrolledLinkedListWriter<'a> {
 
     #[inline]
     pub fn extend_from_slice(&mut self, mut buf: &[u8]) {
+        debug_assert!(buf.len() <= 32); // Check memcpy assumptions.
+        debug_assert!(!buf.is_empty()); // Check memcpy assumptions.
```
fulmicoton:
```suggestion
```

What is this for?

stacker/src/expull.rs
```diff
@@ -108,59 +84,68 @@ impl<'a> ExpUnrolledLinkedListWriter<'a> {
 
     #[inline]
     pub fn extend_from_slice(&mut self, mut buf: &[u8]) {
+        debug_assert!(buf.len() <= 32); // Check memcpy assumptions.
```
fulmicoton:
```
        debug_assert!(buf.len() <= 32); // Check memcpy assumptions.
```

Is this an assumption we need?
If so can we make the method private?

stacker/src/memory_arena.rs
```diff
@@ -134,44 +128,66 @@ impl MemoryArena {
...
 
     #[inline]
     pub fn slice_mut(&mut self, addr: Addr, len: usize) -> &mut [u8] {
-        self.pages[addr.page_id()].slice_mut(addr.page_local_addr(), len)
+        self.get_page_mut(addr.page_id())
+            .slice_mut(addr.page_local_addr(), len)
+    }
+
+    #[inline]
+    fn add_page(&mut self, len: usize) -> usize {
```
fulmicoton: I don't understand the semantics of this method. Can you add some doc?
Why len, why usize?

fulmicoton: Also at 1MB a pop, I don't think optimizing this matters.

stacker/src/memory_arena.rs
```diff
@@ -134,44 +128,66 @@ impl MemoryArena {
...
+    data: Box<[u8; PAGE_SIZE]>,
 }
 
 impl Page {
     fn new(page_id: usize) -> Page {
         Page {
             page_id,
             len: 0,
-            data: vec![0u8; PAGE_SIZE].into_boxed_slice(),
+            data: vec![0u8; PAGE_SIZE].into_boxed_slice().try_into().unwrap(),
```
fulmicoton, line 190: Awesome I didn't know that  worked

stacker/src/expull.rs
PSeitz: `extend_from_slice` is called only for byte slices of 1-18 (?), and the memcpy could be optimized for that, which I had tested.
I removed it, since I didn't want to add too many implicit assumptions.
The debug_asserts are outdated

fulmicoton: Did you change something? I do not see any change yet.

stacker/src/memory_arena.rs
PSeitz: It's inlined to the hot path though. I refactored it a little bit

stacker/src/expull.rs
PSeitz: I updated it now

stacker/src/memory_arena.rs
fulmicoton: I suspect inline is useless too.

stacker/src/expull.rs
fulmicoton: I still see the debug_asserts

PSeitz: They are removed https://github.com/quickwit-oss/tantivy/pull/2058/commits/82ce3245e1adc92f5163cc55a4d5aa610ee26923

stacker/src/memory_arena.rs
PSeitz: Yes, I removed that in the refactoring. I didn't see any impact on performance though

fulmicoton (approved):

PR #2062, opened 2023-05-30
tokenizer-api: reduce Tokenizer overhead

PSeitz:
Previously a new `Token` for each text encountered was created, which
contains `String::with_capacity(200)`
In the new API the token_stream gets mutable access to the tokenizer,
this allows state to be shared (in this PR Token is shared).
Ideally the allocation for the BoxTokenStream would also be removed, but
this may require some lifetime tricks.

https://github.com/quickwit-oss/tantivy/issues/1654

### DIFF
--- a/tokenizer-api/src/lib.rs
+++ b/tokenizer-api/src/lib.rs
@@ -34,7 +34,7 @@ impl Default for Token {
             offset_from: 0,
             offset_to: 0,
             position: usize::MAX,
-            text: String::with_capacity(200),
+            text: String::new(),
             position_length: 1,
         }
     }
--- a/tokenizer-api/src/lib.rs
+++ b/tokenizer-api/src/lib.rs
@@ -40,13 +40,24 @@ impl Default for Token {
     }
 }
 
+impl Token {
+    /// reset to default
+    pub fn reset(&mut self) {
+        self.offset_from = 0;
+        self.offset_to = 0;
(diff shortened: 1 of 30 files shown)


tokenizer-api/src/lib.rs
```diff
@@ -40,13 +40,24 @@ impl Default for Token {
...
+
 /// `Tokenizer` are in charge of splitting text into a stream of token
 /// before indexing.
 pub trait Tokenizer: 'static + Clone + Send + Sync {
     /// The token stream returned by this Tokenizer.
-    type TokenStream<'a>: TokenStream;
+    type TokenStream<'a, 'b>: TokenStream;
     /// Creates a token stream for a given `str`.
-    fn token_stream<'a>(&self, text: &'a str) -> Self::TokenStream<'a>;
+    fn token_stream<'a, 'b>(&'b mut self, text: &'a str) -> Self::TokenStream<'a, 'b>;
```
adamreichold: Is it really helpful to have two lifetimes here instead of binding both to the presumably shorter one, e.g.

```rust
fn token_stream<'a>(&'a mut self, text: &'a str) -> Self::TokenStream<'a>;
```

and let the general case be handled via covariance?

PSeitz: Good point, one lifetime is enough

fulmicoton (approved):

PR #2065, opened 2023-06-01
Readding s in datetime precision variant names

fulmicoton:
There is no clear win and it change some serialization in quickwit.

### DIFF
--- a/common/src/datetime.rs
+++ b/common/src/datetime.rs
@@ -15,21 +15,12 @@ use time::{OffsetDateTime, PrimitiveDateTime, UtcOffset};
 pub enum DateTimePrecision {
     /// Second precision.
     #[default]
-    Second,
-    /// Millisecond precision.
-    Millisecond,
-    /// Microsecond precision.
-    Microsecond,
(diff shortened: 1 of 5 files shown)


common/src/datetime.rs
```diff
@@ -14,23 +14,25 @@ use time::{OffsetDateTime, PrimitiveDateTime, UtcOffset};
...
 pub enum DateTimePrecision {
     /// Second precision.
+    #[serde(alias = "secs")]
+    #[serde(alias = "seconds")]
     #[default]
     Second,
     /// Millisecond precision.
+    #[serde(alias = "millis")]
+    #[serde(alias = "millisecs")]
+    #[serde(alias = "milliseconds")]
```
PSeitz: Is there a reason for all the aliases? I'd prefer just one variant

fulmicoton: it is something we put in quickwit config. I thought it could be nice to be a tad lenient on those.

PSeitz (approved):

PR #2069, opened 2023-06-01
feat: add stop words for Hungarian language

tnxbutno:
Adding the Hungarian Language stop words to stop word filter.


PSeitz (approved):
LGTM Thanks!

PR #2070, opened 2023-06-03
Fix building on windows with mmap

ChillFish8:
Hides the ability to customise the madvice mmap flags to UNIX only. Allowing you to build tantivy successfully on Windows with the mmap directory.

Originally this was going to be an empty enum or an enum with only the `Normal` variant, but that would have been more confusing IMO.

### DIFF
--- a/src/directory/mmap_directory.rs
+++ b/src/directory/mmap_directory.rs
@@ -6,6 +6,7 @@ use std::path::{Path, PathBuf};
 use std::sync::{Arc, RwLock, Weak};
 use std::{fmt, result};
 
+use aho_corasick::Anchored::No;
 use common::StableDeref;
 use fs4::FileExt;
 use memmap2::Mmap;
@@ -105,6 +106,8 @@ impl MmapCache {
         MmapCache {
             counters: CacheCounters::default(),
             cache: HashMap::default(),
+            #[cfg(unix)]
+            madvice_opt: None,
         }
     }
 
@@ -192,7 +195,7 @@ impl MmapDirectoryInner {
 
     #[cfg(unix)]
     fn with_advice(&mut self, madvice_opt: Option<Advice>) {
-        self.mmap_cache.write().with_advice(madvice_opt)
+        self.mmap_cache.write().unwrap().with_advice(madvice_opt)
     }
 
     fn watch(&self, callback: WatchCallback) -> WatchHandle {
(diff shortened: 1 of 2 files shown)


src/directory/mmap_directory.rs
```diff
@@ -31,12 +32,26 @@ pub(crate) fn make_io_err(msg: String) -> io::Error {
...
 
+#[cfg(unix)]
 /// Returns `None` iff the file exists, can be read, but is empty (and hence
-/// cannot be mmapped)
-fn open_mmap(
+/// cannot be mmapped), this method also allows for optional [Advice] to be
+/// specified for the given workload.
+fn open_mmap_with_advice(
     full_path: &Path,
     madvice_opt: Option<Advice>,
```
PSeitz: Shouldn't this be non optional?

ChillFish8: It could be however, I did not want to make this a breaking change so chose to leave it as optional.

src/directory/mmap_directory.rs
```diff
@@ -215,20 +242,24 @@ impl MmapDirectory {
...
         directory_path: P,
         madvice: Advice,
     ) -> Result<MmapDirectory, OpenDirectoryError> {
-        Self::open_with_access_pattern_impl(directory_path.as_ref(), Some(madvice))
+        let dir = Self::open_with_access_pattern_impl(directory_path.as_ref())?;
+        dir.with_advice(Some(madvice));
+        Ok(dir)
     }
 
     fn open_with_access_pattern_impl(
```
fulmicoton, line 261:
```suggestion
    fn open_impl(
```

src/directory/mmap_directory.rs
fulmicoton: This is not a breaking change. That's not a public API.

src/directory/mmap_directory.rs
fulmicoton: we can keep the helper function to avoid the monomorphization on `P`, but the name is nonsense now.

src/directory/mmap_directory.rs
fulmicoton: @PSeitz: Not necessarily. We still need one way or another to handle absence of Advice.
@ChillFish8 : No this is not public API.

PSeitz: There's still `open_mmap` for an API without Advice

fulmicoton: I thought it was marked as not unix.

PSeitz (approved):

Issue #1352, opened 2022-04-29
Document as trait


PSeitz:
Currently the layout and it's behaviour (serialization etc.) for Document is provided by tantivy.

Users of tantivy have to convert their structure into the tantivy Document.

An alternative approach would be to accept Document as a trait and let tantivy work with Documents via its trait methods.

So instead of converting into `Document` the user could just implement the `Document` trait on its structure and also potentially customize serialization into the doc store.

fulmicoton:
That's an interesting idea.

Issue #2015, opened 2023-04-27
Big indexing refactoring


fulmicoton:
# Design: indexing refactoring.

The point of this design doc is present a refactoring of the DocMapper.
That refactoring has several purpose.
- Improve the performance and the memory footprint associated with JSON fields / Quickwit dynamic mode.
- Improve the code quality
- Offer an efficient solution to document validation on the ingestion side in Quickwit.
- In the future, offer a more powerful schema definition.

# Solution

The doc mapper is used to build `IndexingMappingTree`.

Each edge from a node to another node, is labelled with a json segment.
For extra performance, we can use a raw json escaped string directly here.

Each node is labelled with
- a json path (encoded with a \0 as a separator and unescaped)
- an unordered unique id
- a parameter describing what should if value (bool, str, Number) is encounterred at this level.
- if an object is observed at this level.

Finally each Node also has unique `UnorderedId`

The tree is prepopulated with the statically defined fields, and will grow as we
append nodes to the tree.

Indexing a JSON document then just requires to visit the JSON tree as follows.
We start from the root node of the `IndexingMappingTree`.

When we are visiting an object, we follow the edges associated with all of the key of the object.
Because the edges are labelled with an escaped string, this part can be done with borrowed string.

If an edge does not exist yet, we can decide, depending on whether we are part of the current node
accept it or not, to:
- create new node (dynamic mode, or one of our ancestor is a JSON node).
- ignore the JSON node (lenient mode)
- return an error static mode.

Similarly when we reach a JSON literal, depending on the current node, we should perform
different operations.
- return an error if we had a static node.
- index / store / store as a fast field, with all different relevant options.

`squirrel_json` for instance make it possible to implement such a visit efficiently.

# Indexing

Like in the current version of tantivy, the index writer uses the same HashMap for all Terms.
The terms that are stored there are not `(Field, type, JsonPath, Value)` anymore.
They are instead `(UnorderedId, type, Value)`.

# Upon serialization

We first browse the tree to gather all UnorderedId, and create a UnorderedId -> OrderedId bijection.
The OrderedId is simply the ordinal of the sorted (unescaped) json paths. The OrderedId does not
require a global sort, it just requires to order the different edges locally, and do a Deep-First search.

Like for the current segment serialization, the index writer will create an Vec of
`(Term, Addr)` where `Addr` is an address to access our posting/position data.
We start by scanning this Vec and replacing the unordered id by an ordered id.
[...]

PR #2071, opened 2023-06-05
POC: Tantivy documents as a trait

ChillFish8:
## Problem
Building on what #1352 describes, one of Tantivy's biggest limitations/pain points IMO is the fact that you must convert whatever document type you are using in your code, to a Tantivy `Document` type, which often involves re-allocating and a lot of extra code in order to walk through potentially nested objects (I.e. JSON objects) to be able to index it with tantivy.

## Use cases
- Often in high throughput situations you want to avoid needlessly allocating things, the current setup always requires allocating especially with nested JSON objects, in particular, Quickwit could be quite heavily affected by this change as it allows for borrowed keys and values from the original data.
- Users with custom document types (i.e. lnx) can use the same type across the code base without creating a somewhat messy conversion step between tantivy documents and the documents used everywhere else in the code.

## A limited solution
The solution to this issue initially could be thought up as a basic document trait which simply creates an 
`fn field_values(&self) -> impl Iterator<Item = (Field, Value)>` method for accessing the document data. In theory with the use of GATs now we can even make the `Value` take borrowed data avoiding the allocation issue.

```rust
pub trait Document {
    fn field_values(&self) -> impl Iterator<Item = (Field, Value)>;
}
```

#### Problems with this approach
Although the above suggestion works for a basic setup, it doesn't really solve the issue since you still need a set of concrete tantivy values before you can index data, by which point the amount of effort it takes to convert to a `Document` is very small.

## A fully flexible solution
To get around the issue of making document indexing more flexible and more powerful, we can extend how much of the system is represented by traits, in particular, we replace `Document, Value, FieldValue, serde_json::Map<String,  serde_json::Value>, serde_json::Value` with a set of traits as described bellow:

### `DocumentAccess` trait
Using GATs we can avoid a complicated set of lifetimes and unsafe, instead, we simply described the type the document uses for its values which can borrow data (`Value<'a>`), the owned version of this type which can be the same type potentially, but can also be different depending on application (`OwnedValue`) and finally, we describe the `FieldsValuesIter` which is just used to get around not be able to use anonymous types in the form of `impl Iterator`.

As you can see with the code below, we've replaced the `FieldValue` type with a simple tuple instead, technically this could be kept and made generic, but I don't think it's that useful to do so.

#### Compatibility
The original `Document` type has been kept, and simply implements the trait, meaning a user already using tantivy should not experience any direct conflict if they just keep using the original type.

```rust
/// The core trait representing a document within the index.
pub trait DocumentAccess: Send + Sync + 'static {
    /// The value of the field.
    type Value<'a>: DocValue<'a> + Clone
    where Self: 'a;
    /// The owned version of a value type.
    ///
    /// It's possible that this is the same type as the borrowed
    /// variant simply by using things like a `Cow`, but it may
    /// be beneficial to implement them as separate types for
    /// some use cases.
    type OwnedValue: ValueDeserialize + Debug;
    /// The iterator over all of the fields and values within the doc.
    type FieldsValuesIter<'a>: Iterator<Item = (Field, Self::Value<'a>)>
    where Self: 'a;

    /// Returns the number of fields within the document.
    fn len(&self) -> usize;

    /// Returns true if the document contains no fields.
    fn is_empty(&self) -> bool {
        self.len() == 0
    }

    /// Get an iterator iterating over all fields and values in a document.
    fn iter_fields_and_values(&self) -> Self::FieldsValuesIter<'_>;

    /// Create a new document from a given stream of fields.
    fn from_fields(fields: Vec<(Field, Self::OwnedValue)>) -> Self;

    /// Sort and groups the field_values by field.
    ///
    /// The result of this method is not cached and is
    /// computed on the fly when this method is called.
    fn get_sorted_field_values(&self) -> Vec<(Field, Vec<Self::Value<'_>>)> {
         // Code emitted for ease of reading
    }
}
```

### `DocValue<'a>` trait
This trait is fairly simple in what it does, it simply defines the common methods on the old `Document` type as methods of the trait, and then has a generic `JsonVisitor` type which can be used to represent JSON data in more flexible ways and without allocation.

The original `Value` type implements this trait for compatibility and ergonomics, technically speaking if you wanted to do an approach similar to the first solution you can absolutely do this just by using the original tantivy types.

```rust
pub trait DocValue<'a>: Send + Sync + Debug {
    /// The visitor used to walk through the key-value pairs
    /// of the provided JSON object.
    type JsonVisitor: JsonVisitor<'a>;

    /// Get the string contents of the value if applicable.
    ///
    /// If the value is not a string, `None` should be returned.
    fn as_str(&self) -> Option<&str>;
    /// Get the facet contents of the value if applicable.
    ///
    /// If the value is not a string, `None` should be returned.
    fn as_facet(&self) -> Option<&Facet>;
    /// Get the u64 contents of the value if applicable.
    ///
    /// If the value is not a u64, `None` should be returned.
    fn as_u64(&self) -> Option<u64>;
    /// Get the i64 contents of the value if applicable.
    ///
    /// If the value is not a i64, `None` should be returned.
    fn as_i64(&self) -> Option<i64>;
    /// Get the f64 contents of the value if applicable.
    ///
    /// If the value is not a f64, `None` should be returned.
    fn as_f64(&self) -> Option<f64>;
    /// Get the date contents of the value if applicable.
    ///
    /// If the value is not a date, `None` should be returned.
    fn as_date(&self) -> Option<DateTime>;
    /// Get the bool contents of the value if applicable.
    ///
    /// If the value is not a boolean, `None` should be returned.
    fn as_bool(&self) -> Option<bool>;
    /// Get the IP addr contents of the value if applicable.
    ///
    /// If the value is not an IP addr, `None` should be returned.
    fn as_ip_addr(&self) -> Option<Ipv6Addr>;
    /// Get the bytes contents of the value if applicable.
    ///
    /// If the value is not bytes, `None` should be returned.
    fn as_bytes(&self) -> Option<&[u8]>;
    /// Get the pre-tokenized string contents of the value if applicable.
    ///
    /// If the value is not pre-tokenized string, `None` should be returned.
    fn as_tokenized_text(&self) -> Option<&PreTokenizedString>;
    /// Get the JSON contents of the value if applicable.
    ///
    /// If the value is not pre-tokenized string, `None` should be returned.
    fn as_json(&self) -> Option<Self::JsonVisitor>;
}
```

### `JsonVisitor<'a>` trait
The JSON visitor effectively just replaces the `Map<String, Value>` with a trait that allows for walking through the object.

This also means that any type which implements this trait can also be serialized via serde_json.

```rust
/// A trait representing a JSON key-value object.
///
/// This allows for access to the JSON object without having it as a
/// concrete type.
pub trait JsonVisitor<'a> {
    /// The visitor for each value within the object.
    type ValueVisitor: JsonValueVisitor<'a>;

    #[inline]
    /// The size hint of the iterator length.
    fn size_hint(&self) -> usize {
        0
    }

    /// Get the next key-value pair in the object or return `None`
    /// if the element is empty.
    fn next_key_value(&mut self) -> Option<(&'a str, Self::ValueVisitor)>;
}
```

### `JsonValueVisitor<'a>` trait
Similar to the JSON visitor it describes the behaviour of a `serde_json::Value` rather than just an object.

```rust
/// A trait representing a JSON value.
///
/// This allows for access to the JSON value without having it as a
/// concrete type.
///
/// This is largely a subset of the `Value<'a>` trait.
pub trait JsonValueVisitor<'a> {
    /// The iterator for walking through the element within the array.
    type ArrayIter: Iterator<Item = Self>;
    /// The visitor for walking through the key-value pairs within
    /// the object.
    type ObjectVisitor: JsonVisitor<'a>;

    /// Checks if the value is `null` or not.
    fn is_null(&self) -> bool;
    /// Get the string contents of the value if applicable.
    ///
    /// If the value is not a string, `None` should be returned.
    fn as_str(&self) -> Option<&str>;
    /// Get the i64 contents of the value if applicable.
    ///
    /// If the value is not a i64, `None` should be returned.
    fn as_i64(&self) -> Option<i64>;
    /// Get the u64 contents of the value if applicable.
    ///
    /// If the value is not a u64, `None` should be returned.
    fn as_u64(&self) -> Option<u64>;
    /// Get the f64 contents of the value if applicable.
    ///
    /// If the value is not a f64, `None` should be returned.
    fn as_f64(&self) -> Option<f64>;
    /// Get the bool contents of the value if applicable.
    ///
    /// If the value is not a boolean, `None` should be returned.
    fn as_bool(&self) -> Option<bool>;
    /// Get the array contents of the value if applicable.
    ///
    /// If the value is not an array, `None` should be returned.
    fn as_array(&self) -> Option<Self::ArrayIter>;
    /// Get the object contents of the value if applicable.
    ///
    /// If the value is not an object, `None` should be returned.
    fn as_object(&self) -> Option<Self::ObjectVisitor>;
}
```

### Deserialization via `ValueDeserialize` 
Originally I wanted to use something like `serde::DeserializedOwned` for this job, but it became obvious that what it gained in ease of compatibility with serde, it lost in terms of complexity when handling custom deserializer. A single simple trait became better for this purpose.

One thing that could be improved further is passing a JSON deserializer for the JSON values, at the moment we pass in a `Map<String, serde_json::Value>` which is fairly limited at the moment and not the most ergonomic thing in the world.

## Compatibility and generic methods
One of the things the trait approach requires is a set of generic where documents to be handled, which gives us two ways to handle it:

- We can add generics everywhere on things like `Index`, etc... so it's one document type for the whole index, with a default type keeping compatibility with existing code. The problem is it causes *everything* to require generics (See: https://github.com/ChillFish8/tantivy/pull/2 and the thousands of lines changed, which isn't even finished!)
- We add generics only to the methods which require them (i.e. when adding or fetching docs), making it potentially break some existing projects if the type cannot be inferred (although this is fairly unlikely), which is this PR.


src/core/json_utils.rs
```diff
@@ -90,20 +90,20 @@ pub(crate) fn index_json_values<'a>(
     Ok(())
 }
 
-fn index_json_object(
+fn index_json_object<'a, V: JsonVisitor<'a>>(
     doc: DocId,
-    json_value: &serde_json::Map<String, serde_json::Value>,
+    mut json_visitor: V,
```
PSeitz: not sure with the lifetime, but can we write this as `impl JsonVisitor`?

src/query/more_like_this/query.rs
```diff
@@ -22,26 +24,61 @@ use crate::DocAddress;
 ///     .with_stop_words(vec!["for".to_string()])
 ///     .with_document(DocAddress::new(2, 1));
 /// ```
-#[derive(Debug, Clone)]
-pub struct MoreLikeThisQuery {
+pub struct MoreLikeThisQuery<D: DocumentAccess = Document> {
```
PSeitz: Not sure if this worth the added complexity. The trait is mainly for performance reasons during indexing

src/schema/document.rs
```diff
@@ -1,14 +1,327 @@
...
+
+pub trait DocValue<'a>: Send + Sync + Debug {
+    /// The visitor used to walk through the key-value pairs
+    /// of the provided JSON object.
+    type JsonVisitor: JsonVisitor<'a>;
+
+    /// Get the string contents of the value if applicable.
+    ///
+    /// If the value is not a string, `None` should be returned.
+    fn as_str(&self) -> Option<&str> {
```
PSeitz: the usage with the list of `if` is a little unwiedly. Can we use an Enum with a lifetime instead?

`schema::Value` but on referenced data

src/core/json_utils.rs
ChillFish8: I think so, but I don't think we really gain anything having it be an anonymous type rather than a regular generic

src/query/more_like_this/query.rs
ChillFish8: I agree, after using it a bit the convenience when indexing is great but getting a document back from a searcher kind of sucks :/ MoreLikeThisQuery also gets caught up in the generic mess because it takes in a value which now can also be generic is a user is fetching other docs.

I'm considering if it's best to box this to hide the generic as it really did make it pretty annoying to work with.

src/core/json_utils.rs
adamreichold: I think in library interfaces, impl trait in argument position is a bit of an anti-pattern compared to regular generics. The main problem is that user code cannot specific the type using the turbofish syntax. This can be worked through casts, extra bindings with type ascriptions, etc., but as API courtesy, named generic parameter seem preferable.

src/query/more_like_this/query.rs
PSeitz: We could remove the part where the `DocumentAccess` has control of the serialization. So we would serialize as Document via `DocumentAccess`  and Search would return Document.

ChillFish8: The serialization should already be the same, which is why you can serialize as one type with `DocumentAccess` but deserialize as `Document` still.

The problem I see though is that if we only support serialization with a custom type, people will likely be confused as to why you can't then retrieve a custom type back from the search.

As I write this I think because of the first point I mentioned we can simply remove the generic from `MoreLikeThis` and just deserialize it as a Document (Which may be what you're saying already in which case I apologise for the roundabout conversion)

src/schema/document.rs
ChillFish8: I'm a bit on the edge with that approach right now, I agree the multiple `if`'s aren't great, but with the enum it requires we either have everything be within the enum including the array visitor and the JSON visitor which adds more generics to the value to work with. Or we have a value for the primitive types and then arrays and objects are still the `as_x` approach with an `if` statement but that still feels off :/

Also, I think this would require two separate enums, one for the core document value, and one for a JSON value.

```rust
pub enum ReferenceValue<'a, AV, JV> {
   Str(&'a str),
   U64(u64),
   // << Insert other primitives here >>
   Array(AV),
   JSON(JV),
}
```

src/query/more_like_this/query.rs
ChillFish8: Ended up fixing this and just using the default Document types internally.

src/schema/document.rs
PSeitz: I had something like this in mind:

```rust
pub enum ReferenceValue<
    'a,
    ArrIter: Iterator<Item = ReferenceValue<'a, ArrIter, JsonIter>>,
    JsonIter: Iterator<Item = (&'a str, ReferenceValue<'a, ArrIter, JsonIter>)>,
> {
    Str(&'a str),
    U64(u64),
    // << Insert other primitives here >>
    Array(ArrIter),
    JSON(JsonIter),
}
```

PSeitz: The planned change here is also related https://github.com/quickwit-oss/tantivy/issues/2015

fulmicoton:
@ChillFish8 I'll try to have a look at your solution this week. Sorry for the late reaction.

src/schema/document.rs
ChillFish8: In this situation, I'm not sure if I like the ergonomics of the enum approach after trying it out: 

- We still need two enums effectively, one for tantivy data, and one for JSON data as having the same type can be confusing IMO as it's not clear if returning an IpAddr for some JSON value is going to be ignored, converted, or something else. -> Also this means in the situation where you have a separate type for JSON data vs the top-level data, i.e. tantivy::Value vs srede_json::Value, I think(?) you need to create enums locally for each iterator to handle the variants where you have the two types like so (minus all the boilerplate.)

```rust
pub enum SomeIteratorForValues<'a> {
  Tantivy(MyTantivyValueIterator<'a>),
  Serde(MySerdeValueIterator<'a>),
}
```

Because of the two enums situation for JSON and non-JSON data, we end up with a not-so-great set of generics both for the tantivy maintenance side, but also on the sure side, although technically it's the same amount of types as the existing POC, I don't think it's really any nicer than the `if let` based approach.

This could probably be cleaned up but I only quickly typed this out:

```rust
pub enum ReferenceValue<
    'a,
    ArrIter: Iterator<Item = ReferenceValue<'a, ArrIter, JsonArrayIter, JsonIter>>,
    JsonArrayIter: Iterator<Item = JsonReferenceValue<'a, JsonArrayIter, JsonIter>>,
    JsonIter: Iterator<Item = (&'a str, JsonReferenceValue<'a, JsonArrayIter, JsonIter>)>,
> {
    Str(&'a str),
    U64(u64),
    I64(i64),
    F64(f64),
    Facet(&'a Facet),
    IpAddr(Ipv6Addr),
    Bool(bool),
    PreTokStr(&'a PreTokenizedString),
    Array(ArrIter),
    JSON(JsonIter),
}

pub enum JsonReferenceValue<
    'a,
    ArrIter: Iterator<Item = JsonReferenceValue<'a, ArrIter, JsonIter>>,
    JsonIter: Iterator<Item = (&'a str, JsonReferenceValue<'a, ArrIter, JsonIter>)>,
> {
    Str(&'a str),
    U64(u64),
    I64(i64),
    F64(f64),
    Bool(bool),
    Array(ArrIter),
    Object(JsonIter),
}

pub trait DemoAccess: Send + Sync + 'static {
    /// The iterator over all of the fields and values within the doc.
    type ArrIter<'a>: Iterator<Item = ReferenceValue<'a, Self::ArrIter<'a>, Self::JsonArrayIter<'a>, Self::JsonIter<'a>>>
    where Self: 'a;
    /// The iterator over all of the fields and values within the doc.
    type JsonArrayIter<'a>: Iterator<Item = (&'a str, JsonReferenceValue<'a, Self::ArrIter<'a>, Self::JsonIter<'a>>)>
    where Self: 'a;
    /// The iterator over all of the fields and values within the doc.
    type JsonIter<'a>: Iterator<Item = (&'a str, JsonReferenceValue<'a, Self::ArrIter<'a>, Self::JsonIter<'a>>)>
    where Self: 'a;
    /// The iterator over all of the fields and values within the doc.
    type FieldsValuesIter<'a>: Iterator<Item = (Field, ReferenceValue<'a, Self::ArrIter<'a>, Self::JsonArrayIter<'a>, Self::JsonIter<'a>>)>
    where Self: 'a;

    /// Get an iterator iterating over all fields and values in a document.
    fn iter_fields_and_values(&self) -> Self::FieldsValuesIter<'_>;
}

pub struct Interesting(BTreeMap<Field, Value>);
impl DemoAccess for Interesting {
    type ArrIter<'a> = ArrayIter<'a>;
    type JsonArrayIter<'a> = JsonArrayIter<'a>;
    type JsonIter<'a> = JsonIter<'a>;
    type FieldsValuesIter<'a> = InterestingIter<'a>;

    fn iter_fields_and_values(&self) -> Self::FieldsValuesIter<'_> {
        InterestingIter(self.0.iter())
    }
}

pub struct ArrayIter<'a>(&'a std::slice::Iter<'a, ReferenceValue<'a, Self, JsonArrayIter<'a>, JsonIter<'a>>>);
impl<'a> Iterator for ArrayIter<'a> {
    type Item = ReferenceValue<'a, Self, JsonArrayIter<'a>, JsonIter<'a>>;

    fn next(&mut self) -> Option<Self::Item> {
        todo!()
    }
}

pub struct JsonIter<'a>(serde_json::map::Iter<'a>);
impl<'a> Iterator for JsonIter<'a> {
    type Item = (&'a str, JsonReferenceValue<'a, ArrayIter<'a>, Self>);

    fn next(&mut self) -> Option<Self::Item> {
        todo!()
    }
}

pub struct InterestingIter<'a>(std::collections::btree_map::Iter<'a, Field, Value>);
impl<'a> Iterator for InterestingIter<'a> {
    type Item = (Field, ReferenceValue<'a, ArrayIter<'a>, JsonArrayIter<'a>, JsonIter<'a>>);

    fn next(&mut self) -> Option<Self::Item> {
           todo!()
    }
}
```

ChillFish8: I think what would be better if we went with the enum approach is we keep the visitor traits and custom traits rather than the iterators, reason being is it groups and contains the required types nicer making it more ergonomic and easier to read.

PSeitz: > We still need two enums effectively, one for tantivy data, and one for JSON data as having the same type can be confusing IMO as it's not clear if returning an IpAddr for some JSON value is going to be ignored, converted, or something else

Having JSON as a second class citizen is mainly like this for historical reasons, since the JSON capability was added relatively recently. IMO Value itself should be a tree like structure, there's no need for JSON.

As you said, it's confusing if an `IpAddr` is passed as JSON. That's not the case with `Value`.

ChillFish8: >  IMO Value itself should be a tree like structure, there's no need for JSON.

I'm not sure I understand.

PSeitz: This is already very similar to `serde_json::Value`, but more specific for tantivy. There's no need for JSON as `serde_json::Value`.

```rust
pub enum ReferenceValue<
    'a,
    ArrIter: Iterator<Item = ReferenceValue<'a, ArrIter, ObjectIter>>,
    ObjectIter: Iterator<Item = (&'a str, ReferenceValue<'a, ArrIter, ObjectIter>)>,
> {
    Str(&'a str),
    U64(u64),
    IpAddr(Ipv6Addr),
    // << Insert other tantivy types here >>
    Array(ArrIter),
    Object(ObjectIter),
}
```

E.g. the `Document` API `add_json_object`, should be `tantivy::Value::Object` or similar instead
```rust
    pub fn add_json_object(
        &mut self,
        field: Field,
        json_object: serde_json::Map<String, serde_json::Value>,
    )
```

ChillFish8: Right, so we're effectively making the JSON type become just a dynamic structure of tantivy values rather than JSON-specific data types, in which case yes we can probably use the single enum

ChillFish8:
A lot has changed since the last discussion so I'll update everything here:

# Changes

- Moved to trait + `ReferenceValue` approach to make the API a bit cleaner than a bunch of `if let Some() = ` everywhere.
- Arbitrary nested objects are now allowed, not just strictly JSON types.
- Completely reworked how documents are serialized, removing the need for serde_json to serialize JSON objects. This should improve the performance on datasets which are heavily nested.
- New deserializer and serializer implementations based on `serde`'s API make it suitable for dynamic objects and custom document implementations.
- `document.rs` has become it's own sub-module now since there is too much going on for it to be contained in a single file.

# New API design

### `DocumentAccess`
This provides top-level access to the document fields and values. This is what actually owns the data tantivy indexes.

```rust

/// The core trait representing a document within the index.
pub trait DocumentAccess: DocumentDeserialize + Send + Sync + 'static {
    /// The value of the field.
    type Value<'a>: DocValue<'a> + Clone
    where Self: 'a;

    /// The iterator over all of the fields and values within the doc.
    type FieldsValuesIter<'a>: Iterator<Item = (Field, Self::Value<'a>)>
    where Self: 'a;

    /// Returns the number of fields within the document.
    fn len(&self) -> usize;

    /// Returns true if the document contains no fields.
    fn is_empty(&self) -> bool {
        self.len() == 0
    }

    /// Get an iterator iterating over all fields and values in a document.
    fn iter_fields_and_values(&self) -> Self::FieldsValuesIter<'_>;

    /// Sort and groups the field_values by field.
    ///
    /// The result of this method is not cached and is
    /// computed on the fly when this method is called.
    fn get_sorted_field_values(&self) -> Vec<(Field, Vec<Self::Value<'_>>)> {
          // default impl cut out here
    }
}
```

### `DocValue` and `ReferenceValue`
The interface for tantivy to get borrowed data from the document in the form of a `ReferenceValue`

```rust
/// A enum representing a value for tantivy to index.
pub enum ReferenceValue<'a, V>
where V: DocValue<'a> + ?Sized
{
    /// A null value.
    Null,
    /// The str type is used for any text information.
    Str(&'a str),
    /// Unsigned 64-bits Integer `u64`
    U64(u64),
    /// Signed 64-bits Integer `i64`
    I64(i64),
    /// 64-bits Float `f64`
    F64(f64),
    /// Date/time with nanoseconds precision
    Date(DateTime),
    /// Facet
    Facet(&'a Facet),
    /// Arbitrarily sized byte array
    Bytes(&'a [u8]),
    /// IpV6 Address. Internally there is no IpV4, it needs to be converted to `Ipv6Addr`.
    IpAddr(Ipv6Addr),
    /// Bool value
    Bool(bool),
    /// Pre-tokenized str type,
    PreTokStr(&'a PreTokenizedString),
    /// A an array containing multiple values.
    Array(V::ArrayIter),
    /// A nested/dynamic object.
    Object(V::ObjectIter),
}

/// A single field value.
pub trait DocValue<'a>: Send + Sync + Debug {
    /// The iterator for walking through the elements within the array.
    type ArrayIter: Iterator<Item = ReferenceValue<'a, Self>>;
    /// The visitor walking through the key-value pairs within
    /// the object.
    type ObjectIter: Iterator<Item = (&'a str, ReferenceValue<'a, Self>)>;

    /// Returns the field value represented by an enum which borrows its data.
    fn as_value(&self) -> ReferenceValue<'a, Self>;

    // -- Default methods exist but have been omitted here.
}
```

## Document (de)serializing
The doc store (de)serialization has been reworked, providing an API similar to Serde to allow users to efficiently serialize and deserialize custom documents from the doc store.

### Notes on why deserializes are traits rather than concrete types
The reason why deserializers are trait-based rather than concrete types is that it removes the need for users to be aware of the reader and writer generics which just improves the overall ergonomics. This is what Serde also does.

### Some notes on `#[inline]` in places
When testing I noticed a few situations where the compiler wasn't inlining some of the visitor methods and value access
methods, if you have LTO off I don't think it does any inlining if you use a custom type, which is why some of the methods are marked as inline here (and probably want some docs hinting that others may want to make the methods they implement as inline)

### `DocumentDeserialize`
The top-level trait for deserializing a document. This is a required super trait for the `DocumentAccess` trait.

Eventually, if the API becomes that you no longer need to provide a `Field` and instead just the name of the field, this trait can be removed and made more in line with the rest of the deserializer API, but for now it's just special-cased which isn't the end of the world.

```rust
/// The core trait for deserializing a document.
///
/// TODO: Improve docs
pub trait DocumentDeserialize: Sized {
    /// Attempts to deserialize Self from a given document deserializer.
    fn deserialize<'de, D>(deserializer: D) -> Result<Self, DeserializeError>
    where D: DocumentDeserializer<'de>;
}
```

### `DocumentDeserializer`
A deserializer specifically for documents, for the same reasons I mentioned above, is slightly special-cased because of the `Field` requirement of top-level fields.

```rust
/// A deserializer that can walk through each entry in the document.
pub trait DocumentDeserializer<'de> {
    /// A indicator as to how many values are in the document.
    ///
    /// This can be used to pre-allocate entries but should not
    /// be depended on as a fixed size.
    fn size_hint(&self) -> usize;

    /// Attempts to deserialize the next field in the document.
    fn next_field<V: ValueDeserialize>(&mut self) -> Result<Option<(Field, V)>, DeserializeError>;
}
```
### `ValueDeserializer`
A more traditional deserializer trait for arbitrary values, this is the main entry point for all values other than the top-level document.

I've implemented this trait on a few standard types where it makes sense to do so (Strings, ints, etc...)

```rust
/// The core trait for deserializing values.
///
/// TODO: Improve docs
pub trait ValueDeserialize: Sized {
    /// Attempts to deserialize Self from a given value deserializer.
    fn deserialize<'de, D>(deserializer: D) -> Result<Self, DeserializeError>
    where D: ValueDeserializer<'de>;
}
```

### `ValueDeserializer`
The API of the value deserializer is designed to provide some simple convenience APIs if you already know the target type (for example when implementing `ValueDeserialize` for Strings, u64s, etc...) and then a more powerful visitor API for collections
or types which can deserialize multiple types (like with an enum-based approach).

You may notice there is no convenience API for deserializing an array or object, you must instead use a visitor; this is because
the convenience API would require some private (but not really private) methods for sharing their inner readers.

```rust
/// A value deserializer.
pub trait ValueDeserializer<'de> {
    /// Attempts to deserialize a null value from the deserializer.
    fn deserialize_null(self) -> Result<(), DeserializeError>;

    /// Attempts to deserialize a string value from the deserializer.
    fn deserialize_string(self) -> Result<String, DeserializeError>;

    /// Attempts to deserialize a u64 value from the deserializer.
    fn deserialize_u64(self) -> Result<u64, DeserializeError>;

    /// Attempts to deserialize an i64 value from the deserializer.
    fn deserialize_i64(self) -> Result<i64, DeserializeError>;

    /// Attempts to deserialize a f64 value from the deserializer.
    fn deserialize_f64(self) -> Result<f64, DeserializeError>;

    /// Attempts to deserialize a datetime value from the deserializer.
    fn deserialize_datetime(self) -> Result<DateTime, DeserializeError>;

    /// Attempts to deserialize a facet value from the deserializer.
    fn deserialize_facet(self) -> Result<Facet, DeserializeError>;

    /// Attempts to deserialize a bytes value from the deserializer.
    fn deserialize_bytes(self) -> Result<Vec<u8>, DeserializeError>;

    /// Attempts to deserialize an IP address value from the deserializer.
    fn deserialize_ip_address(self) -> Result<Ipv6Addr, DeserializeError>;

    /// Attempts to deserialize a bool value from the deserializer.
    fn deserialize_bool(self) -> Result<bool, DeserializeError>;

    /// Attempts to deserialize a pre-tokenized string value from the deserializer.
    fn deserialize_pre_tokenized_string(self) -> Result<PreTokenizedString, DeserializeError>;

    /// Attempts to deserialize the value using a given visitor.
    fn deserialize_any<V>(self, visitor: V) -> Result<V::Value, DeserializeError>
    where V: ValueVisitor;
}
```

### `ValueVisitor`

Basically just copied from Serde except with more specialised types like datetimes, IPs, etc...

All the methods implement a default method for convenience, you can just choose what types you support or not.

```rust
/// A value visitor for deserializing a document value.
///
/// This is strongly inspired by Serde but has a few extra types.
///
/// TODO: Improve docs
pub trait ValueVisitor {
    /// The value produced by the visitor.
    type Value;

    #[inline]
    /// Called when the deserializer visits a string value.
    fn visit_null(&self) -> Result<Self::Value, DeserializeError> {
        Err(DeserializeError::UnsupportedType(ValueType::Null))
    }

    #[inline]
    /// Called when the deserializer visits a string value.
    fn visit_string(&self, _val: String) -> Result<Self::Value, DeserializeError> {
        Err(DeserializeError::UnsupportedType(ValueType::String))
    }

    #[inline]
    /// Called when the deserializer visits a u64 value.
    fn visit_u64(&self, _val: u64) -> Result<Self::Value, DeserializeError> {
        Err(DeserializeError::UnsupportedType(ValueType::U64))
    }

    #[inline]
    /// Called when the deserializer visits a i64 value.
    fn visit_i64(&self, _val: i64) -> Result<Self::Value, DeserializeError> {
        Err(DeserializeError::UnsupportedType(ValueType::I64))
    }

    #[inline]
    /// Called when the deserializer visits a f64 value.
    fn visit_f64(&self, _val: f64) -> Result<Self::Value, DeserializeError> {
        Err(DeserializeError::UnsupportedType(ValueType::F64))
    }

    #[inline]
    /// Called when the deserializer visits a bool value.
    fn visit_bool(&self, _val: bool) -> Result<Self::Value, DeserializeError> {
        Err(DeserializeError::UnsupportedType(ValueType::Bool))
    }

    #[inline]
    /// Called when the deserializer visits a datetime value.
    fn visit_datetime(&self, _val: DateTime) -> Result<Self::Value, DeserializeError> {
        Err(DeserializeError::UnsupportedType(ValueType::DateTime))
    }

    #[inline]
    /// Called when the deserializer visits an IP address value.
    fn visit_ip_address(&self, _val: Ipv6Addr) -> Result<Self::Value, DeserializeError> {
        Err(DeserializeError::UnsupportedType(ValueType::IpAddr))
    }

    #[inline]
    /// Called when the deserializer visits a facet value.
    fn visit_facet(&self, _val: Facet) -> Result<Self::Value, DeserializeError> {
        Err(DeserializeError::UnsupportedType(ValueType::Facet))
    }

    #[inline]
    /// Called when the deserializer visits a bytes value.
    fn visit_bytes(&self, _val: Vec<u8>) -> Result<Self::Value, DeserializeError> {
        Err(DeserializeError::UnsupportedType(ValueType::Bytes))
    }

    #[inline]
    /// Called when the deserializer visits a pre-tokenized string value.
    fn visit_pre_tokenized_string(
        &self,
        _val: PreTokenizedString,
    ) -> Result<Self::Value, DeserializeError> {
        Err(DeserializeError::UnsupportedType(ValueType::PreTokStr))
    }

    #[inline]
    /// Called when the deserializer visits an array.
    fn visit_array<'de, A>(&self, _access: A) -> Result<Self::Value, DeserializeError>
    where A: ArrayAccess<'de> {
        Err(DeserializeError::UnsupportedType(ValueType::Array))
    }

    #[inline]
    /// Called when the deserializer visits an object value.
    fn visit_object<'de, A>(&self, _access: A) -> Result<Self::Value, DeserializeError>
    where A: ObjectAccess<'de> {
        Err(DeserializeError::UnsupportedType(ValueType::Object))
    }
}
```

### Collection access (`ArrayAccess` and `ObjectAccess`)
These are again very similar to Serde, and allow the user to incrementally deserialize entries without doing several passes over the types. It also allows arrays to have varying types rather than one strict type.

```rust
/// Access to a sequence of values which can be deserialized.
pub trait ArrayAccess<'de> {
    /// A indicator as to how many values are in the object.
    ///
    /// This can be used to pre-allocate entries but should not
    /// be depended on as a fixed size.
    fn size_hint(&self) -> usize;

    /// Attempts to deserialize the next element in the sequence.
    fn next_element<V: ValueDeserialize>(&mut self) -> Result<Option<V>, DeserializeError>;
}

/// TODO: Improve docs
pub trait ObjectAccess<'de> {
    /// A indicator as to how many values are in the object.
    ///
    /// This can be used to pre-allocate entries but should not
    /// be depended on as a fixed size.
    fn size_hint(&self) -> usize;

    /// Attempts to deserialize the next key-value pair in the object.
    fn next_entry<V: ValueDeserialize>(&mut self) -> Result<Option<(String, V)>, DeserializeError>;
}
```

ChillFish8:
I've marked this PR as not a draft now since I'm not planning to change the API design too much now as this PR is already huge (sorry 😅) 

Any future changes are probably going to be documentation and expanding unit tests, although the serialization and deserialization APIs are fairly well tested I think some of the other bits may want to be checked.

fulmicoton:
@ChillFish8 I had a look at the PR. Thank you for the massive investigation.

- I suspect what we want most here is a trait on the input side, and the output is less important. This is true for quickwit at least. Do you need that on the consumption side too for Lnx ?
- Have you managed to get a perf improvement as a result of your PR?

ChillFish8:
> I suspect what we want most here is a trait on the input side, and the output is less important. This is true for quickwit at least. Do you need that on the consumption side too for Lnx ?

lnx definitely cares more about the ingest side of things rather than the deserialization side of it since it stores docs separately to the index, the main thing it wants is to be able to index its zero-copy views without having to actually allocate, the deserialization side of things was largely done so everything is consistent and able to support going back to the type you originally provided tantivy with.

> Have you managed to get a perf improvement as a result of your PR?

I've only done some very light perf testing with this so far, in situations where you can just provide an owned value directly and aren't re-allocating, the performance is about the same as previous, I think on average it was maybe a little bit faster, but not by any significant margin.

Whatever perf gain this will bring will basically come out of whether with the trait-based approach, you can avoid additional allocations or transforms on your data. Which I haven't really tested in a more realistic situation.

src/core/json_utils.rs
```diff
@@ -114,53 +115,75 @@ fn index_json_object(
...
     ctx: &mut IndexingContext,
     positions_per_path: &mut IndexingPositionsPerPath,
 ) {
     match json_value {
-        serde_json::Value::Null => {}
-        serde_json::Value::Bool(val_bool) => {
-            json_term_writer.set_fast_value(*val_bool);
+        ReferenceValue::Null => {}
+        ReferenceValue::Str(val) => {
+            // TODO: Maybe this should be removed now we can explicitly state the type can be a
```
PSeitz: Yes this should probably be covered when deserializing to `ReferenceValue`

src/core/json_utils.rs
```diff
@@ -114,53 +115,75 @@ fn index_json_object(
...
+            postings_writer.subscribe(doc, 0u32, json_term_writer.term(), ctx);
+        }
+        ReferenceValue::Facet(_) => {
+            unimplemented!("Facet support in dynamic fields is not yet implemented")
+        }
+        ReferenceValue::IpAddr(_) => {
+            unimplemented!("IP address support in dynamic fields is not yet implemented")
+        }
+        ReferenceValue::Date(_) => {
+            unimplemented!("Datetime support in dynamic fields is not yet implemented")
```
PSeitz: We can easily support datetime

```rust
json_term_writer.set_fast_value(dt);
postings_writer.subscribe(doc, 0u32, json_term_writer.term(), ctx);
```

src/core/searcher.rs
```diff
@@ -83,7 +84,7 @@ impl Searcher {
     ///
     /// The searcher uses the segment ordinal to route the
     /// request to the right `Segment`.
-    pub fn doc(&self, doc_address: DocAddress) -> crate::Result<Document> {
+    pub fn doc<D: DocumentAccess>(&self, doc_address: DocAddress) -> crate::Result<D> {
```
PSeitz: Should this generic be moved to the searcher so we can have a default?

src/directory/mmap_directory.rs
```diff
@@ -31,12 +32,26 @@ pub(crate) fn make_io_err(msg: String) -> io::Error {
     io::Error::new(io::ErrorKind::Other, msg)
 }
 
+#[cfg(unix)]
 /// Returns `None` iff the file exists, can be read, but is empty (and hence
-/// cannot be mmapped)
-fn open_mmap(
+/// cannot be mmapped), this method also allows for optional [Advice] to be
+/// specified for the given workload.
+fn open_mmap_with_advice(
```
PSeitz: Merge conflict? There's already an `open_mmap_with_advice` on main

src/fastfield/writer.rs
```diff
@@ -308,31 +295,61 @@ fn record_json_value_to_columnar_writer(
...
+                doc,
+                json_path_writer.as_str(),
+                NumericalValue::from(val),
+            );
+        }
+        ReferenceValue::Bool(val) => {
+            columnar_writer.record_bool(doc, json_path_writer, val);
+        }
+        ReferenceValue::Date(_) => {
+            unimplemented!("Datetime support in dynamic fields is not yet implemented")
```
PSeitz: We can support that. There's a precision config missing on JSON, so we could just store it with the highest precision

src/indexer/segment_writer.rs
```diff
@@ -157,15 +155,17 @@ impl SegmentWriter {
             + self.segment_serializer.mem_usage()
     }
 
-    fn index_document(&mut self, doc: &Document) -> crate::Result<()> {
+    fn index_document<D: DocumentAccess>(&mut self, doc: &D) -> crate::Result<()> {
         let doc_id = self.max_doc;
+        // TODO: This should be removed...
```
PSeitz: Why should this be removed?
There was a reason for the grouping, but I don't remember

src/schema/document/core.rs
```diff
@@ -0,0 +1,310 @@
+use std::collections::{BTreeMap, HashMap, HashSet};
+use std::net::Ipv6Addr;
```
PSeitz: Not a fan of calling anything `core`, since it's typically not well defined what's core and what's not

In this case I'd rather move it to `mod.rs` or pick a more concrete name

src/directory/mmap_directory.rs
ChillFish8: This is because I built this pr on top of my windows fix originally

src/core/searcher.rs
ChillFish8: If I remember right, adding the generic to the searcher itself caused a whole world of pain with us referencing it everywhere and using it in places like the segment readers

src/schema/document/core.rs
```diff
@@ -0,0 +1,310 @@
...
+impl From<Vec<FieldValue>> for Document {
+    fn from(field_values: Vec<FieldValue>) -> Self {
+        Self { field_values }
+    }
+}
+
+impl PartialEq for Document {
+    fn eq(&self, other: &Self) -> bool {
+        // super slow, but only here for tests
+        let convert_to_comparable_map = |field_values: &[FieldValue]| {
```
PSeitz: If we iterate like in `segment_writer` it would be faster

src/schema/document/de.rs
```diff
@@ -0,0 +1,1023 @@
...
+impl From<io::Error> for DeserializeError {
+    fn from(error: io::Error) -> Self {
+        Self::CorruptedValue(Arc::new(error))
+    }
+}
+
+/// The core trait for deserializing a document.
+///
+/// TODO: Improve docs
+pub trait DocumentDeserialize: Sized {
```
PSeitz, line 64: We we only store `Document` in the doc store, we could  drop this, right?

src/schema/document/de.rs
```diff
@@ -0,0 +1,1023 @@
...
+    /// This can be used to pre-allocate entries but should not
+    /// be depended on as a fixed size.
+    fn size_hint(&self) -> usize;
+
+    /// Attempts to deserialize the next key-value pair in the object.
+    fn next_entry<V: ValueDeserialize>(&mut self) -> Result<Option<(String, V)>, DeserializeError>;
+}
+
+/// TODO: Improve docs
+pub struct GenericDocumentDeserializer<'de, R> {
```
PSeitz: Why is this called generic? On the format side it seems to assume Vint and BinarySerializable

src/schema/document/helpers.rs
```diff
@@ -0,0 +1,208 @@
+//! Implementations of some of the core traits on varius types to improve the ergonomics
```
PSeitz: I'd rename it from `helpers.rs` to something more concrete like`serde_json_impl.rs` or similar

src/schema/document/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+    where Self: 'a;
+
+    /// Get an iterator iterating over all fields and values in a document.
+    fn iter_fields_and_values(&self) -> Self::FieldsValuesIter<'_>;
+
+    /// Sort and groups the field_values by field.
+    ///
+    /// The result of this method is not cached and is
+    /// computed on the fly when this method is called.
+    fn get_sorted_field_values(&self) -> Vec<(Field, Vec<Self::Value<'_>>)> {
```
PSeitz, line 192: I think we can return a slice instead of the `Vec`

src/schema/facet.rs
```diff
@@ -40,6 +40,7 @@ pub enum FacetParseError {
 /// belonging to a facet also belongs to the ancestor of
 /// its facet. In the example above, `/electronics/tv_and_video/`
 /// and `/electronics`.
+#[repr(transparent)]
```
PSeitz: Why do we need this?

PSeitz:
@ChillFish8 I had a look at it and it looks pretty good. Having `ReferenceValue` and `DocValue` is slightly confusing though.

On the deserialization part, I think it adds considerable complexity without a lot of performance gain. I tend to prefer to drop it, and be able to convert from the doc stored `Document` to your original value instead.

ChillFish8:
> Having ReferenceValue and DocValue is slightly confusing though.

I agree it's a little confusing, but it's somewhat of a necessary evil, I'm not sure if there is a better name for a custom document value, but the API for lazily creating the reference values is quite an important feature (for me at least)

> On the deserialization part, I think it adds considerable complexity without a lot of performance gain. I tend to prefer to drop it, and be able to convert from the doc stored Document to your original value instead.

While I don't have any strong opinions on this, I do think it's potentially even more confusing if we go "Ah yes you can serialize with a customs value but you can't deserialize back to a custom value, you have to use the concrete `Document` type" and also you do lose a little bit of efficiency having to re-construct your types from the `Document` type rather than walking through with the deserializer, which I think can become more apparent at higher QPS.

src/schema/document/core.rs
ChillFish8: Probably, this was from the original code so I didn't bother changing it.

src/indexer/segment_writer.rs
ChillFish8: My original thought was this seems to be a bit of optimisation around the `Document` type storing it's multi-value fields individually, so my original thinking was this could just get removed and we simply go "If you actually give an array iterator for the field then you can probably a bit more efficient and not need to sort and group and do those small allocations" That being said, having used this PR for a bit, I've found myself not really using the array iterator that much just due to how lnx processes documents.

src/schema/facet.rs
ChillFish8: This got left over from when I was trying to work around the facet being an owned value making it somewhat awkward to use on borrowed data without allocating each time.

I ended up giving up because making it use borrowed data made things worse, so this can be removed.

PSeitz:
> While I don't have any strong opinions on this, I do think it's potentially even more confusing if we go "Ah yes you can serialize with a customs value but you can't deserialize back to a custom value, you have to use the concrete `Document` type" and also you do lose a little bit of efficiency having to re-construct your types from the `Document` type rather than walking through with the deserializer, which I think can become more apparent at higher QPS.

We don't need to serialize with a custom value, but just provide an API over the data.
The same API can be provided back via a `Document` (maybe also deserialize directly from the Document format is possible).
If the user serializes the value, it would also move things like format handling to the user.
I tend to think that this is not worth the added complexity. I don't have data on the cost of deserialization for high QPS cases.

ChillFish8:
> We don't need to serialize with a custom value, but just provide an API over the data.
The same API can be provided back via a Document (maybe also deserialize directly from the Document format is possible).
If the user serializes the value, it would also move things like format handling to the user.
I tend to think that this is not worth the added complexity. I don't have data on the cost of deserialization for high QPS cases.

I see your point, but I don't think removing it is adding a huge amount of additional complexity, it makes the code a bit more sparse and abstracted, but the way we serialize and deserialize data is exactly the same, I just re-wrote it from a somewhat dense block of code on the `Document` type to avoid make it work well with custom types when indexing.

The only thing I can see happening is if we provide custom documents for indexing but not retrieval, the question will inevitably come up as "Why?" which I can understand, also it can be quite convenient deserializing into a custom type but it's not necessarily the end of the world I suppose.

src/schema/document/de.rs
ChillFish8: By generic I meant in the literal sense, i.e. it's generic over a given reader. Probably can be renamed to something better.

PSeitz:
> 
> I see your point, but I don't think removing it is adding a huge amount of additional complexity, it makes the code a bit more sparse and abstracted, but the way we serialize and deserialize data is exactly the same, I just re-wrote it from a somewhat dense block of code on the `Document` type to avoid make it work well with custom types when indexing.
> 
> The only thing I can see happening is if we provide custom documents for indexing but not retrieval, the question will inevitably come up as "Why?" which I can understand, also it can be quite convenient deserializing into a custom type but it's not necessarily the end of the world I suppose.

I had another look and misunderstood it before. There's no custom serialization format, right?
In that case I think it's fine.

Overall the PR looks good, nice job!

ChillFish8:
> I had another look and misunderstood it before. There's no custom serialization format, right?
In that case I think it's fine.

Format is effectively the exact same as it was before, just with the additional codes to handle collections and objects, but it is not user defined.

PSeitz (approved):

src/schema/document/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+            Some(val)
+        } else {
+            None
+        }
+    }
+
+    #[inline]
+    /// If the Value is a pre-tokenized string, returns the associated string. Returns None
+    /// otherwise.
+    fn as_tokenized_text(&self) -> Option<&'a PreTokenizedString> {
```
fulmicoton: `as_pretokenized_text`...

fulmicoton:
@PSeitz You can merge whenever you see fit.

PSeitz:
`test_json_indexing` fails because `"complexobject": {"field.with.dot": 1}` is indexed as u64 and then queried with i64.
The current order is i64,u64,f64. I'm not sure what's the best way to handle this when the type is user provided. Maybe it should not be user provided.

src/schema/value.rs
```diff
@@ -298,188 +322,71 @@ impl From<PreTokenizedString> for Value {
...
-                panic!("Expected a json object.");
+    fn from(value: serde_json::Value) -> Self {
+        match value {
+            serde_json::Value::Null => Self::Null,
+            serde_json::Value::Bool(val) => Self::Bool(val),
+            serde_json::Value::Number(number) => {
+                if let Some(val) = number.as_u64() {
+                    Self::U64(val)
+                } else if let Some(val) = number.as_i64() {
+                    Self::I64(val)
```
PSeitz:
```suggestion
                if let Some(val) = number.as_i64() {
                    Self::I64(val)
                } else if let Some(val) = number.as_u64() {
                    Self::U64(val)
```

src/schema/value.rs
```diff
@@ -298,188 +322,71 @@ impl From<PreTokenizedString> for Value {
...
+                } else if let Some(val) = number.as_f64() {
+                    Self::F64(val)
+                } else {
+                    panic!("Unsupported serde_json number {number}");
+                }
             }
+            serde_json::Value::String(val) => Self::Str(val),
```
PSeitz:
```rust
                if can_be_rfc3339_date_time(&text) {
                    match OffsetDateTime::parse(&text, &Rfc3339) {
                        Ok(dt) => {
                            let dt_utc = dt.to_offset(time::UtcOffset::UTC);
                            Self::Date(DateTime::from_utc(dt_utc))
                        }
                        Err(_) => Self::Str(text),
                    }
                } else {
                    Self::Str(text)
                }
```
```rust
fn can_be_rfc3339_date_time(text: &str) -> bool {
    if let Some(&first_byte) = text.as_bytes().get(0) {
        if first_byte >= b'0' && first_byte <= b'9' {
            return true;
        }
    }

    false
}
```

src/query/fuzzy_query.rs
```diff
@@ -202,18 +202,20 @@ mod test {
...
         }"#,
         )?;
         index_writer.add_document(doc)?;
-        let doc = schema.parse_document(
+        let doc = Document::parse_json(
+            &schema,
             r#"{
             "attributes": {
-                "aa": "japan"
+                "as": "japan"
```
PSeitz:
```suggestion
                "aa": "japan"
```

PSeitz:
I had an extra pass and fixed the remaining issues. Thanks for the really nice PR!

softhuafei:
Thank you for your work. I have a question. The current document serialization interface `BinaryDocumentSerializer::serialize_doc`only serializes stored fields. Is there any other interface that allows us to completely serialize and deserialize the document?

PSeitz:
No, but I think you could add a stored bytes field and put your custom serialized doc there

softhuafei:
> No, but I think you could add a stored bytes field and put your custom serialized doc there

Thank you for your answer. Is there any example of custom document and related serialization and deserialization?

PSeitz:
There is no custom document serialization and deserialization, having your real doc nested in a bytes field would be a workaround.

softhuafei:
Sorry, my previous statement was not clear enough. I want to be able to fully serialize/deserialize TantivyDocument, just like the old version of document:
```
impl BinarySerializable for Document {
    fn serialize<W: Write + ?Sized>(&self, writer: &mut W) -> io::Result<()> {
        let field_values = self.field_values();
        VInt(field_values.len() as u64).serialize(writer)?;
        for field_value in field_values {
            field_value.serialize(writer)?;
        }
        Ok(())
    }

    fn deserialize<R: Read>(reader: &mut R) -> io::Result<Self> {
        let num_field_values = VInt::deserialize(reader)?.val() as usize;
        let field_values = (0..num_field_values)
            .map(|_| FieldValue::deserialize(reader))
            .collect::<io::Result<Vec<FieldValue>>>()?;
        Ok(Document::from(field_values))
    }
}
```

Should I modify `BinaryDocumentSerializer` and add a new method to it, such as `serialize_entire_doc`, which behaves like `serialize_doc`, except that it will serialize all fields?

PSeitz:
`TantivyDocument` has serde support. Would that work for you?

```rust
#[derive(Clone, Debug, serde::Serialize, serde::Deserialize, Default)]
pub struct TantivyDocument {
    field_values: Vec<FieldValue>,
}
```

Issue #1333, opened 2022-04-13
Bytes FastField and FilterCollector


danielhstahl:
**Describe the bug**
The docs state [here](https://docs.rs/tantivy/latest/tantivy/fastfield/index.html) that bytes are supported as a FastField.  Indeed, this code works: `schema_builder.add_bytes_field("file_path", INDEXED | FAST);`

The docs [here](https://docs.rs/tantivy/latest/tantivy/collector/struct.FilterCollector.html) describe using a FilterCollector on FastFields.  However, when I try to implement the FilterCollector for a bytes FastField I cannot get the `predicate` to accept a parameter of type `[u8]` or `Vec<u8>`: "the trait bound `Vec<u8>: FastValue` is not satisfied
the trait `FastValue` is not implemented for `Vec<u8>`"

**Which version of tantivy are you using?**
0.17.0

fulmicoton:
Ah yes... We should probably fix the documentation.
You cannot use the FilterCollector with a bytes field.

Do you need help working around this?
You could implement your own custom FilterCollector, it is quite straightforward.

danielhstahl:
I'm looking into doing a custom filter.  This seems to be where the filtering occurs: https://github.com/quickwit-oss/tantivy/blob/main/src/collector/filter_collector_wrapper.rs#L178-L180. How would I get around the FastField constraint since it seems like bytes are not implemented for FastFields?

fulmicoton:
I'm not sure I understand the question. Here is an example usage of a bytes fastfield reader. https://github.com/quickwit-oss/tantivy/blob/6656a70d1b1d88600fee993c371fdece0cf4caa9/src/fastfield/bytes/mod.rs#L15-L38

danielhstahl:
Ok, I got it.  

The use of the static lifetime seems to be an issue though...I am using a filter based on values that I'm passing in to Tantivy which obviously aren't static so I'm getting a compiler error.

```
fn filter_results(files_lookup: &HashMap<&[u8], &std::string::String>, key: &[u8]) -> bool {
    match files_lookup.get(key) {
        Some(_result) => true,
        None => false,
    }
}
//in another function
//`search_input.search_files` does not live long enough borrowed value does not live long enough
let files_lookup: HashMap<_, _> = search_input
        .search_files
        .iter()
        .map(|v| (v.as_bytes(), v))
        .collect(); //hashmap for O(1) lookups
let filter_files_collector =
[...]

PR #2075, opened 2023-06-06
Add BytesFilterCollector to support filtering based on a bytes fast field

adamreichold:
We ran into #1333 today and after implementing a specialized filtering collector for our use case, I thought it might be a nice community service if fewer people would have to copy & paste this code in the future by providing a variant of `FilterCollector` specialised to work on a bytes fast field.

This seems especially useful as this has gotten a bit more complicated since the introduction of Columnar and `BytesColumn`. (I am not sure if I have used to API optimally myself.)

I also tried to boy-scout the existing `FilterCollector` a little bit while I was working on the module.

### DIFF
--- a/src/collector/filter_collector_wrapper.rs
+++ b/src/collector/filter_collector_wrapper.rs
@@ -6,21 +6,24 @@
 //
 // Of course, you can have a look at the tantivy's built-in collectors
 // such as the `CountCollector` for more examples.
-
-// ---
-// Importing tantivy...
+use std::fmt::Debug;
 use std::marker::PhantomData;
-use std::sync::Arc;
 
-use columnar::{BytesColumn, ColumnValues, DynamicColumn, HasAssociatedColumnType};
+use columnar::{BytesColumn, Column, DynamicColumn, HasAssociatedColumnType};
 
 use crate::collector::{Collector, SegmentCollector};
 use crate::schema::Field;
-use crate::{Score, SegmentReader, TantivyError};
+use crate::{DocId, Score, SegmentReader, TantivyError};
 
 /// The `FilterCollector` filters docs using a fast field value and a predicate.
-/// Only the documents for which the predicate returned "true" will be passed on to the next
-/// collector.
+///
+/// Only the documents containing at least one value for which the predicate returns `true`
+/// will be passed on to the next collector.
+///
+/// In other words,
+/// - documents with no values are filtered out.
+/// - documents with several values are accepted if at least one value matches the predicate.
+///
 ///
 /// ```rust
 /// use tantivy::collector::{TopDocs, FilterCollector};
@@ -116,16 +119,16 @@ where
             )));
         }
 
-        let column_values = segment_reader
+        let column_opt = segment_reader
             .fast_fields()
-            .column_first_or_default(schema.get_field_name(self.field))?;
+            .column_opt(field_entry.name())?;
 
         let segment_collector = self
             .collector
             .for_segment(segment_local_id, segment_reader)?;
 
         Ok(FilterSegmentCollector {
-            column_values,
+            column_opt,
             segment_collector,
             predicate: self.predicate.clone(),
             t_predicate_value: PhantomData,
@@ -144,15 +147,30 @@ where
     }
 }
 
-pub struct FilterSegmentCollector<TSegmentCollector, TPredicate, TPredicateValue>
-where
-    TPredicate: 'static,
-    DynamicColumn: Into<Option<columnar::Column<TPredicateValue>>>,
-{
-    column_values: Arc<dyn ColumnValues<TPredicateValue>>,
+pub struct FilterSegmentCollector<TSegmentCollector, TPredicate, TPredicateValue> {
+    column_opt: Option<Column<TPredicateValue>>,
     segment_collector: TSegmentCollector,
     predicate: TPredicate,
     t_predicate_value: PhantomData<TPredicateValue>,
+}
+
+impl<TSegmentCollector, TPredicate, TPredicateValue>
+    FilterSegmentCollector<TSegmentCollector, TPredicate, TPredicateValue>
+where
+    TPredicateValue: PartialOrd + Copy + Debug + Send + Sync + 'static,
+    TPredicate: 'static + Fn(TPredicateValue) -> bool + Send + Sync,
+{
+    #[inline]
+    fn accept_document(&self, doc_id: DocId) -> bool {
+        if let Some(column) = &self.column_opt {
+            for val in column.values_for_doc(doc_id) {
+                if (self.predicate)(val) {
+                    return true;
+                }
+            }
+        }
+        false
+    }
 }
 
 impl<TSegmentCollector, TPredicate, TPredicateValue> SegmentCollector
@@ -160,15 +178,13 @@ impl<TSegmentCollector, TPredicate, TPredicateValue> SegmentCollector
 where
     TSegmentCollector: SegmentCollector,
     TPredicateValue: HasAssociatedColumnType,
-    TPredicate: 'static + Fn(TPredicateValue) -> bool + Send + Sync,
-    DynamicColumn: Into<Option<columnar::Column<TPredicateValue>>>,
+    TPredicate: 'static + Fn(TPredicateValue) -> bool + Send + Sync, /* DynamicColumn: Into<Option<columnar::Column<TPredicateValue>>> */
 {
     type Fruit = TSegmentCollector::Fruit;
 
     fn collect(&mut self, doc: u32, score: Score) {
-        let value = self.column_values.get_val(doc);
-        if (self.predicate)(value) {
-            self.segment_collector.collect(doc, score)
+        if self.accept_document(doc) {
+            self.segment_collector.collect(doc, score);
         }
     }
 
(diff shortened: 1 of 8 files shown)


adamreichold:
(The example is silly of course, but I was unable to come up with something nicer on the spot. In our use case, we have serialized geometry stored in that bytes fast field which I think would be too involved for the example here especially since it also serves as a unit test.)

src/collector/filter_collector_wrapper.rs
```diff
@@ -168,13 +166,155 @@ where
...
+impl<TSegmentCollector, TPredicate> SegmentCollector
+    for BytesFilterSegmentCollector<TSegmentCollector, TPredicate>
+where
+    TSegmentCollector: SegmentCollector,
+    TPredicate: 'static + Fn(&[u8]) -> bool + Send + Sync,
+{
+    type Fruit = TSegmentCollector::Fruit;
+
+    fn collect(&mut self, doc: u32, score: Score) {
+        let ord = self.column.ords().values.get_val(doc);
```
fulmicoton: !?
this line is quite the footgun if you are dealing with a multivalued column, or an optional column.

The ords() stuff returns a Column. 
A column object is something that contains
- values
- an index that maps efficient a docid to a range of Values.

If you have always one value per doc, the mapping is trivial
and your code works.
Otherwise it is plain wrong.

adamreichold: > If you have always one value per doc, the mapping is trivial
and your code works.
Otherwise it is plain wrong.

Indeed, one value per doc is my use case. I think this restriction applies to the existing `FilterCollector` as well?

Would you prefer to document this restriction or to try to extend the code to handle multiple values? (I think this would raise the issue if the result of applying the predicate to the values should be combined using "logical or" or "logical and"? Or to pass in the an iterator/vector of values to the predicate?)

adamreichold: > Or to pass in the an iterator/vector of values to the predicate?

Noticed that this is somewhat unwieldy without generic closure as the concrete type of the iterator would have to be exposed. So if multiple values should be supported, a boolean flag like `conjunction_of_multiple_values` seems more appropriate. Possibly as constant generic argument `const CONJUNCTION_OF_MULTIPLE_VALUES: bool` to avoid the runtime overhead.

adamreichold: I modified the code to at least handle missing values by passing an empty slice to the predicate since this should correspond to passing the default value for the existing `FilterCollector` if I understand things correctly.

adamreichold: @fulmicoton I guess I am still doing something wrong in the latest version of the code posted here: I just bumped my project from Tantivy 0.19 to 0.20 and changed our filter based on `BytesFastFieldReader` to one based on `BytesColumn` as done here and this regressed a single search query from **40ms** to **920ms**. Uff...

I would be grateful for any advice on how to fix the code here, my benchmark or otherwise avoid this.

For example, is it at all possible to avoid the dictionary encoding for high-cardinality fields like mine?

(Basically the same benchmark without accessing the bytes fast field improves from 40ms to 20ms. The code in question is [here](https://gitlab.opencode.de/umwelt-info/metadaten/-/merge_requests/247). I can also share the index itself if there is interest in that, it is just slightly too large to attach here at 43MB.)

adamreichold: (With Tantivy 0.19, there version with the filter is basically as fast as the version without it because it does yield fewer documents which completely offsets the extra work done for the filtering. But the cardinalities are exactly the same for 0.19 and 0.20.)

adamreichold: ((Of course, the filter in question is not the most efficient way to do spatial filtering and we were planning to switch to `rstar` as a spatial index backing this functionality. But with Tantivy 0.19 we would not even need to do that because filtering on a bytes fast field was sufficiently fast.))

fulmicoton:
@adamreichold The filter collector semantics were a bit obsolete now that we handle any cardinality for doc values. I tweaked it a little. (not your BytesFilterCollector, the original one too)

Can you have a look at my commit?

Also, are you using this in production? The BytesCollector is now using dictionary encoding...
I wouldn't be surprised if you experience a strong performance with the new Columnar implementation.
Let me know if this is used for something "serious", in which case we should offer the possibility to have a different encoding for Bytes fields.

adamreichold:
> @adamreichold The filter collector semantics were a bit obsolete now that we handle any cardinality for doc values. I tweaked it a little. (not your BytesFilterCollector, the original one too)
> 
> Can you have a look at my commit?

Are you sure you pushed this somewhere? I cannot see any new commits on the PR here or on the main branch. The last change to this module appears to be bd5eea98527e74d7f57dd39d62e4ab25ce8a5e5b on which this PR is based.

> Also, are you using this in production? The BytesCollector is now using dictionary encoding... I wouldn't be surprised if you experience a strong performance with the new Columnar implementation. Let me know if this is used for something "serious", in which case we should offer the possibility to have a different encoding for Bytes fields.

Well, we are using the equivalent functionality for Tantivy 0.19 in production. (We cannot upgrade to 0.20 until #2085 is resolved.)

I think for now, I can work around this because our geometry is simple enough (just bounding boxes, i.e. uniformly four floats) that I can replace the serialization of the geometry by just pulling out the individual coordinates of a multi-valued f64 fast field.

This still means I cannot use `FilterCollector` but rather have to write a bespoke collector to handle the cardinality and this approach breaks down when the geometry becomes an arbitrary feature collections which cannot be encoded as just a list of coordinates.

fulmicoton:
@adamreichold oops sorry i just saw you already reported the perf regression. 
I opened #2090 accordingly.

Apprarently I cannot push, so I created `issue/2075-bytes-filter-collector`. Note it does not solve your performance problems at all.

adamreichold:
> I opened https://github.com/quickwit-oss/tantivy/issues/2090 accordingly.

Thanks for considering this.

> Apprarently I cannot push, so I created issue/2075-bytes-filter-collector. Note it does not solve your performance problems at all.

Strange, I did check that box:

image

Will pull the commit into this PR manually...

adamreichold:
> Will pull the commit into this PR manually...

I pulled the commit in. I also fixed some nits throughout the code base as the Clippy build was getting somewhat noisy.

Finally, I took the liberty of adding one more commit adding a `CONJUNCTION` const parameter to the two collectors which determine whether they use `any` (as in your commit) or `all` semantics (by constructing the filter via `new_conjunction` instead of `new`).

The changes does make the type signatures even more unwieldy and the complexity might be confusing. So let me know what you think I can just drop that commit from the PR if it is too much.

src/collector/filter_collector_wrapper.rs
```diff
@@ -146,35 +173,240 @@ where
...
+    #[inline]
+    fn accept_document(&self, doc_id: DocId) -> bool {
+        if let Some(column) = &self.column_opt {
+            for val in column.values_for_doc(doc_id) {
+                if CONJUNCTION ^ (self.predicate)(val) {
+                    return !CONJUNCTION;
+                }
+            }
+        }
+        CONJUNCTION
```
fulmicoton: I don't think this logic is correct.

src/indexer/stamper.rs
```diff
@@ -101,6 +101,7 @@ mod test {
 
     use super::Stamper;
 
+    #[allow(clippy::redundant_clone)]
```
fulmicoton, line 104: Yeah... that's a bug of clippy isn't it?

fulmicoton:
@adamreichold I invited you to the tantivy-dev org, so you can push PR to tantivy-search in the future.

fulmicoton:
> Finally, I took the liberty of adding one more commit adding a CONJUNCTION const parameter to the two collectors > > which determine whether they use any (as in your commit) or all semantics (by constructing the filter via
> new_conjunction instead of new).

> The changes does make the type signatures even more unwieldy and the complexity might be confusing. So let me 
> know what you think I can just drop that commit from the PR if it is too much.

Can we remove the conjunction stuff?

src/indexer/stamper.rs
adamreichold: Yes, the lint does not consider interior mutability.

adamreichold:
> Can we remove the conjunction stuff?

Sure enough, I put it into separate commit because I was not sure whether it was a good thing to add myself.

src/collector/filter_collector_wrapper.rs
adamreichold: I dropped the commit, but would be glad to understand the issue.

I was thinking this would expand to

```rust
if let Some(column) = &self.column_opt {
    for val in column.values_for_doc(doc_id) {
        if false ^ (self.predicate)(val) {
            return !false;
        }
    }
}
false
```

meaning

```rust
if let Some(column) = &self.column_opt {
    for val in column.values_for_doc(doc_id) {
        if (self.predicate)(val) {
            return true;
        }
    }
}
false
```

for `CONJUNCTION == false` and similarly

```rust
if let Some(column) = &self.column_opt {
    for val in column.values_for_doc(doc_id) {
        if true ^ (self.predicate)(val) {
            return !true;
        }
    }
}
true
```

meaning

```rust
if let Some(column) = &self.column_opt {
    for val in column.values_for_doc(doc_id) {
        if !(self.predicate)(val) {
            return false;
        }
    }
}
true
```

for `CONJUNCTION == true`. Don't these loops implement "any" resp. "all" semantics?

fulmicoton: ah sorry yes. Somehow I read `^` as `&` my bad.

fulmicoton (approved):

PR #2080, opened 2023-06-09
remove allocations in split compound words

PSeitz:


src/tokenizer/split_compound_words.rs
```diff
@@ -94,29 +96,31 @@ impl TokenFilter for SplitCompoundWords {
...
-    parts: Vec<Token>,
+    cuts: &'a mut Vec<usize>,
+    parts: &'a mut Vec<Token>,
 }
 
-impl<T: TokenStream> SplitCompoundWordsTokenStream<T> {
+impl<'a, T: TokenStream> SplitCompoundWordsTokenStream<'a, T> {
     // Will use `self.cuts` to fill `self.parts` if `self.tail.token()`
     // can fully be split into consecutive matches against `self.dict`.
     fn split(&mut self) {
```
adamreichold, line 128: I think that if the stream is not fully consumed, `parts` could start out non-empty for the next stream? Hence I would suggest calling `self.parts.clear()` next to `self.cuts.clear()` within `split`.

PSeitz: good catch, but it should be done in `token_stream`

adamreichold: Ah right, `advance` already requires `parts` to be empty to actually end up calling `split`. So doing it up front is necessary. 👍🏽

fulmicoton (approved):

PR #2082, opened 2023-06-09
improvement: make WhitespaceTokenizer also handle unicode whitespaces

Tony-X:


adamreichold:
I wonder if this should be two distinct tokenizers then, e.g. `AsciiWhitespaceTokenizer` and `WhitespaceTokenizer`, or a flag whether to limit this to ASCII. Or is the performance impact of this negligible?

Tony-X:
>  Or is the performance impact of this negligible?

I guess it depends on the context in which we talk about performance. As compared to the actual indexing this should be quite small of a cost. But if we're just talking about checking ascii space v.s. unicode space, for sure there are quite a bit more work per `char` (extra comparisons and lookup).

In general this is a form of `SplitTokenizer` that works with a set of delimiters. We could build distinct types the deals with different set of delimiters (according to the proper definition of WhiteSpace). This is at the cost of a larger API surface.

FWIW, I believe splitting on unicode space is a better default behavior.

fulmicoton:
The whitespace tokenizer does not do much. It does not break on punctuation or anything, making it only useful on generated data. I think we can stick to ascii whitespaces only here.

@Tony-X did you have an actual use case for this?

fulmicoton:
Closing due to lack of response. If we have a good use case we can reopen it.

PR #2086, opened 2023-06-10
Fix is_child_of function not considering the root facet.

### LINKED ISSUE #2085: Cannot collect facets starting from the root in Tantivy 0.20
Applying the following diff

```diff
diff --git a/src/collector/facet_collector.rs b/src/collector/facet_collector.rs
index 6861a9faa..597e4ba12 100644
--- a/src/collector/facet_collector.rs
+++ b/src/collector/facet_collector.rs
@@ -119,17 +119,16 @@ fn facet_depth(facet_bytes: &[u8]) -> usize {
 ///
 ///     {
 ///         let mut facet_collector = FacetCollector::for_field("facet");
-///         facet_collector.add_facet("/lang");
-///         facet_collector.add_facet("/category");
+///         facet_collector.add_facet("/");
 ///         let facet_counts = searcher.search(&AllQuery, &facet_collector)?;
 ///
 ///         // This lists all of the facet counts
 ///         let facets: Vec<(&Facet, u64)> = facet_counts
-///             .get("/category")
+///             .get("/")
 ///             .collect();
 ///         assert_eq!(facets, vec![
-///             (&Facet::from("/category/biography"), 1),
-///             (&Facet::from("/category/fiction"), 3)
+///             (&Facet::from("/lang"), 4),
+///             (&Facet::from("/category"), 4)
 ///         ]);
 ///     }
 ///
```

to the example for `FacetCollector` makes it fail with

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `[(Facet(/lang), 4)]`,
 right: `[(Facet(/lang), 4), (Facet(/category), 4)]`', src/collector/facet_collector.rs:59:9
```

using the just released Tantitvy version 0.20, i.e. it is not possible to collect facets starting with the root.

I am not sure what the problem is, but I suspect that `compute_collapse_mapping_one` does not handle `facet_bytes.len() == 0` which is the case when collecting facets from the root.

adamreichold:
I think this is basically the same issue as #1448 at another place in the code base.

Closes #2085

### DIFF
--- a/src/collector/facet_collector.rs
+++ b/src/collector/facet_collector.rs
@@ -161,6 +161,21 @@ fn facet_depth(facet_bytes: &[u8]) -> usize {
 ///         ]);
 ///     }
 ///
+///     {
+///         let mut facet_collector = FacetCollector::for_field("facet");
+///         facet_collector.add_facet("/");
+///         let facet_counts = searcher.search(&AllQuery, &facet_collector)?;
+///
+///         // This lists all of the facet counts
+///         let facets: Vec<(&Facet, u64)> = facet_counts
+///             .get("/")
+///             .collect();
+///         assert_eq!(facets, vec![
+///             (&Facet::from("/category"), 4),
+///             (&Facet::from("/lang"), 4)
+///         ]);
+///     }
+///
 ///     Ok(())
 /// }
 /// # assert!(example().is_ok());
@@ -284,6 +299,9 @@ impl Collector for FacetCollector {
 fn is_child_facet(parent_facet: &[u8], possible_child_facet: &[u8]) -> bool {
     if !possible_child_facet.starts_with(parent_facet) {
         return false;
+    }
+    if parent_facet.is_empty() {
+        return true;
     }
     possible_child_facet.get(parent_facet.len()).copied() == Some(0u8)
 }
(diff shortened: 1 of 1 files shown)


src/collector/facet_collector.rs
```diff
@@ -789,6 +807,15 @@ mod tests {
         );
         Ok(())
     }
+
+    #[test]
+    fn is_child_facet() {
+        assert!(super::is_child_facet(&b"foo"[..], &b"foo\0bar"[..]));
```
PSeitz, line 813: Can you rename it to `is_child_or_root_facet`

adamreichold: You mean the test or the original function `is_child_facet`?

In case of the original function, I think the suggest name is a bit confusing because it supposedly returns true if the given facet is a child facet, i.e. information on the possible child facet, not the parent facet. `is_parent_or_root_facet` would be better if the case should be noted explicitly in the name but then again every facet is a child facet of the root facet, so the existing name fits IMHO, just the implementation lacks the one crucial check.

PSeitz: ah yes, you are right, I misunderstood how the function is used

PSeitz (approved):

PR #2088, opened 2023-06-11
Align numerical type priority order on the search side.

### LINKED ISSUE #1956: Align type priority order for json numbers
i64, u64, f64

fmassot:
#1978 partially fixes #1956: we should infer number in the following order i64, u64, f64. But when building a query, we call `convert_to_fast_value_and_get_term` to infer the type and this function has the following u64, i64, f64.

This PR fixes the ordering and starts with `i64`.

TODO: 
- [x] fix test
- [x] add a test to check the consistency between ordering at indexing and search.

### DIFF
--- a/src/core/json_utils.rs
+++ b/src/core/json_utils.rs
@@ -212,11 +212,11 @@ pub fn convert_to_fast_value_and_get_term(
             DateTime::from_utc(dt_utc),
         ));
     }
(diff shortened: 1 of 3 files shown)


fulmicoton (approved):

fulmicoton:
thanks! sorry for messing up

PR #2093, opened 2023-06-15
fix: #2078 return error when tokenizer not found while indexing

### LINKED ISSUE #2078: writing and searching shall throw error when the field tokenizer is missing
**Describe the bug**
- What did you do?
See To Reproduce section
- What happened?
Any custom tokenizers are not available to the index object created via Index::open_in_dir(). The missing tokenizer is not treated as an error during writing and searching. From the tantivy user perspective,  the writer can add documents and commit without any error, but fields using these custom tokenizers are not correctly indexed (so are the search query string of these fields)
- What was expected?
Error is thrown during writing and searching.

More background is here https://discord.com/channels/908281611840282624/908286403086024724/1111669445635616819 

0.19.2

**To Reproduce**
1. create an index and use a custom tokenizer for a field
2. load that index via open_in_dir and create a writer
3. writer add document to the index
4. the text for that field is not properly tokenized, because the custom tokenizer is not registered to the loaded index

naveenann:
This is to fix #2078

### DIFF
--- a/src/indexer/segment_writer.rs
+++ b/src/indexer/segment_writer.rs
@@ -458,10 +458,7 @@ mod tests {
     use crate::store::{Compressor, StoreReader, StoreWriter};
     use crate::time::format_description::well_known::Rfc3339;
     use crate::time::OffsetDateTime;
-    use crate::tokenizer::{
-        Language, PreTokenizedString, RemoveLongFilter, Stemmer, TextAnalyzer, Token,
-        WhitespaceTokenizer,
-    };
+    use crate::tokenizer::{PreTokenizedString, Token};
     use crate::{
         DateTime, Directory, DocAddress, DocSet, Document, Index, Postings, Term, TERMINATED,
     };
@@ -912,7 +909,6 @@ mod tests {
         assert_eq!(positions, &[4]); //< as opposed to 3 if we had a position length of 1.
     }
 
-    // ISSUE-#2078 - writing and searching shall throw error when the field tokenizer is missing
     #[test]
     fn test_show_error_when_tokenizer_not_registered() {
         let text_field_indexing = TextFieldIndexing::default()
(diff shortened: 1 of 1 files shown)


src/indexer/segment_writer.rs
```diff
@@ -900,4 +911,42 @@ mod tests {
...
+            .build();
+        let mut schema_builder = Schema::builder();
+        schema_builder.add_text_field("title", text_options);
+        let schema = schema_builder.build();
+        let tempdir = TempDir::new().unwrap();
+        let tempdir_path = PathBuf::from(tempdir.path());
+        let index = Index::create_in_dir(&tempdir_path, schema).unwrap();
+        index
+            .tokenizers()
+            .register("custom_en", custom_en_tokenizer);
```
PSeitz: I think that can be removed, it's unused

src/indexer/segment_writer.rs
```diff
@@ -900,4 +911,42 @@ mod tests {
...
+        let index = Index::open_in_dir(tempdir_path).unwrap();
+        let schema = index.schema();
+        let mut index_writer = index.writer(50_000_000).unwrap();
+        let title = schema.get_field("title").unwrap();
+        let mut document = Document::default();
+        document.add_text(title, "The Old Man and the Sea");
+        index_writer.add_document(document).unwrap();
+        match index_writer.commit() {
```
PSeitz: you can use `unwrap_err()` which will panic if it's not an error

src/indexer/segment_writer.rs
```diff
@@ -900,4 +911,42 @@ mod tests {
         postings.positions(&mut positions);
         assert_eq!(positions, &[4]); //< as opposed to 3 if we had a position length of 1.
     }
+
+    // ISSUE-#2078 - writing and searching shall throw error when the field tokenizer is missing
```
PSeitz: the comment implies this also tests search which is not the case

src/indexer/segment_writer.rs
naveenann: I guess we can remove the tokenizer registration but we may need the index creation else `open_in_dir` will fail

PSeitz (approved):

PR #2094, opened 2023-06-16
move fail dep to dev only

PSeitz:
Changed fail dependency to make it dev-dep only.

fail prevents compilation on wasm target

### DIFF
--- a/src/core/segment_reader.rs
+++ b/src/core/segment_reader.rs
@@ -2,8 +2,6 @@ use std::collections::HashMap;
 use std::sync::{Arc, RwLock};
 use std::{fmt, io};
 
-use fail::fail_point;
-
 use crate::core::{InvertedIndexReader, Segment, SegmentComponent, SegmentId};
 use crate::directory::{CompositeFile, FileSlice};
 use crate::error::DataCorruption;
@@ -151,7 +149,7 @@ impl SegmentReader {
 
         let store_file = segment.open_read(SegmentComponent::Store)?;
 
-        fail_point!("SegmentReader::open#middle");
+        crate::fail_point!("SegmentReader::open#middle");
 
         let postings_file = segment.open_read(SegmentComponent::Postings)?;
         let postings_composite = CompositeFile::open(&postings_file)?;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ census = "0.4.0"
 rustc-hash = "1.1.0"
 thiserror = "1.0.30"
 htmlescape = "0.3.1"
-fail = "0.5.0"
+fail = { version = "0.5.0", optional = true }
 murmurhash32 = "0.3.0"
 time = { version = "0.3.10", features = ["serde-well-known"] }
 smallvec = "1.8.0"
@@ -112,7 +112,7 @@ lz4-compression = ["lz4_flex"]
 snappy-compression = ["snap"]
 zstd-compression = ["zstd"]
 
-failpoints = ["fail/failpoints"]
+failpoints = ["fail", "fail/failpoints"]
 unstable = [] # useful for benches.
 
 quickwit = ["sstable", "futures-util"]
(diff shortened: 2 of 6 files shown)


src/core/segment_reader.rs
```diff
@@ -151,7 +152,10 @@ impl SegmentReader {
 
         let store_file = segment.open_read(SegmentComponent::Store)?;
 
-        fail_point!("SegmentReader::open#middle");
+        #[cfg(test)]
```
adamreichold: Not sure if it is worth it, but you could define your own `fail_point!` macro in say `src/lib.rs` which forwards to `fail::fail_point!` for `#[cfg(test)]` and is a no-op otherwise. This would simplify these call sites and avoid the extra blocks.

PSeitz: Yes that seems cleaner. Also I can't depend on `#[cfg(test)]` because test binaries don't get compiled with it (not sure why)

fulmicoton:
Why though?
The failpoint macro is already no-op if the feature fail/failpoints is disabled.

PSeitz:
> Why though? The failpoint macro is already no-op if the feature fail/failpoints is disabled.

But the crate still gets compiled

fulmicoton:
> But the crate still gets compiled

It is definitely a downside, (it is not a big one and not one we recompile over and over) but a minor one. I prefer that to juggling with the `cfg!()` everywhere.

PSeitz:
> 
> It is definitely a downside, (it is not a big one and not one we recompile over and over) but a minor one. I prefer that to juggling with the `cfg!()` everywhere.

It's centralized at one place now. failpoints is a blocker when compiling to `wasm`

fulmicoton (approved):

fulmicoton:
@PSeitz Approved. When you merge can you tell in the commit message failpoint prevented wasm compilation?

maddanki95:
This commit is causing following failures:
```
running 5 tests
test test_fail_on_commit_segment has been running for over 60 seconds
test test_fail_on_flush_segment has been running for over 60 seconds
test test_fail_on_flush_segment_but_one_worker_remains has been running for over 60 seconds
test test_failpoints_managed_directory_gc_if_delete_fails has been running for over 60 seconds
test test_write_commit_fails has been running for over 60 seconds
test test_fail_on_flush_segment_but_one_worker_remains ... FAILED
test test_write_commit_fails ... FAILED
test test_failpoints_managed_directory_gc_if_delete_fails ... FAILED
test test_fail_on_commit_segment ... FAILED
test test_fail_on_flush_segment ... FAILED

failures:

---- test_fail_on_flush_segment_but_one_worker_remains stdout ----
thread 'test_fail_on_flush_segment_but_one_worker_remains' panicked at 'add_document should have returned an error', tests/failpoints/mod.rs:107:5

---- test_write_commit_fails stdout ----
thread 'test_write_commit_fails' panicked at 'assertion failed: index_writer.commit().is_err()', tests/failpoints/mod.rs:57:5

---- test_failpoints_managed_directory_gc_if_delete_fails stdout ----
thread 'test_failpoints_managed_directory_gc_if_delete_fails' panicked at 'assertion failed: managed_directory.exists(test_path).unwrap()', tests/failpoints/mod.rs:31:5

---- test_fail_on_commit_segment stdout ----
thread 'test_fail_on_commit_segment' panicked at 'assertion failed: index_writer.commit().is_err()', tests/failpoints/mod.rs:122:5

---- test_fail_on_flush_segment stdout ----
thread 'test_fail_on_flush_segment' panicked at 'add_document should have returned an error', tests/failpoints/mod.rs:88:5


failures:
    test_fail_on_commit_segment
    test_fail_on_flush_segment
    test_fail_on_flush_segment_but_one_worker_remains
    test_failpoints_managed_directory_gc_if_delete_fails
    test_write_commit_fails

test result: FAILED. 0 passed; 5 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2892.20s
```
Steps:
1. Cloned the git repo
2. cargo build
3. cargo test

Also `tests::test_validate_checksum ` is long running now in 0.21 version

PR #2096, opened 2023-06-19
Add boxed token filter to ease the building of TextAnalyzer with...

fmassot:
...a vec of filters.

All those additional traits to get rid of GATs are a bit confusing for the reader, unfortunately. 

Example:

```rust
    #[test]
    fn test_text_analyzer_with_filters_boxed() {
        let mut analyzer = TextAnalyzer::build(
            WhitespaceTokenizer::default(),
            vec![
                BoxTokenFilter::from(AlphaNumOnlyFilter),
                BoxTokenFilter::from(LowerCaser),
                BoxTokenFilter::from(RemoveLongFilter::limit(6)),
            ],
        );
        let mut stream = analyzer.token_stream("- first bullet point");
        assert_eq!(stream.next().unwrap().text, "first");
        assert_eq!(stream.next().unwrap().text, "point");
    }
```

PR #2101, opened 2023-06-25
Change TokenFilter trait to simplify (a bit) the boxing of filters

fmassot:
I have a preference for this PR instead of this one: #2096

The main difference is that I changed the `TokenFilter` to make it independent of the `Tokenizer` trait as a `TokenFilter` should be defined only by the input token stream and its output token stream.

The main benefit compared to #2096 is that it avoids creating a `BoxTokenizer` and implementing the `Tokenizer` trait on it... it does not seem to be a substantial benefit but... except it is cleaner. For example in #2096, I tried to bypass `BoxTokenizer` and implemented the `Tokenizer` trait directly on `Box<dyn BoxableTokenizer>`, but I  got stack overflow as I was calling recursively the same method `box_token_stream` but the compiler did not tell me anything... 

Bench results:

```console
cargo bench --bench analyzer
    Finished bench [optimized + debuginfo] target(s) in 0.12s
warning: the following packages contain code that will be rejected by a future version of Rust: quick-xml v0.22.0
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
     Running benches/analyzer.rs (target/release/deps/analyzer-2a32537b8a3b3efa)
Gnuplot not found, using plotters backend
default-tokenize-alice  time:   [874.40 µs 875.34 µs 876.63 µs]
                        change: [-2.0579% -1.8172% -1.5706%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 16 outliers among 200 measurements (8.00%)
  8 (4.00%) high mild
  8 (4.00%) high severe

default-dynamic-tokenize-alice
                        time:   [1.1158 ms 1.1165 ms 1.1173 ms]
                        change: [-1.9267% -1.6803% -1.4354%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 200 measurements (4.00%)
  2 (1.00%) high mild
  6 (3.00%) high severe
```

I also ran the benchmark on tantivy `0.19` before the introduction of GATs, box token streams were used everywhere at that time.

```console
default-tokenize-alice  time:   [1.0371 ms 1.0376 ms 1.0383 ms]
```

So we have a performance regression on the box token filters, unfortunately. I tried to change the code to have something equivalent to what we had on 0.19 but... this bench results did not change.

### DIFF
--- a/src/tokenizer/ascii_folding_filter.rs
+++ b/src/tokenizer/ascii_folding_filter.rs
@@ -5,34 +5,35 @@ use super::{Token, TokenFilter, TokenStream};
 /// This class converts alphabetic, numeric, and symbolic Unicode characters
 /// which are not in the first 127 ASCII characters (the "Basic Latin" Unicode
 /// block) into their ASCII equivalents, if one exists.
-#[derive(Clone)]
-pub struct AsciiFoldingFilter;
+#[derive(Clone, Default)]
+pub struct AsciiFoldingFilter(String);
 
 impl TokenFilter for AsciiFoldingFilter {
-    type OutputTokenStream<T: TokenStream> = AsciiFoldingFilterTokenStream<T>;
+    type OutputTokenStream<'a, T: TokenStream> = AsciiFoldingFilterTokenStream<'a, T>;
 
-    fn filter<T: TokenStream>(&self, token_stream: T) -> Self::OutputTokenStream<T> {
+    fn filter<'a, T: TokenStream>(&'a mut self, token_stream: T) -> Self::OutputTokenStream<'a, T> {
+        self.0.clear();
         AsciiFoldingFilterTokenStream {
-            buffer: String::new(),
+            buffer: &mut self.0,
             tail: token_stream,
         }
     }
 }
 
-pub struct AsciiFoldingFilterTokenStream<T> {
-    buffer: String,
+pub struct AsciiFoldingFilterTokenStream<'a, T> {
+    buffer: &'a mut String,
     tail: T,
 }
 
-impl<'a, T: TokenStream> TokenStream for AsciiFoldingFilterTokenStream<T> {
+impl<'a, T: TokenStream> TokenStream for AsciiFoldingFilterTokenStream<'a, T> {
     fn advance(&mut self) -> bool {
         if !self.tail.advance() {
             return false;
         }
         if !self.token_mut().text.is_ascii() {
             // ignore its already ascii
-            to_ascii(&self.tail.token().text, &mut self.buffer);
-            mem::swap(&mut self.tail.token_mut().text, &mut self.buffer);
+            to_ascii(&self.tail.token().text, self.buffer);
+            mem::swap(&mut self.tail.token_mut().text, self.buffer);
         }
         true
     }
@@ -1563,7 +1564,7 @@ mod tests {
     fn folding_helper(text: &str) -> Vec<String> {
         let mut tokens = Vec::new();
         TextAnalyzer::builder(SimpleTokenizer::default())
-            .filter(AsciiFoldingFilter)
+            .filter(AsciiFoldingFilter::default())
             .build()
             .token_stream(text)
             .process(&mut |token| {
@@ -1574,7 +1575,7 @@ mod tests {
 
     fn folding_using_raw_tokenizer_helper(text: &str) -> String {
         let mut tokenizer = TextAnalyzer::builder(RawTokenizer::default())
-            .filter(AsciiFoldingFilter)
+            .filter(AsciiFoldingFilter::default())
             .build();
         let mut token_stream = tokenizer.token_stream(text);
         token_stream.advance();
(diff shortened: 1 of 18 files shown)


fmassot:
HDFS benchmark


tantivy 0.19:

```
index-hdfs/index-hdfs-no-commit
                        time:   [329.88 ms 332.48 ms 335.11 ms]
                        thrpt:  [63.766 MiB/s 64.272 MiB/s 64.777 MiB/s]
index-hdfs/index-hdfs-with-commit
                        time:   [473.94 ms 478.03 ms 482.20 ms]
                        thrpt:  [44.315 MiB/s 44.703 MiB/s 45.088 MiB/s]
```

tantivy main-branch, with default tokenizer built with boxed token filters, this is mostly equivalent to what was in tantivy 0.19

```
index-hdfs/index-hdfs-no-commit
                        time:   [286.20 ms 288.50 ms 290.75 ms]
                        thrpt:  [73.495 MiB/s 74.068 MiB/s 74.665 MiB/s]
index-hdfs/index-hdfs-with-commit
                        time:   [419.36 ms 422.07 ms 424.78 ms]
                        thrpt:  [50.306 MiB/s 50.629 MiB/s 50.956 MiB/s]
```

tantivy main-branch, with default tokenizer

```
index-hdfs/index-hdfs-no-commit
                        time:   [221.96 ms 224.18 ms 226.59 ms]
                        thrpt:  [94.309 MiB/s 95.320 MiB/s 96.276 MiB/s]
index-hdfs/index-hdfs-with-commit
                        time:   [352.42 ms 354.94 ms 357.60 ms]
                        thrpt:  [59.756 MiB/s 60.205 MiB/s 60.635 MiB/s]
```

fulmicoton:
If I understand the PR correctly, this attempts to get a best of two world situation.
The `TextAnalyzerBuilder` makes it possible to have a statically linked chain of tokenizers... but it also makes it possible to append a list of dynamic Box<dyn Filter>. Correct?

You mentioned you are seeing a regression, how do we explain it?

src/tokenizer/tokenizer.rs
```diff
@@ -1,37 +1,97 @@
...
+    ///     vec![
+    ///        BoxTokenFilter::from(RemoveLongFilter::limit(40)),
+    ///        BoxTokenFilter::from(LowerCaser),
+    ///        BoxTokenFilter::from(Stemmer::default()),
+    ///     ]);
+    /// ```
+    pub fn new<T: Tokenizer>(tokenizer: T, token_filters: Vec<BoxTokenFilter>) -> TextAnalyzer {
         TextAnalyzer {
-            tokenizer: self.tokenizer.box_clone(),
+            tokenizer: Box::new(tokenizer),
```
fulmicoton, line 83: I would have preferred to not have this public constructor, and instead have an extra

`TextAnalyzerBuilder::build_with_dynamic_token_filters(token_filters: Vec<BoxTokenFilter>)`, with some documentation explaining about the purpose of two ways to add filters.

src/tokenizer/ascii_folding_filter.rs
```diff
@@ -9,48 +9,30 @@ use super::{Token, TokenFilter, TokenStream, Tokenizer};
...
-    }
-}
-
-pub struct AsciiFoldingFilterTokenStream<'a, T> {
-    buffer: &'a mut String,
     tail: T,
 }
 
-impl<'a, T: TokenStream> TokenStream for AsciiFoldingFilterTokenStream<'a, T> {
+impl<'a, T: TokenStream> TokenStream for AsciiFoldingFilterTokenStream<T> {
```
fulmicoton: clippy...

src/tokenizer/lower_caser.rs
```diff
@@ -1,42 +1,24 @@
...
-    type Tokenizer<T: Tokenizer> = LowerCaserFilter<T>;
+    type OutputTokenStream<T: TokenStream> = LowerCaserTokenStream<T>;
 
-    fn transform<T: Tokenizer>(self, tokenizer: T) -> Self::Tokenizer<T> {
-        LowerCaserFilter {
-            tokenizer,
+    fn filter<T: TokenStream>(&self, token_stream: T) -> Self::OutputTokenStream<T> {
+        LowerCaserTokenStream {
+            tail: token_stream,
             buffer: String::new(),
```
fulmicoton: you are reverting here the work done by Pascal to reduce the number of allocation we have. Is this necessary?

fulmicoton: @Pseitz can you confirm?

fmassot: Yes indeed :/

On each `tokenizer.token_stream`, it will call `filter` and thus will do a `String::new()`.

That was not the case before.

PSeitz: I'd prefer if we can have an instance with a buffer in the pipeline like before

fmassot: Yes, of course. I should have fixed it by adding the allocation at the `TokenFilter` level.

fmassot:
closing in favor of #2110

PR #2102, opened 2023-06-25
Ngram tokenizer now returns an error with invalid arguments.

### LINKED ISSUE #2100: NgramTokenizer::new should return a result
On invalid arguments, it should return `crate::Result<NgramTokenizer>` with proper `TantivyError::InvalidArgument`.

fmassot:
Fixes #2100


trinity-1686a (approved):
lgtm

Issue #1789, opened 2023-01-14
Add `missing` parameter on relevant aggregations


guilload:
This value will be used for sparse fields and ignored otherwise.

Relevant aggs (non-exhaustive):
`avg`, `stats`, `sum`, ...

PR #2103, opened 2023-06-26
add missing parameter to term agg

PSeitz:
missing parameter on term aggregation

Limitations:
Mixed types columns are not supported.
Missing key of type string is not supported on numerical fields 

It requires a special missing aggregation to address those limitations

https://github.com/quickwit-oss/tantivy/issues/1913
https://github.com/quickwit-oss/tantivy/issues/1789

### DIFF
--- a/columnar/src/block_accessor.rs
+++ b/columnar/src/block_accessor.rs
@@ -1,3 +1,5 @@
+use std::cmp::Ordering;
+
 use crate::{Column, DocId, RowId};
 
 #[derive(Debug, Default, Clone)]
@@ -61,14 +63,18 @@ where F: FnMut(u32) {
     let mut hit = hits_iter.next();
 
     while let (Some(&current_doc), Some(&current_hit)) = (doc, hit) {
-        if current_doc < current_hit {
-            callback(current_doc);
-            doc = docs_iter.next();
-        } else if current_doc == current_hit {
-            doc = docs_iter.next();
-            hit = hits_iter.next();
-        } else {
-            hit = hits_iter.next();
+        match current_doc.cmp(&current_hit) {
+            Ordering::Less => {
+                callback(current_doc);
+                doc = docs_iter.next();
+            }
+            Ordering::Equal => {
+                doc = docs_iter.next();
+                hit = hits_iter.next();
+            }
+            Ordering::Greater => {
+                hit = hits_iter.next();
+            }
         }
     }
 
--- a/columnar/src/block_accessor.rs
+++ b/columnar/src/block_accessor.rs
@@ -4,6 +4,7 @@ use crate::{Column, DocId, RowId};
 pub struct ColumnBlockAccessor<T> {
     val_cache: Vec<T>,
     docid_cache: Vec<DocId>,
+    missing_docids_cache: Vec<DocId>,
     row_id_cache: Vec<RowId>,
 }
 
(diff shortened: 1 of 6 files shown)


src/aggregation/agg_req_with_accessor.rs
```diff
@@ -80,10 +83,79 @@ impl AggregationWithAccessor {
...
+                                    };
+
+                                    col.0 = new_col;
+                                }
+                                // Allow fallback to number on text fields
+                                Key::F64(_) if !match_type && col.1 == ColumnType::Str => {
+                                    let new_col = Column {
+                                        index: columnar::ColumnIndex::Full,
+                                        // Replace u64::MAX later with actual term
+                                        values: col.0.first_or_default_col(u64::MAX),
```
fulmicoton: is u64::MAX mapping to NaN? Why is this working?

columnar/src/block_accessor.rs
```diff
@@ -34,3 +49,76 @@ impl<T: PartialOrd + Copy + std::fmt::Debug + Send + Sync + 'static + Default>
             .zip(self.val_cache.iter().cloned())
     }
 }
+
+fn find_missing_docs<F>(docs: &[u32], hits: &[u32], mut callback: F)
```
fulmicoton, line 57: needs doc explaining the two arguments.

PSeitz: added some docs

src/aggregation/bucket/term_agg.rs
```diff
@@ -379,9 +392,16 @@ impl SegmentAggregationCollector for SegmentTermCollector {
 
         let mem_pre = self.get_memory_consumption();
 
-        bucket_agg_accessor
-            .column_block_accessor
-            .fetch_block(docs, &bucket_agg_accessor.accessor);
+        if let Some(missing) = bucket_agg_accessor.missing_accessor1 {
```
fulmicoton: I don''t understand what happens when we have two columns of type string and int
For instance, if we have 3 docs.

doc1 -> "hello"
doc2 -> 3
doc3 -> missing

And we want a term aggregation with missing = "N/A".
With logic below, don't we end up with 

```
"hello" -> 1
3 -> 1
"N/A" -> 3 (instead of 1)
```

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -35,6 +36,9 @@ pub struct AggregationWithAccessor {
     /// based on search terms. That is not that case currently, but eventually this needs to be
     /// Option or moved.
     pub(crate) accessor: Column<u64>,
+    /// Load insert u64 for missing use case
+    pub(crate) missing_accessor1: Option<u64>,
```
fulmicoton: this is a strange name.

`missing_value_for_accessor1` maybe?

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -80,10 +88,51 @@ impl AggregationWithAccessor {
...
+                    Some(&allowed_column_types),
+                    fallback_type,
+                )?;
                 let first = columns.pop().unwrap();
                 accessor2 = columns.pop();
+                if let Some(missing) = missing {
+                    let assign_col_with_missing =
+                        |col: &ColumnType, missing: &Key, match_type: bool| {
+                            // Set missing only if type matches (match_type). This will help to
+                            // avoid duplicates for mixed type
```
fulmicoton: is it correct though in mixed types scenario?

And for non mixed type if the type does not match what happens?

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -80,10 +88,51 @@ impl AggregationWithAccessor {
...
+                    reader,
+                    field_name,
+                    Some(&allowed_column_types),
+                    fallback_type,
+                )?;
                 let first = columns.pop().unwrap();
                 accessor2 = columns.pop();
+                if let Some(missing) = missing {
+                    let assign_col_with_missing =
+                        |col: &ColumnType, missing: &Key, match_type: bool| {
```
fulmicoton: can we make this closure a proper function?

src/aggregation/bucket/term_agg.rs
PSeitz: I added a comment for this on the aggregation request

fulmicoton (approved):

Issue #1676, opened 2022-11-11
order_by_u64_field and order_by_fast_field should allow sorting in ascending order


vikigenius:
**Is your feature request related to a problem? Please describe.**
Currently the `order_by_u64_field` and `order_by_fast_field` functions sort only in descending order. This is a problem and is confusing especially considering that `sort_by_field` does take Order as an argument

**Describe the solution you'd like**
Maybe let the function take Order as an additional argument

**[Optional] describe alternatives you've considered**
The current workaround I am considering is to use i64 fields with negative numbers. But this might not be easy for all use cases.

GregoireDucharme:
Looking for the same kind of solution.
How would you do that using `sort_by_field`, I can't manage to do it

JPDye:
I'd be happy to take this issue on.

PSeitz:
Fixed via https://github.com/quickwit-oss/tantivy/pull/2111

PR #2111, opened 2023-07-04
fixes order_by_u64_field and order_by_fast_field should allow sorting in ascending order #1676

naveenann:


src/collector/top_score_collector.rs
```diff
@@ -1106,4 +1129,44 @@ mod tests {
...
+                ))
+                .unwrap();
+            index_writer
+                .add_document(doc!(
+                    title => "growler of beer",
+                    size => 64u64,
+                ))
+                .unwrap();
+            index_writer
+                .add_document(doc!(
```
PSeitz, line 1152: can you add an empty doc to the test
```rust
index_writer.add_document(doc!()).unwrap();
```

PSeitz: sorry I meant a doc without a value

```rust
index_writer.add_document(doc!(title => "empty beer"))
```

naveenann: ahh ok. without the size field.

src/collector/top_score_collector.rs
```diff
@@ -157,8 +171,13 @@ impl CustomScorer<u64> for ScorerByField {
             sort_column_opt.ok_or_else(|| FastFieldNotAvailableError {
                 field_name: self.field.clone(),
             })?;
+        let mut value = 0u64;
```
PSeitz:
```suggestion
        let mut default_value = 0u64;
```

PSeitz (approved):
LGTM! Thanks!

PR #2120, opened 2023-07-12
Add missing query_terms impl for TermSetQuery.

### LINKED ISSUE #2116: TermSetQuery does not have an impl for Query::query_terms
**Describe the bug**
`Query::query_terms` has a default impl (which is a no-op)

I believe this is why it was forgotten or overlooked to impl it for `TermSetQuery`

0.20.2 & master

**To Reproduce**
https://docs.rs/tantivy/latest/src/tantivy/query/set_query.rs.html#71-75
https://github.com/quickwit-oss/tantivy/blob/main/src/query/set_query.rs#L71C1-L75

adamreichold:
Closes #2116


trinity-1686a (approved):
LGTM!

PR #2121, opened 2023-07-12
Include only built-in compression algorithms as enum variants

### LINKED ISSUE #2119: Compression structs etc should be behind a feature gate
**Describe the bug**
code such as
```rs
Index::builder()
    .schema(schema)
    .settings(IndexSettings {
        sort_by_field: None,
        docstore_compression: Compressor::Zstd(ZstdCompressor {
            compression_level: Some(9),
        }),
        docstore_blocksize: size::KiB as usize * 4,
        docstore_compress_dedicated_thread: true,
    })
```

should refuse to compile if the feature is not pulled in, because otherwise I get a runtime panic.

0.20.2

**To Reproduce**

adamreichold:
This enables compile-time errors when a compression algorithm is requested which is not actually enabled for the current Cargo project. The cost is that indexes using other compression algorithms cannot even be loaded (even though they are not fully accessible in any case).

As a drive-by, this also fixes `--no-default-features` on `cfg(unix)`.

Closes #2119

### DIFF
--- a/src/store/compressors.rs
+++ b/src/store/compressors.rs
@@ -17,12 +17,16 @@ pub enum Compressor {
     /// No compression
     None,
     /// Use the lz4 compressor (block format)
+    #[cfg(feature = "lz4-compression")]
     Lz4,
     /// Use the brotli compressor
+    #[cfg(feature = "brotli-compression")]
     Brotli,
     /// Use the snap compressor
+    #[cfg(feature = "snappy-compression")]
     Snappy,
     /// Use the zstd compressor
+    #[cfg(feature = "zstd-compression")]
     Zstd(ZstdCompressor),
 }
 
@@ -31,9 +35,13 @@ impl Serialize for Compressor {
     where S: serde::Serializer {
         match *self {
             Compressor::None => serializer.serialize_str("none"),
+            #[cfg(feature = "lz4-compression")]
             Compressor::Lz4 => serializer.serialize_str("lz4"),
+            #[cfg(feature = "brotli-compression")]
             Compressor::Brotli => serializer.serialize_str("brotli"),
+            #[cfg(feature = "snappy-compression")]
             Compressor::Snappy => serializer.serialize_str("snappy"),
+            #[cfg(feature = "zstd-compression")]
             Compressor::Zstd(zstd) => serializer.serialize_str(&zstd.ser_to_string()),
         }
     }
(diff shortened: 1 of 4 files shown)


adamreichold:
Just an idea I would to propose for discussion: I think having one really fast compression algorithm (LZ4) and one really flexible one (Zstd) might actually be enough and neither Brotli nor Snappy add anything substantially different.

src/store/compressors.rs
```diff
@@ -45,27 +53,28 @@ impl<'de> Deserialize<'de> for Compressor {
...
-                            "none",
-                            "lz4",
-                            "brotli",
-                            "snappy",
-                            "zstd",
-                            "zstd(compression_level=5)",
-                        ],
-                    ));
-                }
+                return Err(serde::de::Error::unknown_variant(
```
fulmicoton, line 94: that's misleading though.

If the store is compressed with zstd but we don't have the feature flag, we will get an error message saying something like:

It is "zstd" but we expected something in "none", "lz4", "zstd"... etc.

fulmicoton:
I am ok with hiding the enums. Can we however have an explicit error message about what feature is missing when opening an index with a compression format not compiled in the binary?

fulmicoton:
> might actually be enough and neither Brotli nor Snappy add anything substantially different.

I agree.
+1 for ditching Brotli and Snappy.

src/store/compressors.rs
adamreichold: I will just adjust the deserialization error message to mention only those which were built-in and also point to those which could be enabled via additional feature flags.

adamreichold: > I will just adjust the deserialization error message to mention only those which were built-in and also point to those which could be enabled via additional feature flags.

Ah, the error message itself is generated by Serde. I did update the list of expected variants to match what is built-in though.

Should we keep it that way, or should I switch to custom error message to include reference to those which could be enable but are currently not?

fulmicoton: I suggest using a serde custom error, to tell that the compression format xxx is not available and suggestion that tantivy is maybe missing a compilation flag.

Rationale:
The unknown_variant seem appropriate here, but we really want a correct error message surfaced to the user.

adamreichold: Done.

adamreichold:
> > might actually be enough and neither Brotli nor Snappy add anything substantially different.
> 
> I agree. +1 for ditching Brotli and Snappy.

I will propose this as a separate PR then so please see the PR title and can comment further on this.

fulmicoton (approved):

fulmicoton:
@adamreichold thank you!

PR #2122, opened 2023-07-13
add failing test check_num_columnar_fields

PSeitz:


fulmicoton:
The bug is not about the resulting number of column.

The behavior was caused because the user defined a type (Str) for a column and then pushed a value with a different type.
In order to simplify client code, we ensure that a column for the schema field is always created even if it has no values.


The problem is that we accepted the value with a type that is not matching the schema.

When working with json we do not have this issue, as we return an error or coerce the json value that does not match the schema.

It only happens for people using tantivy as a library and pushing values that are not matching the schema type.

A better behavior would be to return an error at index time.
This behavior probably affects stored documents too.

We should probably add some code to check the documents value types before processing them.
One minor downside is that this operation will be pretty useless for documents coming from json parsing. It should be ok since the operation is relatively light.

@PSeitz can you open a ticket about this?

PR #2123, opened 2023-07-13
RFC: Remove support for Brotli and Snappy compression

adamreichold:
This is a follow-up to https://github.com/quickwit-oss/tantivy/pull/2121#issuecomment-1633375542

LZ4 provides fast and simple compression whereas Zstd is exceptionally flexible so that the additional support for Brotli and Snappy does not really add any distinct functionality on top of those two algorithms.

Removing them reduces our maintenance burden and reduces the number of choices users have to make when setting up their project based on Tantivy.


fulmicoton (approved):

fulmicoton:
@adamreichold can you rebase?

adamreichold:
> @adamreichold can you rebase?

Done.

fulmicoton:
Thank you @adamreichold

PR #2126, opened 2023-07-17
Add preserve original on ascii folding filter.

fmassot:
Same as lucene filter: https://lucene.apache.org/core/6_4_1/analyzers-common/org/apache/lucene/analysis/miscellaneous/ASCIIFoldingFilter.html

There is a minor difference between the two filters with `preserve_original` set to `true` and when the original token is different from the folded token: the lucene filter first emits the folded token and then the original token, whereas tantivy first emits the original token and then the folded one.

### DIFF
--- a/src/tokenizer/ascii_folding_filter.rs
+++ b/src/tokenizer/ascii_folding_filter.rs
@@ -5,8 +5,8 @@ use super::{Token, TokenFilter, TokenStream, Tokenizer};
 /// `AsciiFoldingFilter` converts alphabetic, numeric, and symbolic Unicode characters
 /// which are not in the first 127 ASCII characters (the "Basic Latin" Unicode
 /// block) into their ASCII equivalents, if one exists.
-/// If `preserve_original` is `true`, the filter emits both folded token and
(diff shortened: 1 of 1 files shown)


src/tokenizer/ascii_folding_filter.rs
```diff
@@ -31,25 +44,40 @@ impl<T: Tokenizer> Tokenizer for AsciiFoldingFilterWrapper<T> {
...
+            return true;
+        }
         if !self.tail.advance() {
             return false;
         }
+        let mut text_has_changed = false;
         if !self.token_mut().text.is_ascii() {
-            // ignore its already ascii
-            to_ascii(&self.tail.token().text, self.buffer);
+            text_has_changed = to_ascii(&self.tail.token().text, self.buffer);
```
adamreichold, line 74: Can `text_has_changed == false` happen here even though the `is_ascii` test above already failed?

fmassot: just checked the `to_ascii` and as you can expect it will not change a full ascii text... I will fix that, thanks.

PR #2128, opened 2023-07-18
Change in the default fast field tokenizer manager.

fulmicoton:
`{ fast: true }` now results in the use of a the default fast field tokenizer. (instead of no tokenizer) The default tokenizer lowercases.

Fast field gets a different default tokenizer manager than the normal tokenizer.

The serialization of the fast field options is unchanged.

### DIFF
--- a/src/schema/json_object_options.rs
+++ b/src/schema/json_object_options.rs
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 
 use super::text_options::{FastFieldTextOptions, TokenizerName};
 use crate::schema::flags::{FastFlag, SchemaFlagList, StoredFlag};
-use crate::schema::{TextFieldIndexing, TextOptions};
+use crate::schema::{TextFieldIndexing, TextOptions, DEFAULT_FAST_FIELD_TOKENIZER};
 
 /// The `JsonObjectOptions` make it possible to
 /// configure how a json object field should be indexed and stored.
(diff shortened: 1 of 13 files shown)


src/schema/json_object_options.rs
```diff
@@ -130,15 +129,11 @@ impl JsonObjectOptions {
...
-    pub fn set_fast(mut self, tokenizer_name: Option<&str>) -> Self {
-        if let Some(tokenizer) = tokenizer_name {
-            let tokenizer = TokenizerName::from_name(tokenizer);
-            self.fast = FastFieldTextOptions::EnabledWithTokenizer {
-                with_tokenizer: tokenizer,
-            }
-        } else {
-            self.fast = FastFieldTextOptions::IsEnabled(true);
-        }
+    pub fn set_fast(mut self, tokenizer_name: &str) -> Self {
```
fmassot, line 132: question: is it possible to set fast to Disabled?

src/schema/text_options.rs
```diff
@@ -24,19 +24,68 @@ pub struct TextOptions {
...
-    EnabledWithTokenizer { with_tokenizer: TokenizerName },
+    Enabled { tokenizer: TokenizerName },
 }
 
-impl Default for FastFieldTextOptions {
-    fn default() -> Self {
-        FastFieldTextOptions::IsEnabled(false)
+/// Enum used to control the way we serialize fast field text options.
+///
+/// For backward compatiblity reasons, we folow the format introduce in tantivy 0.19.
```
fmassot, line 44:
```suggestion
/// For backward compatibility reasons, we follow the format introduced in tantivy 0.19.
```

src/schema/mod.rs
```diff
@@ -153,6 +153,8 @@ pub use self::term::{Term, ValueBytes, JSON_END_OF_PATH};
 pub use self::text_options::{TextFieldIndexing, TextOptions, STRING, TEXT};
 pub use self::value::Value;
 
+pub(crate) const DEFAULT_FAST_FIELD_TOKENIZER: &str = "default";
```
fmassot, line 156: nitpick: don't we want to call it by its real name `lower`?

fmassot (approved):

src/schema/mod.rs
adamreichold: I think `TokenizerManager::default_for_fast_fields` registers it both under `lower` and under `default`. If I understand things correctly, using `default` would mean the concrete choice is subject to change whereas `lower` should always yield the lower case even in future Tantivy versions?

PSeitz: I would prefer `default_fast_field`, to avoid confusing it with `default`

fulmicoton: @fmassot @adamreichold "default" is the one that will be used when a user is not specifying any tokenizer.
For isntance, when using the flags.

`.add_text(my_field, STRING | FAST)`

It is then possible in tantivy to pass your own tokenizer manager for indexing and for fast fields.
So not putting the specific tokenizer directly here makes it possible for users to pass a tokenizer manager with the default tokenizer they prefer.

PR #2129, opened 2023-07-19
implement lenient parser

### LINKED ISSUE #5: Add a lenient mode to QueryParser

trinity-1686a:
fix #5 

Most of the grammar part is done, however it's not reporting hints just yet (this isn't really required, but would be a nice addition, to help someone understand how their query was understood, and why the parser think it's somewhat wrong).

At the moment, only the part in query-grammar is done. More work is required to convert the AST to a query without failing. This part hasn't been explored yet, but should be before the ticket is considered fixed.

I'm sending this as a draft to get initial comments on the parser part

### DIFF
--- a/query-grammar/src/infallible.rs
+++ b/query-grammar/src/infallible.rs
@@ -1,0 +1,291 @@
+//! nom combinators for infallible operations
+
+use nom::{IResult, InputLength};
+pub type ErrorList = Vec<(usize, String)>;
+pub type JResult<I,O> = IResult<I, (O, ErrorList), std::convert::Infallible>;
+
+// when rfcs#1733 get stabilized, this can make things clearer
+// trait InfallibleParser<I, O> = nom::Parser<I, (O, ErrorList), std::convert::Infallible>;
+
+// TODO space0 and space1: space0 can't fail, space1 can parse nothing but reports missing space
+
+
+/// A variant of the classical `opt` parser, except it returns an infallible error type.
+///
+/// It's less generic than the original to ease type resolution in the rest of the code.
+pub fn opt_i<I: Clone, O, F>(mut f: F) -> impl FnMut(I) -> JResult<I, Option<O>> where
+  F: nom::Parser<I, O, nom::error::Error<I>>,
+{
+  move |input: I| {
+    let i = input.clone();
+    match f.parse(input) {
+      Ok((i, o)) => Ok((i, (Some(o), Vec::new()))),
+      Err(_) => Ok((i, (None, Vec::new()))),
+    }
+  } 
+}
+
+pub fn fallible<I, O, E: nom::error::ParseError<I>, F>(mut f: F) -> impl FnMut(I) -> IResult<I, O, E>
+where F: nom::Parser<I, (O, ErrorList), std::convert::Infallible>
+{   
+    use nom::Err;
+    move |input: I|
+    match f.parse(input) {
+        Ok((input, (output, _err))) => Ok((input, output)),
+        Err(Err::Incomplete(needed)) => Err(Err::Incomplete(needed)),
+        Err(Err::Error(val)) | Err(Err::Failure(val)) => match val {},
+    }
+}
+
+pub fn delimited_infallible<I, O1, O2, O3, F, G, H>(
+  mut first: F,
+  mut second: G,
+  mut third: H,
+) -> impl FnMut(I) -> JResult<I, O2>
+where
+  F: nom::Parser<I, (O1, ErrorList), std::convert::Infallible>,
+  G: nom::Parser<I, (O2, ErrorList), std::convert::Infallible>,
+  H: nom::Parser<I, (O3, ErrorList), std::convert::Infallible>,
+{
+  move |input: I| {
+    let (input, (_, mut err)) = first.parse(input)?;
+    let (input, (o2, mut err2)) = second.parse(input)?;
+    err.append(&mut err2);
+    let (input, (_, mut err3)) = third.parse(input)?;
+    err.append(&mut err3);
+    Ok((input, (o2, err)))
+  }
+}
+
+// Parse nothing. Just a lazy way to not implement terminated/preceded and use delimited instead
+pub fn nothing(i: &str) -> JResult<&str, ()> {
+    Ok((i, ((), Vec::new())))
+}
+
+pub trait TupleInfallible<I, O> {
+  /// Parses the input and returns a tuple of results of each parser.
+  fn parse(&mut self, input: I) -> JResult<I, O>;
+}
+
+impl<Input, Output, F: nom::Parser<Input, (Output, ErrorList), std::convert::Infallible>>
+  TupleInfallible<Input, (Output,)> for (F,)
+{
+  fn parse(&mut self, input: Input) -> JResult<Input, (Output,)> {
+    self.0.parse(input).map(|(i, (o,e))| (i, ((o,),e)))
+  }
+}
+
+
+// these macros are heavily copied from nom, with some minor adaptations for our type
+macro_rules! tuple_trait(
+  ($name1:ident $ty1:ident, $name2: ident $ty2:ident, $($name:ident $ty:ident),*) => (
+    tuple_trait!(__impl $name1 $ty1, $name2 $ty2; $($name $ty),*);
+  );
+  (__impl $($name:ident $ty: ident),+; $name1:ident $ty1:ident, $($name2:ident $ty2:ident),*) => (
+    tuple_trait_impl!($($name $ty),+);
+    tuple_trait!(__impl $($name $ty),+ , $name1 $ty1; $($name2 $ty2),*);
+  );
+  (__impl $($name:ident $ty: ident),+; $name1:ident $ty1:ident) => (
+    tuple_trait_impl!($($name $ty),+);
+    tuple_trait_impl!($($name $ty),+, $name1 $ty1);
+  );
+);
+
+macro_rules! tuple_trait_impl(
+  ($($name:ident $ty: ident),+) => (
+    impl<
+      Input: Clone, $($ty),+ ,
+      $($name: nom::Parser<Input, ($ty, ErrorList), std::convert::Infallible>),+
+    > TupleInfallible<Input, ( $($ty),+ )> for ( $($name),+ ) {
+
+      fn parse(&mut self, input: Input) -> JResult<Input, ( $($ty),+ )> {
+        let mut error_list = Vec::new();
+        tuple_trait_inner!(0, self, input, (), error_list, $($name)+)
+      }
+    }
+  );
+);
+
+macro_rules! tuple_trait_inner(
+  ($it:tt, $self:expr, $input:expr, (), $error_list:expr, $head:ident $($id:ident)+) => ({
+    let (i, (o, mut err)) = $self.$it.parse($input.clone())?;
+    $error_list.append(&mut err);
+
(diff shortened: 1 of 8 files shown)


adamreichold:
The main question that came to mind reading this was whether the lenient parser should fully replace the existing parser. I think as an API between systems, the existing - let's say "strict" - parser is still useful. Of course, as a user interface, a lenient parser is very much preferable.

How do people think about having both parsers and choosing between them at compile time or via a runtime flag? I understand that maintaining two parsers is a significant commitment but I also feel that replacing the existing behaviour in a "flag day" type of change could trigger a certain amount of user backlash.

trinity-1686a:
the goal isn't to remove the strict parser. It was rewritten so both could share some code, but it will definitely be kept for the foreseeable future.

adamreichold:
> the goal isn't to remove the strict parser. It was rewritten so both could share some code, but it will definitely be kept for the foreseeable future.

Ah sorry, then I misunderstood the PR in my first reading and the concern does not apply at all. Thanks for clarifying!

query-grammar/src/query_grammar.rs
```diff
@@ -20,249 +19,359 @@ use crate::Occur;
...
-/// need to be escaped with a backslash character '\'.
-fn field_name<'a>() -> impl Parser<&'a str, Output = String> {
-    static ESCAPED_SPECIAL_CHARS_RE: Lazy<Regex> =
-        Lazy::new(|| Regex::new(ESCAPED_SPECIAL_CHARS_PATTERN).unwrap());
-
-    recognize::<String, _, _>(escaped(
-        (
-            take_while1(|c| !SPECIAL_CHARS.contains(&c) && c != '-'),
-            take_while(|c| !SPECIAL_CHARS.contains(&c)),
+/// consume a field name followed by collon. Return the field name with escape sequence
```
fulmicoton:
```suggestion
/// consume a field name followed by colon. Return the field name with escape sequence
```

query-grammar/src/query_grammar.rs
```diff
@@ -488,56 +888,67 @@ mod test {
...
+        let query_lenient = parse_to_ast_lenient(query).0;
+        let query_lenient_str = format!("{query_lenient:?}");
+        assert_eq!(query_lenient_str, expected, "lenient parser failed");
     }
 
-    fn test_is_parse_err(query: &str) {
-        assert!(parse_to_ast().parse(query).is_err());
+    #[track_caller]
+    fn test_is_parse_err(query: &str, lenient_expected: &str) {
+        assert!(parse_to_ast(query).is_err(), "strict parser succeeded");
```
fulmicoton:
```suggestion
        assert!(parse_to_ast(query).is_err(), "strict parser succeeded where an error was expected.");
```

query-grammar/src/lib.rs
```diff
@@ -1,19 +1,25 @@
...
 
+/// Parse a query
 pub fn parse_query(query: &str) -> Result<UserInputAst, Error> {
-    let (user_input_ast, _remaining) = parse_to_ast().parse(query).map_err(|_| Error)?;
+    let (_remaining, user_input_ast) = parse_to_ast(query).map_err(|_| Error)?;
     Ok(user_input_ast)
 }
+
+/// Parse a query, trying to recover from syntax errors, and giving hints toward fixing errors.
+pub fn parse_query_leniet(query: &str) -> (UserInputAst, Vec<(usize, String)>) {
```
fulmicoton:
```suggestion
pub fn parse_query_lenient(query: &str) -> (UserInputAst, Vec<(usize, String)>) {
```

query-grammar/src/query_grammar.rs
```diff
@@ -413,38 +735,121 @@ fn aggregate_binary_expressions(
...
+    map(delimited(space0, opt(ast), eof), |opt_ast| {
+        rewrite_ast(opt_ast.unwrap_or_else(UserInputAst::empty_query))
+    })(i)
+}
+
+pub fn parse_to_ast_lenient(i: &str) -> (UserInputAst, ErrorList) {
+    // TODO remove front space and detect final eof.
+    let (res, mut error) = ast_infallible(i).unwrap().1;
+
+    // convert end-based index to start-based index.
```
fulmicoton, line 956: for loop please.

```rust
for error in errors {
     error.0 = query_str.len() - error.0;
}
```

query-grammar/src/query_grammar.rs
```diff
@@ -413,38 +735,121 @@ fn aggregate_binary_expressions(
...
-        .map(rewrite_ast)
+pub fn parse_to_ast(i: &str) -> IResult<&str, UserInputAst> {
+    map(delimited(space0, opt(ast), eof), |opt_ast| {
+        rewrite_ast(opt_ast.unwrap_or_else(UserInputAst::empty_query))
+    })(i)
+}
+
+pub fn parse_to_ast_lenient(i: &str) -> (UserInputAst, ErrorList) {
+    // TODO remove front space and detect final eof.
+    let (res, mut error) = ast_infallible(i).unwrap().1;
```
fulmicoton:
```suggestion
    let (res, mut errors) = ast_infallible(i).unwrap().1;
```

query-grammar/src/query_grammar.rs
```diff
@@ -413,38 +735,121 @@ fn aggregate_binary_expressions(
...
-        .with(optional(ast()).skip(eof()))
-        .map(|opt_ast| opt_ast.unwrap_or_else(UserInputAst::empty_query))
-        .map(rewrite_ast)
+pub fn parse_to_ast(i: &str) -> IResult<&str, UserInputAst> {
+    map(delimited(space0, opt(ast), eof), |opt_ast| {
+        rewrite_ast(opt_ast.unwrap_or_else(UserInputAst::empty_query))
+    })(i)
+}
+
+pub fn parse_to_ast_lenient(i: &str) -> (UserInputAst, ErrorList) {
```
fulmicoton:
```suggestion
pub fn parse_to_ast_lenient(query_str: &str) -> (UserInputAst, ErrorList) {
```

query-grammar/src/infallible.rs
```diff
@@ -0,0 +1,286 @@
+//! nom combinators for infallible operations
+
+use nom::{IResult, InputLength};
+pub type ErrorList = Vec<(usize, String)>;
```
fulmicoton: No clue what the first `usize` argument is, nor the String.

fulmicoton: Can we introduce type to make it less confusing?

It seems like the first argument changes meaning halfway through. (offset to the end to offset from the start)

query-grammar/src/infallible.rs
```diff
@@ -0,0 +1,286 @@
...
+                res.push(o);
+                i = i1;
+            }
+        }
+
+        loop {
+            let len = i.input_len();
+            match sep.parse(i.clone()) {
+                Err(_) => unreachable!(),
+                Ok((i1, (_, mut err))) => {
```
fulmicoton: The `let .. else ...` pattern is a good candidate here.
```
let Ok((i1, (_, mut err))) = sep.parse(i.clone()) else { unreachable!() };
```

You can also add a function

```
fn unwrap_unfallible<T>(res: Result<T, Infallible>) -> T {}
```

if it ends up useful in several places.

query-grammar/src/infallible.rs
```diff
@@ -0,0 +1,286 @@
...
+    mut f: F,
+) -> impl FnMut(I) -> JResult<I, Vec<O>>
+where
+    I: Clone + InputLength,
+    F: nom::Parser<I, (O, ErrorList), std::convert::Infallible>,
+    G: nom::Parser<I, (O2, ErrorList), std::convert::Infallible>,
+{
+    move |mut i: I| {
+        let mut res = Vec::new();
+        let mut errors = Vec::new();
```
fulmicoton: can you add the type of these two please?

query-grammar/src/infallible.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        loop {
+            let len = i.input_len();
+            match sep.parse(i.clone()) {
+                Err(_) => unreachable!(),
+                Ok((i1, (_, mut err))) => {
+                    errors.append(&mut err);
+
+                    match f.parse(i1.clone()) {
+                        Err(_) => unreachable!(),
```
fulmicoton: same as above.

query-grammar/src/query_grammar.rs
```diff
@@ -413,38 +735,121 @@ fn aggregate_binary_expressions(
...
+            .map(|(_, occur, ast)| (occur, ast))
+            .collect();
+        return (UserInputAst::Clause(clauses), Vec::new());
+    }
+
+    if leafs.iter().all(|(_, occur, _)| occur.is_none()) {
+        let first = leafs.remove(0).2;
+        let rest = leafs
+            .into_iter()
+            .map(|(operand, _, ast)| (operand.unwrap_or(BinaryOperand::Or), ast))
```
fulmicoton: should the operand depend on the query parser default behavior?

fulmicoton (changes requested):
Nice approach. 

But you really should put more effort in making your code readable.
Here are a bunch of concrete things you can do to improve this.


# Have proper variable names

This is short and well contained, so `i` and `o` might be ok variable names.
I would have preferred `input` and `output` but I guess this can be argued.
`i1`, `i2` however, is really not helpful.

# Stop abusing iterator style. 

For instance, if you do it to mutate stuff, that's probably a bad idea.
If you start using rare iterator functions, that's probably a bad idea too.
If you end up struggling with dealing with error handling, using more than one trick like transpose and `collect::<Result<T,_>>`, it is probably a bad idea.
If your code seems to suggest a variable is a vec but it is in fact an Option or a Result, this is a bad idea.

# Consider using struct instead of tuple

`type Error = (usize, String);`
That kind of tuple makes code easy to write, but hard to understand.
You actually changed the semantics of the first argument, which is making things even worse.

If possible, I would have preferred two structs.
This however comes with a verbosity cost, so this may not be pertinent:  I leave you judge of whether the change should be done or not.

# Comment more

Something like
`type Error = (usize, String);`
is really hard for the reader.

# Show types in your code when it is not easy to guess.

We sometimes have to read rust without type hints.
For instance in github or in an editor for which rust analyzer has not finished crunching a
freshly checked out branch.
It is useful to add explicit types here in strategic places.

query-grammar/src/query_grammar.rs
trinity-1686a: this part needs to be improved to work with a mix of binary operand and occurs, however I'm yet to find how to mix both in a way that makes sens and transition coherently from full-occurs and full-binary-operand

fulmicoton (approved):

PR #2130, opened 2023-07-20
RFC: Add optional support for in-memory spatial index

adamreichold:
This adds optional support for spatial queries backed by per-segment in-memory spatial indexes using the `rstar` crate. The type of geometry stored in the indexes is generic with the only requirement that it can be built based on a segment reader. Similarly, queries are internally generic so that it should be straight forward to support all query types which the underlying `rstar` crate supports.

I am not sure whether this is reasonable to include upstream as it is somewhat limited in scope by being based on in-memory spatial indexes. I do think it is useful in the generic form as the code does take care of all the necessary wiring around warmers and queries, so projects using Tantivy as a library can concentrate on their geometry. Finally, this could obviously also live in a separate crate, e.g. `tantivy-spatial`, but it is not enough code to warrant the extra ceremony IMHO.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ itertools = "0.11.0"
 measure_time = "0.8.2"
 async-trait = "0.1.53"
 arc-swap = "1.5.0"
+rstar = { version = "0.11", optional = true }
 
 columnar = { version= "0.1", path="./columnar", package ="tantivy-columnar" }
 sstable = { version= "0.1", path="./sstable", package ="tantivy-sstable", optional = true }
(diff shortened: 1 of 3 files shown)


fulmicoton:
@adamreichold I don't think we want to merge this.

Several users have been asking for a way to extend tantivy with datastructures. (For ANN) usually.
I wonder if we could provide a way for users to plug their own `SegmentComponent` as if it was a plugin.
It would require writing, merging, and some way to access those.

@bloodbare

adamreichold:
> I don't think we want to merge this.

Understood. As written if I really desperately want to share this, I can create a separate crate.

>  I wonder if we could provide a way for users to plug their own SegmentComponent as if it was a plugin.

You something that is persisted with the rest of the index instead of being created on-demand by a warmer? This sounds much more intrusive (w.r.t. the internals Tantivy needs to expose). Couldn't a warmer just handle its own serialization if required? Or maybe making the underlying directory available to it would already be enough?

PR #2132, opened 2023-07-21
delayed column opening during merge

PSeitz:
This is the first part of addressing https://github.com/quickwit-oss/quickwit/issues/3633
Instead of loading all Column into memory for the merge, only the current `column_name, column_category` group is loaded. 

Second part would be to replace mmap calls with file reads, for better control on releasing memory

### DIFF
--- a/columnar/src/columnar/merge/mod.rs
+++ b/columnar/src/columnar/merge/mod.rs
@@ -2,14 +2,12 @@ mod merge_dict_column;
 mod merge_mapping;
 mod term_merger;
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeMap, HashSet};
 use std::io;
 use std::net::Ipv6Addr;
-use std::rc::Rc;
 use std::sync::Arc;
 
-use common::GroupByIteratorExtended;
-use itertools::{EitherOrBoth, Itertools};
+use itertools::Itertools;
 pub use merge_mapping::{MergeRowOrder, ShuffleMergeOrder, StackMergeOrder};
 
 use super::writer::ColumnarSerializer;
@@ -31,14 +29,16 @@ use crate::{
 /// In practise, today, only Numerical colummns are coerced into one type today.
 ///
 /// See also [README.md].
-#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+///
+/// The ordering has to match the ordering of the variants in [ColumnType].
+#[derive(Copy, Clone, Eq, PartialOrd, Ord, PartialEq, Hash, Debug)]
 pub(crate) enum ColumnTypeCategory {
-    Bool,
-    Str,
     Numerical,
-    DateTime,
     Bytes,
+    Str,
+    Bool,
     IpAddr,
+    DateTime,
 }
 
 impl From<ColumnType> for ColumnTypeCategory {
@@ -87,11 +87,18 @@ pub fn merge_columnar(
         .map(|reader| reader.num_rows())
         .collect::<Vec<u32>>();
 
-    let columns_to_merge_iter =
-        group_columns_for_merge_iter(columnar_readers, required_columns, &merge_row_order)?;
-    for res in columns_to_merge_iter {
-        let (column_name, column_type, grouped_columns) = res?;
-        let columns = grouped_columns.columns;
+    let columns_to_merge =
+        group_columns_for_merge(columnar_readers, required_columns, &merge_row_order)?;
+    for res in columns_to_merge {
+        let ((column_name, _column_type_category), grouped_columns) = res;
+        let grouped_columns = grouped_columns.open(&merge_row_order)?;
+        if grouped_columns.is_empty() {
+            continue;
+        }
+
+        let column_type = grouped_columns.column_type_after_merge();
+        let mut columns = grouped_columns.columns;
+        coerce_columns(column_type, &mut columns)?;
 
         let mut column_serializer =
             serializer.start_serialize_column(column_name.as_bytes(), column_type);
@@ -221,35 +228,9 @@ struct GroupedColumns {
 }
 
 impl GroupedColumns {
-    fn new(num_columnars: usize) -> Self {
-        GroupedColumns {
-            required_column_type: None,
-            columns: vec![None; num_columnars],
-        }
-    }
-
-    /// Set the dynamic column for a given columnar.
-    fn set_column(&mut self, columnar_id: usize, column: DynamicColumn) {
-        self.columns[columnar_id] = Some(column);
-    }
-
-    /// Force the existence of a column, as well as its type.
-    fn require_type(&mut self, required_type: ColumnType) -> io::Result<()> {
-        if let Some(existing_required_type) = self.required_column_type {
-            if existing_required_type == required_type {
-                // This was just a duplicate in the `required_columns`.
-                // Nothing to do.
-                return Ok(());
-            } else {
-                return Err(io::Error::new(
-                    io::ErrorKind::InvalidInput,
-                    "Required column conflicts with another required column of the same type \
-                     category.",
-                ));
-            }
-        }
-        self.required_column_type = Some(required_type);
-        Ok(())
+    /// Check is column group can be skipped during serialization.
+    fn is_empty(&self) -> bool {
+        self.required_column_type.is_none() && self.columns.iter().all(Option::is_none)
     }
 
     /// Returns the column type after merge.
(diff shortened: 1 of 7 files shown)


common/src/file_slice.rs
```diff
@@ -32,6 +34,42 @@ pub trait FileHandle: 'static + Send + Sync + HasLen + fmt::Debug {
...
+        // Read the data into the buffer
+        file.read_exact(&mut buffer)?;
+
+        Ok(OwnedBytes::new(buffer))
+    }
+    // todo implement async
+}
+impl HasLen for WrapFile {
+    fn len(&self) -> usize {
+        self.0.metadata().unwrap().len() as usize
```
adamreichold: Unwrapping on results of doing I/O seems dangerous? Maybe better to return zero as the length instead?

common/src/file_slice.rs
```diff
@@ -32,6 +34,42 @@ pub trait FileHandle: 'static + Send + Sync + HasLen + fmt::Debug {
     }
 }
 
+#[derive(Debug)]
+pub struct WrapFile(pub File);
+
+#[async_trait]
+impl FileHandle for WrapFile {
+    fn read_bytes(&self, range: Range<usize>) -> io::Result<OwnedBytes> {
+        let mut file = self.0.try_clone()?; // Clone the file to read from it separately
```
adamreichold: This is probably a premature optimization for elsewhen, but at least on Unix, positioned I/O using e.g. [`read_exact_at`](https://doc.rust-lang.org/stable/std/os/unix/fs/trait.FileExt.html#method.read_exact_at) might avoid this duplication of the file descriptor.

columnar/src/columnar/reader/mod.rs
```diff
@@ -102,30 +102,41 @@ impl ColumnarReader {
...
-            let column_code: u8 = key_bytes.last().cloned().unwrap();
-            let column_type: ColumnType = ColumnType::try_from_code(column_code)
-                .map_err(|_| io_invalid_data(format!("Unknown column code `{column_code}`")))?;
-            let range = stream.value().clone();
-            let column_name =
+        Ok(std::iter::from_fn(move || {
+            if stream.advance() {
+                let key_bytes: &[u8] = stream.key();
+                let column_code: u8 = key_bytes.last().cloned().unwrap();
+                // TODO Error Handling. The API gets quite ugly when returning the error here, so
```
adamreichold, line 114: Does "quite ugly" refer to wrapping the iterator item in a `io::Result`? Is there really any alternative to doing that?

columnar/src/columnar/merge/mod.rs
```diff
@@ -348,56 +356,115 @@ fn is_empty_after_merge(
...
-            })
-            .require_type(column_type)?;
-    }
+/// Iterates over the columns of the columnar readers, grouped by column name.
+/// Key functionality is that `open` of the Columns is done lazy per group.
+fn group_columns_for_merge_iter<'a>(
+    columnar_readers: &'a [&'a ColumnarReader],
+    required_columns: &'a [(String, ColumnType)],
+    merge_row_order: &'a MergeRowOrder,
+) -> io::Result<impl Iterator<Item = io::Result<(Rc<String>, ColumnType, GroupedColumns)>> + 'a> {
```
adamreichold: Usually, if `Rc<String>` can be used, the simpler `Rc<str>` can be used as well.

columnar/src/columnar/merge/mod.rs
```diff
@@ -348,56 +356,115 @@ fn is_empty_after_merge(
...
+    let column_iters: Vec<_> = columnar_readers
+        .iter()
+        .enumerate()
+        .map(|(reader_ord, reader)| {
+            Ok(reader
+                .iter_columns()?
+                .map(move |el| (Rc::new(el.0), reader_ord, el.1)))
```
adamreichold: From a readability point of view, spelling out the tuple bindings here and below would probably be helpful.

columnar/src/columnar/merge/mod.rs
```diff
@@ -348,56 +356,115 @@ fn is_empty_after_merge(
...
-        }
-    }
+    // Weave in the required columns into the sorted by column name iterator.
+    let groups_with_required = required_columns_list
+        .into_iter()
+        .merge_join_by(group_iter, |a, b| a.cmp(&b.0));
 
-    let mut merge_columns: BTreeMap<(String, ColumnType), Vec<Option<DynamicColumn>>> =
-        Default::default();
+    Ok(groups_with_required.flat_map(move |either| {
```
adamreichold: I wonder if this large and nested iterator construction can be untangled somewhat, at least visually.

IIUC, there is first the construction of `force_type`, `key` and `group`. And there is the the `group_iter.map(..).filter(..)` step based on that.

Maybe these could be separate named functions so that the composition is easier to see? Or at least named let bindings? Admittedly, pulling things apart can actually decrease readability, but it has a bit of a "wall of text" vibe as it is.

common/src/file_slice.rs
PSeitz: zero may have similar bad side effects. Probably best to preload it or change the trait

common/src/file_slice.rs
PSeitz: For now, I'd just stick with one implementation. In the merge cases this is not really relevant, where this is planned to be used

columnar/src/columnar/merge/mod.rs
PSeitz: switched to `Rc<str>`

columnar/src/columnar/merge/mod.rs
PSeitz: I refactored it to be more readable

columnar/src/columnar/reader/mod.rs
PSeitz: Yes, the wrapping and the usage, since it won't be simply collected in a `Result<>`

common/src/file_slice.rs
PSeitz: moved to constructor

columnar/src/columnar/merge/mod.rs
```diff
@@ -348,56 +358,157 @@ fn is_empty_after_merge(
...
-            if is_empty_after_merge(merge_row_order, &column, columnar_id) {
-                continue;
+/// Iterates over the columns of the columnar readers, grouped by column name.
+/// Key functionality is that `open` of the Columns is done lazy per group.
+fn group_columns_for_merge_iter<'a>(
+    columnar_readers: &'a [&'a ColumnarReader],
+    required_columns: &'a [(String, ColumnType)],
+    merge_row_order: &'a MergeRowOrder,
+) -> io::Result<impl Iterator<Item = io::Result<(Rc<str>, ColumnType, GroupedColumns)>> + 'a> {
+    let column_iters: Vec<_> = columnar_readers
```
fulmicoton:
```suggestion
    let column_iters: Vec<MergeColumn> = columnar_readers
```

columnar/src/columnar/merge/mod.rs
```diff
@@ -348,56 +358,157 @@ fn is_empty_after_merge(
...
+        .iter()
+        .map(|(col_name, _)| col_name.to_string())
+        .collect();
+    required_columns_list.sort();
+
+    // Kmerge and group on column_name.
+    let group_iter = GroupByIteratorExtended::group_by(
+        column_iters
+            .into_iter()
+            .kmerge_by(|a, b| a.column_name < b.column_name),
```
fulmicoton: nitpick... I prefer `lhs, rhs` or `left`,`right` for that comparisons.

```suggestion
            .kmerge_by(|a, b| a.column_name < b.column_name),
```

columnar/src/columnar/merge/mod.rs
```diff
@@ -348,56 +358,157 @@ fn is_empty_after_merge(
...
+        .map(|(col_name, typ)| (col_name.to_string(), *typ))
+        .collect::<HashMap<String, _>>();
+    let mut required_columns_list: Vec<String> = required_columns
+        .iter()
+        .map(|(col_name, _)| col_name.to_string())
+        .collect();
+    required_columns_list.sort();
+
+    // Kmerge and group on column_name.
+    let group_iter = GroupByIteratorExtended::group_by(
```
fulmicoton: Can you put the kmerge outside? There is not need to put the groupby and the kmerge in the same expression.

columnar/src/columnar/merge/mod.rs
PSeitz: it's `Vec<impl Iter<Item=MergeColumn>>`, but we can't use impl iter here

columnar/src/columnar/merge/mod.rs
PSeitz: done

columnar/src/columnar/merge/mod.rs
fulmicoton: ah I see.

columnar/src/columnar/merge/mod.rs
```diff
@@ -348,56 +358,157 @@ fn is_empty_after_merge(
...
-        // creating empty columns.
-
-        for (column_name, handle) in column_name_and_handle {
-            let column_category: ColumnTypeCategory = handle.column_type().into();
-            let column = handle.open()?;
-            if is_empty_after_merge(merge_row_order, &column, columnar_id) {
-                continue;
+/// Iterates over the columns of the columnar readers, grouped by column name.
+/// Key functionality is that `open` of the Columns is done lazy per group.
+fn group_columns_for_merge_iter<'a>(
```
fulmicoton: The code is much more complicated. Could we have kept the previous code, and worked of `DynamicColumnHandle` instead of `DynamicColumn` and gotten the same benefit?

common/src/file_slice.rs
```diff
@@ -32,6 +33,59 @@ pub trait FileHandle: 'static + Send + Sync + HasLen + fmt::Debug {
     }
 }
 
+#[derive(Debug)]
+pub struct WrapFile {
```
fulmicoton, line 38: Rustdoc please.

columnar/src/columnar/merge/mod.rs
PSeitz: The columns need to be loaded to access things like min max values. The alternative would be to change the code so we can access only the metadata with something like `open_metadata`

fulmicoton: We can lighten the work of `group_columns_for_merge` and change it so that it does not:
- filter out columns that end up being empty.
- identify the resulting ColumnType.

It would just stops at returning a 

```
fn group_columns_for_merge(
    columnar_readers: &[&ColumnarReader],
    required_columns: &[(String, ColumnType)],
    merge_row_order: &MergeRowOrder,
) -> io::Result<BTreeMap<(String, ColumnTypeCategory), Vec<Option<DynamicColumnHandle>>>> {
```

On the callsite, we would then iterate through these handles, and open the `DynamicColumn`,
and do the conversion of the `(ColumnTypeCategory, Vec<Option<DynamicColumnHandle>>)` into a 
`(ColumnType, Vec<Option<DynamicHandle>>)` where all of the columns that are Some contain at least one element.

The rest is then similar to what exists today.
The code should not be more complicated than it is today.

One tricky part to consider is how to deal with required columns. 
If all columns for this `(String, ColumnTypeCategory)` are empty (following deletes), whether or not we want a column depends on whether the column is marked as required or not.

PSeitz: In the high cardinality num column case, the BTreeMap variant may cost more than 100MB heap allocation, which is why I opted to the Iterator version.
From the code complexity the main difference is the mixing of the required columns into the (sorted by column name) Column stream instead of simply inserting them into the `BTreeMap`

fulmicoton: This is too complex. Can you go with the BTreeMap solution?

PSeitz: I switched to a BtreeMap

fulmicoton (approved):

common/src/file_slice.rs
```diff
@@ -32,6 +33,61 @@ pub trait FileHandle: 'static + Send + Sync + HasLen + fmt::Debug {
...
+        }
+
+        #[cfg(not(unix))]
+        {
+            let mut file = self.file.try_clone()?; // Clone the file to read from it separately
+                                                   // Seek to the start position in the file
+            file.seek(io::SeekFrom::Start(start as u64))?;
+            // Read the data into the buffer
+            file.read_exact(&mut buffer)?;
+        }
```
SteveLauC, line 79: Any reason why we are not using [`std::os::windows::fs::FileExt`](https://doc.rust-lang.org/stable/std/os/windows/fs/trait.FileExt.html) here?

PSeitz: I wasn't aware of it. 
Windows is not a priority though, except that it should work

SteveLauC: > Windows is not a priority though, except that it should work

Fair enough

PR #2135, opened 2023-07-25
alternative mixed field aggregation collection

PSeitz:
instead of having multiple accessor in one AggregationWithAccessor split it into
multiple independent AggregationWithAccessor

Performance seems unchanged
```
Before
test aggregation::agg_bench::bench::bench_aggregation_terms_many_json_mixed_type_with_sub_agg                            ... bench: 173,939,311 ns/iter (+/- 35,629,524)
test aggregation::agg_bench::bench::bench_aggregation_terms_many_json_mixed_type_with_sub_agg_multi                      ... bench: 227,312,130 ns/iter (+/- 34,251,990)
test aggregation::agg_bench::bench::bench_aggregation_terms_many_json_mixed_type_with_sub_agg_opt                        ... bench: 216,645,922 ns/iter (+/- 72,184,100)
test aggregation::agg_bench::bench::bench_aggregation_terms_many_json_mixed_type_with_sub_agg_sparse                     ... bench:  43,224,161 ns/iter (+/- 11,300,519)

After
test aggregation::agg_bench::bench::bench_aggregation_terms_many_json_mixed_type_with_sub_agg                            ... bench: 171,519,343 ns/iter (+/- 76,513,404)
test aggregation::agg_bench::bench::bench_aggregation_terms_many_json_mixed_type_with_sub_agg_multi                      ... bench: 228,347,835 ns/iter (+/- 12,738,742)
test aggregation::agg_bench::bench::bench_aggregation_terms_many_json_mixed_type_with_sub_agg_opt                        ... bench: 218,119,685 ns/iter (+/- 72,979,994)
test aggregation::agg_bench::bench::bench_aggregation_terms_many_json_mixed_type_with_sub_agg_sparse                     ... bench:  41,439,166 ns/iter (+/- 10,048,723)


```


src/aggregation/agg_req_with_accessor.rs
```diff
@@ -95,33 +99,37 @@ impl AggregationWithAccessor {
...
+                        reader,
+                        &limits,
+                    )?,
+                    agg: agg.clone(),
+                    str_dict_column: str_dict_column.clone(),
+                    limits: limits.new_guard(),
+                    column_block_accessor: Default::default(),
+                })
+            })
+            .collect::<crate::Result<_>>()?;
```
adamreichold, line 131: I think the return type fixes the types here, so that `let aggs: ...`, `Ok(aggs)` and the `?` are unnecessary, i.e. it could just end in `}).collect()`.

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -141,15 +149,15 @@ pub(crate) fn get_aggs_with_segment_accessor_and_validate(
 ) -> crate::Result<AggregationsWithAccessor> {
     let mut aggss = Vec::new();
```
adamreichold, line 150: Is this a good candidate for a call to `reserve`?

PSeitz: no, the cardinality is close to the user request

src/aggregation/agg_req_with_accessor.rs
PSeitz: I don't understand, we need to collect to Result to forward the error

adamreichold: Certainly, but the extra binding and type hints seem unnecessary because the type that is collected already matches the return type, does it not?

So in think the function could just end in

```rust
        acc_field_types
            .into_iter()
            .map(|(accessor, field_type)| {
                Ok(AggregationWithAccessor {
                    accessor,
                    field_type,
                    sub_aggregation: get_aggs_with_segment_accessor_and_validate(
                        sub_aggregation,
                        reader,
                        &limits,
                    )?,
                    agg: agg.clone(),
                    str_dict_column: str_dict_column.clone(),
                    limits: limits.new_guard(),
                    column_block_accessor: Default::default(),
                })
            })
            .collect()
```

src/aggregation/agg_req_with_accessor.rs
adamreichold: But isn't it already known here because the size of `aggss` will usually match that of `aggs` or at most twice that, so that it would usually help to `aggss.reserve(aggs.len())` even if it isn't exact?

(I also wonder whether `VecWithNames::from_entries` wants to use [`Iterator::unzip`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.unzip) but admittedly unrelated to this change.)

adamreichold: Looking at this again, I wonder why the intermediate `aggss` is built at all instead of directly collecting into a `VecWithNames`?

src/aggregation/agg_req_with_accessor.rs
PSeitz: we return a `Result<Vec>` not `Vec<Result>`

adamreichold: Yes, which is what you are collecting by giving the type hint `::<crate::Result<_>>` but that type hint just matches the return type. I am not suggesting to change any types, just to drop the unnecessary temporary binding and error forwarding because the collection already produces exactly the required type.

(Turning `Iterator<Item=Result<S>>` into `Result<T> where T: FromIterator<Item=S>` is the usual [`FromIterator` impl](https://doc.rust-lang.org/stable/std/iter/trait.FromIterator.html#impl-FromIterator%3CResult%3CA,+E%3E%3E-for-Result%3CV,+E%3E) which you invoke either way.)

src/aggregation/agg_req_with_accessor.rs
PSeitz: This is not performance relevant, it's only a couple of objects

src/aggregation/agg_req_with_accessor.rs
PSeitz: I think the binding with type assignment increases readability

src/aggregation/agg_req_with_accessor.rs
adamreichold: I agree that this is certainly not measurable as an individual change so I agree with your point that this is certainly not a necessary change. I would like to add two observations though: The cost of small memory allocations is usually not dominated by their size, but by the constant cost of the synchronization necessary to call into the allocator implementation (which is why thread-local caches help so much in the average case). There is also a notion called "distributed fat" which suggests that even though these changes are not measurable individually, they can add up to measurable costs when considered in aggregate.

Of course, this is not really a good reason to fight every single allocation. Only, to avoid them if there is no loss in readability and functionality implied by avoiding them. Personally, I see calling `reserve` (or better yet using `Extend` or `FromIterator` which does this automatically) in this category but this admittedly more of a style question.

src/aggregation/agg_req_with_accessor.rs
adamreichold: You can still spell out the full return type if desired in the turbo-fish if you want to, e.g.

```rust
.collect::<crate::Result<Vec<AggregationWithAccessor>>>()
```

but the extra binding, error forwarding and ok wrapping are unnecessary.

(There is even a Clippy lint for unnecessary let bindings but IIRC it does not firing when ok wrapping is involved because this is sometimes necessary to change the error type via the `From` invocation hidding in `?`.)

PSeitz: I don't agree, the binding and type serves the purpose of increased readability by giving it a name and a type, which is important if the transformation gets nested

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -52,20 +49,31 @@ impl AggregationWithAccessor {
         sub_aggregation: &Aggregations,
         reader: &SegmentReader,
         limits: AggregationLimits,
-    ) -> crate::Result<AggregationWithAccessor> {
+    ) -> crate::Result<Vec<AggregationWithAccessor>> {
         let mut str_dict_column = None;
-        let mut accessor2 = None;
         use AggregationVariants::*;
-        let (accessor, field_type) = match &agg.agg {
+        let acc_field_types: Vec<(Column, ColumnType)> = match &agg.agg {
```
fulmicoton, line 55: Thanks for putting the type!

src/aggregation/agg_req_with_accessor.rs
fulmicoton: I agree with @PSeitz . It is more readable when we put the type on the variable side.

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -95,33 +99,37 @@ impl AggregationWithAccessor {
...
-                &sub_aggregation,
-                reader,
-                &limits,
-            )?,
-            agg: agg.clone(),
-            str_dict_column,
-            limits: limits.new_guard(),
-            column_block_accessor: Default::default(),
-        })
+        let aggs: Vec<_> = acc_field_types
```
fulmicoton:
```suggestion
        let aggs: Vec<AggregationWithAccessor> = acc_field_types
```

fulmicoton (approved):

PR #2142, opened 2023-08-13
Fix explanation of boost queries seeking beyond query result.

### LINKED ISSUE #2137: Query.explain throws a panic
I tried to use query.explain and it threw this error:
```
thread 'main' panicked at 'assertion failed: self.doc() <= target', tantivy-0.20.2\src\postings\segment_postings.rs:173:9
```

My query was:
```
title:sea^20 body:whale^70
```

0.20.2

**To Reproduce**

To reproduce, just use the sample example in the tantivy docs:
https://tantivy-search.github.io/examples/basic_search.html

But use this query:
```
let query = query_parser.parse_query("title:sea^20 body:whale^70")?;
```

Then use this to get the explanation:
```
let explanation = query.explain(&searcher, doc_address)?;
println!("exp {:?}", explanation.to_pretty_json());
```

The searcher returns correct response, but the explanation throws an error.

The explanation works fine if I remove the weights from query. Or if I remove the field prefixes (e.g. `title:`). But using both doesn't work

adamreichold:
Closes #2137

### DIFF
--- a/bitpacker/src/blocked_bitpacker.rs
+++ b/bitpacker/src/blocked_bitpacker.rs
@@ -64,10 +64,8 @@ fn mem_usage<T>(items: &Vec<T>) -> usize {
 
 impl BlockedBitpacker {
     pub fn new() -> Self {
-        let mut compressed_blocks = vec![];
-        compressed_blocks.resize(8, 0);
         Self {
-            compressed_blocks,
+            compressed_blocks: vec![0; 8],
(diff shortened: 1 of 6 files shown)


fulmicoton (approved):

fulmicoton:
Thank you!

aliflux-omo:
Thanks for the fix. Any idea when will this be released in the next Tantivy version?

PR #2144, opened 2023-08-14
DOC: show the escape backslash

cjrh:
Before this PR, the char doesn't render:

image

I tested that there are no further shenanigans with backquotes:

image


PSeitz (approved):
Thanks!

PR #2149, opened 2023-08-18
add missing aggregation part 2

### LINKED ISSUE #1789: Add `missing` parameter on relevant aggregations

PSeitz:
Add missing support for:
- Mixed types columns
- Key of type string on numerical fields

The special aggregation is slower than the integrated one in TermsAggregation and therefore not
chosen by default, although it can cover all use cases.

closes #1789

### DIFF
--- a/columnar/src/column/dictionary_encoded.rs
+++ b/columnar/src/column/dictionary_encoded.rs
@@ -30,10 +30,10 @@ impl fmt::Debug for BytesColumn {
 }
 
 impl BytesColumn {
-    pub fn empty() -> BytesColumn {
+    pub fn empty(num_docs: u32) -> BytesColumn {
         BytesColumn {
             dictionary: Arc::new(Dictionary::empty()),
-            term_ord_column: Column::build_empty_column(0),
+            term_ord_column: Column::build_empty_column(num_docs),
         }
     }
 
--- a/columnar/src/column/dictionary_encoded.rs
+++ b/columnar/src/column/dictionary_encoded.rs
@@ -30,6 +30,13 @@ impl fmt::Debug for BytesColumn {
 }
 
 impl BytesColumn {
+    pub fn empty() -> BytesColumn {
+        BytesColumn {
+            dictionary: Arc::new(Dictionary::empty()),
+            term_ord_column: Column::build_empty_column(0),
+        }
+    }
+
     /// Fills the given `output` buffer with the term associated to the ordinal `ord`.
     ///
     /// Returns `false` if the term does not exist (e.g. `term_ord` is greater or equal to the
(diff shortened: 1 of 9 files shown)


src/aggregation/agg_req_with_accessor.rs
```diff
@@ -54,38 +59,68 @@ impl AggregationWithAccessor {
         reader: &SegmentReader,
         limits: AggregationLimits,
     ) -> crate::Result<Vec<AggregationWithAccessor>> {
-        let mut missing_value_term_agg = None;
-        let mut str_dict_column = None;
+        let get_agg_with_accessor = |aggs: &mut Vec<AggregationWithAccessor>,
```
fulmicoton: does this have to be inlined? can it be a proper funciton?

fulmicoton: nitpick:
I suggest we try to stick to a rule for argument lists: the mut stuff at the end of the list of arguments.
This is arbitrary. (that's the rule enforced at Google)

fulmicoton: It sounds like it could have been a function with a clearer prototype:

`Fn(.., .., .., .., ..) -> crate::Result<Option<AggregationWithAccessor>> {}`

columnar/src/column/dictionary_encoded.rs
```diff
@@ -30,6 +30,13 @@ impl fmt::Debug for BytesColumn {
 }
 
 impl BytesColumn {
+    pub fn empty() -> BytesColumn {
+        BytesColumn {
+            dictionary: Arc::new(Dictionary::empty()),
+            term_ord_column: Column::build_empty_column(0),
```
fulmicoton: This `num_docs` given by this column is incorrect. I can see it causing bugs.

src/aggregation/agg_req_with_accessor.rs
PSeitz: The closure has access to the parameters and is mainly there to remove code duplication, moving it to its own function has too many parameters imo

```rust
fn add_agg_with_accessor(
    aggs: &mut Vec<AggregationWithAccessor>,
    accessor: Column<u64>,
    column_type: ColumnType,
    str_dict_column: Option<StrColumn>,
    missing_value_term_agg: Option<Key>,
    agg: &Aggregation,
    sub_aggregation: &Aggregations,
    reader: &SegmentReader,
    limits: AggregationLimits,
) -> crate::Result<()> {
```

I refactored it a little bit and added types to the closure instead.


> nitpick:
I suggest we try to stick to a rule for argument lists: the mut stuff at the end of the list of arguments.
This is arbitrary. (that's the rule enforced at Google)

I'm not really in favor of that, since this would cause churn when refactoring something to mut for very little gain (at least for how I read code).
I changed it in this case though.

columnar/src/column/dictionary_encoded.rs
PSeitz: It's fine in the current usage, but yes this could cause bugs. Fixed

fulmicoton (approved):

PR #2150, opened 2023-08-18
Improve aggregation error message

### LINKED ISSUE #2143: Fix aggregation deserialization error message.
Right now giving a mispelled aggregation variant gives back an non-helpful error message.

```
~/git/quickwit/quickwit/target/debug/quickwit tool local-search --index gharchive --query "*" --max-hits 0 --aggregation "{\"agg\": {\"term\": {\"field\": \"c\"}}}"
❯ Searching directly on the index storage (without calling REST API)...
✘ Command failed: Invalid aggregation request: no variant of enum AggregationVariants found in flattened data at line 1 column 29
```

PSeitz:
Improve aggregation error message by wrapping the deserialization with a
custom struct. This deserialization variant is slower, since we need to
keep the deserialized data around twice with this approach.
For now the valid variants list is manually updated. This could be
replaced with a proc macro.
closes #2143

### DIFF
--- a/src/aggregation/agg_req.rs
+++ b/src/aggregation/agg_req.rs
@@ -44,82 +44,45 @@ use super::metric::{
 /// The key is the user defined name of the aggregation.
 pub type Aggregations = HashMap<String, Aggregation>;
 
-#[derive(Clone, Debug, PartialEq, Serialize)]
 /// Aggregation request.
 ///
 /// An aggregation is either a bucket or a metric.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(try_from = "AggregationForDeserialization")]
 pub struct Aggregation {
     /// The aggregation variant, which can be either a bucket or a metric.
     #[serde(flatten)]
     pub agg: AggregationVariants,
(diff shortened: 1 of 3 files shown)


src/aggregation/agg_req.rs
```diff
@@ -60,6 +60,70 @@ pub struct Aggregation {
...
 }
 
+impl<'de> serde::Deserialize<'de> for Aggregation {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where D: serde::Deserializer<'de> {
+        #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+        /// Aggregation request.
+        ///
+        /// An aggregation is either a bucket or a metric.
+        pub struct AggregationWrapper {
```
fulmicoton: why pub and why inline the struct?

fulmicoton (approved):

fulmicoton:
@PSeitz I pushed a commit with a less brittle approach. I leave it to you to modify if necessary and merge.

PSeitz:
> @PSeitz I pushed a commit with a less brittle approach. I leave it to you to modify if necessary and merge.

Nice, I didn't know about `try_from` in serde

PR #2151, opened 2023-08-25
add missing parameter for stats,min,max,count,sum,avg

### LINKED ISSUE #1913: Handle missing columns or multiple columns in aggregation

PSeitz:
add missing parameter for stats,min,max,count,sum,avg  #1789
add tests for #1913

### DIFF
--- a/src/aggregation/metric/average.rs
+++ b/src/aggregation/metric/average.rs
@@ -20,12 +20,21 @@ use super::{IntermediateStats, SegmentStatsCollector};
 pub struct AverageAggregation {
     /// The field name to compute the average on.
     pub field: String,
+    /// The missing parameter defines how documents that are missing a value should be treated.
+    /// By default they will be ignored but it is also possible to treat them as if they had a
+    /// value. Examples in JSON format:
+    /// { "field": "my_numbers", "missing": "10.0" }
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub missing: Option<f64>,
 }
 
 impl AverageAggregation {
     /// Creates a new [`AverageAggregation`] instance from a field name.
     pub fn from_field_name(field_name: String) -> Self {
-        Self { field: field_name }
+        Self {
+            field: field_name,
+            missing: None,
+        }
     }
     /// Returns the field name the aggregation is computed on.
     pub fn field_name(&self) -> &str {
--- a/src/aggregation/metric/average.rs
+++ b/src/aggregation/metric/average.rs
@@ -24,7 +24,7 @@ pub struct AverageAggregation {
     /// By default they will be ignored but it is also possible to treat them as if they had a
     /// value. Examples in JSON format:
     /// { "field": "my_numbers", "missing": "10.0" }
-    #[serde(skip_serializing_if = "Option::is_none", default)]
+    #[serde(default)]
     pub missing: Option<f64>,
 }
 
--- a/src/aggregation/metric/count.rs
+++ b/src/aggregation/metric/count.rs
@@ -20,12 +20,21 @@ use super::{IntermediateStats, SegmentStatsCollector};
 pub struct CountAggregation {
     /// The field name to compute the count on.
     pub field: String,
+    /// The missing parameter defines how documents that are missing a value should be treated.
+    /// By default they will be ignored but it is also possible to treat them as if they had a
+    /// value. Examples in JSON format:
+    /// { "field": "my_numbers", "missing": "10.0" }
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub missing: Option<f64>,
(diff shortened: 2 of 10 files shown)


src/aggregation/metric/average.rs
```diff
@@ -20,12 +20,21 @@ use super::{IntermediateStats, SegmentStatsCollector};
 pub struct AverageAggregation {
     /// The field name to compute the average on.
     pub field: String,
+    /// The missing parameter defines how documents that are missing a value should be treated.
+    /// By default they will be ignored but it is also possible to treat them as if they had a
+    /// value. Examples in JSON format:
+    /// { "field": "my_numbers", "missing": "10.0" }
+    #[serde(skip_serializing_if = "Option::is_none", default)]
```
fulmicoton:
```suggestion
    #[serde(default)]
```

src/aggregation/metric/count.rs
```diff
@@ -20,12 +20,21 @@ use super::{IntermediateStats, SegmentStatsCollector};
 pub struct CountAggregation {
     /// The field name to compute the count on.
     pub field: String,
+    /// The missing parameter defines how documents that are missing a value should be treated.
+    /// By default they will be ignored but it is also possible to treat them as if they had a
+    /// value. Examples in JSON format:
+    /// { "field": "my_numbers", "missing": "10.0" }
+    #[serde(skip_serializing_if = "Option::is_none", default)]
```
fulmicoton:
```suggestion
    #[serde(default)]
```

src/aggregation/metric/max.rs
```diff
@@ -20,12 +20,21 @@ use super::{IntermediateStats, SegmentStatsCollector};
 pub struct MaxAggregation {
     /// The field name to compute the maximum on.
     pub field: String,
+    /// The missing parameter defines how documents that are missing a value should be treated.
+    /// By default they will be ignored but it is also possible to treat them as if they had a
+    /// value. Examples in JSON format:
+    /// { "field": "my_numbers", "missing": "10.0" }
+    #[serde(skip_serializing_if = "Option::is_none", default)]
```
fulmicoton:
```suggestion
    #[serde(default)]
```

src/aggregation/metric/min.rs
```diff
@@ -20,12 +20,21 @@ use super::{IntermediateStats, SegmentStatsCollector};
 pub struct MinAggregation {
     /// The field name to compute the minimum on.
     pub field: String,
+    /// The missing parameter defines how documents that are missing a value should be treated.
+    /// By default they will be ignored but it is also possible to treat them as if they had a
+    /// value. Examples in JSON format:
+    /// { "field": "my_numbers", "missing": "10.0" }
+    #[serde(skip_serializing_if = "Option::is_none", default)]
```
fulmicoton:
```suggestion
    #[serde(default)]
```

src/aggregation/metric/stats.rs
```diff
@@ -29,12 +29,21 @@ use crate::{DocId, TantivyError};
 pub struct StatsAggregation {
     /// The field name to compute the stats on.
     pub field: String,
+    /// The missing parameter defines how documents that are missing a value should be treated.
+    /// By default they will be ignored but it is also possible to treat them as if they had a
+    /// value. Examples in JSON format:
+    /// { "field": "my_numbers", "missing": "10.0" }
+    #[serde(skip_serializing_if = "Option::is_none", default)]
```
fulmicoton:
```suggestion
    #[serde(default)]
```

src/aggregation/metric/sum.rs
```diff
@@ -20,12 +20,21 @@ use super::{IntermediateStats, SegmentStatsCollector};
 pub struct SumAggregation {
     /// The field name to compute the minimum on.
     pub field: String,
+    /// The missing parameter defines how documents that are missing a value should be treated.
+    /// By default they will be ignored but it is also possible to treat them as if they had a
+    /// value. Examples in JSON format:
+    /// { "field": "my_numbers", "missing": "10.0" }
+    #[serde(skip_serializing_if = "Option::is_none", default)]
```
fulmicoton:
```suggestion
    #[serde(default)]
```

fulmicoton (approved):

PR #2152, opened 2023-08-26
POC: Setting thread affinity on indexer threads

ChillFish8:
Recently I wanted to test some smarter ways of balancing a large number of indexers in a multi-tenant situation without affecting searchers and decided to test out specifying specifically what CPU set the indexers could run on, giving searchers their own unaffected CPU set.

This is mostly an idea rather than a super useful feature, I've not included specifying the searcher CPU set as I'm not sure the best way to implement that in a way that doesn't result in extra calls when it doesn't need to be (i.e. SingleThreaded)


fulmicoton:
Have you measured a benefit?

ChillFish8:
@fulmicoton I have yet to get around to testing the general performance change this has, I wanted this particular feature to allow some better control over the indexing threads preventing indexers from interfering with searchers since tantivy spawns its own set of threads internally unlike the searchers.

The main advantage of this change is that you could spawn say `30` indexers all using 1 thread but constrain them to `6` CPU cores letting the OS load balance the indexers.

PR #2155, opened 2023-08-26
Fix testing on windows

ChillFish8:
After manually doing this fix on several local instances when doing a fix or change, I finally broke and have just made this PR to fix testing tantivy on Windows.

The error is because of the deps used in the `bench` directory aren't supported on Windows.


fulmicoton (approved):

fulmicoton:
Thank you!

PR #2160, opened 2023-08-28
Initial Implementation of ExistsQuery

### LINKED ISSUE #2159: Add ExistsQuery
In order to implement quickwit-oss/quickwit#3721 we should add an exists query that works with fast fields to tantivy. 

Add exists query in a way similar to range query.

Implement it fully directly in quickwit.

imotov:
Implements ExistsQuery that returns all document where a field has some value.

Fixes #2159

### DIFF
--- a/src/query/exist_query.rs
+++ b/src/query/exist_query.rs
@@ -1,0 +1,345 @@
+use core::fmt::Debug;
+
+use columnar::{ColumnIndex, DynamicColumn};
+
+use super::{ConstScorer, EmptyScorer};
+use crate::core::SegmentReader;
+use crate::docset::{DocSet, TERMINATED};
+use crate::query::explanation::does_not_match;
+use crate::query::{EnableScoring, Explanation, Query, Scorer, Weight};
+use crate::{DocId, Score, TantivyError};
+
+/// Query that matches all documents with a non-null value in the specified field.
+///
+/// All of the matched documents get the score 1.0.
+#[derive(Clone, Debug)]
+pub struct ExistsQuery {
+    field_name: String,
+}
+
+impl ExistsQuery {
+    /// Creates a new `ExistQuery` from the given field.
+    ///
+    /// This query matches all documents with at least one non-null value in the specified field.
+    /// This constructor never fails, but executing the search with this query will return an
+    /// error if the specified field doesn't exists or is not a fast field.
+    pub fn new_exists_query(field: String) -> ExistsQuery {
+        ExistsQuery { field_name: field }
+    }
+}
+
+impl Query for ExistsQuery {
+    fn weight(&self, enable_scoring: EnableScoring) -> crate::Result<Box<dyn Weight>> {
+        let schema = enable_scoring.schema();
+        let Some((field, _path)) = schema.find_field(&self.field_name) else {
+            return Err(TantivyError::FieldNotFound(self.field_name.clone()));
+        };
+        let field_type = schema.get_field_entry(field).field_type();
+        if !field_type.is_fast() {
+            return Err(TantivyError::SchemaError(format!(
+                "Field {} is not a fast field.",
+                self.field_name
+            )));
+        }
+        Ok(Box::new(ExistsWeight {
+            field_name: self.field_name.clone(),
+        }))
+    }
+}
+
+/// Weight associated with the `ExistsQuery` query.
+pub struct ExistsWeight {
+    field_name: String,
+}
+
+impl Weight for ExistsWeight {
+    fn scorer(&self, reader: &SegmentReader, boost: Score) -> crate::Result<Box<dyn Scorer>> {
+        let fast_field_reader = reader.fast_fields();
+        let dynamic_columns: crate::Result<Vec<DynamicColumn>> = fast_field_reader
+            .dynamic_column_handles(&self.field_name)?
+            .into_iter()
+            .map(|handle| handle.open().map_err(|io_error| io_error.into()))
+            .collect();
+        let mut non_empty_columns = Vec::new();
+        for column in dynamic_columns? {
+            if !matches!(column.column_index(), ColumnIndex::Empty { .. }) {
+                non_empty_columns.push(column)
+            }
+        }
+        // TODO: we can optimizer more here since in most cases we will have only one index
+        if !non_empty_columns.is_empty() {
+            let docset = ExistsDocSet::new(non_empty_columns, reader.max_doc());
+            Ok(Box::new(ConstScorer::new(docset, boost)))
+        } else {
+            Ok(Box::new(EmptyScorer))
+        }
+    }
+
+    fn explain(&self, reader: &SegmentReader, doc: DocId) -> crate::Result<Explanation> {
+        let mut scorer = self.scorer(reader, 1.0)?;
+        if scorer.seek(doc) != doc {
+            return Err(does_not_match(doc));
+        }
+        Ok(Explanation::new("ExistsQuery", 1.0))
+    }
+}
+
+pub(crate) struct ExistsDocSet {
+    columns: Vec<DynamicColumn>,
+    doc: DocId,
+    max_doc: DocId,
+}
+
+impl ExistsDocSet {
+    pub(crate) fn new(columns: Vec<DynamicColumn>, max_doc: DocId) -> Self {
+        let mut set = Self {
+            columns,
+            doc: 0u32,
+            max_doc,
+        };
+        set.find_next();
+        set
+    }
+
+    fn find_next(&mut self) -> DocId {
+        while self.doc < self.max_doc {
+            if self
+                .columns
+                .iter()
+                .any(|col| col.column_index().has_value(self.doc))
+            {
+                return self.doc;
+            }
+            self.doc += 1;
+        }
+        self.doc = TERMINATED;
+        TERMINATED
+    }
+}
+
+impl DocSet for ExistsDocSet {
+    fn advance(&mut self) -> DocId {
+        self.seek(self.doc + 1)
+    }
+
+    fn size_hint(&self) -> u32 {
+        0
+    }
+
+    fn doc(&self) -> DocId {
+        self.doc
+    }
+
+    #[inline(always)]
+    fn seek(&mut self, target: DocId) -> DocId {
+        self.doc = target;
+        self.find_next()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::net::Ipv6Addr;
+    use std::ops::Bound;
+
+    use common::DateTime;
+    use time::OffsetDateTime;
+
+    use crate::collector::Count;
+    use crate::query::exist_query::ExistsQuery;
+    use crate::query::{BooleanQuery, RangeQuery};
+    use crate::schema::{Facet, FacetOptions, Schema, FAST, INDEXED, STRING, TEXT};
+    use crate::{doc, Index, Searcher};
+
+    #[test]
+    fn test_exists_query_simple() -> crate::Result<()> {
+        let mut schema_builder = Schema::builder();
+        let all_field = schema_builder.add_u64_field("all", INDEXED | FAST);
+        let even_field = schema_builder.add_u64_field("even", INDEXED | FAST);
+        let odd_field = schema_builder.add_text_field("odd", STRING | FAST);
+        let multi_field = schema_builder.add_text_field("multi", FAST);
+        let _never_field = schema_builder.add_u64_field("never", INDEXED | FAST);
+        let schema = schema_builder.build();
+
+        let index = Index::create_in_ram(schema);
+        {
+            let mut index_writer = index.writer_for_tests()?;
+            for i in 0u64..100u64 {
+                if i % 2 == 0 {
+                    if i % 10 == 0 {
+                        index_writer.add_document(doc!(all_field => i, even_field => i, multi_field => i.to_string(), multi_field => (i + 1).to_string()))?;
+                    } else {
+                        index_writer.add_document(doc!(all_field => i, even_field => i))?;
+                    }
+                } else {
+                    index_writer.add_document(doc!(all_field => i, odd_field => i.to_string()))?;
+                }
+            }
+            index_writer.commit()?;
+        }
+        let reader = index.reader()?;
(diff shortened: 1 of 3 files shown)


src/query/exist_query.rs
```diff
@@ -0,0 +1,287 @@
...
+    ///
+    /// If the value type is not correct, something may go terribly wrong when
+    /// the `Weight` object is created.
+    pub fn new_exists_query(field: String) -> ExistsQuery {
+        ExistsQuery { field }
+    }
+}
+
+impl Query for ExistsQuery {
+    fn weight(&self, enable_scoring: EnableScoring<'_>) -> crate::Result<Box<dyn Weight>> {
```
fulmicoton: Nitpick: I suspect you can drop the lifetime here. (no?)

```suggestion
    fn weight(&self, enable_scoring: EnableScoring) -> crate::Result<Box<dyn Weight>> {
```

src/query/exist_query.rs
```diff
@@ -0,0 +1,287 @@
...
+            .get_field_entry(schema.get_field(&self.field)?)
+            .field_type();
+        if !field_type.is_fast() {
+            return Err(TantivyError::SchemaError(format!(
+                "Field {:?} is not a fast field.",
+                self.field
+            )));
+        }
+        Ok(Box::new(ExistsWeight {
+            field: self.field.to_string(),
```
fulmicoton: nitpick

```suggestion
            field: self.field.clone(),
```

Does not change anything, but gives the false impression that `self.field` is not a String.

src/query/exist_query.rs
```diff
@@ -0,0 +1,287 @@
...
+            FieldType::U64(_) => Some(ColumnType::U64),
+            FieldType::I64(_) => Some(ColumnType::I64),
+            FieldType::F64(_) => Some(ColumnType::F64),
+            FieldType::Bool(_) => Some(ColumnType::Bool),
+            FieldType::Date(_) => Some(ColumnType::DateTime),
+            FieldType::Bytes(_) => Some(ColumnType::Bytes),
+            FieldType::IpAddr(_) => Some(ColumnType::IpAddr),
+            _ => None,
+        }
+        .expect("Should be here");
```
fulmicoton: Wait no we need to support Json too here. Besides we really don't care about the column type in the case of existence.

src/query/exist_query.rs
```diff
@@ -0,0 +1,287 @@
...
+        if let Some(dynamic_column) =
+            fast_field_reader.dynamic_column_handle(&self.field, column_type)?
+        {
+            let docset = ExistsDocSet::new(dynamic_column.open()?);
+            return Ok(Box::new(ConstScorer::new(docset, boost)));
+        } else {
+            return Err(TantivyError::SchemaError(format!(
```
fulmicoton: I wonder if we should return an empty doc set in that case. Either way the behavior is something for which we need a unit test and a documentation.

columnar/src/column_index/mod.rs
```diff
@@ -159,6 +159,18 @@ impl ColumnIndex {
             }
         }
     }
+
+    pub fn get_docids_with_existing_value(
+        &self,
+        selected_docid_range: Range<u32>,
+        doc_ids: &mut Vec<u32>,
+    ) {
+        for doc_id in selected_docid_range {
```
fulmicoton: With the given implementation of get_docids_with_existing_value, I don't think the buffering is worth it.
@PSeitz may have a better insight on this.

I'd rather have a simple first implementation that does not fill a buffer with a dynamic horizon, but instead just seeks for the next doc that has a value in a dumb for loop.

The buffering could help if we have a more sophisticated implem of get_docs_with_existing_values.

fulmicoton:
@imotov The approach is ok. I would
- [x] ditch the doc ids buffering. It makes your code more complex, and I suspect it makes it slower than the trivial implementation. @PSeitz may have a different view on this.
- [x] possibly change the spec of what to do if the column is not found
- [x] add unit tests

The other half of the ticket is to integrate this into the ExistsQuery of quickwit:
- [ ] disabling exists query indexing for fast fields
- [ ] plug that implementation instead.

One tricky thing will be to test what happens when facing a fast field or when facing a dynamic field.

src/query/exist_query.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+}
+
+impl Query for ExistsQuery {
+    fn weight(&self, enable_scoring: EnableScoring<'_>) -> crate::Result<Box<dyn Weight>> {
+        let schema = enable_scoring.schema();
+        let field_type = schema
+            .get_field_entry(schema.get_field(&self.field)?)
+            .field_type();
+        if !field_type.is_fast() {
```
fulmicoton: this part won't work for json fields. In quickwit we use json fields for the dynamic mode.

We can probably be a little bit ugly here.

```rust
let dynamic_field_opt: Option<Field> = if cfg!(feature = "quickwit") {
            self.schema.get_field("_dynamic").ok()
} else {
    None
};
let field: Option<Field> = schema.find_field_with_default(&self.field, default_field_opt);
let field_type = schema
    .get_field_entry(schema.get_field(&self.field)?)
    .field_type();
```

src/query/exist_query.rs
```diff
@@ -0,0 +1,287 @@
...
+use crate::query::{EnableScoring, Explanation, Query, Scorer, Weight};
+use crate::schema::FieldType;
+use crate::{DocId, Score, TantivyError};
+
+/// Query that matches all of the documents.
+///
+/// All of the document get the score 1.0.
+#[derive(Clone, Debug)]
+pub struct ExistsQuery {
+    field: String,
```
fulmicoton: Renaming could make things a little bit less confusing with the `Field` type
```suggestion
    field_name: String,
```

src/query/exist_query.rs
```diff
@@ -0,0 +1,287 @@
...
+            field: self.field.to_string(),
+            field_type: field_type.clone(),
+        }))
+    }
+}
+
+/// Weight associated with the `ExistsQuery` query.
+pub struct ExistsWeight {
+    field: String,
+    field_type: FieldType,
```
fulmicoton: I don't think we need to have the `field_type` here.

```suggestion
```

we can just get the column index straight from the DynamicColumn

columnar/src/column_index/mod.rs
PSeitz: The buffering mainly reduced the overhead of single calls, e.g. dynamic dispatch, translating into value space and SIMD.

Only dynamic dispatch applies here, but since every doc is called separately, there's no gain here.

A push down of the check to the index implementation still may make sense, but we should back it with benchmarks and I would start with the simple version.

src/query/exist_query.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// full scan.
+    fetch_horizon: u32,
+    /// Current batch of loaded docs.
+    loaded_docs: VecCursor,
+    last_seek_pos_opt: Option<u32>,
+}
+
+const DEFAULT_FETCH_HORIZON: u32 = 128;
+impl ExistsDocSet {
+    pub(crate) fn new(column: DynamicColumn) -> Self {
```
PSeitz: We can have multiple Columns on JSON fields (numerical, bool, ip, text, bytes). They would all need to be checked for existence

fulmicoton: Excellent point!

imotov:
> possibly change the spec of what to do if the column is not found

I have implemented it to be consistent with other queries that I have seen. So, if field doesn't exists or not fast - it returns an error. I can see how returning nothing if the field doesn't exist might make sense, but it might also be pretty confusing. Returning nothing if field is not fast can be really confusing to users.

src/query/exist_query.rs
imotov: I have changed this part quite a bit. It seems to work.

src/query/exist_query.rs
```diff
@@ -0,0 +1,306 @@
...
+/// Query that matches all documents with a non-null value in the specified field.
+///
+/// All of the matched documents get the score 1.0.
+#[derive(Clone, Debug)]
+pub struct ExistsQuery {
+    field_name: String,
+}
+
+impl ExistsQuery {
+    #![allow(dead_code)]
```
fulmicoton:
```suggestion
```

fulmicoton: Why would we allow dead code? The compiler is informing you that there is no way to call this method.

src/query/exist_query.rs
```diff
@@ -0,0 +1,306 @@
...
+
+impl Query for ExistsQuery {
+    fn weight(&self, enable_scoring: EnableScoring) -> crate::Result<Box<dyn Weight>> {
+        let schema = enable_scoring.schema();
+        let Some((field, _path)) = schema.find_field(&self.field_name) else {
+            return Err(TantivyError::FieldNotFound(self.field_name.clone()));
+        };
+        let field_type = schema.get_field_entry(field).field_type();
+        if !field_type.is_fast() {
+            return Err(TantivyError::SchemaError(format!(
```
fulmicoton, line 39: Yeah I think this is a good idea actually. My bad.

src/query/exist_query.rs
```diff
@@ -0,0 +1,306 @@
...
+        }))
+    }
+}
+
+/// Weight associated with the `ExistsQuery` query.
+pub struct ExistsWeight {
+    field_name: String,
+}
+
+impl ExistsWeight {}
```
fulmicoton:
```suggestion
```

columnar/src/column_index/mod.rs
```diff
@@ -159,6 +159,18 @@ impl ColumnIndex {
...
+            if self.has_value(doc_id) {
+                doc_ids.push(doc_id)
+            }
+        }
+    }
```
fulmicoton:
```suggestion
```

src/query/exist_query.rs
imotov: I am not sure I follow. It is not hooked up to grammar at the moment, so the only users of this will be quickwit's ExistsQuery for now.

imotov: Oh, yeah, I keep forgetting that `pub use` thing.

src/query/exist_query.rs
```diff
@@ -0,0 +1,306 @@
...
+            if !matches!(column.column_index(), ColumnIndex::Empty { .. }) {
+                non_empty_columns.push(column)
+            }
+        }
+        if !non_empty_columns.is_empty() {
+            let docset = ExistsDocSet::new(non_empty_columns, reader.max_doc());
+            Ok(Box::new(ConstScorer::new(docset, boost)))
+        } else {
+            Ok(Box::new(EmptyScorer))
+        }
```
fulmicoton, line 75: That will work... Can you add a TODO and open a ticket to optimize this?

Since we returned a Boxed scorer it is worth being a tad more precise here.
Most of the time we will have only one column index.
Often this column index will be containing all documents.

For optional indexes we also know how to be more accurate.

Ideally the code should have a function like 

```
pub fn doc_set_from_index(column_index: ColumnIndex) -> Box<dyn DocSet>;
```

imotov: Opened #2166

PSeitz:
> I have implemented it to be consistent with other queries that I have seen. So, if field doesn't exists or not fast - it returns an error. I can see how returning nothing if the field doesn't exist might make sense, but it might also be pretty confusing. Returning nothing if field is not fast can be really confusing to users.

Non-Existence should return an empty result on JSON fields, since they may be non-existing only on some segments.

fulmicoton:
@PSeitz @imotov there is a bit of a semantic problem in our discussion. When we say the field is absent, we are talking of `Field` in the sense of tantivy schema. Not `(Field, Path)`.

If it is not available, meaning resolve path has return `None`, I think it makes sense to return an error.
In quickwit, we will need to have some logic to be a little bit more lenient, in order to handle changes of schema.
That logic is already pretty solid, and exists in the conversion from QueryAst to tantivy query.

src/query/exist_query.rs
```diff
@@ -0,0 +1,304 @@
...
+    }
+}
+
+impl DocSet for ExistsDocSet {
+    fn advance(&mut self) -> DocId {
+        self.doc += 1;
+        self.find_next()
+    }
+
+    fn size_hint(&self) -> u32 {
```
fulmicoton, line 125: We need an implementation and a unit test for  `DocSet::seek`?

The implementation is rather trivial.

imotov: I think at the moment it will be pretty much the same as default implementation. We could optimize it a tiny bit but I would rather do that in #2166 where we can add `pub fn doc_set_from_index(column_index: ColumnIndex) -> Box<dyn DocSet>;` and then implement something like AndDocSet, which can then perhaps have a more optimized advanced.

imotov:
@PSeitz, @fulmicoton  so, the way it is implemented right now if you have a json field called `foo` and you try to run presence query on `foo.bar`, it will return an error if `foo` doesn't exists and it will return an empty result set if `foo` exists and is defined as `json` field but none of the json documents have `bar` inside of them. I updated the test in [e4e5225](https://github.com/quickwit-oss/tantivy/pull/2160/commits/e4e5225561c43a793091f0f7a3df2b7444b68aad) to demonstrate these scenarios.

src/query/exist_query.rs
fulmicoton: There is a trivial to get something much much faster than the default implementation.


```rust
 fn advance(&mut self) -> DocId {
    self.seek(self.doc + 1)
 }

 #[inline(always)]
 fn seek(&mut self, target: DocId) -> DocId {
        self.doc = target;
        self.find_next()
 }
```

imotov: Oh, I totally misinterpreted what default seek is actually doing. I see your point now.

fulmicoton (approved):

imotov:
~Can somebody merge this in? I am not authorized to merge pull requests in this repository.~

Not ready to be merged yet. Found some issues while trying to integrate it into quickwit.

imotov:
Turned out that the issue was in quickwit and not here. OK to merge.

fulmicoton (approved):

PR #2170, opened 2023-09-04
add support for exists query syntax in query parser

trinity-1686a:
tantivy side of https://github.com/quickwit-oss/quickwit/issues/3714

### DIFF
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -302,7 +302,9 @@ fn term_group_infallible(i: &str) -> JResult<&str, UserInputAst> {
 
 fn exists(i: &str) -> IResult<&str, UserInputLeaf> {
     value(
-        UserInputLeaf::Exists { field: None },
+        UserInputLeaf::Exists {
+            field: String::new(),
+        },
         tuple((space0, char('*'))),
     )(i)
 }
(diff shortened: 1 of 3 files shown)


query-grammar/src/user_input_ast.rs
```diff
@@ -16,6 +16,9 @@ pub enum UserInputLeaf {
         field: Option<String>,
         elements: Vec<String>,
     },
+    Exists {
+        field: Option<String>,
```
fulmicoton: I don't understand the reason for the `Option`  here.
 Can you add a comment explaining the meaning of `Exists { field: None }` and when it can be built?

query-grammar/src/query_grammar.rs
```diff
@@ -300,10 +300,42 @@ fn term_group_infallible(i: &str) -> JResult<&str, UserInputAst> {
     }
 }
 
+fn exists(i: &str) -> IResult<&str, UserInputLeaf> {
```
fulmicoton: Can we stop the one char variable? If it is `i` for input, let's at least go for `inp`

query-grammar/src/user_input_ast.rs
trinity-1686a: it is just an intermediary state that should never make it out to the world. I've changed so it's always a String, which can temporarily be empty

fulmicoton (approved):

PR #2173, opened 2023-09-07
extend FuzzyTermQuery to support json field

PingXia-at:
## Summary

Discord discussion here: https://discord.com/channels/908281611840282624/908286403086024724/1148718777651970126

## Test Plan

- [x] Added test for fuzzyTermQuery

### DIFF
--- a/src/query/automaton_weight.rs
+++ b/src/query/automaton_weight.rs
@@ -16,10 +16,9 @@ use super::phrase_prefix_query::prefix_end;
 pub struct AutomatonWeight<A> {
     field: Field,
     automaton: Arc<A>,
-    // This is specifically used for JSON fields.
-    // Context: For a JSON field, the term's value bytes include the JSON path, JSON path type, and the actual value bytes.
-    // When performing fuzzy or regex search on a JSON field's terms dictionary, it's essential to filter out terms that don't match the JSON path.
-    // Otherwise, a term might be mistakenly matched because of the JSON path, even if the actual value bytes don't match the search criteria.
+    // For JSON fields, the term dictionary include terms from all paths.
+    // We apply additional filtering based on the given JSON path, when searching within the term dictionary.
+    // This prevents terms from unrelated paths from matching the search criteria.
     json_path_bytes: Option<Box<[u8]>>,
 }
 
--- a/src/query/automaton_weight.rs
+++ b/src/query/automaton_weight.rs
@@ -10,10 +10,14 @@ use crate::schema::{Field, IndexRecordOption};
 use crate::termdict::{TermDictionary, TermStreamer};
 use crate::{DocId, Score, TantivyError};
 
+use super::phrase_prefix_query::prefix_end;
+
 /// A weight struct for Fuzzy Term and Regex Queries
 pub struct AutomatonWeight<A> {
     field: Field,
     automaton: Arc<A>,
+    // Used to filter out terms that don't match the json path
+    json_path_bytes: Option<Vec<u8>>,
 }
 
 impl<A> AutomatonWeight<A>
@@ -26,6 +30,20 @@ where
         AutomatonWeight {
             field,
             automaton: automaton.into(),
+            json_path_bytes: None,
+        }
+    }
+
+    /// Create a new AutomationWeight for a json path
+    pub fn new_with_json_path<IntoArcA: Into<Arc<A>>>(
+        field: Field,
+        automaton: IntoArcA,
+        json_path_bytes: Option<&[u8]>,
+    ) -> AutomatonWeight<A> {
+        AutomatonWeight {
+            field,
+            automaton: automaton.into(),
+            json_path_bytes: json_path_bytes.map(|bytes_ref| bytes_ref.to_vec()),
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/query/automaton_weight.rs
```diff
@@ -10,10 +10,14 @@ use crate::schema::{Field, IndexRecordOption};
...
 use crate::{DocId, Score, TantivyError};
 
+use super::phrase_prefix_query::prefix_end;
+
 /// A weight struct for Fuzzy Term and Regex Queries
 pub struct AutomatonWeight<A> {
     field: Field,
     automaton: Arc<A>,
+    // Used to filter out terms that don't match the json path
+    json_path_bytes: Option<Vec<u8>>,
```
adamreichold: It would be nice to use `Option<Box<[u8]>>` instead here to keep the size increase minimal when this is not used.

src/query/automaton_weight.rs
```diff
@@ -34,7 +52,15 @@ where
         term_dict: &'a TermDictionary,
     ) -> io::Result<TermStreamer<'a, &'a A>> {
         let automaton: &A = &self.automaton;
-        let term_stream_builder = term_dict.search(automaton);
+        let mut term_stream_builder = term_dict.search(automaton);
+
+        if self.json_path_bytes.is_some() {
```
adamreichold: This should use pattern matching, e.g.

```rust
if let Some(json_path_bytes) = &self.json_path_bytes {
```

instead of `.is_some()` follow by `.as_ref().unwrap()`.

fulmicoton:
@PSeitz can you review?

src/query/fuzzy_query.rs
```diff
@@ -132,18 +132,45 @@ impl FuzzyTermQuery {
...
         } else {
             automaton_builder.build_dfa(term_text)
         };
-        Ok(AutomatonWeight::new(
-            self.term.field(),
-            DfaWrapper(automaton),
-        ))
+
+        if let Some(json_path_bytes) = term_value.as_json_path_bytes() {
+            return Ok(AutomatonWeight::new_for_json_path(
```
DavidZbarsky-at: nit: drop `return` and semicolon

src/schema/term.rs
```diff
@@ -413,6 +422,21 @@ where B: AsRef<[u8]>
         Some((json_path, ValueBytes::wrap(&term[1..])))
     }
 
+    /// Returns the json path bytes (including the JSON_END_OF_PATH byte)
+    ///
+    /// Returns `None` if the value is not JSON.
+    pub(crate) fn as_json_path_bytes(&self) -> Option<&[u8]> {
```
PSeitz, line 428: I think the return type should be `ValueBytes`

PSeitz: I don't think we need an extra method, `as_json` already covers the path. `as_json` seems unused currently, you can edit it to include JSON_END_OF_PATH if needed

PingXia-at: The json_path_bytes are used [here](https://github.com/PingXia-at/tantivy/blob/4d054e6541f695d1c61e64788b40e7f32fe8cf59/src/query/automaton_weight.rs#L51)

```
fn automaton_stream<'a>(
        &'a self,
        term_dict: &'a TermDictionary,
    ) -> io::Result<TermStreamer<'a, &'a A>> {
        let automaton: &A = &self.automaton;
        let mut term_stream_builder = term_dict.search(automaton);

        if let Some(json_path_bytes) = &self.json_path_bytes {
            term_stream_builder = term_stream_builder.ge(json_path_bytes);
            if let Some(end) = prefix_end(json_path_bytes) {
                term_stream_builder = term_stream_builder.lt(&end);
            }
        }

        term_stream_builder.into_stream()
    }
```

Two reasons for this new method
1. We need to include the `JSON_END_OF_PATH ` byte. Otherwise, the automaton_stream will include path `aa` while we're only interested in path `a`
2. The StreamBuilder.ge / lt methods require byte references, not a string for the path

PingXia-at: >as_json seems unused currently, you can edit it to include JSON_END_OF_PATH if needed

cool, I'll change this method

PingXia-at: >as_json seems unused currently, you can edit it to include JSON_END_OF_PATH if needed

FYI it is used in debug_value_bytes.

src/query/phrase_prefix_query/mod.rs
```diff
@@ -6,7 +6,7 @@ pub use phrase_prefix_query::PhrasePrefixQuery;
 pub use phrase_prefix_scorer::PhrasePrefixScorer;
 pub use phrase_prefix_weight::PhrasePrefixWeight;
 
-fn prefix_end(prefix_start: &[u8]) -> Option<Vec<u8>> {
+pub(crate) fn prefix_end(prefix_start: &[u8]) -> Option<Vec<u8>> {
```
PSeitz, line 9: I don't understand the `u8::MAX` logic here

PingXia-at: I referred to the [phrasePrefixQuery implementation](https://github.com/quickwit-oss/tantivy/blob/6564e0c467106ef8570a0fc4f469017347e3e8ef/src/query/phrase_prefix_query/phrase_prefix_weight.rs#L85), which also filters the term based on the term value prefix.

PingXia-at: I believe it attempts to find the next larger prefix. Typically, this involves incrementing the last u8 value by 1. However, there is an edge case to consider when the last u8 value is u8::MAX.

PSeitz (approved):

PSeitz:
LGTM. Thanks!

PR #2183, opened 2023-09-21
Fix DateHistogram bucket gap

### LINKED ISSUE #3837: Aborting aggregation because memory limit was exceeded
The quickwit grafana plugin looks great, but there are also some issues with its use：

Through the Grafana plugin, I only searched for data within an hour, which is around a few million, but the volumes logs reported an error：

Failed to load log volume for this query
Internal error: `Aborting aggregation because memory limit was exceeded. Limit: 500.00 MB, Current: 1.36 PB`.

because I added a condition to the Lucene query. If I didn't add a condition, it would be fine

other ： 
1、Can the level fields and their values in the data source configuration be customized so that I can distinguish between other types of data, rather than just general log levels

2、Can I use Quickwit as a data source to create reports similar to curve and pie charts,

3、Or support aggregation in languages such as count, group by, etc. on Granafa
   and  Comparison of some calculations： > 、>=、 < 、!=  and so on

PSeitz:
Fixes a computation issue of the number of buckets needed in the DateHistogram.

This is due to a missing normalization from request values (ms) to fast field
values (ns), when converting an intermediate result to the final result.
This results in a wrong computation by a factor 1_000_000.
The Histogram normalizes values to nanoseconds, to make the user input like
extended_bounds (ms precision) and the values from the fast field (ns precision for date type) compatible.
This normalization happens only for date type fields, as other field types don't have precision settings.
The normalization does not happen due a missing `column_type`, which is not
correctly passed after merging an empty aggregation (which does not have a `column_type` set), with a regular aggregation.

Another related issue is an empty aggregation, which will not have
`column_type` set, will not convert the result to human readable format.

This PR fixes the issue by:
- Limit the allowed field types of DateHistogram to DateType
- Instead of passing the column_type, which is only available on the segment level, we flag the aggregation as `is_date_agg`.
- Fix the merge logic

Add a flag to do normalization only once. This is not an issue currently, but it could become easily one.

closes https://github.com/quickwit-oss/quickwit/issues/3837

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -17,11 +17,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install Rust
-        run: rustup toolchain install nightly --profile minimal --component llvm-tools-preview
+        run: rustup toolchain install nightly-2023-09-10 --profile minimal --component llvm-tools-preview
       - uses: Swatinem/rust-cache@v2
       - uses: taiki-e/install-action@cargo-llvm-cov
       - name: Generate code coverage
-        run: cargo +nightly llvm-cov --all-features --workspace --doctests --lcov --output-path lcov.info
+        run: cargo +nightly-2023-09-10 llvm-cov --all-features --workspace --doctests --lcov --output-path lcov.info
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
         continue-on-error: true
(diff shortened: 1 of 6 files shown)


PSeitz:
Failing coverage depends on: https://github.com/time-rs/time/issues/624

(or we could switch temporarily to an older nightly version)

fulmicoton (approved):

PR #2185, opened 2023-09-21
encode some part of posting list as -1 instead of direct values

### LINKED ISSUE #2180: when recording term frequency and position on json field, tf and position for number fields are generated and stored
When indexing on a json field configured to store TF+position, synthetic TF are created and stored for integer fields, despite these values not making much sense for non-string fields.
The value are created by [these lines](https://github.com/quickwit-oss/tantivy/blob/c520ac46fc88dd9e06121cdd6d1d701a5678247e/src/postings/recorder.rs#L136-L140)

This poses issues when trying to encode TF-1 instead of tf (which makes for a more compact format). In this specific case, changing the default TF to be 1 instead of 0 make things works, but it would be better to not encode unused information at all.

related to https://github.com/quickwit-oss/quickwit/issues/3740

### LINKED ISSUE #3740: Fix the +1 bitpacking scheme
Right now the posting lists are compressed using delta encoding + bitpacking.

In reality the documents are strictly monotonous, so the delta is always >0.
We could bitpack

`(doc_n+1 - doc_n - 1)` instead of `(doc_n+1 - doc_n)`

The benefit is usually tiny. We save 128 when this change made it possible to change the bitwidth used for the block.
For a block with bitwidth b, assuming a uniform distirbution (a suspicious but useful hypothesis), we have roughly speaking `1 / 2^(b - 1)` to gain 128 bits.

(8 + 128*(b-1))/(8 + 128*b)

So we have `1/ 2^(b-1)` chances to improve our compression rate by a factor of `16 / (1 + 16*b)`.

- For a bitwidth of 1, a full saturation that's 100% chance to bring our block of 136bits down to 8 bits.
- For a bitwidth of 2, (roughly > 90% saturation) this gives us a chance to bring our block of 264bits down to 136 bits.
- Above it gets rapidly not as interesting.

This should not hurt performance too much provided we do this when we have the info in a bitpacking register.

Currently we have two unused bits in the bitwidths  that we could use to ensure backward compatibility.
bit 7 could for instance encode whether we want this scheme or not.

The implementation is trickier than it may seem unfortunately. 
(In particular what should be the initial value we use)

trinity-1686a:
fix https://github.com/quickwit-oss/quickwit/issues/3740 on next tantivy pull
fix #2180 

tested with quickwit on 100k documents from github-archive, using a json term covering everything, indexed+stored+record-position+fast, all put in a single split.
Here are the size of the different sections of the split, at each step of the changes:

image

the code should be compatible with old indexes, however it generate posting lists that can't be read by old version.
Not emitting TF cause the decoding to become somewhat of a guesswork, we know this field has frequency, but does it really has frequency for *this* term? We now decide that based on whether the section is lengthily enough to contain those. As far as I can tell it's correct, but not beautiful, and rather fragile. I think I prefer encoding dummy TF/position than actually not emitting those

### DIFF
--- a/src/postings/block_segment_postings.rs
+++ b/src/postings/block_segment_postings.rs
@@ -30,7 +30,7 @@ pub struct BlockSegmentPostings {
     block_max_score_cache: Option<Score>,
     doc_freq: u32,
     data: OwnedBytes,
-    pub(crate) skip_reader: SkipReader,
+    skip_reader: SkipReader,
 }
 
 fn decode_bitpacked_block(
@@ -303,7 +303,6 @@ impl BlockSegmentPostings {
 
     pub(crate) fn load_block(&mut self) {
         let offset = self.skip_reader.byte_offset();
-        // I'm not sure why it's there, but it breaks with blocks that can be encoded on zero byte.
         if self.block_is_loaded() {
             return;
         }
@@ -372,6 +371,10 @@ impl BlockSegmentPostings {
             data: OwnedBytes::empty(),
             skip_reader: SkipReader::new(OwnedBytes::empty(), 0, IndexRecordOption::Basic),
         }
+    }
+
+    pub(crate) fn skip_reader(&self) -> &SkipReader {
+        &self.skip_reader
     }
 }
 
--- a/src/postings/block_segment_postings.rs
+++ b/src/postings/block_segment_postings.rs
@@ -45,7 +45,7 @@ fn decode_bitpacked_block(
     let num_consumed_bytes =
         doc_decoder.uncompress_block_sorted(data, doc_offset, doc_num_bits, strict_delta);
     if let Some(freq_decoder) = freq_decoder_opt {
-        freq_decoder.uncompress_block_unsorted(&data[num_consumed_bytes..], tf_num_bits);
+        freq_decoder.uncompress_block_unsorted(&data[num_consumed_bytes..], tf_num_bits, strict_delta);
     }
 }
 
--- a/src/postings/block_segment_postings.rs
+++ b/src/postings/block_segment_postings.rs
@@ -24,7 +24,7 @@ fn max_score<I: Iterator<Item = Score>>(mut it: I) -> Option<Score> {
 #[derive(Clone)]
 pub struct BlockSegmentPostings {
     pub(crate) doc_decoder: BlockDecoder,
-    loaded_offset: usize,
+    block_loaded: bool,
     freq_decoder: BlockDecoder,
     freq_reading_option: FreqReadingOption,
     block_max_score_cache: Option<Score>,
@@ -88,6 +88,12 @@ fn split_into_skips_and_postings(
 }
 
 impl BlockSegmentPostings {
+    /// Opens a `BlockSegmentPostings`.
+    /// `doc_freq` is the number of documents in the posting list.
+    /// `record_option` represents the amount of data available according to the schema.
+    /// `requested_option` is the amount of data requested by the user.
+    /// If for instance, we do not request for term frequencies, this function will not decompress
+    /// term frequency blocks.
     pub(crate) fn open(
         doc_freq: u32,
         data: FileSlice,
@@ -102,7 +108,10 @@ impl BlockSegmentPostings {
                 // 8 is the minimum size of a block with frequency (can be more if pos are stored
                 // too)
                 if skip_data.len() < 8 * block_count {
-                    // the field might be encoded with frequency, but this term in particular isn't
+                    // the field might be encoded with frequency, but this term in particular isn't.
+                    // This can happen for JSON field with term frequencies:
+                    // - text terms are encoded with term freqs.
+                    // - numerical terms are encoded without term freqs.
                     record_option = IndexRecordOption::Basic;
                 }
                 SkipReader::new(skip_data, doc_freq, record_option)
@@ -118,7 +127,7 @@ impl BlockSegmentPostings {
 
         let mut block_segment_postings = BlockSegmentPostings {
             doc_decoder: BlockDecoder::with_val(TERMINATED),
-            loaded_offset: usize::MAX,
+            block_loaded: false,
             freq_decoder: BlockDecoder::with_val(1),
             freq_reading_option,
             block_max_score_cache: None,
@@ -188,7 +197,7 @@ impl BlockSegmentPostings {
             split_into_skips_and_postings(doc_freq, postings_data)?;
         self.data = postings_data;
         self.block_max_score_cache = None;
-        self.loaded_offset = usize::MAX;
+        self.block_loaded = false;
         if let Some(skip_data) = skip_data_opt {
             self.skip_reader.reset(skip_data, doc_freq);
         } else {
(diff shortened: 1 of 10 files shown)


src/postings/compression/mod.rs
```diff
@@ -33,14 +33,34 @@ impl BlockEncoder {
...
+        let num_bits = self.bitpacker.num_bits_strictly_sorted(offset, block);
         let written_size =
             self.bitpacker
-                .compress_sorted(offset, block, &mut self.output[..], num_bits);
+                .compress_strictly_sorted(offset, block, &mut self.output[..], num_bits);
         (num_bits, &self.output[..written_size])
     }
 
-    pub fn compress_block_unsorted(&mut self, block: &[u32]) -> (u8, &[u8]) {
+    pub fn compress_block_unsorted(
```
fulmicoton, line 53: I think this deserve a comment about the extra comment.

src/postings/compression/mod.rs
```diff
@@ -76,16 +96,41 @@ impl BlockDecoder {
         compressed_data: &[u8],
         offset: u32,
         num_bits: u8,
+        strict_delta: bool,
     ) -> usize {
-        self.output_len = COMPRESSION_BLOCK_SIZE;
-        self.bitpacker
-            .decompress_sorted(offset, compressed_data, &mut self.output, num_bits)
+        if strict_delta {
+            let offset = std::num::NonZeroU32::new(offset).map(std::num::NonZeroU32::get);
```
fulmicoton, line 112: Is it an artistic version of :
```
    if offset == 0u32 {
         None
    } else {
         Some(offset)
    }
```

src/postings/block_segment_postings.rs
```diff
@@ -274,13 +293,15 @@ impl BlockSegmentPostings {
 
     pub(crate) fn load_block(&mut self) {
         let offset = self.skip_reader.byte_offset();
-        if self.loaded_offset == offset {
-            return;
-        }
+        // I'm not sure why it's there, but it breaks with blocks that can be encoded on zero byte.
+        // if self.loaded_offset == offset {
```
fulmicoton: the idea is simply to avoid decompressing stuff again if we are hitting the same block.

Is this something that cannot happen, because of what's happening at a higher level?

src/postings/skip.rs
```diff
@@ -6,6 +6,13 @@ use crate::query::Bm25Weight;
 use crate::schema::IndexRecordOption;
 use crate::{DocId, Score, TERMINATED};
 
+// doc num bits between 0..=32 are used for legacy, delta without -1 encoding
+// 33..=65 are used for current delta-1 encoding
+// 66..=255 is currently unused
+//
+// when ysing strict delta, we also encode term frequency as value-1 instead of value.
```
fulmicoton:
```suggestion
// when using strict delta, we also encode term frequency as value-1 instead of value.
```

src/postings/skip.rs
```diff
@@ -6,6 +6,13 @@ use crate::query::Bm25Weight;
 use crate::schema::IndexRecordOption;
 use crate::{DocId, Score, TERMINATED};
 
+// doc num bits between 0..=32 are used for legacy, delta without -1 encoding
+// 33..=65 are used for current delta-1 encoding
```
fulmicoton: I suspect bit 7 would have been nicer. Of course it is not super natural since our values go from 0 to 32 included. (hence not fitting on 5 bits)

```
let delta_1 = (encoding_flag >> 7) & 1;
let num_bits = (encoding_flag & MASK);
```

src/postings/skip.rs
```diff
@@ -172,12 +180,17 @@ impl SkipReader {
         let bytes = self.owned_read.as_slice();
         let advance_len: usize;
         self.last_doc_in_block = read_u32(bytes);
-        let doc_num_bits = bytes[4];
+        let mut doc_num_bits = bytes[4];
```
fulmicoton: `doc_num_bits` changes meaning here.
First it is something that encodes both `num_bits` and whether we need to do -1 stuff.
Then it is something that does encode only `num_bits`.

Can we make it two variables with different names?

src/postings/block_segment_postings.rs
```diff
@@ -81,22 +91,31 @@ impl BlockSegmentPostings {
     pub(crate) fn open(
```
fulmicoton, line 97:
```suggestion
    /// Opens a `BlockSegmentPostings`.
    /// `doc_freq` is the number of documents in the posting list.
    /// `record_option` represents the amount of data available according to the schema.
    /// `requested_option` is the amount of data requested by the user.
    /// If for instance, we do not request for term frequencies, this function will not decompress
    /// term frequency blocks.
```

src/postings/block_segment_postings.rs
```diff
@@ -81,22 +91,31 @@ impl BlockSegmentPostings {
...
         let bytes = data.read_bytes()?;
         let (skip_data_opt, postings_data) = split_into_skips_and_postings(doc_freq, bytes)?;
         let skip_reader = match skip_data_opt {
-            Some(skip_data) => SkipReader::new(skip_data, doc_freq, record_option),
+            Some(skip_data) => {
+                let block_count = doc_freq as usize / COMPRESSION_BLOCK_SIZE;
+                // 8 is the minimum size of a block with frequency (can be more if pos are stored
+                // too)
+                if skip_data.len() < 8 * block_count {
+                    // the field might be encoded with frequency, but this term in particular isn't
```
fulmicoton:
```suggestion
                    // the field might be encoded with frequency, but this term in particular isn't.
                    // This can happen for JSON field with term frequencies:
                    // - text terms are encoded with term freqs.
                    // - numerical terms are encoded without term freqs.
```

src/postings/block_segment_postings.rs
fulmicoton: If not, this could actually be a big regression on some queries
(e.g. query interseciton of two terms that are strongly correlated. e.g. `Barrack AND Obama`

trinity-1686a: I believe it could happen. I changed how tracking of whether the block is tracked is done so it doesn't fail with zero-sized blocks

src/postings/block_segment_postings.rs
```diff
@@ -265,22 +293,24 @@ impl BlockSegmentPostings {
...
 
     pub(crate) fn block_is_loaded(&self) -> bool {
-        self.loaded_offset == self.skip_reader.byte_offset()
+        self.block_loaded
     }
 
     pub(crate) fn load_block(&mut self) {
         let offset = self.skip_reader.byte_offset();
-        if self.loaded_offset == offset {
+        // I'm not sure why it's there, but it breaks with blocks that can be encoded on zero byte.
```
fulmicoton:
```suggestion
        // I'm not sure why it's there, but it breaks with blocks that can be encoded on zero byte.
```

do we still need the comment?

src/postings/block_segment_postings.rs
```diff
@@ -24,7 +24,7 @@ fn max_score<I: Iterator<Item = Score>>(mut it: I) -> Option<Score> {
 #[derive(Clone)]
 pub struct BlockSegmentPostings {
```
fulmicoton, line 25: the `skip_reader` member in `BlockSegmentPostings` is `pub(crate)`, but calling next on it can break the invariant you rephrased ("if `block_loaded` is set to true the block loaded in `doc_decoder` and `freq_decoder` is matching the block of skip_reader).

Can you see if it is possible to make it private, and if needed expose it with an inline non mut accessor.

fulmicoton (changes requested):
the skip_reader member in BlockSegmentPostings is pub(crate), but calling next on it can break the invariant you rephrased ("if block_loaded is set to true the block loaded in doc_decoder and freq_decoder is matching the block of skip_reader).

Can you see if it is possible to make it private, and if needed expose it with an inline non mut accessor.

fulmicoton (approved):

PR #2186, opened 2023-09-24
replace BinaryHeap for TopN

PSeitz:
replace BinaryHeap for TopN with variant that selects the median with QuickSort,
which runs in O(n) time.

#### Top 10
Query | tantivy-0.20 | tantivy-0.21 | tantivy-topn
-- | -- | -- | --
AVERAGE | 2,882 μs | 2,844 μs | 2,469 μs

#### Top 100
Query | tantivy-0.20 | tantivy-0.21 | tantivy-topn
-- | -- | -- | --
AVERAGE | 2,846 μs | 2,902 μs | 2,504 μs

#### Top 1000
Query | tantivy-0.20 | tantivy-0.21 | tantivy-topn
-- | -- | -- | --
AVERAGE | 2,848 μs | 3,169 μs | 2,518 μs

### Variant with `push` 

topn_1
topn_2
topn_3

### DIFF
--- a/src/collector/top_collector.rs
+++ b/src/collector/top_collector.rs
@@ -20,6 +20,14 @@ pub(crate) struct ComparableDoc<T, D> {
     pub feature: T,
     pub doc: D,
 }
+impl<T: std::fmt::Debug, D: std::fmt::Debug> std::fmt::Debug for ComparableDoc<T, D> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ComparableDoc")
+            .field("feature", &self.feature)
+            .field("doc", &self.doc)
+            .finish()
+    }
+}
 
 impl<T: PartialOrd, D: PartialOrd> PartialOrd for ComparableDoc<T, D> {
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
@@ -99,7 +107,8 @@ where T: PartialOrd + Clone
         }
 
         Ok(top_collector
-            .into_iter_sorted()
+            .into_sorted_vec()
+            .into_iter()
             .skip(self.offset)
             .map(|cdoc| (cdoc.feature, cdoc.doc))
             .collect())
@@ -151,7 +160,8 @@ impl<T: PartialOrd + Clone> TopSegmentCollector<T> {
     pub fn harvest(self) -> Vec<(T, DocAddress)> {
         let segment_ord = self.segment_ord;
         self.topn_computer
-            .into_iter_sorted()
+            .into_sorted_vec()
+            .into_iter()
             .map(|comparable_doc| {
                 (
                     comparable_doc.feature,
--- a/src/collector/top_collector.rs
+++ b/src/collector/top_collector.rs
@@ -107,9 +107,7 @@ where T: PartialOrd + Clone
                     .take(self.limit)
                     .collect());
             }
-            if child.len() > self.limit {
-                child.truncate(self.limit);
-            }
+            child.truncate(self.limit);
 
             return Ok(child);
         }
--- a/src/collector/top_collector.rs
+++ b/src/collector/top_collector.rs
@@ -60,8 +60,7 @@ pub(crate) struct TopCollector<T> {
 }
 
 impl<T> TopCollector<T>
-where
-    T: PartialOrd + Clone,
+where T: PartialOrd + Clone
 {
     /// Creates a top collector, with a number of documents equal to "limit".
     ///
--- a/src/collector/top_collector.rs
+++ b/src/collector/top_collector.rs
@@ -60,7 +60,8 @@ pub(crate) struct TopCollector<T> {
 }
 
 impl<T> TopCollector<T>
-where T: PartialOrd + Clone
+where
+    T: PartialOrd + Clone,
 {
     /// Creates a top collector, with a number of documents equal to "limit".
     ///
@@ -86,30 +87,10 @@ where T: PartialOrd + Clone
 
     pub fn merge_fruits(
         &self,
-        mut children: Vec<Vec<(T, DocAddress)>>,
+        children: Vec<Vec<(T, DocAddress)>>,
     ) -> crate::Result<Vec<(T, DocAddress)>> {
         if self.limit == 0 {
             return Ok(Vec::new());
-        }
-        if children.len() == 1 {
-            let mut child = children.pop().unwrap();
-            child.sort_by(|a, b| {
-                if a.0 == b.0 {
-                    a.1.cmp(&b.1)
-                } else {
-                    b.0.partial_cmp(&a.0).unwrap_or(std::cmp::Ordering::Equal)
-                }
-            });
-            if self.offset != 0 {
-                return Ok(child
-                    .into_iter()
-                    .skip(self.offset)
-                    .take(self.limit)
-                    .collect());
-            }
-            child.truncate(self.limit);
-
-            return Ok(child);
         }
         let mut top_collector = TopNComputer::new(self.limit + self.offset);
         for child_fruit in children {
@@ -150,9 +131,9 @@ where T: PartialOrd + Clone
 /// The Top Collector keeps track of the K documents
 /// sorted by type `T`.
 ///
-/// The implementation is based on a `BinaryHeap`.
+/// The implementation is based on a repeatedly truncating on the median after K * 2 documents
 /// The theoretical complexity for collecting the top `K` out of `n` documents
-/// is `O(n log K)`.
+/// is `O(n + K)`.
 pub(crate) struct TopSegmentCollector<T> {
     topn_computer: TopNComputer<T, DocId>,
     segment_ord: u32,
--- a/src/collector/top_collector.rs
+++ b/src/collector/top_collector.rs
@@ -1,7 +1,7 @@
 use std::cmp::Ordering;
-use std::collections::BinaryHeap;
 use std::marker::PhantomData;
 
+use super::top_score_collector::TopNComputer;
 use crate::{DocAddress, DocId, SegmentOrdinal, SegmentReader};
 
 /// Contains a feature (field, score, etc.) of a document along with the document address.
@@ -86,32 +86,48 @@ where T: PartialOrd + Clone
 
     pub fn merge_fruits(
         &self,
-        children: Vec<Vec<(T, DocAddress)>>,
+        mut children: Vec<Vec<(T, DocAddress)>>,
     ) -> crate::Result<Vec<(T, DocAddress)>> {
         if self.limit == 0 {
             return Ok(Vec::new());
         }
-        let mut top_collector = BinaryHeap::new();
+        if children.len() == 1 {
+            let mut child = children.pop().unwrap();
+            child.sort_by(|a, b| {
+                if a.0 == b.0 {
+                    a.1.cmp(&b.1)
+                } else {
+                    b.0.partial_cmp(&a.0).unwrap_or(std::cmp::Ordering::Equal)
+                }
+            });
+            if self.offset != 0 {
+                return Ok(child
+                    .into_iter()
+                    .skip(self.offset)
+                    .take(self.limit)
+                    .collect());
+            }
+            if child.len() > self.limit {
+                child.truncate(self.limit);
+            }
+
+            return Ok(child);
+        }
+        let mut top_collector = TopNComputer::new(self.limit + self.offset);
         for child_fruit in children {
             for (feature, doc) in child_fruit {
-                if top_collector.len() < (self.limit + self.offset) {
-                    top_collector.push(ComparableDoc { feature, doc });
-                } else if let Some(mut head) = top_collector.peek_mut() {
-                    if head.feature < feature {
-                        *head = ComparableDoc { feature, doc };
-                    }
-                }
+                top_collector.push(ComparableDoc { feature, doc });
             }
         }
+
         Ok(top_collector
-            .into_sorted_vec()
-            .into_iter()
+            .into_iter_sorted()
             .skip(self.offset)
             .map(|cdoc| (cdoc.feature, cdoc.doc))
             .collect())
     }
 
-    pub(crate) fn for_segment<F: PartialOrd>(
+    pub(crate) fn for_segment<F: PartialOrd + Clone>(
         &self,
         segment_id: SegmentOrdinal,
         _: &SegmentReader,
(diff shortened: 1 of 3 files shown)


src/collector/top_collector.rs
```diff
@@ -86,32 +86,48 @@ where T: PartialOrd + Clone
...
             return Ok(Vec::new());
         }
-        let mut top_collector = BinaryHeap::new();
+        if children.len() == 1 {
+            let mut child = children.pop().unwrap();
+            child.sort_by(|a, b| {
+                if a.0 == b.0 {
+                    a.1.cmp(&b.1)
+                } else {
+                    b.0.partial_cmp(&a.0).unwrap_or(std::cmp::Ordering::Equal)
```
adamreichold: Isn't this redundant to the call of `PartialEq` above? Something like

```rust
match b.0.partial_cmp(&a.0) {
  Some(Ordering::Equal) => a.1.cmp(&b.1),
  Some(other) => other,
  None => Ordering::Equal,
}
```

Or is it important to definitely call `PartialEq` first and not rely on `PartialEq` and `PartialCmp` agreeing on equality?

src/collector/top_collector.rs
```diff
@@ -86,32 +86,48 @@ where T: PartialOrd + Clone
...
+                }
+            });
+            if self.offset != 0 {
+                return Ok(child
+                    .into_iter()
+                    .skip(self.offset)
+                    .take(self.limit)
+                    .collect());
+            }
+            if child.len() > self.limit {
```
adamreichold: I think `Vec::truncate` already does this check internally, so the call could be unconditional.

src/collector/top_score_collector.rs
```diff
@@ -86,12 +85,15 @@ where
 /// The `TopDocs` collector keeps track of the top `K` documents
 /// sorted by their score.
 ///
-/// The implementation is based on a `BinaryHeap`.
-/// The theoretical complexity for collecting the top `K` out of `n` documents
-/// is `O(n log K)`.
+/// The implementation is based on a repeatedly truncating on the median after K * 2 documents
```
adamreichold, line 88: So this basically trades having to store `2K` instead of `K` documents against less computational sorting work?

src/collector/top_score_collector.rs
```diff
@@ -736,6 +722,64 @@ impl SegmentCollector for TopScoreSegmentCollector {
...
+
+        median_score
+    }
+
+    pub(crate) fn into_iter_sorted(self) -> impl Iterator<Item = ComparableDoc<Score, DocId>> {
+        let mut sorted_buffer = self.buffer;
+        sorted_buffer.sort_unstable();
+
+        // Return the sorted top N elements
+        sorted_buffer.into_iter().take(self.max_size / 2)
```
adamreichold: Couldn't this use `sorted_buffer.capacity()` and thereby avoid storing `max_size` at all?

I also wonder if truncation before calling `.into_iter()` isn't preferable to usage of `.take(..)` as it simplifies the returned iterator, including returning `impl ExactSizeIterator<Item = ...>`.

src/collector/top_score_collector.rs
```diff
@@ -736,6 +722,64 @@ impl SegmentCollector for TopScoreSegmentCollector {
...
+        TopNComputer {
+            buffer: Vec::with_capacity(max_size),
+            max_size,
+            threshold: None,
+        }
+    }
+
+    #[inline]
+    pub(crate) fn push(&mut self, doc: ComparableDoc<Score, DocId>) {
+        if let Some(last_median) = self.threshold.clone() {
```
adamreichold, line 753: Isn't `Option<Score>: Copy`?

adamreichold: Ah, sorry `Score` is a type parameter, not the type alias.

adamreichold, line 753: If `new` were to take a `initial_threshold` argument to initialize `self.threshold` to e.g. `Score::MIN`, could the `Option` layer be avoided completely?

trinity-1686a:
if it makes that much of a difference, we'll probably want to do the same in quickwit's [topk module](https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-common/src/binary_heap.rs)

src/collector/top_score_collector.rs
trinity-1686a: `Vec::with_capacity` contract is
> The vector will be able to hold at least capacity elements without reallocating.

Current implementation seems to always give exactly the requested capacity (except for ZST), but it could in theory give more, in which case we could return more docs than initially requested

adamreichold: I think if the implementation is to be robust against that, the check line 753 should also use `max_size` instead of `capacity`, shouldn't it?

(It would lead to more than `2K` documents being tracked and hence the first `K` could be out of order w.r.t. the median at `len() / 2 == capacity() / 2 > K`? Or does sorting the whole buffer at the end take care of that?)

(Wouldn't it actually suffice to call `truncate_median` in `into_sorted_iter` and then sort only the remaining `K` elements generally (thereby also avoiding the `take` adaptor)?)

src/collector/top_score_collector.rs
PSeitz: `2K`, so we can drop elements when truncating the median

src/collector/top_collector.rs
PSeitz: I don't think that's important. I replaced that part reuse the comparison in `ComparableDoc`

src/collector/top_score_collector.rs
PSeitz: Having the `truncate_median` logic work on a little bit more capacity is no issue.

The check in line 753 did not work, so I replaced it with some `unsafe`

> (Wouldn't it actually suffice to call truncate_median in into_sorted_iter and then sort only the remaining K elements generally (thereby also avoiding the take adaptor)?)

Yes, but I don't think the performance is worth it the lower readability

src/collector/top_score_collector.rs
PSeitz: Yes, but it's already a generic without that function on the caller side

PSeitz:
> if it makes that much of a difference, we'll probably want to do the same in quickwit's [topk module](https://github.com/quickwit-oss/quickwit/blob/main/quickwit/quickwit-common/src/binary_heap.rs)

I made `TopNComputer` pub, so we can just call it from Quickwit

src/collector/top_score_collector.rs
adamreichold: I am sorry, but I just don't understand. I meant to have

```rust
pub struct TopNComputer<Score, DocId> {
    buffer: Vec<ComparableDoc<Score, DocId>>,
    top_n: usize,
    pub(crate) threshold: Score,
}
```

so that the `if let Some(threshold) = self.threshold.clone()` can be omitted, i.e. branching on the `Option` tag is removed.

src/collector/top_score_collector.rs
adamreichold: > Yes, but I don't think the performance is worth it the lower readability

To be honest, I am not sure that readability actually suffers:

```rust
pub(crate) fn into_iter_sorted(mut self) -> impl ExactSizeIterator<Item = ComparableDoc<Score, DocId>> {
    self.truncate_median();
    self.buffer.sort_unstable();
    self.buffer.into_iter()
}
```

src/collector/top_score_collector.rs
```diff
@@ -736,9 +722,81 @@ impl SegmentCollector for TopScoreSegmentCollector {
...
+            }
+        }
+        if self.buffer.len() == self.buffer.capacity() {
+            let median = self.truncate_median();
+            self.threshold = Some(median);
+        }
+
+        // This is faster since it avoids the buffer resizing to be inlined from vec.push()
+        // (this is in the hot path)
+        let uninit = self.buffer.spare_capacity_mut();
```
adamreichold, line 766: Maybe add a `FIXME` comment to use `push_within_capacity` when that stabilizes.

src/collector/top_score_collector.rs
PSeitz: `Score` is a generic, we don't have access to `Score::MIN`

src/collector/top_score_collector.rs
PSeitz: anyone proofreading would need to check `truncate_median` truncates exact to top n ... which it doesn't

src/collector/top_score_collector.rs
adamreichold: Not in `TopNComputer`, yes. But the caller of `TopNComputer::new` has access and hence could pass this value on like

```rust
let mut top_collector = TopNComputer::new(self.limit + self.offset, Score::MIN);
```

src/collector/top_score_collector.rs
adamreichold: > which it doesn't

If capacity is over-allocated, but since you went for `spare_capacity` anyway, you could actually enforce that by checking `self.buffer.len() == self.max_size` instead of `self.buffer.capacity()`.

I agree that this entangles this function with the invariants of `TopNComputer`, but I am not sure that `TopNCompuer` can be reviewed piece by piece in any case as the contents of `self.buffer` is really only meaningful here due to these invariants. 

That said, I would still suggest doing at least

```rust
pub(crate) fn into_iter_sorted(self) -> impl ExactSizeIterator<Item = ComparableDoc<Score, DocId>> {
        let mut sorted_buffer = self.buffer;
        sorted_buffer.sort_unstable();

        // Return the sorted top N elements
        sorted_buffer.truncate(self.top_n);
        sorted_buffer.into_iter()
}
```

to simplify the returned iterator.

src/collector/top_score_collector.rs
PSeitz: No, as I wrote before it's not available on the caller side, there's just `T`

src/collector/top_score_collector.rs
PSeitz: I don't think that simplifies the code or the returned Iterator

src/collector/top_score_collector.rs
adamreichold: Okay, then I this was what I did not parse from

> Yes, but it's already a generic without that function on the caller side

src/collector/top_score_collector.rs
adamreichold: It definitely avoids "injecting" the code for the `Take` adaptor into the caller, c.f. https://doc.rust-lang.org/stable/src/core/iter/adapters/take.rs.html#35

src/collector/top_score_collector.rs
```diff
@@ -661,27 +663,21 @@ impl Collector for TopDocs {
...
 
         if let Some(alive_bitset) = reader.alive_bitset() {
             let mut threshold = Score::MIN;
-            weight.for_each_pruning(threshold, reader, &mut |doc, score| {
+            top_n.threshold = Some(threshold);
+            weight.for_each_pruning(Score::MIN, reader, &mut |doc, score| {
                 if alive_bitset.is_deleted(doc) {
                     return threshold;
                 }
                 let heap_item = ComparableDoc {
```
fulmicoton: `heap_item` ? We need to update the naming here.

src/collector/top_score_collector.rs
```diff
@@ -736,9 +722,82 @@ impl SegmentCollector for TopScoreSegmentCollector {
...
+        unsafe {
+            self.buffer.set_len(self.buffer.len() + 1);
+        }
+    }
+
+    #[inline(never)]
+    fn truncate_median(&mut self) -> Score {
+        let len = self.buffer.len();
+        // Use select_nth_unstable to find the median score
+        let (_, median_el, _) = self.buffer.select_nth_unstable(len / 2);
```
fulmicoton: It is more solid to write it use `self.top_n` here I think.
```suggestion
        let (_, median_el, _) = self.buffer.select_nth_unstable(self.top_n);
```

src/collector/top_score_collector.rs
```diff
@@ -736,9 +722,82 @@ impl SegmentCollector for TopScoreSegmentCollector {
...
+
+        let median_score = median_el.feature.clone();
+        // Remove all elements below the median
+        self.buffer.truncate(len / 2);
```
fulmicoton:
```suggestion
        self.buffer.truncate(self.top_n);
```

src/collector/top_score_collector.rs
```diff
@@ -736,9 +722,82 @@ impl SegmentCollector for TopScoreSegmentCollector {
...
+
+        median_score
+    }
+
+    pub(crate) fn into_iter_sorted(self) -> impl Iterator<Item = ComparableDoc<Score, DocId>> {
+        let mut sorted_buffer = self.buffer;
```
fulmicoton: nitpick

```suggestion
        if self.buffer.len() > self.top_n {
            self.truncate_median();
        }
        let mut sorted_buffer = self.buffer;
```

Note this change requires to self.top_n change in truncate_median.

src/collector/top_score_collector.rs
```diff
@@ -736,9 +722,82 @@ impl SegmentCollector for TopScoreSegmentCollector {
...
+        let (_, median_el, _) = self.buffer.select_nth_unstable(len / 2);
+
+        let median_score = median_el.feature.clone();
+        // Remove all elements below the median
+        self.buffer.truncate(len / 2);
+
+        median_score
+    }
+
+    pub(crate) fn into_iter_sorted(self) -> impl Iterator<Item = ComparableDoc<Score, DocId>> {
```
fulmicoton: nitpick
I think we can return the Vec directly.

src/collector/top_score_collector.rs
```diff
@@ -736,9 +722,82 @@ impl SegmentCollector for TopScoreSegmentCollector {
...
+        }
+        if self.buffer.len() == self.buffer.capacity() {
+            let median = self.truncate_median();
+            self.threshold = Some(median);
+        }
+
+        // This is faster since it avoids the buffer resizing to be inlined from vec.push()
+        // (this is in the hot path)
+        // TODO: Replace with `push_within_capacity` when it's stabilized
+        let uninit = self.buffer.spare_capacity_mut();
```
fulmicoton, line 766: If we go this way, we might want to remove the main branch of this code? 
`if doc.feature < last_median`

We can do it by making by converting this bool into 1 or 0, and 

```
self.buffer.set_len(self.buffer.len() + inc_from_bool);
```

fulmicoton (changes requested):
See comments inline

src/collector/top_score_collector.rs
PSeitz: The `if doc.feature < last_median` early exit is quite important for performance I think

src/collector/top_score_collector.rs
```diff
@@ -736,9 +723,84 @@ impl SegmentCollector for TopScoreSegmentCollector {
...
+        uninit[0].write(doc);
+        // This is safe because it would panic in the line above
+        unsafe {
+            self.buffer.set_len(self.buffer.len() + 1);
+        }
+    }
+
+    #[inline(never)]
+    fn truncate_top_n(&mut self) -> Score {
+        let truncate_pos = self.top_n.min(self.buffer.len().saturating_sub(1));
```
fulmicoton: I think there is still an off-by-1 error here no?

```
self.top_n.min(self.buffer.len()).saturating_sub(1)
```
seems more correct?

Alternatively, this could also be a branch...

PSeitz: The logic was correct I think, either truncate top n or don't truncate if the buffer is smaller than top n

I could remove that special case, since it only appeared when called from `into_sorted_vec`

fulmicoton (approved):

PR #2192, opened 2023-09-26
chore: clarify the docs that atomic_read may read a previously not ex…

ethever:
Clarify the docs that the atomic_read function in `Directory` trait may read a previously not existing file.

### DIFF
--- a/src/directory/directory.rs
+++ b/src/directory/directory.rs
@@ -171,6 +171,7 @@ pub trait Directory: DirectoryClone + fmt::Debug + Send + Sync + 'static {
     /// This should only be used for small files.
     ///
     /// You should only use this to read files create with [`Directory::atomic_write()`].
+    /// The file may or may not previously exist.
     fn atomic_read(&self, path: &Path) -> Result<Vec<u8>, OpenReadError>;
 
     /// Atomically replace the content of a file with data.
@@ -234,7 +235,8 @@ pub trait DirectoryClone {
 }
 
 impl<T> DirectoryClone for T
-where T: 'static + Directory + Clone
+where
+    T: 'static + Directory + Clone,
 {
     fn box_clone(&self) -> Box<dyn Directory> {
         Box::new(self.clone())


ethever:
Add `println!` to those function:
```rust
fn atomic_read(&self, path: &Path) -> Result<Vec<u8>, OpenReadError> {
        println!("atomic_read: {path:?}");
        let bytes =
            self.open_read(path)?
                .read_bytes()
                .map_err(|io_error| OpenReadError::IoError {
                    io_error: Arc::new(io_error),
                    filepath: path.to_path_buf(),
                })?;
        Ok(bytes.as_slice().to_owned())
    }

    fn atomic_write(&self, path: &Path, data: &[u8]) -> io::Result<()> {
        println!("atomic_write: {path:?}");
        let path_buf = PathBuf::from(path);
        self.fs.write().unwrap().write(path_buf, data);
        if path == *META_FILEPATH {
            drop(self.fs.write().unwrap().watch_router.broadcast());
        }
        Ok(())
    }
```
Then we run the test in `index_writer`: 
```rust
#[test]
    fn test_operations_group() {
        // an operations group with 2 items should cause 3 opstamps 0, 1, and 2.
        let mut schema_builder = schema::Schema::builder();
        let text_field = schema_builder.add_text_field("text", TEXT);
        let index_writer = index.writer_for_tests().unwrap();
        let operations = vec![
            UserOperation::Add(doc!(text_field=>"a")),
            UserOperation::Add(doc!(text_field=>"b")),
        ];
        let batch_opstamp1 = index_writer.run(operations).unwrap();
        assert_eq!(batch_opstamp1, 2u64);
    }
```
with the output:
```shell
cargo test --lib  -- test_operations_group --nocapture
   Compiling tantivy v0.21.0 (/root/projects/tantivy)
    Finished test [unoptimized + debuginfo] target(s) in 12.48s
     Running unittests src/lib.rs (target/debug/deps/tantivy-7283131eb69df149)

running 1 test
atomic_read: ".managed.json"
atomic_write: ".managed.json"
atomic_write: "meta.json"
atomic_read: "meta.json"
atomic_read: "meta.json"
atomic_read: "meta.json"
atomic_write: ".managed.json"
atomic_write: ".managed.json"
atomic_write: ".managed.json"
atomic_write: ".managed.json"
atomic_write: ".managed.json"
atomic_write: ".managed.json"
test indexer::index_writer::tests::test_operations_group ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 813 filtered out; finished in 0.02s

```
We can see that the tantivy first use the `atomic_read` to read a previously not exisiting file `.managed.json`.
But this line(`Reads the full content file that has been written using`) of the comments of the `atomic_read` can easily be misunderstood.

fulmicoton:
I don't understand what you mean.

fulmicoton:
Can we be a little bit more accurate then and say that if the file does not exists this method returns `OpenReadError::FileDoesNotExist`?

ethever:
> Can we be a little bit more accurate then and say that if the file does not exists this method returns `OpenReadError::FileDoesNotExist`?

I apologize for the confusion in this...

PR #2198, opened 2023-10-02
feat(aggregators/metric): Add a top_hits aggregator

ditsuke:
## Summary

Implements the [`top_hits`][elastic-search-top-hits] aggregator. The aggregator is backed by ~~a BinaryHeap based on prior discussions on [Discord#tantivy-dev][tantivy-dev-top-hits-discussion]~~ the new `TopNComputer`.

[elastic-search-top-hits]: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-top-hits-aggregation.html
[tantivy-dev-top-hits-discussion]: https://discord.com/channels/908281611840282624/915785344396439552/1154687081180823612

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,6 @@ tokenizer-api = { version= "0.2", path="./tokenizer-api", package="tantivy-token
 sketches-ddsketch = { version = "0.2.1", features = ["use_serde"] }
 futures-util = { version = "0.3.28", optional = true }
 fnv = "1.0.7"
-serde-tuple-vec-map = "1.0.1"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.3.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ thiserror = "1.0.30"
 htmlescape = "0.3.1"
 fail = { version = "0.5.0", optional = true }
 murmurhash32 = "0.3.0"
-time = { version = "0.3.10", features = ["serde-well-known", "macros"] }
+time = { version = "0.3.10", features = ["serde-well-known"] }
 smallvec = "1.8.0"
 rayon = "1.5.2"
 lru = "0.12.0"
@@ -80,6 +80,7 @@ futures = "0.3.21"
 paste = "1.0.11"
 more-asserts = "0.3.1"
 rand_distr = "0.4.3"
+time = { version = "0.3.10", features = ["serde-well-known", "macros"] }
 
 [target.'cfg(not(windows))'.dev-dependencies]
 criterion = "0.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,7 @@ tokenizer-api = { version= "0.2", path="./tokenizer-api", package="tantivy-token
 sketches-ddsketch = { version = "0.2.1", features = ["use_serde"] }
 futures-util = { version = "0.3.28", optional = true }
 fnv = "1.0.7"
+serde-tuple-vec-map = "1.0.1"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.3.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ thiserror = "1.0.30"
 htmlescape = "0.3.1"
 fail = { version = "0.5.0", optional = true }
 murmurhash32 = "0.3.0"
-time = { version = "0.3.10", features = ["serde-well-known"] }
+time = { version = "0.3.10", features = ["serde-well-known", "macros"] }
 smallvec = "1.8.0"
 rayon = "1.5.2"
 lru = "0.12.0"
--- a/src/aggregation/agg_req.rs
+++ b/src/aggregation/agg_req.rs
@@ -35,7 +35,7 @@ use super::bucket::{
 };
 use super::metric::{
     AverageAggregation, CountAggregation, MaxAggregation, MinAggregation,
-    PercentilesAggregationReq, StatsAggregation, SumAggregation,
+    PercentilesAggregationReq, StatsAggregation, SumAggregation, TopHitsAggregation,
 };
 
 /// The top-level aggregation request structure, which contains [`Aggregation`] and their user
@@ -147,6 +147,9 @@ pub enum AggregationVariants {
     /// Computes the sum of the extracted values.
     #[serde(rename = "percentiles")]
     Percentiles(PercentilesAggregationReq),
+    /// Finds the top k values matching some order
+    #[serde(rename = "top_hits")]
+    TopHits(TopHitsAggregation),
 }
 
 impl AggregationVariants {
@@ -164,6 +167,9 @@ impl AggregationVariants {
             AggregationVariants::Stats(stats) => stats.field_name(),
             AggregationVariants::Sum(sum) => sum.field_name(),
             AggregationVariants::Percentiles(per) => per.field_name(),
+            // FIXME: a single fast field doesn't make sense for top hits
+            // What should we do here?
+            AggregationVariants::TopHits(top_hits) => top_hits.field_name(),
         }
     }
 
--- a/src/aggregation/agg_req.rs
+++ b/src/aggregation/agg_req.rs
@@ -158,7 +158,7 @@ pub enum AggregationVariants {
 }
 
 impl AggregationVariants {
-    /// Returns the name of the field used by the aggregation.
+    /// Returns the name of the field(s) used by the aggregation.
     pub fn get_fast_field_names(&self) -> Vec<&str> {
         match self {
             AggregationVariants::Terms(terms) => vec![terms.field.as_str()],
--- a/src/aggregation/agg_req.rs
+++ b/src/aggregation/agg_req.rs
@@ -158,7 +158,7 @@ pub enum AggregationVariants {
 }
 
 impl AggregationVariants {
-    /// Returns the name of the field(s) used by the aggregation.
+    /// Returns the name of the fields used by the aggregation.
     pub fn get_fast_field_names(&self) -> Vec<&str> {
         match self {
             AggregationVariants::Terms(terms) => vec![terms.field.as_str()],
--- a/src/aggregation/agg_req.rs
+++ b/src/aggregation/agg_req.rs
@@ -93,7 +93,12 @@ impl Aggregation {
     }
 
     fn get_fast_field_names(&self, fast_field_names: &mut HashSet<String>) {
-        fast_field_names.insert(self.agg.get_fast_field_name().to_string());
+        fast_field_names.extend(
+            self.agg
+                .get_fast_field_names()
+                .iter()
+                .map(|s| s.to_string()),
+        );
         fast_field_names.extend(get_fast_field_names(&self.sub_aggregation));
     }
 }
@@ -154,22 +159,20 @@ pub enum AggregationVariants {
 
 impl AggregationVariants {
     /// Returns the name of the field used by the aggregation.
-    pub fn get_fast_field_name(&self) -> &str {
+    pub fn get_fast_field_names(&self) -> Vec<&str> {
         match self {
-            AggregationVariants::Terms(terms) => terms.field.as_str(),
-            AggregationVariants::Range(range) => range.field.as_str(),
-            AggregationVariants::Histogram(histogram) => histogram.field.as_str(),
-            AggregationVariants::DateHistogram(histogram) => histogram.field.as_str(),
-            AggregationVariants::Average(avg) => avg.field_name(),
-            AggregationVariants::Count(count) => count.field_name(),
-            AggregationVariants::Max(max) => max.field_name(),
-            AggregationVariants::Min(min) => min.field_name(),
-            AggregationVariants::Stats(stats) => stats.field_name(),
-            AggregationVariants::Sum(sum) => sum.field_name(),
-            AggregationVariants::Percentiles(per) => per.field_name(),
-            // FIXME: a single fast field doesn't make sense for top hits
-            // What should we do here?
-            AggregationVariants::TopHits(top_hits) => top_hits.field_name(),
+            AggregationVariants::Terms(terms) => vec![terms.field.as_str()],
+            AggregationVariants::Range(range) => vec![range.field.as_str()],
+            AggregationVariants::Histogram(histogram) => vec![histogram.field.as_str()],
+            AggregationVariants::DateHistogram(histogram) => vec![histogram.field.as_str()],
+            AggregationVariants::Average(avg) => vec![avg.field_name()],
+            AggregationVariants::Count(count) => vec![count.field_name()],
+            AggregationVariants::Max(max) => vec![max.field_name()],
+            AggregationVariants::Min(min) => vec![min.field_name()],
+            AggregationVariants::Stats(stats) => vec![stats.field_name()],
+            AggregationVariants::Sum(sum) => vec![sum.field_name()],
+            AggregationVariants::Percentiles(per) => vec![per.field_name()],
+            AggregationVariants::TopHits(top_hits) => top_hits.field_names(),
         }
     }
 
--- a/src/aggregation/agg_req_with_accessor.rs
+++ b/src/aggregation/agg_req_with_accessor.rs
@@ -53,7 +53,6 @@ pub struct AggregationWithAccessor {
     /// If this needs to used by other aggregations, we need to refactor this.
     // NOTE: we can make all other aggregations use this instead of the `accessor` and `field_type`
     // (making them obsolete) But will it have a performance impact?
-    // pub(crate) accessors: HashMap<String, (Column<u64>, ColumnType)>,
     pub(crate) accessors: Vec<(Column<u64>, ColumnType)>,
     pub(crate) value_accessors: HashMap<String, DynamicColumn>,
     pub(crate) agg: Aggregation,
--- a/src/aggregation/agg_req_with_accessor.rs
+++ b/src/aggregation/agg_req_with_accessor.rs
@@ -54,6 +54,8 @@ pub struct AggregationWithAccessor {
     // NOTE: we can make all other aggregations use this instead of the `accessor` and `field_type`
     // (making them obsolete) But will it have a performance impact?
     pub(crate) accessors: Vec<(Column<u64>, ColumnType)>,
+    /// Map field names to dynamic column accessors.
+    /// This field is used for `docvalue_fields`, which is currently only supported for `top_hits`.
     pub(crate) value_accessors: HashMap<String, DynamicColumn>,
     pub(crate) agg: Aggregation,
 }
@@ -304,7 +306,10 @@ impl AggregationWithAccessor {
                     .value_field_names()
                     .iter()
                     .map(|field_name| {
-                        Ok((field_name.to_string(), get_dyn_column(reader, field_name)?))
+                        Ok((
+                            field_name.to_string(),
+                            get_dynamic_column(reader, field_name)?,
+                        ))
                     })
                     .collect::<crate::Result<_>>()?;
 
@@ -389,7 +394,7 @@ fn get_ff_reader(
     Ok(ff_field_with_type)
 }
 
-fn get_dyn_column(
+fn get_dynamic_column(
     reader: &SegmentReader,
     field_name: &str,
 ) -> crate::Result<columnar::DynamicColumn> {
--- a/src/aggregation/agg_req_with_accessor.rs
+++ b/src/aggregation/agg_req_with_accessor.rs
@@ -196,7 +196,7 @@ impl AggregationWithAccessor {
                         .map(|m| matches!(m, Key::Str(_)))
                         .unwrap_or(false);
 
-                // Actually we could convert the text to a number and have the fast path, if it
+                // Actually we could convert the text to a number and have the fast path, if it is
                 // provided in Rfc3339 format. But this use case is probably common
                 // enough to justify the effort.
                 let text_on_date_col = column_and_types.len() == 1
--- a/src/aggregation/agg_req_with_accessor.rs
+++ b/src/aggregation/agg_req_with_accessor.rs
@@ -16,7 +16,7 @@ use super::metric::{
 use super::segment_agg_result::AggregationLimits;
 use super::VecWithNames;
 use crate::aggregation::{f64_to_fastfield_u64, Key};
-use crate::{SegmentOrdinal, SegmentReader};
+use crate::{SegmentOrdinal, SegmentReader, TantivyError};
 
 #[derive(Default)]
 pub(crate) struct AggregationsWithAccessor {
@@ -210,7 +210,7 @@ impl AggregationWithAccessor {
 
                     let accessors = column_and_types
                         .iter()
-                        .map(|c_t| (c_t.0.clone(), c_t.1.clone()))
+                        .map(|c_t| (c_t.0.clone(), c_t.1))
                         .collect();
                     add_agg_with_accessors(accessors, &mut res, Default::default(), None)?;
                 }
@@ -282,23 +282,22 @@ impl AggregationWithAccessor {
                 add_agg_with_accessor(accessor, column_type, &mut res)?;
             }
             TopHits(top_hits) => {
-                let agg = {
+                let (agg, top_hits) = {
                     let top_hits =
                         top_hits.validate_and_resolve(reader.fast_fields().columnar())?;
-                    Aggregation {
-                        agg: AggregationVariants::TopHits(top_hits),
-                        sub_aggregation: agg.sub_aggregation().clone(),
-                    }
+                    (
+                        Aggregation {
+                            agg: AggregationVariants::TopHits(top_hits.clone()),
+                            sub_aggregation: agg.sub_aggregation().clone(),
+                        },
+                        top_hits,
+                    )
                 };
                 let accessors: Vec<(Column<u64>, ColumnType)> = top_hits
                     .field_names()
                     .iter()
                     .map(|field| {
-                        Ok(get_ff_reader(
-                            reader,
-                            field,
-                            Some(get_numeric_or_date_column_types()),
-                        )?)
+                        get_ff_reader(reader, field, Some(get_numeric_or_date_column_types()))
                     })
                     .collect::<crate::Result<_>>()?;
 
@@ -399,7 +398,9 @@ fn get_dyn_column(
     Ok(ff_fields
         .dynamic_column_handles(field_name)?
         .get(0)
-        .expect("column must exist")
+        .ok_or(TantivyError::FieldNotFound(format!(
+            "field {field_name} not found"
+        )))?
         .open()?)
 }
 
--- a/src/aggregation/agg_req_with_accessor.rs
+++ b/src/aggregation/agg_req_with_accessor.rs
@@ -207,12 +207,13 @@ impl AggregationWithAccessor {
                         missing.clone()
                     };
 
-                    let missing_value_for_accessor =
-                        if let Some(missing) = missing_value_term_agg.as_ref() {
-                            get_missing_val(column_type, missing, agg.agg.get_fast_field_name())?
-                        } else {
-                            None
-                        };
+                    let missing_value_for_accessor = if let Some(missing) =
+                        missing_value_term_agg.as_ref()
+                    {
+                        get_missing_val(column_type, missing, agg.agg.get_fast_field_names()[0])?
+                    } else {
+                        None
+                    };
 
                     let agg = AggregationWithAccessor {
                         segment_id,
@@ -266,7 +267,7 @@ impl AggregationWithAccessor {
             }
             TopHits(top_hits) => {
                 let accessors: Vec<(Column<u64>, ColumnType)> = top_hits
-                    .get_fields()
+                    .field_names()
                     .iter()
                     .map(|field| {
                         get_ff_reader(reader, field, Some(get_numeric_or_date_column_types()))
--- a/src/aggregation/agg_req_with_accessor.rs
+++ b/src/aggregation/agg_req_with_accessor.rs
@@ -14,7 +14,7 @@ use super::metric::{
 use super::segment_agg_result::AggregationLimits;
 use super::VecWithNames;
 use crate::aggregation::{f64_to_fastfield_u64, Key};
-use crate::SegmentReader;
+use crate::{SegmentOrdinal, SegmentReader};
 
 #[derive(Default)]
 pub(crate) struct AggregationsWithAccessor {
@@ -32,6 +32,7 @@ impl AggregationsWithAccessor {
 }
 
 pub struct AggregationWithAccessor {
+    pub(crate) segment_id: SegmentOrdinal,
     /// In general there can be buckets without fast field access, e.g. buckets that are created
     /// based on search terms. That is not that case currently, but eventually this needs to be
     /// Option or moved.
@@ -44,10 +45,13 @@ pub struct AggregationWithAccessor {
     pub(crate) limits: ResourceLimitGuard,
     pub(crate) column_block_accessor: ColumnBlockAccessor<u64>,
     /// Used for missing term aggregation, which checks all columns for existence.
+    /// And also for `top_hits` aggregation, which may sort on multiple fields.
     /// By convention the missing aggregation is chosen, when this property is set
     /// (instead bein set in `agg`).
     /// If this needs to used by other aggregations, we need to refactor this.
-    pub(crate) accessors: Vec<Column<u64>>,
+    // NOTE: we can make all other aggregations use this instead of the `accessor` and `field_type`
+    // (making them obsolete) But will it have a performance impact?
+    pub(crate) accessors: Vec<(Column<u64>, ColumnType)>,
     pub(crate) agg: Aggregation,
 }
 
@@ -57,6 +61,7 @@ impl AggregationWithAccessor {
         agg: &Aggregation,
         sub_aggregation: &Aggregations,
         reader: &SegmentReader,
+        segment_id: SegmentOrdinal,
         limits: AggregationLimits,
     ) -> crate::Result<Vec<AggregationWithAccessor>> {
         let add_agg_with_accessor = |accessor: Column<u64>,
@@ -64,12 +69,38 @@ impl AggregationWithAccessor {
                                      aggs: &mut Vec<AggregationWithAccessor>|
          -> crate::Result<()> {
             let res = AggregationWithAccessor {
+                segment_id,
                 accessor,
                 accessors: Vec::new(),
                 field_type: column_type,
                 sub_aggregation: get_aggs_with_segment_accessor_and_validate(
                     sub_aggregation,
                     reader,
+                    segment_id,
+                    &limits,
+                )?,
+                agg: agg.clone(),
+                limits: limits.new_guard(),
+                missing_value_for_accessor: None,
+                str_dict_column: None,
+                column_block_accessor: Default::default(),
+            };
+            aggs.push(res);
+            Ok(())
+        };
+
+        let add_agg_with_accessors = |accessors: Vec<(Column<u64>, ColumnType)>,
+                                      aggs: &mut Vec<AggregationWithAccessor>|
+         -> crate::Result<()> {
+            let res = AggregationWithAccessor {
+                segment_id,
+                accessor: accessors[0].0.clone(),
+                field_type: accessors[0].1,
+                accessors,
+                sub_aggregation: get_aggs_with_segment_accessor_and_validate(
+                    sub_aggregation,
+                    reader,
+                    segment_id,
                     &limits,
                 )?,
                 agg: agg.clone(),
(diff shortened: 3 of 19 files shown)


adamreichold:
> The aggregator is backed by a `BinaryHeap` based on prior discussions

Did those discussions already consider the recently introduced [`TopNComputer`](https://github.com/quickwit-oss/tantivy/blob/82d91271910e622b86b26158158e03c5435aef60/src/collector/top_score_collector.rs#L729) which might be a better choice?

src/collector/top_score_collector.rs
```diff
@@ -718,7 +718,7 @@ impl SegmentCollector for TopScoreSegmentCollector {
         self.0.collect(doc, score);
     }
 
-    fn harvest(self) -> Vec<(Score, DocAddress)> {
```
fulmicoton: this is not helpful

Cargo.toml
```diff
@@ -63,6 +63,7 @@ common = { version= "0.6", path = "./common/", package = "tantivy-common" }
 tokenizer-api = { version= "0.2", path="./tokenizer-api", package="tantivy-tokenizer-api" }
 sketches-ddsketch = { version = "0.2.1", features = ["use_serde"] }
 futures-util = { version = "0.3.28", optional = true }
+serde-tuple-vec-map = "1.0.1"
```
fulmicoton: can we live without the extra deps?

fulmicoton:
Hello @ditsuke, is this something you actually have a need for? Can you describe the use case?

src/collector/top_score_collector.rs
ditsuke: Whoops sorry about that. Will revert in a later commit

Cargo.toml
ditsuke: For sure, on it.

ditsuke:
> > The aggregator is backed by a `BinaryHeap` based on prior discussions
> 
> Did those discussions already consider the recently introduced [`TopNComputer`](https://github.com/quickwit-oss/tantivy/blob/82d91271910e622b86b26158158e03c5435aef60/src/collector/top_score_collector.rs#L729) which might be a better choice?

Yes, the idea behind the new algorithm came up in that discussion thanks to @fulmicoton. I didn't use that here since it wasn't found suitable for pagination, but I'm happy to reconsider if that's wrong

ditsuke:
> Hello @ditsuke, is this something you actually have a need for? Can you describe the use case?

Hi @fulmicoton, we discussed the use-case on #tantivy-help some weeks back. Our use-case is querying for the top docs in each bucket (really the most recent post by some source).

For [reference](https://discord.com/channels/908281611840282624/908286403086024724/1151390584167206933).

adamreichold:
> I didn't use that here since it wasn't found suitable for pagination, but I'm happy to reconsider if that's wrong

I might be misunderstanding things since I wasn't part of that discussion, but my understanding was that the stable sort order needed for pagination is not so much a question of `BinaryHeap` versus `TopNComputer`, but rather whether ties are resolved in a stable manner by comparing document addresses when sorting the final result set?

ditsuke:
> > I didn't use that here since it wasn't found suitable for pagination, but I'm happy to reconsider if that's wrong
> 
> I might be misunderstanding things since I wasn't part of that discussion, but my understanding was that the stable sort order needed for pagination is not so much a question of `BinaryHeap` versus `TopNComputer`, but rather whether ties are resolved in a stable manner by comparing document addresses when sorting the final result set? 

You're actually right! My initial impression was that the elimination step would be unstable for distributions with > n/2 elements ≤ median (so when you can potentially eliminate more than n/2 elements and have to make a choice to retain some of the conflicting values, given that we need to cap the elimination to n/2 elements). But in fact this isn't a problem anymore with our comparator falling back to the doc address on conflicts, so we should be able to use `TopNComputer` here safely.

Cargo.toml
ditsuke: And done

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,350 @@
...
+/// ```
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
+pub struct TopHitsAggregation {
+    sort: Vec<KeyOrder>,
+    size: usize,
+    from: Option<usize>,
+}
+
+#[derive(Debug, Clone, PartialEq, Default)]
+struct KeyOrder(String, Order);
```
PSeitz: can't we use Derive here?

src/aggregation/agg_req.rs
```diff
@@ -164,6 +167,9 @@ impl AggregationVariants {
             AggregationVariants::Stats(stats) => stats.field_name(),
             AggregationVariants::Sum(sum) => sum.field_name(),
             AggregationVariants::Percentiles(per) => per.field_name(),
+            // FIXME: a single fast field doesn't make sense for top hits
+            // What should we do here?
+            AggregationVariants::TopHits(top_hits) => top_hits.field_name(),
```
PSeitz: Return a Vec with multiple field names

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,350 @@
...
+        let features: Vec<ComparableDocFeature> = agg_with_accessor.aggs.values[self.accessor_idx]
+            .accessors
+            .iter()
+            .zip(self.inner_collector.req.sort.iter())
+            .map(|((c, _), KeyOrder(_, order))| {
+                let order = *order;
+                match c.values_for_doc(doc_id).next() {
+                    Some(value) => ComparableDocFeature { value, order },
+                    // TODO: confirm if this default 0-value is correct
+                    None => ComparableDocFeature { value: 0, order },
```
PSeitz: This mixes docs with value 0 and non-existence which will be wrong in some cases

ditsuke: How do you suggest handling missing fields then? Perhaps with a None and rank them below docs with extant values?

src/aggregation/metric/top_hits.rs
ditsuke: Is there built-in derive to map `(f, o)` => `{"f": "o"}` or `{field: f, order: o}` => `{"f": "o"}`?

src/lib.rs
```diff
@@ -290,7 +290,7 @@ impl DocAddress {
 ///
 /// The id used for the segment is actually an ordinal
 /// in the list of `Segment`s held by a `Searcher`.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
fulmicoton, line 339: Do we need serialize for `DocAddress`?

fulmicoton:
This PR is starting to look pretty good! Thank you @ditsuke.
Can you add unit tests?
@PSeitz can you take care with the rest of review?

src/aggregation/metric/top_hits.rs
PSeitz: Yes, including `None` and scoring it lower than anything with a value would be the right approach

src/aggregation/metric/top_hits.rs
PSeitz: Ah I see, custom serialization is probably the best way to do this.

Can you change it into a struct with field names?

PSeitz:
## Fetching Docs
One tricky part that is missing is fetching the actual content of the document. So far aggregations are limited to fast fields due to the way they operate.
Question is when to fetch documents and from which data source.

## Document's data source
Generally they are two data-sources for a documents data in top hits: The doc store and [fast fields](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html#docvalue-fields).

Doc store access is relatively expensive, fast fields (doc values) access are cheap.

Fast field terms may be limited to a certain length, so some long texts may be missing there.

## When to Fetch
Aggregation works roughly like that

Segment Collection => Intermediate Result
Segment Collection => Intermediate Result
Segment Collection => Intermediate Result
... Could be 100 of Intermediate Result

Intermediate results can be de/serialized, merged and converted to a final result. 

Fetching a documents data could happen when converting to a intermediate result or final result.

When the data-source is a fast field, fetching at the intermediate result step will be fine.

### Fetch at Intermediate Result
This will cause some overhead as, we will fetch e.g. Top 10 for each Intermediate Result, but after merging 100 Intermediate Result only keep the Top 10. So we would fetch 10_000 documents. From the doc store that would be expensive. From the fast field this should be fine (except maybe very long texts).

### Fetch at Final Result
This would require passing additionally metadata like the segment id. In distributed scenarios like quickwit this may require more additional metadata to be able to resolve at the end, as final result conversion may happen at a different node.

## Conclusion
These approaches are quite different and require a different approach. Managing both would be possible but definitely adds some complexity. 

The main problem I see with fetching the final result is that is requires to have access to all the `Index` at the node that assembles the final result, but I think in that kind of aggregation most users are interested in only one, two few fields anyways not the whole doc.

We could limit the aggregation to only handle them with the [docvalue fields](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html#docvalue-fields) parameter.

I'm not sure which variant is the best approach, it very much depends on user queries. Supporting only `docvalue fields` is the simplest variant, so we could just start with that and change it to fetch at the final result if we run into issues.

adamreichold:
> Supporting only docvalue fields is the simplest variant, so we could just start with that and change it to fetch at the final result if we run into issues.

I think this sounds like the most reasonable approach from a risk management and incremental feature development perspective.

src/lib.rs
ditsuke: At the moment yes, since the aggregators return `DocAddress` and associated scores. Is there another way to go about it?

ditsuke:
> This PR is starting to look pretty good! Thank you @ditsuke. Can you add unit tests?

Thank you! I'm working on the tests

ditsuke:
> > Supporting only docvalue fields is the simplest variant, so we could just start with that and change it to fetch at the final result if we run into issues.
> 
> I think this sounds like the most reasonable approach from a risk management and incremental feature development perspective.

Agreed, that sounds reasonable. @PSeitz do we handle the docvalue fields support in this PR or with a follow-up?

src/aggregation/metric/top_hits.rs
ditsuke: Done with [74654bf](https://github.com/quickwit-oss/tantivy/pull/2198/commits/74654bf48b7f61793aebaaacedcafc03f7c78c9d)

PSeitz:
> 
> Agreed, that sounds reasonable. @PSeitz do we handle the docvalue fields support in this PR or with a follow-up?

It should be in this PR, or we can't add tests for the aggregation.

src/aggregation/metric/top_hits.rs
ditsuke: Looks much better now [7e507be](https://github.com/quickwit-oss/tantivy/pull/2198/commits/7e507bea6b7a9a7cb3a6de7c24e74aed727443c4)

ditsuke:
@PSeitz rough dig at docvalue_fields in [a26a353](https://github.com/quickwit-oss/tantivy/pull/2198/commits/a26a3535a2275f88dcda679886385b79cdc415a9). Could you have a look and validate if its heading in the right direction?

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -51,7 +53,8 @@ pub struct AggregationWithAccessor {
     /// If this needs to used by other aggregations, we need to refactor this.
     // NOTE: we can make all other aggregations use this instead of the `accessor` and `field_type`
     // (making them obsolete) But will it have a performance impact?
-    pub(crate) accessors: Vec<(Column<u64>, ColumnType)>,
+    pub(crate) accessors: HashMap<String, (Column<u64>, ColumnType)>,
```
ditsuke: Because we need named access to fields. Alternatively we could leave the `accessors` as a vec and have a new `HashMap<String, ValueAccessor>` that could be used for the docvalue query. I think the second way can be useful because the query will otherwise need access to the segment reader to query term_id => Str/byte. More:
https://github.com/quickwit-oss/tantivy/pull/2198/commits/a26a3535a2275f88dcda679886385b79cdc415a9#diff-da09c2e16f884b4f3436a59897de837ff6b5b49003f54ed3e7d9e6687d7229d6R135-R146

PSeitz: Why do we need the field name?

ditsuke: - To map field name in docvalue_fields to a (accessor, type) pair
- Since we have 2 places — the aggregation itself and the field query, where we need the accessors, relying solely on order is not an option unless we have separate accessor vectors for each case.

Is there an alternative solution I might be missing?

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,827 @@
...
+        assert_eq!(
+            agg_res
+                .0
+                .get("top_hits_req")
+                .expect("top_hits_req must exist"),
+            &AggregationResult::MetricResult(MetricResult::TopHits(TopHitsMetricResult {
+                hits: vec![
+                    TopHitsVecEntry {
+                        id: crate::DocAddress {
+                            segment_ord: 0,
```
ditsuke: Here relying on the segment ordinal is causing a flake in the tests. I'm not sure why this is, since `get_test_index_from_docs` seems to commit segments in order of declaration. Any clues?

src/aggregation/agg_req_with_accessor.rs
ditsuke: I created a new accessor field to hold `DynamicColumn` instead

ditsuke:
Rebased to main

ditsuke:
This PR is ready for overall review. Expecting CI to pass this time

ditsuke:
The failing test stems from flake as discussed [here](https://github.com/quickwit-oss/tantivy/pull/2198#discussion_r1365840050). I can use help in identifying why the segments aren't committed in order.

src/aggregation/metric/top_hits.rs
PSeitz: The merge_segments flag merges the segments,so `segment_ord` depends on that flag

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,848 @@
...
+        let date_2016 = datetime!(2016-01-02 00:00:00 UTC);
+
+        assert_eq!(
+            agg_res
+                .0
+                .get("top_hits_req")
+                .expect("top_hits_req must exist"),
+            &AggregationResult::MetricResult(MetricResult::TopHits(TopHitsMetricResult {
+                hits: vec![
+                    TopHitsVecEntry {
```
PSeitz: This is hard to read. Can you assert on the json format instead?

similar to:
```rust
        let res = exec_request_with_query(agg_req, &index, None)?;

        assert_eq!(
            res["my_stats"],
            json!({
                "avg":  2.5,
                "count": 4,
                "max": 10.0,
                "min": 0.0,
                "sum": 10.0
            })
        );

```

ditsuke: done, much better!

src/aggregation/metric/top_hits.rs
ditsuke: I've already accounted for that condition. Now it appears that the ordinal of the segments is non-determinate from their declaration order in the initialization vector: so the segment declared [first](https://github.com/quickwit-oss/tantivy/pull/2198/files#diff-da09c2e16f884b4f3436a59897de837ff6b5b49003f54ed3e7d9e6687d7229d6R731-R734) might have ord `1` instead of the expected `0` when `merge_segments` is `false`. My initial hypothesis was a race condition between segment writers but that doesn't seem to be the case here.

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,813 @@
...
+        let date_2017 = datetime!(2017-06-15 00:00:00 UTC);
+        let date_2016 = datetime!(2016-01-02 00:00:00 UTC);
+
+        assert_eq!(
+            agg_res["top_hits_req"],
+            json!({
+                "hits": [
+                    {
+                        "id": {
+                            "segment_ord": 0,
```
PSeitz: why do you return the segment and docid here?

ditsuke: Is there something else we can use to uniquely identify/retrieve documents?

PSeitz: I mean do we do something with it or is it just for the user?

ditsuke: Ah, yes it's for the user

PSeitz: I think in the context of a searcher this is valid information, but outside to the user not so much (e.g. the segment ord may be deleted from a merge, or there are multiple indices)

ditsuke: Makes sense. In the standard use-case for term-bucketing and retrieving top-x documents from each bucket, this information should be valid and useful, for example when the user wants to retrieve the entire document rather than a set of fast fields from each. On the other hand we should make it explicit that the identifier isn't useful outside of that context perhaps with a leading `_`, how would you like it to be done?

PSeitz: I think I would remove it and if there is some demand add more search field features like `docvalue fields`

ditsuke: Our current use-case needs access to a large in-store text field, so it would be problematic if it is removed without an alternative like `fields` :/

Perhaps we can do a workaround where we exclude the DocAddress from the serialized form?

Alternatively I could also work on `fields` support with some help, but preferably in a follow-up PR sometime next month (exam time at uni!)

PSeitz: Can you make it a fast field or would that be too expensive?
Otherwise I would remove it from the serialized final result, I think

ditsuke: I'll evaluate making it a fast field before we proceed. Generally speaking, would  you draw the line considering 10,000-100,000 documents where this field can be 20-50KB? I'm not very familiar with how the term dictionary stores, loads and queries at the moment so I'm curious if there's an approach to estimating the memory expense.

PSeitz: The term fastfield uses an sstable with compressed blocks. The blocks are relatively small, that means may we get many blocks. On the top of my head I'm not sure if that may cause problems, maybe @trinity-1686a knows more

trinity-1686a: iirc, when building a block, only the size of the key portion of the table is taken into account for sizing blocks, not the value part. If I understand how fastfields uses the sstable, (key=docid, value=the entire field value, correct me if I'm mistaken), it means you'll get unusually large blocks, but they will each contain as many docs as they would if the values were smaller.

If I understand correctly everything, assuming doc_ids are sequential, we can store about 2k docs in a block, for an upper bound a little under 100MiB of raw data. So I'd say it would use in the order of 100MiB per loaded sstable at a given instant

ditsuke: Thank you, that was helpful! I have removed DocAddress from top hits' results.

src/aggregation/metric/top_hits.rs
ditsuke: Resolving since top_hits' tests don't have reliance on addresses anymore, but this is still potentially an issue for other tests in the future.

Cargo.toml
```diff
@@ -44,7 +44,7 @@ thiserror = "1.0.30"
 htmlescape = "0.3.1"
 fail = { version = "0.5.0", optional = true }
 murmurhash32 = "0.3.0"
-time = { version = "0.3.10", features = ["serde-well-known"] }
+time = { version = "0.3.10", features = ["serde-well-known", "macros"] }
```
PSeitz: why do do we need macros?

ditsuke: I'm using the time! macro in the tests

src/aggregation/agg_req.rs
```diff
@@ -147,23 +152,27 @@ pub enum AggregationVariants {
...
     Percentiles(PercentilesAggregationReq),
+    /// Finds the top k values matching some order
+    #[serde(rename = "top_hits")]
+    TopHits(TopHitsAggregation),
 }
 
 impl AggregationVariants {
-    /// Returns the name of the field used by the aggregation.
-    pub fn get_fast_field_name(&self) -> &str {
+    /// Returns the name of the field(s) used by the aggregation.
```
PSeitz:
```suggestion
    /// Returns the name of the fields used by the aggregation.
```

Cargo.toml
PSeitz: No test only dependencies as regular dependencies please

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -321,6 +389,20 @@ fn get_ff_reader(
     Ok(ff_field_with_type)
 }
 
+fn get_dyn_column(
```
PSeitz:
```suggestion
fn get_dynamic_column(
```

dyn is a already taken word in rust

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -44,10 +47,14 @@ pub struct AggregationWithAccessor {
...
     /// Used for missing term aggregation, which checks all columns for existence.
+    /// And also for `top_hits` aggregation, which may sort on multiple fields.
     /// By convention the missing aggregation is chosen, when this property is set
     /// (instead bein set in `agg`).
     /// If this needs to used by other aggregations, we need to refactor this.
-    pub(crate) accessors: Vec<Column<u64>>,
+    // NOTE: we can make all other aggregations use this instead of the `accessor` and `field_type`
+    // (making them obsolete) But will it have a performance impact?
+    pub(crate) accessors: Vec<(Column<u64>, ColumnType)>,
+    pub(crate) value_accessors: HashMap<String, DynamicColumn>,
```
PSeitz: This is missing a comment on what is String, and what's it's used for

Cargo.toml
ditsuke: Fair, moved the feature to dev-dependencies

src/aggregation/agg_req_with_accessor.rs
ditsuke: Done

PSeitz:
Failing test is a rust nightly bug

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,820 @@
...
+    fn get_fields(
+        &self,
+        accessors: &HashMap<String, DynamicColumn>,
+        doc_id: DocId,
+    ) -> FieldRetrivalResult {
+        let dvf = self
+            .doc_value_fields
+            .iter()
+            .map(|field| {
+                let accessor = accessors.get(field).expect("field accessor must exist");
```
PSeitz:
```suggestion
                let accessor = accessors.get(field).expect("could not find field in accessors");
```

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,820 @@
...
+                        .next()
+                        .map_or(OwnedValue::Null, OwnedValue::F64),
+                    DynamicColumn::Bytes(accessor) => 'l: {
+                        let Some(term_ord) = accessor.term_ords(doc_id).next() else {
+                            break 'l OwnedValue::Null;
+                        };
+                        let mut buffer = vec![];
+                        match accessor
+                            .ord_to_bytes(term_ord, &mut buffer)
+                            .expect("term dictionary read must succeed")
```
PSeitz:
```suggestion
                            .expect("could not find term_ord in term dictionary")
```

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,820 @@
...
+                    DynamicColumn::Str(accessor) => 'l: {
+                        let Some(term_ord) = accessor.term_ords(doc_id).next() else {
+                            break 'l OwnedValue::Null;
+                        };
+                        let mut buffer = vec![];
+                        match accessor
+                            .ord_to_bytes(term_ord, &mut buffer)
+                            .expect("term dictionary read must succeed")
+                        {
+                            false => OwnedValue::Null,
```
PSeitz: I think all stored term ords do exist

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,820 @@
...
+    pub doc_value_fields: HashMap<String, OwnedValue>,
+}
+
+impl RetrievalFields {
+    fn get_field_names(&self) -> Vec<&str> {
+        self.doc_value_fields.iter().map(|s| s.as_str()).collect()
+    }
+
+    fn resolve_field_names(self, reader: &ColumnarReader) -> crate::Result<Self> {
+        let globbed_string_to_regex = |globbed: &str| {
```
PSeitz: Can you add a comment on the regex and it's purpose here?

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,820 @@
...
+        let dvf = self
+            .doc_value_fields
+            .iter()
+            .map(|field| {
+                let accessor = accessors.get(field).expect("field accessor must exist");
+
+                let value: OwnedValue = match accessor {
+                    DynamicColumn::U64(accessor) => accessor
+                        .values_for_doc(doc_id)
+                        .next()
```
PSeitz: We should fetch all the data, not just the first one

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,820 @@
...
+            .into_iter()
+            .flatten()
+            .collect();
+
+        Ok(Self {
+            doc_value_fields: resolved_fields,
+        })
+    }
+
+    fn get_fields(
```
PSeitz:
```suggestion
    fn get_document_field_data(
```

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,820 @@
...
+        &self,
+        accessors: &HashMap<String, DynamicColumn>,
```
PSeitz: There may be multiple columns per column name, so it should be a
tuple `(String, ColumnType)`

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,820 @@
...
+                r#"{ "date": "2017-06-15T00:00:00Z", "text": "ccc", "text2": "ddd" }"#,
+            ],
+            vec![
+                r#"{ "text": "aaa", "text2": "bbb", "date": "2018-01-02T00:00:00Z" }"#,
+                r#"{ "text": "aaa", "text2": "bbb", "date": "2016-01-02T00:00:00Z" }"#,
+            ],
+        ];
+
+        println!("docs: {:?}", docs);
+        println!("merge_segments: {:?}", merge_segments);
```
PSeitz:
```suggestion
```

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,820 @@
...
+}
+
+/// Search query result for each matched document
+/// TODO: move this to a common module
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
+pub struct FieldRetrivalResult {
+    /// The fast fields returned for each hit.
+    #[serde(rename = "docvalue_fields")]
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    pub doc_value_fields: HashMap<String, OwnedValue>,
```
PSeitz, line 122: nested objects are flattened. I didn't see a step to unflatten them (we could decide to not include that in this PR)

ditsuke:
@PSeitz thanks for the latest round of review, I'll get back to the PR early next week!

src/aggregation/metric/top_hits.rs
PSeitz: I just saw that elastic doesn't do unflattening, so that's fine

src/aggregation/metric/top_hits.rs
ditsuke: I think this will return an Error on an IO failure. Should be bubbled up instead?

PSeitz: I mean we never get back false, or that's an error

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,822 @@
...
+                            .ord_to_bytes(term_ord, &mut buffer)
+                            .expect("could not find term_ord in the term dictionary")
+                        {
+                            false => OwnedValue::Null,
+                            true => OwnedValue::Bytes(buffer),
+                        }
+                    }
+                    DynamicColumn::Str(accessor) => 'l: {
+                        let Some(term_ord) = accessor.term_ords(doc_id).next() else {
+                            break 'l OwnedValue::Null;
```
PSeitz: This is hard to read

ditsuke: Removed with [d760c6c](https://github.com/quickwit-oss/tantivy/pull/2198/commits/d760c6c840b050f1ca3053af5b6f68c2f4afadcb)

src/aggregation/metric/top_hits.rs
ditsuke: Sounds good

src/aggregation/metric/top_hits.rs
ditsuke: [d760c6c](https://github.com/quickwit-oss/tantivy/pull/2198/commits/d760c6c840b050f1ca3053af5b6f68c2f4afadcb)

src/aggregation/metric/top_hits.rs
ditsuke: Replaced with plain `.unwrap()` as elsewhere in the codebase

src/aggregation/metric/top_hits.rs
ditsuke: Just to confirm, result `docvalue_fields` should still be grouped by just name?
```jsonc
{
	"docvalue_fields": {
             // Corresponds to ctypes date and u64 for example
		"created": [
			"2023-07-08T...",
			091303333333
		]
	}
}
```

src/aggregation/metric/top_hits.rs
ditsuke: > I mean we never get back false, or that's an error

Oh right, good catch!

PSeitz: `.expect` is definitely better than unwrap as you can provide context

src/aggregation/metric/top_hits.rs
ditsuke: See [33b12a8](https://github.com/quickwit-oss/tantivy/pull/2198/commits/33b12a8cf88d5e0034554a593432a60c3b8515b6)

src/aggregation/metric/top_hits.rs
ditsuke: Good point [5162e14](https://github.com/quickwit-oss/tantivy/pull/2198/commits/5162e140da94929fcea042697aaa741d116e89ed)

src/aggregation/metric/top_hits.rs
PSeitz: > Just to confirm, result `docvalue_fields` should still be grouped by just name?
> 
> ```js
> {
> 	"docvalue_fields": {
>              // Corresponds to ctypes date and u64 for example
> 		"created": [
> 			"2023-07-08T...",
> 			091303333333
> 		]
> 	}
> }
> ```

I think yes. We can't be 100% compatible with elastic search here, as they don't support mixed types.

ditsuke: Fair, I've added the functionality to top_hits. I was considering covering this in the tests but `SchemaBuilder::add_field` does not allow duplicate field names, is there another way to create one?

PSeitz: Yes via JSON fields, you can check the `test_aggregation_on_json_object_mixed_types` test

ditsuke: Ran into a blocker while writing the test. For nested JSON fields
1. `ColumnarReader` seems to remove the `.` separators from field names (as they exist in the `column_dictionary`.
2. `SegmentReader` retains the `.` separator.

What this means: Let's say we need a nested field `json.f`, we start with either a `json.f` or `json.*` for globbed queries. Now we resolve field names from `ColumnarReader`, and depending on whether we started with a glob we're either unable to find the field or unable to find the matched field (`jsonf`) from the `SegmentReader`.

Is `1.` expected behavior? If yes, what could be an alternative way to resolve column names before the aggregation.

Relevant function: https://github.com/quickwit-oss/tantivy/blob/33b12a8cf88d5e0034554a593432a60c3b8515b6/src/aggregation/metric/top_hits.rs#L129-L164

ditsuke: Resolved with [46d1cf7](https://github.com/quickwit-oss/tantivy/pull/2198/commits/46d1cf7fa66a74daa441408e394e3f2e3579baa4), this was due to `ColumnarReader`'s `\0` separated json key representation.

ditsuke:
@PSeitz I believe all review points have been resolved.

ditsuke:
Hi @PSeitz, I don't want to rush the review but I'm just tagging you in case this slipped through your notifications earlier. Please let me know if there are issues with any of the new updates here.

fulmicoton:
@PSeitz can you resume review?

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -244,6 +283,39 @@ impl AggregationWithAccessor {
                 )?;
                 add_agg_with_accessor(accessor, column_type, &mut res)?;
             }
+            TopHits(top_hits) => {
+                let (agg, top_hits) = {
+                    let top_hits =
+                        top_hits.validate_and_resolve(reader.fast_fields().columnar())?;
+                    (
+                        Aggregation {
```
PSeitz: Why is top hits special and does need this overload mechanism

ditsuke: Because of `docvalue_fields`, which needs a step for resolution and validation for field names and wildcard fields. To be fair, should be moved to a common module and made available to other Top collectors too.

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,834 @@
...
+                "Expected exactly one key-value pair in KeyOrder, found more",
+            ));
+        }
+        Ok(Self { field: k, order: v })
+    }
+}
+
+impl TopHitsAggregation {
+    /// Validate and resolve field retrieval parameters
+    pub fn validate_and_resolve(&self, reader: &ColumnarReader) -> crate::Result<Self> {
```
PSeitz: this should be a mutable method, clone or not should be decided by the caller

src/aggregation/metric/top_hits.rs
```diff
@@ -0,0 +1,834 @@
...
+        let cf = self.clone(); // :(
```
PSeitz: what is cf?

src/aggregation/agg_req_with_accessor.rs
PSeitz: you can clone `agg` at the function start and make it mutable instead

ditsuke: Done

src/aggregation/metric/top_hits.rs
ditsuke: Makes sense, done

src/aggregation/metric/top_hits.rs
ditsuke: I wish I remembered, but it anyways doesn't exist now that the method doesn't clone :sweat_smile: :joy:

src/aggregation/agg_req_with_accessor.rs
```diff
@@ -146,7 +196,7 @@ impl AggregationWithAccessor {
                         .map(|m| matches!(m, Key::Str(_)))
                         .unwrap_or(false);
 
-                // Actually we could convert the text to a number and have the fast path, if it is
+                // Actually we could convert the text to a number and have the fast path, if it
```
PSeitz: "if it is"  was correct

src/aggregation/collector.rs
```diff
@@ -135,10 +145,11 @@ impl AggregationSegmentCollector {
     pub fn from_agg_req_and_reader(
         agg: &Aggregations,
         reader: &SegmentReader,
+        segment_local_id: SegmentOrdinal,
```
PSeitz:
```suggestion
        segment_ordinal: SegmentOrdinal,
```
segment_local_id can easily be confused with SegmentId

PSeitz: Why is it not SegmentId?

ditsuke: > segment_local_id can easily be confused with SegmentId

Good point. Looking back, this was probably because a function parameter took in `_segment_id: SegmentOrdinal` as a remnant of a past `SegmentLocalId` => `SegmentOrdinal` rename. I've renamed parameters along the entire hierarchy.

> Why is it not SegmentId

As the use-case is constructing a `DocAddress` that can be used to retrieve a document within the context of a searcher.

src/aggregation/agg_req_with_accessor.rs
ditsuke: Ouch My bad

src/aggregation/collector.rs
PSeitz: I would avoid ordinal addressing if it's not necessary performance wise, as this is a common source of bugs

PSeitz:
Looks good so far, except the segment ordinal part

src/aggregation/collector.rs
ditsuke: As a precautionary measure, doc addresses were excluded from serialized results after we discussed these concerns back in November. If it's still concerning for some reason we can explore switching to SegmentId, but I'll probably need some pointers to start.

ditsuke:
> Looks good so far, except the segment ordinal part

Thank you, I dropped a comment about the SegmentOrdinal in the review thread.

src/aggregation/collector.rs
PSeitz: Ah I just saw this is part of the `Collector` trait, which may be performance relevant in other cases. In that case we leave it like that

PSeitz (approved):

PSeitz:
Thanks for the PR!

(and sorry for the slow Review)

ditsuke:
> Thanks for the PR!
> 
> (and sorry for the slow Review)

Thanks for merging and the thorough review, very educational!

Issue #2200, opened 2023-10-03
Deregister FileWatcher in MmapDirectory


naveenann:
**Is your feature request related to a problem? Please describe.**

On Creating the Index the File Watcher starts a watch on meta.json and cannot be deregister or unwatched. This is a problem when you have to delete the index. The File Watcher will still be active and will try to read the file and throw an Warning.

`thread-tantivy-meta-file-watcher ThreadId(17) tantivy::directory::file_watcher: Failed to open meta file "/data/85ee930e/meta.json": Os { code: 2, kind: NotFound, message: "No such file or directory" }
2023-09-30T07:09:25.687061Z  WARN thread-tantivy-meta-file-watcher ThreadId(17) tantivy::directory::file_watcher: Failed to open meta file "/data/85ee930e/meta.json": Os { code: 2, kind: NotFound, message: "No such file or directory" }`

**Describe the solution you'd like**

A way to drop the File watcher after we delete the index. A fn in `mmap_directory.rs` to drop/ unregister the file watcher.

PSeitz:
Unregister on drop should be done. Having an interface to unregister could be done when there's a use case for it.

PR #2201, opened 2023-10-03
fixes Deregister FileWatcher in MmapDirectory #2200

naveenann:


src/directory/file_watcher.rs
```diff
@@ -88,6 +88,9 @@ impl FileWatcher {
 
         Ok(hasher.finalize())
     }
+    pub fn unsubscribe(&self) {
```
PSeitz: this should be terminate, as there is no resubcribe after that

naveenann: Make sense

fulmicoton:
@PSeitz @naveenann  what is the status of this PR?

PSeitz:
The change caused some weird side-effects (I think it was recursion) and the tests to fail. I didn't investigate yet

PR #2207, opened 2023-10-04
add AsRef, expose object and array iter on Value

PSeitz:
add AsRef
expose object and array iter
move to_json to Document

### DIFF
--- a/src/schema/value.rs
+++ b/src/schema/value.rs
@@ -49,6 +49,13 @@ pub enum OwnedValue {
     IpAddr(Ipv6Addr),
 }
 
+impl AsRef<OwnedValue> for OwnedValue {
+    #[inline]
+    fn as_ref(&self) -> &OwnedValue {
+        self
+    }
+}
+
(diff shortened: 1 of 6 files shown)


src/schema/value.rs
```diff
@@ -49,6 +49,13 @@ pub enum OwnedValue {
     IpAddr(Ipv6Addr),
 }
 
+impl AsRef<OwnedValue> for OwnedValue {
+    #[inline]
+    fn as_ref(&self) -> &OwnedValue {
+        self
+    }
+}
+
```
ChillFish8, line 58: Not against this, but what is this allowing? Just ergonomics?

PSeitz: Yes, just ergonomics, because the Trait is implemented on the reference and the compiler does not auto infer a value to its reference to access the trait

ChillFish8: Fair enough

ChillFish8 (approved):

PR #2209, opened 2023-10-05
fix numeric order, refactor Document

PSeitz:
fix numeric order to prefer i64
rename and move Document stuff


src/schema/document/existing_type_impls.rs
```diff
@@ -25,10 +25,10 @@ impl<'a> Value<'a> for &'a serde_json::Value {
...
-                if let Some(val) = number.as_u64() {
-                    ReferenceValue::U64(val)
-                } else if let Some(val) = number.as_i64() {
+                if let Some(val) = number.as_i64() {
                     ReferenceValue::I64(val)
+                } else if let Some(val) = number.as_u64() {
+                    ReferenceValue::U64(val)
                 } else if let Some(val) = number.as_f64() {
                     ReferenceValue::F64(val)
                 } else {
```
ChillFish8, line 34: I wonder if we should document that the `i64` will take priority, as I can imagine this might be confusing for some users using this impl.

PSeitz: Yes, we should document that somewhere, or do a cast when indexing maybe

ChillFish8 (approved):

PR #2210, opened 2023-10-06
replace ReferenceValue with Self in Value

PSeitz:
Returning `Self` instead of `ReferenceValue` to lower the complexity

### DIFF
--- a/src/schema/document/value.rs
+++ b/src/schema/document/value.rs
@@ -9,12 +9,11 @@ use crate::tokenizer::PreTokenizedString;
 /// A single field value.
 pub trait Value<'a>: Send + Sync + Debug {
     /// The child value type returned by this doc value.
-    type ChildValue: Value<'a>;
     /// The iterator for walking through the elements within the array.
-    type ArrayIter: Iterator<Item = ReferenceValue<'a, Self::ChildValue>>;
+    type ArrayIter: Iterator<Item = Self>;
     /// The visitor walking through the key-value pairs within
     /// the object.
-    type ObjectIter: Iterator<Item = (&'a str, ReferenceValue<'a, Self::ChildValue>)>;
+    type ObjectIter: Iterator<Item = (&'a str, Self)>;
 
     /// Returns the field value represented by an enum which borrows it's data.
     fn as_value(&self) -> ReferenceValue<'a, Self>;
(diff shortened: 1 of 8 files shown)


fulmicoton (approved):

src/schema/document/value.rs
```diff
@@ -9,12 +9,11 @@ use crate::tokenizer::PreTokenizedString;
...
 pub trait Value<'a>: Send + Sync + Debug {
     /// The child value type returned by this doc value.
-    type ChildValue: Value<'a>;
     /// The iterator for walking through the elements within the array.
-    type ArrayIter: Iterator<Item = ReferenceValue<'a, Self::ChildValue>>;
+    type ArrayIter: Iterator<Item = Self>;
     /// The visitor walking through the key-value pairs within
     /// the object.
-    type ObjectIter: Iterator<Item = (&'a str, ReferenceValue<'a, Self::ChildValue>)>;
+    type ObjectIter: Iterator<Item = (&'a str, Self)>;
```
ChillFish8, line 16: The reason why this was done is because it creates huge ergonomic issues when you are building more complex indexing structures, or you have anything touching inline values, i.e. SmallVec. I don't really agree with this change as it significantly hinders ergonomics and any optimisations you might want to do when working with custom types.

PSeitz: This still compiles. The PR removes a case where your value's recursive ChildValue is not `Self`. I don't know if that's really relevant.

```rust
#[derive(Clone, Debug, Default)]
pub struct TantivyDocument2 {
    field_values: SmallVec<[FieldValue; 4]>,
}

impl Document for TantivyDocument2 {
    type Value<'a> = &'a OwnedValue;
    type FieldsValuesIter<'a> = FieldValueIter<'a>;

    fn iter_fields_and_values(&self) -> Self::FieldsValuesIter<'_> {
        FieldValueIter(self.field_values.iter())
    }
}

impl DocumentDeserialize for TantivyDocument2 {
    fn deserialize<'de, D>(mut _deserializer: D) -> Result<Self, DeserializeError>
    where
        D: DocumentDeserializer<'de>,
    {
        unimplemented!()
    }
}
```

ChillFish8: The above example doesn't touch on the recursive structure of the `Value` which is why it compiles. The `ChildValue` is useful in places where you have something like a flattened document where you may want a SmallVec to store the single value fields inline, which you can't do without creating a new type to prevent the recursive type error.

PSeitz: Can you give an example?

ChillFish8: I will try to do an example later today if I have a chance, I admit it's a somewhat niche situation.

PR #2211, opened 2023-10-06
remove Document: DocumentDeserialize dependency

PSeitz:
The dependency requires users to implement an API they may not use.

### DIFF
--- a/src/core/searcher.rs
+++ b/src/core/searcher.rs
@@ -5,7 +5,7 @@ use std::{fmt, io};
 use crate::collector::Collector;
 use crate::core::{Executor, SegmentReader};
 use crate::query::{Bm25StatisticsProvider, EnableScoring, Query};
-use crate::schema::document::Document;
+use crate::schema::document::{Document, DocumentDeserialize};
 use crate::schema::{Schema, Term};
 use crate::space_usage::SearcherSpaceUsage;
 use crate::store::{CacheStats, StoreReader};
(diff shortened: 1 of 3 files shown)


src/core/searcher.rs
```diff
@@ -84,7 +84,10 @@ impl Searcher {
     ///
     /// The searcher uses the segment ordinal to route the
     /// request to the right `Segment`.
-    pub fn doc<D: Document>(&self, doc_address: DocAddress) -> crate::Result<D> {
+    pub fn doc<D: Document + DocumentDeserialize>(
```
fulmicoton: Could it be?

```
pub fn doc<D: DocumentDeserialize>(
```

src/core/searcher.rs
```diff
@@ -104,7 +107,10 @@ impl Searcher {
 
     /// Fetches a document in an asynchronous manner.
     #[cfg(feature = "quickwit")]
-    pub async fn doc_async<D: Document>(&self, doc_address: DocAddress) -> crate::Result<D> {
+    pub async fn doc_async<D: Document + DocumentDeserialize>(
```
fulmicoton: same question as above

fulmicoton (approved):
See comments inline

src/core/searcher.rs
PSeitz: yep, removed the unnecessary trait

PR #2212, opened 2023-10-06
run coverage only after merge

PSeitz:
coverage is a quite slow step in CI. It can be run only after merging

Add compile benchmarks to regular CI test

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,8 +43,6 @@ jobs:
     - name: Check Stable Compilation
       run: cargo build --all-features
 
-    - name: Check Nightly Compilation
-      run: cargo +nightly build --all-features
     
     - name: Check Bench Compilation
       run: cargo +nightly bench --no-run --profile=dev --all-features
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,6 +39,15 @@ jobs:
 
     - name: Check Formatting
       run: cargo +nightly fmt --all -- --check
+    
+    - name: Check Stable Compilation
+      run: cargo build --all-features
+
+    - name: Check Nightly Compilation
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -39,6 +39,15 @@ jobs:
 
     - name: Check Formatting
       run: cargo +nightly fmt --all -- --check
+    
+    - name: Check Stable Compilation
+      run: cargo build --all-features
+
+    - name: Check Nightly Compilation
```
fulmicoton, line 46: Can we remove this!?

.github/workflows/test.yml
```diff
@@ -39,6 +39,15 @@ jobs:
...
+      run: cargo +nightly build --all-features
```
fulmicoton:
```suggestion
```

.github/workflows/test.yml
```diff
@@ -39,6 +39,15 @@ jobs:
...
+    
+    - name: Check Bench Compilation
+      run: cargo +nightly bench --no-run --profile=dev --all-features
```
fulmicoton, line 48: Will we look at the result? It won't give a diff of the last run will it?
It seems too much for a PR especially.

adamreichold: Note the `--no-run` argument. This will not execute the benchmarks, just check that they compile (and even use the `dev` profile to get fast build time instead of fast builds).

PSeitz: We broke the benchmarks compilation previously, since they are not covered in the CI. This did break as a consequence https://pseitz.github.io/tantivy_bench_graph/index.html

fulmicoton: ah got it!

fulmicoton (approved):

PR #2222, opened 2023-10-19
add fields_metadata to SegmentReader, add columnar docs

PSeitz:


src/core/segment_reader.rs
```diff
@@ -280,6 +282,94 @@ impl SegmentReader {
...
+    /// that have been indexed as a part of a JSON field.
+    /// The returned field name is the full field name, including the name of the JSON field.
+    pub fn fields_metadata(&self) -> crate::Result<Vec<FieldMetadata>> {
+        let mut indexed_fields: Vec<(String, Type)> = Vec::new();
+        for (field, field_entry) in self.schema().fields() {
+            let field_name = field_entry.name().to_string();
+            let is_indexed = field_entry.is_indexed();
+
+            if is_indexed {
+                indexed_fields.push((
```
fulmicoton: I don't think it makes sense to add the name field name for json field. Shouldn't this code be part of an else statement?

src/core/segment_reader.rs
```diff
@@ -280,6 +282,94 @@ impl SegmentReader {
...
+                    field_entry.field_type().value_type(),
+                ));
+                let is_json = field_entry.field_type().value_type() == Type::Json;
+                if is_json {
+                    let inv_index = self.inverted_index(field)?;
+                    let encoded_fields_in_index = inv_index.list_encoded_fields()?;
+                    indexed_fields.extend(
+                        encoded_fields_in_index
+                            .into_iter()
+                            .map(|(name, typ)| (format!("{}.{}", field_name, name), typ)),
```
fulmicoton: I suspect having picked `(String, Type)` here, forcing the concatenation of the field and the field path was not a good idea. This is my first pass reviewing so I might edit this comment later.

src/core/segment_reader.rs
```diff
@@ -280,6 +282,94 @@ impl SegmentReader {
...
+            .map(|(name, handle)| (name.to_string(), Type::from(handle.column_type())))
+            .collect();
+        // Since the type is encoded differently in the fast field and in the inverted index,
+        // the order of the fields is not guaranteed to be the same. Therefore, we sort the fields.
+        // If we are sure that the order is the same, we can remove this sort.
+        indexed_fields.sort_unstable();
+        fast_fields.sort_unstable();
+
+        // Maybe too slow for the high cardinality case
+        fn is_field_stored(field_name: &str, schema: &Schema) -> bool {
```
fulmicoton: Of course this will be missing some fields that are just STORED and in the dynamic mode, but I think this is the right trade-off!

src/core/segment_reader.rs
fulmicoton: Second pass here.
I think keep the `field_name` and the `field_path` separated might be helpful. The trouble is that the fast fields are all in the same bag so I am not sure this is a good idea.
So your Vec would become `Vec<(&str, Option<&str>, Type)>`.
The stored field logic would become computationally simpler.
The fast field logic would then get the extra split logic.

Also quite frankly, I am not entirely sure this is better... You can ignore this comment.

src/core/segment_reader.rs
fulmicoton: could `Schema::find_field_with_default<'a>` be used? I am a bit scared by spreading the logic in a non-trivial way.
for instance here we don't support escaped "dot".

PSeitz: Yes, I'll switch to calling into schema

src/core/segment_reader.rs
PSeitz: We need to return the full name, as this is used e.g. to select a field by a user

src/core/segment_reader.rs
PSeitz: We need to return the full name, as this is used e.g. to select a field by a user

src/core/index.rs
```diff
@@ -489,6 +490,23 @@ impl Index {
...
 
+    /// Returns the list of fields that have been indexed in the Index.
+    /// The field list includes the field defined in the schema as well as the fields
+    /// that have been indexed as a part of a JSON field.
+    /// The returned field name is the full field name, including the name of the JSON field.
+    ///
+    /// The returned field names can be used in queries.
+    ///
+    /// Notice: If your data contains JSON fields this is **very expensive**.
+    pub fn fields_metadata(&self) -> crate::Result<Vec<FieldMetadata>> {
```
fulmicoton, line 506: very clean...

src/core/segment_reader.rs
fulmicoton: I understand we want _dynamic.attributes.color in the list. I don't understand why we want _dynamic.

src/core/segment_reader.rs
```diff
@@ -280,6 +286,117 @@ impl SegmentReader {
...
     }
 
+    /// Returns the list of fields that have been indexed in the segment.
+    /// The field list includes the field defined in the schema as well as the fields
+    /// that have been indexed as a part of a JSON field.
+    /// The returned field name is the full field name, including the name of the JSON field.
+    ///
+    /// The returned field names can be used in queries.
+    ///
+    /// Notice: If your data contains JSON fields this is **very expensive**.
```
fulmicoton: /// Disclaimer
///
/// Some fields may not be listed here. For instance, if the schema contains a json field that is not
// indexed nor a fast field but is stored, it is possible for the field to not be listed.

fulmicoton:
```suggestion
    /// Notice: If your data contains JSON fields this is **very expensive**, as it requires browsing through
    // the inverted index term dictionary and the columnar field dictionary.
    /// Disclaimer: Some fields may not be listed here. For instance, if the schema contains a json field that 
    /// is not indexed nor a fast field but is stored, it is possible for the field to not be listed.
```

src/core/segment_reader.rs
```diff
@@ -280,6 +286,117 @@ impl SegmentReader {
...
+        fn is_field_stored(field_name: &str, schema: &Schema) -> bool {
+            schema
+                .find_field(field_name)
+                .map(|(field, _path)| schema.get_field_entry(field).is_stored())
+                .unwrap_or(false)
+        }
+
+        let merged: Vec<FieldMetadata> = indexed_fields
+            .into_iter()
+            .merge_join_by(fast_fields, |i, j| i.cmp(j))
```
fulmicoton: nitpick: we usually use `left / right` in the code base, or `lhs, rhs`.
```suggestion
            .merge_join_by(fast_fields, |i, j| i.cmp(j))
```

src/core/segment_reader.rs
```diff
@@ -330,6 +447,75 @@ impl SegmentReader {
...
+        schema
+            .find_field(field_name)
+            .map(|(field, _path)| schema.get_field_entry(field).is_stored())
+            .unwrap_or(false)
+    }
+
+    let mut out = Vec::new();
+    for (_key, mut group) in &input
+        .into_iter()
+        .kmerge_by(|a, b| a < b)
```
fulmicoton: left / right
```suggestion
        .kmerge_by(|a, b| a < b)
```

src/core/segment_reader.rs
```diff
@@ -330,6 +447,75 @@ impl SegmentReader {
...
+impl PartialOrd for FieldMetadata {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for FieldMetadata {
+    fn cmp(&self, other: &Self) -> Ordering {
+        match self.field_name.cmp(&other.field_name) {
+            Ordering::Equal => self.typ.cmp(&other.typ),
```
fulmicoton: Let's avoid inconsistent Ord / PartialOrd / Eq/ PartialEq / Hash implementation. This is a very common footgun.
(Here your Ord impl is inconsistent with Eq)

src/core/segment_reader.rs
fulmicoton: Can you. remove the Ord implementation of FieldMetadata, create a function extracting the key you want (returning the pair field name field type ).
Put that in that file, and call it here instead?

src/core/segment_reader.rs
```diff
@@ -330,6 +447,75 @@ impl SegmentReader {
...
+    }
+
+    let mut out = Vec::new();
+    for (_key, mut group) in &input
+        .into_iter()
+        .kmerge_by(|a, b| a < b)
+        // TODO: Remove allocation
+        .group_by(|el| (el.field_name.to_string(), el.typ))
+    {
+        let mut merged = group.next().unwrap();
```
fulmicoton:
```suggestion
        let mut merged: FieldMetadata = group.next().unwrap();
```

src/core/segment_reader.rs
```diff
@@ -330,6 +447,75 @@ impl SegmentReader {
...
+        for el in group {
+            merged |= el;
+        }
+        merged.stored = is_field_stored(&merged.field_name, schema);
```
fulmicoton, line 488: can you explain the special treatment of stored in a comment?

src/core/segment_reader.rs
```diff
@@ -330,6 +447,75 @@ impl SegmentReader {
...
+) -> Vec<FieldMetadata> {
+    // Maybe too slow for the high cardinality case
+    fn is_field_stored(field_name: &str, schema: &Schema) -> bool {
+        schema
+            .find_field(field_name)
+            .map(|(field, _path)| schema.get_field_entry(field).is_stored())
+            .unwrap_or(false)
+    }
+
+    let mut out = Vec::new();
```
fulmicoton:
```suggestion
    let mut merged_field_metadatas: Vec<FieldMetadata> = Vec::new();
```

src/core/segment_reader.rs
```diff
@@ -280,6 +286,117 @@ impl SegmentReader {
...
+                .find_field(field_name)
+                .map(|(field, _path)| schema.get_field_entry(field).is_stored())
+                .unwrap_or(false)
+        }
+
+        let merged: Vec<FieldMetadata> = indexed_fields
+            .into_iter()
+            .merge_join_by(fast_fields, |i, j| i.cmp(j))
+            .map(|either| match either {
+                EitherOrBoth::Left((field_name, typ)) => FieldMetadata {
```
fulmicoton: Nitpick... I think the code could have been simpler:
You could have had a function creating 
- a list of indexed fields with true only for `indexed`.
- a list of fast fields with true only for whether they are fast or not.
Used the general merge routine to merge the two.

fulmicoton (changes requested):
the main comment is to remove the PartialOrd / Ord, but please have a look at the other comments too.

src/core/segment_reader.rs
PSeitz: I'll switch to a derive with a comment to not reorder

src/core/segment_reader.rs
PSeitz: I prefer to have the Ord implementation, so it will behave consistent with `sort_by` etc.

src/core/segment_reader.rs
PSeitz: removed the JSON field from the list

src/core/segment_reader.rs
PSeitz: What special treatment do you mean?

src/core/segment_reader.rs
```diff
@@ -330,6 +432,64 @@ impl SegmentReader {
...
+        assert!(self.typ == rhs.typ);
+        self.indexed |= rhs.indexed;
+        self.stored |= rhs.stored;
+        self.fast |= rhs.fast;
+    }
+}
+
+/// Helper to merge the field metadata from multiple segments.
+pub fn merge_field_meta_data(
+    input: Vec<Vec<FieldMetadata>>,
```
fulmicoton: `input`   :(

src/core/segment_reader.rs
```diff
@@ -330,6 +432,64 @@ impl SegmentReader {
...
+    schema: &Schema,
+) -> Vec<FieldMetadata> {
+    // Maybe too slow for the high cardinality case
+    fn is_field_stored(field_name: &str, schema: &Schema) -> bool {
```
fulmicoton: nitpick: Can we have the function outside of this?

I personally find it less readable.
I assume your point is that you want to namespace the function as much as possible.
One trouble with these is that it makes it non trivial to know whether they capture their environment or not.

So answering the question, "what does thjis function does" actually needs to look at its impl, as opposed to looking at the prototype of the function. It could hide any surprises.

fulmicoton (approved):

src/core/segment_reader.rs
PSeitz: I think `fn` can't capture anything?
I moved it outside

PR #2224, opened 2023-10-24
json path writer

PSeitz:
Create JsonPathWriter to have a canonical way to build tantivy JSON paths. `json_utils` is kept as it is for now, as this may be replaced anyways if we pass `(UnorderedId, type, Value)` instead having the full path in term.

- refactor logic to JsonPathWriter
- use in encode_column_name
- add inlines

### DIFF
--- a/common/src/json_path_writer.rs
+++ b/common/src/json_path_writer.rs
@@ -29,11 +29,13 @@ impl JsonPathWriter {
     /// write queries  like `k8s.node.id:5`.
     /// On the other, enabling that feature can lead to
     /// ambiguity.
+    #[inline]
     pub fn set_expand_dots(&mut self, expand_dots: bool) {
         self.expand_dots = expand_dots;
     }
 
     /// Push a new segment to the path.
+    #[inline]
     pub fn push(&mut self, segment: &str) {
         let len_path = self.path.len();
         self.indices.push(len_path);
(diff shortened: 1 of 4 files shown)


common/src/json_path_writer.rs
```diff
@@ -0,0 +1,112 @@
...
+            self.path.push_str(JSON_PATH_SEGMENT_SEP_STR);
+        }
+        self.path.push_str(segment);
+        if self.expand_dots {
+            // This might include the separation byte, which is ok because it is not a dot.
+            let appended_segment = &mut self.path[len_path..];
+            // The unsafe below is safe as long as b'.' and JSON_PATH_SEGMENT_SEP are
+            // valid single byte ut8 strings.
+            // By utf-8 design, they cannot be part of another codepoint.
+            replace_in_place(b'.', JSON_PATH_SEGMENT_SEP, unsafe {
```
fulmicoton: nitpick: 

I prefer:
```suggestion
            unsafe { replace_in_place(b'.', JSON_PATH_SEGMENT_SEP, appended_segment.as_bytes_mut() ) };
```

The point being that the block within the unsafe contains all of the justification for the correctness.

PSeitz: done

fulmicoton (approved):

Issue #2215, opened 2023-10-12
Dynamic Schema


PSeitz:
# Problem Outline
1. tantivy root fields can't be indexed without explicit configuration, e.g. there's no index everything mode. Workaround is to copy into a seperate JSON field.
2. Nested fields can't be configured differently from their parent for JSON fields. Workaround is be to copy into a seperate field. (https://github.com/quickwit-oss/quickwit/issues/3896)

That's a lot of copying and also difficult to untangle when searching the data. E.g. there's a special code behind a feature flag for quickwit `_dynamic` field.

Core Issues:

1. Configuration
2. Indexing Performance (affected by copying).
3. Untangle copied data on retrieval

# Requirements

1. Allow configurations for any nested field
2. Enable schemaless indexing of root level fields
3. Ensure quick access to a value's current config without compromising indexing performance
4. Avoid overhead of field name->config resolution where possible

# Proposal
The proposed changes affect schema definition, indexing and the `Index` data-structures.

## Schema API
Schema allows to configure a root aka catch all config.
Instead of configuring fields, we define configurations on paths. Each configuration would get an Configuration ID, that can be referenced.

### Example JSON
```JSON
{
    "a": "b", // will be indexed with root config
    "severity": "INFO",
    {"attributes": 
        { "price": 100, "tags": ["blue"] }
    }
}
```

```rust
let mut schema_builder = tantivy::schema::SchemaBuilder::new();
schema_builder.set_dynamic_schema(JsonObjectOptions{fast: true, indexed: true}); // All fields will be fast and indexed
schema_builder.add_json_field("attributes", DISABLED); // ignore (also nested fields). Do we need this feature?
schema_builder.add_u64_field("attributes.price", FAST | INDEXED); // overrule nested `attribute.price` to be fast and indexed
schema_builder.add_text_field("severity", STRING);
let schema = schema_builder.build();

let attr_config_id = schema.get_config_id("attributes.price"); // FAST | INDEXED | U64

```

## Document API

Documents would return `(Path, Config ID, Value)` pairs.

`fn iter_fields_and_values(&self, schema: &Schema) -> impl Iterator<Item = (&str, Config ID, Value)>`

Returning a Config ID would keep the limitation that we only iterate configured fields.
The Problem is that the API will get very ugly with regards to Object and Array Iterator types on `Value`.
To mitigate this, we can switch to only accept flatten paths, so nested structures would be flattened. 
We flatten them anyways while indexing.

[...]

PR #2231, opened 2023-10-27
collect json paths in indexing

PSeitz:
Partially implementing https://github.com/quickwit-oss/tantivy/issues/2015 and preparation for https://github.com/quickwit-oss/tantivy/issues/2215

Next step would be to remove the `JsonTermWriter` and use the same technique in fast field json paths

### DIFF
--- a/src/indexer/path_to_unordered_id.rs
+++ b/src/indexer/path_to_unordered_id.rs
@@ -1,0 +1,69 @@
+use fnv::FnvHashMap;
+
+#[derive(Default)]
+pub(crate) struct PathToUnorderedId {
+    map: FnvHashMap<String, u32>,
+}
+
+impl PathToUnorderedId {
+    #[inline]
+    pub(crate) fn get_id(&mut self, path: &str) -> u32 {
+        if let Some(id) = self.map.get(path) {
+            return *id;
+        }
+        self.insert_new_path(path)
+    }
+    #[cold]
+    fn insert_new_path(&mut self, path: &str) -> u32 {
+        let next_id = self.map.len() as u32;
+        self.map.insert(path.to_string(), next_id);
+        next_id
+    }
+
+    /// Retuns ids which reflect the lexical order of the paths.
+    ///
+    /// The returned vec can be indexed with the unordered id to get the ordered id.
+    pub(crate) fn unordered_id_to_ordered_id(&self) -> Vec<u32> {
+        let mut sorted_ids: Vec<(&String, &u32)> = self.map.iter().collect();
+        sorted_ids.sort_unstable_by_key(|(path, _)| *path);
+        let mut result = vec![sorted_ids.len() as u32; sorted_ids.len()];
+        for (ordered, unordered) in sorted_ids.iter().map(|(_k, v)| v).enumerate() {
+            result[**unordered as usize] = ordered as u32;
+        }
+        result
+    }
+
+    /// Retuns the paths so they can be queried by the ordered id (which is the index).
+    pub(crate) fn ordered_id_to_path(&self) -> Vec<&String> {
+        let mut paths = self.map.keys().collect::<Vec<_>>();
+        paths.sort_unstable();
+        paths
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn path_to_unordered_test() {
+        let mut path_to_id = PathToUnorderedId::default();
+        let terms = vec!["b", "a", "b", "c"];
+        let ids = terms
+            .iter()
+            .map(|term| path_to_id.get_id(term))
+            .collect::<Vec<u32>>();
+        assert_eq!(ids, vec![0, 1, 0, 2]);
+        let ordered_ids = ids
+            .iter()
+            .map(|id| path_to_id.unordered_id_to_ordered_id()[*id as usize])
+            .collect::<Vec<u32>>();
+        assert_eq!(ordered_ids, vec![1, 0, 1, 2]);
+        // Fetch terms
+        let terms_fetched = ordered_ids
+            .iter()
+            .map(|id| path_to_id.ordered_id_to_path()[*id as usize])
+            .collect::<Vec<&String>>();
+        assert_eq!(terms_fetched, terms);
+    }
+}
--- a/src/indexer/path_to_unordered_id.rs
+++ b/src/indexer/path_to_unordered_id.rs
@@ -1,5 +1,27 @@
 use fnv::FnvHashMap;
 
+/// `Field` is represented by an unsigned 32-bit integer type.
+/// The schema holds the mapping between field names and `Field` objects.
+#[derive(Copy, Default, Clone, Debug, PartialEq, PartialOrd, Eq, Ord, Hash)]
+pub struct OrderedPathId(u32);
+
+impl OrderedPathId {
+    /// Create a new field object for the given PathId.
+    pub const fn from_ordered_id(field_id: u32) -> OrderedPathId {
+        OrderedPathId(field_id)
+    }
+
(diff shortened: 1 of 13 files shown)


stacker/src/arena_hashmap.rs
```diff
@@ -235,6 +235,29 @@ impl ArenaHashMap {
         }
     }
 
+    #[inline]
+    /// This will invalidate the hashmaps get and insert methods.
+    /// Only iter() is still valid afterwards
+    ///
+    /// # Safety
+    /// Any call to get or mutate_or_create after this call is undefined behavior.
+    pub unsafe fn iter_mut_keys<F: Fn(&mut [u8])>(&mut self, cb: F) {
```
adamreichold: Could this also be iterator-based, i.e. return `impl Iterator<Item=&mut [u8]>` or do the lifetimes not work out?

src/indexer/path_to_unordered_id.rs
```diff
@@ -0,0 +1,76 @@
...
+        let next_id = self.map.len() as u32;
+        self.map.insert(path.to_string(), next_id);
+        next_id
+    }
+
+    /// Retuns ids which reflect the lexical order of the paths.
+    ///
+    /// The returned vec can be indexed with the unordered id to get the ordered id.
+    pub(crate) fn unordered_id_to_ordered_id(&self) -> Vec<u32> {
+        let mut sorted_ids: Vec<(&String, &u32)> = self.map.iter().collect::<_>();
```
adamreichold: The turbofish at the end looks ineffectual.

src/indexer/path_to_unordered_id.rs
```diff
@@ -0,0 +1,76 @@
...
+            result[**unordered as usize] = ordered as u32;
+        }
+        result
+    }
+
+    /// Retuns ids which reflect the lexical order of the paths.
+    ///
+    /// The returned vec can be indexed with the unordered id to get the ordered id.
+    pub(crate) fn ordered_id_to_path(&self) -> Vec<&String> {
+        let unordered_id_to_ordered_id = self.unordered_id_to_ordered_id();
```
adamreichold: I am likely missing something, but wouldn't it be simpler to just do

```rust
let mut paths = self.map.keys().collect::<Vec<_>>();
paths.sort_unstable_key();
paths
```

than to undo what `unordered_id_to_ordered_id` did?

adamreichold: (I may also just be confused because the doc comment seems to be copied from the previous function without adjusting it to this one.)

PSeitz: yes that's more straightforward

stacker/src/arena_hashmap.rs
PSeitz: Yes, lifetimes on mutable iterators are really messy and not worth here

src/indexer/path_to_unordered_id.rs
```diff
@@ -0,0 +1,69 @@
+use fnv::FnvHashMap;
+
+#[derive(Default)]
+pub(crate) struct PathToUnorderedId {
+    map: FnvHashMap<String, u32>,
+}
+
+impl PathToUnorderedId {
+    #[inline]
+    pub(crate) fn get_id(&mut self, path: &str) -> u32 {
```
fulmicoton: get_or_allocated_unordered_id

src/indexer/path_to_unordered_id.rs
```diff
@@ -0,0 +1,69 @@
...
+        let next_id = self.map.len() as u32;
+        self.map.insert(path.to_string(), next_id);
+        next_id
+    }
+
+    /// Retuns ids which reflect the lexical order of the paths.
+    ///
+    /// The returned vec can be indexed with the unordered id to get the ordered id.
+    pub(crate) fn unordered_id_to_ordered_id(&self) -> Vec<u32> {
+        let mut sorted_ids: Vec<(&String, &u32)> = self.map.iter().collect();
```
fulmicoton: I think `&str` might be mildly better.

```suggestion
        let mut sorted_ids: Vec<(&str, &u32)> = self.map.iter().map(String::as_str).collect();
```

src/postings/postings_writer.rs
```diff
@@ -43,12 +43,27 @@ fn make_field_partition(term_offsets: &[(Term<&[u8]>, Addr)]) -> Vec<(Field, Ran
...
-    ctx: IndexingContext,
+    mut ctx: IndexingContext,
+    schema: Schema,
     per_field_postings_writers: &PerFieldPostingsWriter,
     fieldnorm_readers: FieldNormReaders,
     doc_id_map: Option<&DocIdMapping>,
     serializer: &mut InvertedIndexSerializer,
 ) -> crate::Result<()> {
+    // Replace unordered ids by ordered ids to be able to sort
+    let unordered_id_to_ordered_id = ctx.path_to_unordered_id.unordered_id_to_ordered_id();
```
fulmicoton: please add the type here.
```suggestion
    let unordered_id_to_ordered_id = ctx.path_to_unordered_id.unordered_id_to_ordered_id();
```

stacker/src/arena_hashmap.rs
fulmicoton: Why do we need to do the remap here?

COuld that be done in a safe not mut way upon serialization? (in serialize postings)

fulmicoton:
This PR surely saves memory, but i have no clue what impact on indexing throughput is.
Did you measure that?

(Of course, the next step is to push the ids at a higher level to avoid building the json path construction, but you were right to not ship that in that PR.)

stacker/src/arena_hashmap.rs
PSeitz: | Why do we need to do the remap here?

We need the keys in lexicographical order, for the serialization.

I think I can change it to consume the hashmap and remove the unsafe

PSeitz:
> This PR surely saves memory, but i have no clue what impact on indexing throughput is. Did you measure that?
> 
> (Of course, the next step is to push the ids at a higher level to avoid building the json path construction, but you were right to not ship that in that PR.)

I didn't measure a lot, but it seemed relatively unchanged. The extra look-ups seem to be set of by the less copying and faster hash calculation

src/postings/postings_writer.rs
```diff
@@ -44,19 +47,40 @@ fn make_field_partition(term_offsets: &[(Term<&[u8]>, Addr)]) -> Vec<(Field, Ran
...
+            let path_id = unordered_id_to_ordered_id[unordered_id as usize];
+            (field, path_id, &key[byte_range_path.end..], addr)
+        } else {
+            (field, 0.into(), &key[5..], addr)
+        }
+    }));
+    // Sort by field, path, and term
+    term_offsets.sort_unstable_by(
+        |(field1, path_id1, bytes1, _), (field2, path_id2, bytes2, _)| match field1.cmp(field2) {
+            std::cmp::Ordering::Equal => match path_id1.cmp(path_id2) {
```
fulmicoton: I think you can simplify that into

(field1, path_id1, bytes1).cmp(&(field2, path_id2, bytes2))

src/postings/postings_writer.rs
```diff
@@ -98,7 +124,8 @@ pub(crate) trait PostingsWriter: Send + Sync {
     /// The actual serialization format is handled by the `PostingsSerializer`.
     fn serialize(
         &self,
-        term_addrs: &[(Term<&[u8]>, Addr)],
+        term_addrs: &[(Field, OrderedPathId, &[u8], Addr)],
+        ordered_id_to_path: &[&String],
```
fulmicoton:
```suggestion
        ordered_id_to_path: &[&str],
```

fulmicoton (approved):

Issue #1824, opened 2023-01-23
Missing search results with in memory index


EntilZha:
I've been working on a small command line app to merge bibtex files and have been using tantivy to index paper titles. I noticed some strange behavior where if I (from python extension):
1. Create an in memory index
2. Call `writer.add_document` in a for loop
3. Call `writer.commit()` outside the for loop
The search results I'd expect to be exact matches (duplicate detection) are not showing up. I reran the same code several times and every so often I do get the correct results, which suggested to me that on commit, some index entries were getting lost. If I have `writer.commit` inside the for loop, the issue goes away. Likewise, if I use a file-based index, the issue also goes away.

Reading the docs, it looks like even with an in memory index, calling `writer.commit()` after the for loop and not in it should work, is there something I'm missing or might this be a bug?

If it's not a known issue already and you cant reproduce it, once I finish with my cli I can share the relevant code.

PSeitz:
Thanks for the report. Can you provide something to reproduce that behavior in rust? I'm not very familiar with python

Do you reload your `IndexReader` after committing? It may have a view on the old segments.

K0UR05H:
Hello @PSeitz
I've run into a similar problem. here is a code snippet to reproduce the behavior:

```rust
use tantivy::collector::TopDocs;
use tantivy::query::QueryParser;
use tantivy::schema::*;
use tantivy::{doc, Index, ReloadPolicy};

fn main() -> tantivy::Result<()> {
    let mut schema_builder = Schema::builder();
    let title = schema_builder.add_text_field("title", TEXT | STORED);
    let schema = schema_builder.build();

    let index = Index::create_in_ram(schema);

    let reader = index
        .reader_builder()
        .reload_policy(ReloadPolicy::OnCommit)
        .try_into()?;

    let mut index_writer = index.writer(10_000_000)?;
    index_writer.add_document(doc!(
        title => "The Old Man and the Sea"
    ))?;
    index_writer.commit()?;

    let query_parser = QueryParser::for_index(&index, vec![title]);
    let query = query_parser.parse_query("sea")?;

    let searcher = reader.searcher();
    let top_docs = searcher.search(&query, &TopDocs::with_limit(10))?;

    assert!(!top_docs.is_empty());

    Ok(())
}
```

I guess the problem is `ReloadPolicy::OnCommit` taking a few milliseconds to reload. I solved this problem by changing the reload policy to manual and calling `reader.reload()` after `index_writer.commit()`.

[...]

PR #2235, opened 2023-10-30
rename ReloadPolicy onCommit to onCommitWithDelay

giovannicuccu:
Here is the pull request for issue [#1824](https://github.com/quickwit-oss/tantivy/issues/1824) that we dicussed on discord.
I simply renamed the reload policy from OnCommit to OnCommitWithDelay and updated the reference in the documentation and tests.

I tried to implement an OnCommit policy for synchronous reload but I did not succeed. 
Here are the main issues I faced:
1. IndexReader and IndexWriter can be created indipendently (see `test_index_on_commit_reload_policy_different_directories` in `core/test.rs` ) and as far as I understand they share the access to the same directory. When an IndexWriter commits the changes are propagated to the file system, if an IndexReader is executed in a different process there is no simple way to notify the commit in a synchronous way because "there isn't a  channel" between them (aside from the file system). I checked if there are some notifications for updating mmapped files but I did not find anything
2. There semantic of the reload is "lost" as the callback is passed to the directory object that implements the commit notification in a different way (worker thread for `mmap_directory`, direct call inside atomic_write for `ram_directory`). If the ReloadPolicy is "directory implementation indipendent" there is the need to associate it to the callback.

If you think it'd feasible to implement a sync OnCommit policy I ask for some advice and I'll start to dig further into the problem.

### DIFF
--- a/examples/basic_search.rs
+++ b/examples/basic_search.rs
@@ -164,7 +164,7 @@ fn main() -> tantivy::Result<()> {
     // will reload the index automatically after each commit.
     let reader = index
         .reader_builder()
-        .reload_policy(ReloadPolicy::OnCommit)
+        .reload_policy(ReloadPolicy::OnCommitWithDelay)
         .try_into()?;
 
     // We now need to acquire a searcher.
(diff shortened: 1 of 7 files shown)


PSeitz (approved):

PSeitz:
Thanks! As of now you can implement sync behavior by manually reloading, which should be fine for most use cases

PR #2237, opened 2023-11-02
bump index version

### LINKED ISSUE #2236: Bump the index format
We changed the index in a backward compatible way. We still should have incremented the index version to make sure we get a significant error msg if the problem is just a version mismatch (old version opening an index created with a new version.)
and have slightly nicer logic to match the version.

(allow version N-1 and version N, for the current version of tantivy.)

trinity-1686a:
fix #2236 

> (allow version N-1 and version N, for the current version of tantivy.)

we already have some logic to allow older index revision, but it's an unnamed constant hidden somewhere. I moved it alongside the current version

### DIFF
--- a/src/directory/footer.rs
+++ b/src/directory/footer.rs
@@ -7,7 +7,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::directory::error::Incompatibility;
 use crate::directory::{AntiCallToken, FileSlice, TerminatingWrite};
-use crate::{Version, INDEX_FORMAT_VERSION};
+use crate::{Version, INDEX_FORMAT_OLDEST_SUPPORTED_VERSION, INDEX_FORMAT_VERSION};
 
 const FOOTER_MAX_LEN: u32 = 50_000;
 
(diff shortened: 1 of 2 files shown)


src/directory/footer.rs
```diff
@@ -103,7 +103,7 @@ impl Footer {
     /// Has to be called after `extract_footer` to make sure it's not accessing uninitialised memory
     pub fn is_compatible(&self) -> Result<(), Incompatibility> {
         let library_version = crate::version();
-        if self.version.index_format_version < 4
```
fulmicoton, line 106: nitpick

```
const SUPPORTED_INDEX_FORMAT_VERSION_RANGE: Range<u32> = INDEX_FORMAT_OLDEST_SUPPORTED_VERSION..=INDEX_FORMAT_VERSION;

if !SUPPORTED_VERSION_RANGE.contains(&self.version.index_format_version) {
   // ... 
}
```

fulmicoton (approved):

PR #2238, opened 2023-11-03
remove JsonTermWriter

PSeitz:


common/src/json_path_writer.rs
```diff
@@ -40,7 +54,9 @@ impl JsonPathWriter {
         let len_path = self.path.len();
         self.indices.push(len_path);
         if !self.path.is_empty() {
-            self.path.push_str(JSON_PATH_SEGMENT_SEP_STR);
+            unsafe {
+                self.path.as_mut_vec().push(JSON_PATH_SEGMENT_SEP);
```
fulmicoton: is the unsafe changing anything? If the assembly is the same i would rather use the non unsafe code.

fulmicoton: https://godbolt.org/z/s1bf4s9zW

append_char does the same as byte.

PSeitz: good point, removed the unsafe

src/core/json_utils.rs
```diff
@@ -369,275 +359,106 @@ pub(crate) fn encode_column_name(
...
-    pub fn set_str(&mut self, text: &str) {
-        self.close_path_and_set_type(Type::Str);
-        self.term_buffer.append_bytes(text.as_bytes());
+pub fn term_from_json_paths<'a>(
+    json_field: Field,
+    paths: impl Iterator<Item = &'a str>,
+    expand_dots_enabled: bool,
+) -> Term {
+    let mut json_path = JsonPathWriter::with_expand_dots(expand_dots_enabled);
+    for patho in paths {
```
fulmicoton: patho?

src/core/json_utils.rs
```diff
@@ -248,79 +247,70 @@ fn index_json_value<'a, V: Value<'a>>(
...
+/// Tries to infer a JSON type from a string.
+pub fn convert_to_fast_value_and_set(term: &Term, phrase: &str) -> Option<Term> {
+    let mut term = term.clone();
     if let Ok(dt) = OffsetDateTime::parse(phrase, &Rfc3339) {
         let dt_utc = dt.to_offset(UtcOffset::UTC);
-        return Some(set_fastvalue_and_get_term(
-            json_term_writer,
-            DateTime::from_utc(dt_utc),
-        ));
+        term.append_type_and_fast_value(DateTime::from_utc(dt_utc));
```
fulmicoton, line 275: I suspect this is a bug or a bug in waiting.

For this to work, we need term to not contain any value at this point.

src/core/json_utils.rs
```diff
@@ -248,79 +245,57 @@ fn index_json_value<'a, V: Value<'a>>(
     }
 }
 
-// Tries to infer a JSON type from a string.
-pub fn convert_to_fast_value_and_get_term(
-    json_term_writer: &mut JsonTermWriter,
-    phrase: &str,
-) -> Option<Term> {
+/// Tries to infer a JSON type from a string and append it to the term.
```
fulmicoton, line 257:
```suggestion
/// Tries to infer a JSON type from a string and append it to the term.
///
/// Disclaimer
/// The term is expected to not contain the type byte nor the value.
```

src/core/json_utils.rs
```diff
@@ -248,79 +245,57 @@ fn index_json_value<'a, V: Value<'a>>(
...
+pub(crate) fn convert_to_fast_value_and_append(term: &Term, phrase: &str) -> Option<Term> {
```
fulmicoton: the prototype and/or the impl is a bit silly / misleading.
We pass a reference but always start by cloning it... even though we sometime return None in the end.

src/schema/term.rs
```diff
@@ -182,6 +178,15 @@ impl Term {
         self.0.extend(value.to_be_bytes().as_ref());
     }
 
+    /// Append a string type marker + string to a term.
+    /// This is used in JSON type to append a str after the path.
+    ///
+    /// It will not clear existing bytes.
+    pub(crate) fn append_str(&mut self, val: &str) {
```
fulmicoton: this is a footgun.

It is named like `append_bytes` but acts as `append_type_and_fast_value`

fulmicoton (changes requested):
This PR contains several pitfalls. It cannot be merged as is.

src/core/json_utils.rs
PSeitz: I cloned it, because it's not performance critical and simplifies the code

src/core/json_utils.rs
```diff
@@ -248,79 +245,50 @@ fn index_json_value<'a, V: Value<'a>>(
...
 
-// Tries to infer a JSON type from a string.
-pub fn convert_to_fast_value_and_get_term(
-    json_term_writer: &mut JsonTermWriter,
-    phrase: &str,
-) -> Option<Term> {
+/// Tries to infer a JSON type from a string and append it to the term.
+///
+/// The term must be json + JSON path.
+pub(crate) fn convert_to_fast_value_and_append(mut term: Term, phrase: &str) -> Option<Term> {
```
fulmicoton:
```suggestion
pub(crate) fn convert_to_fast_value_and_append_to_json_term(mut term: Term, phrase: &str) -> Option<Term> {
```

src/core/json_utils.rs
```diff
@@ -248,79 +245,50 @@ fn index_json_value<'a, V: Value<'a>>(
...
+    assert_eq!(
```
fulmicoton, line 264: if it hurts perf, you can use debug_assert

src/schema/term.rs
```diff
@@ -170,6 +162,10 @@ impl Term {
         self.set_bytes(val.to_u64().to_be_bytes().as_ref());
     }
 
+    /// Append a type marker + fast value to a term.
+    /// This is used in JSON type to append a fast value after the path.
+    ///
+    /// It will not clear existing bytes.
     pub(crate) fn append_type_and_fast_value<T: FastValue>(&mut self, val: T) {
         self.0.push(T::to_type().to_code());
```
fulmicoton, line 169: can we have debug assert here too?

src/schema/term.rs
```diff
@@ -182,6 +178,15 @@ impl Term {
         self.0.extend(value.to_be_bytes().as_ref());
     }
 
+    /// Append a string type marker + string to a term.
+    /// This is used in JSON type to append a str after the path.
+    ///
+    /// It will not clear existing bytes.
+    pub(crate) fn append_type_and_str(&mut self, val: &str) {
+        self.0.push(Type::Str.to_code());
```
fulmicoton, line 185: can we have debug assert here too?

src/core/json_utils.rs
fulmicoton: I'd prefer to have the assert in the inner `append_type_and_...` methods.

src/schema/term.rs
PSeitz: In indexing we don't have a path, but a path id, so the debug_assert doesn't work here. It's also atypical for the Term methods, they are just building blocks, and the caller has to know the Term state

fulmicoton: the comment made me think it was only for json. So it is not?

PSeitz: It is, but there are two JSON currently 
Indexing: [FieldId][PathId][Value]
Query: [FieldId][JsonPath][JsonPathEnd][Value]

When we switch to `[PathId][Value]` for all data during Indexing, we can split `Term` into `Term` and `IndexingTerm` with a clearer API.

fulmicoton (approved):

PR #2239, opened 2023-11-05
Fixing functional tests.

fulmicoton:


adamreichold:
The failure to build `SplitCompoundWordsFilter` seems to be a Rust nightly bug, i.e. it does build using stable and the type `AhoCorasick` is still definitely `Send + Sync` in the current stable version of that crate. I reported this upstream at https://github.com/rust-lang/rust/issues/117602

fulmicoton:
ah thanks @adamreichold !

PR #2242, opened 2023-11-08
add sstable ord_to_term benchmark

PSeitz:
```
ord_to_term_suffix      
                        time:   [6.3816 µs 6.4004 µs 6.4169 µs]

open_and_ord_to_term_suffix
                        time:   [15.715 ms 15.766 ms 15.822 ms

ord_to_term             
                        time:   [25.292 µs 25.307 µs 25.323 µs]

open_and_ord_to_term    
                        time:   [3.5459 ms 3.5549 ms 3.5643 ms]

```

### DIFF
--- a/sstable/Cargo.toml
+++ b/sstable/Cargo.toml
@@ -24,3 +24,8 @@ rand = "0.8"
 [[bench]]
 name = "stream_bench"
 harness = false
+
+[[bench]]
+name = "ord_to_term"
+harness = false
+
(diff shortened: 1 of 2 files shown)


trinity-1686a:
I assume this is related to https://github.com/quickwit-oss/quickwit/issues/3964

the `open_and_*` benches are suspiciously fast compared to what I saw there (an order of magnitude faster). My first guess as to why is that the key is on average much smaller (that's an sstable on json field, so the keys are pretty long), but just making the prefix longer doesn't impact the result enough that I'm confident it's the explanation

trinity-1686a (approved):

fulmicoton (approved):
Awesome! This is not exactly the most relevant dataset but still better than nothing.

PR #2246, opened 2023-11-10
multilayer sstable

### LINKED ISSUE #1948: SSTableIndex could be a multi layer SSTable

trinity-1686a:
fix #1948 
related to https://github.com/quickwit-oss/quickwit/issues/3964

implement a new file format for sstable which allows much faster opening of very large table, at the cost of somewhat slower individual queries to the table
current results:
```
before:
        ord_to_term_suffix           time:   [15.442 µs 15.452 µs 15.464 µs]
        open_and_ord_to_term_suffix  time:   [37.591 ms 37.638 ms 37.688 ms]
        ord_to_term                  time:   [65.466 µs 65.483 µs 65.503 µs]
        open_and_ord_to_term         time:   [8.6174 ms 8.6311 ms 8.6474 ms]
after:
        ord_to_term_suffix           time:   [41.148 µs 41.222 µs 41.320 µs]
        open_and_ord_to_term_suffix  time:   [81.646 µs 81.663 µs 81.679 µs]
        ord_to_term                  time:   [92.304 µs 92.331 µs 92.360 µs]
        open_and_ord_to_term         time:   [99.245 µs 99.653 µs 100.41 µs]
```

### DIFF
--- a/columnar/src/tests.rs
+++ b/columnar/src/tests.rs
@@ -26,7 +26,7 @@ fn test_dataframe_writer_str() {
     assert_eq!(columnar.num_columns(), 1);
     let cols: Vec<DynamicColumnHandle> = columnar.read_columns("my_string").unwrap();
     assert_eq!(cols.len(), 1);
-    assert_eq!(cols[0].num_bytes(), 87);
+    assert_eq!(cols[0].num_bytes(), 99);
 }
 
 #[test]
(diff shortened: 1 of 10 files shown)


fulmicoton:
Can you try an fst table? (It should faster... build times should be impacted however.)

trinity-1686a:
superseded by #2268

Issue #1787, opened 2023-01-13
Implement extended stats aggregation.


guilload:
[Extended stats aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-extendedstats-aggregation.html)

Look into using the [parallel version of Welford's online algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm) for computing the unbiased variance.

ExpressGradient:
Hi there, I'm interested in working on this issue and would appreciate some guidance on getting started. I have some experience with Rust and am familiar with the concept of extended stats aggregation. Please let me know how I can contribute to this project. Thanks!

PSeitz:
Hi,

all the aggregations in tantivy are in the aggregation folder:
https://github.com/quickwit-oss/tantivy/tree/main/src/aggregation

The aggregation docs (latest release, not main branch. better build that locally with `cargo doc`) 
https://docs.rs/tantivy/latest/tantivy/aggregation/index.html

For the extended stats, it's probably a good idea to use the normal stats as a blueprint. There are also tests in that file.
https://github.com/quickwit-oss/tantivy/blob/main/src/aggregation/metric/stats.rs

Connecting a new aggregation is pretty straightforward, just extend the request and result enums ... and then fix all the compile errors.
https://github.com/quickwit-oss/tantivy/blob/main/src/aggregation/agg_req.rs#L229
https://github.com/quickwit-oss/tantivy/blob/main/src/aggregation/agg_result.rs#L84

Also feel free to drop by in discord to ask questions: https://discord.gg/MT27AG5EVE

PSeitz-dd:
https://github.com/quickwit-oss/tantivy/pull/2247

PR #2247, opened 2023-11-11
Issue 1787 extended stats

giovannicuccu:
Here is a pull request for #1787 
Some notes:
- the "simple" stats are calculated via ExtendedStats (with the same logic used for Min, Max, etc)
- the tests use an approximate comparison using an extended versione of the macro `assert_nearly_equals` which accepts an extra parameter specifying the max allowed difference between the two values
- The algorithm for calculating the variance is Welford's online algorithm with a modification for calculating the mean since the code can use the sum (there is a comment in the code)
- The sum is computed using the Kahan algorithm (as in elastic search) 
- The value sum_of_squares of extended stats is the same value computed by elastic search even if  [sum of squares](https://en.wikipedia.org/wiki/Total_sum_of_squares) has a different meaning. The "real" sum of squares is in the code so if you decide to expose it the change is straightforward
- For a specific data set I made a comparison with the numbers returned by elastic search, here is a simple summary:

|   |  variance | variance_sampling |  std_deviation | std_deviation_sampling | lower | lower_sampling | upper | upper_sampling | sum of squares |
| - |  -------- | ------------------ | -------------- | ------------------------ | ----- | ---------------- | ------- | ---------------- | -------- |
| ES | 9.138888888888891 | 10.966666666666669 | 3.023059524536176 | 3.3115957885386114 | -0.8794523824056855 | -1.4565249104105558 | 11.21278571573902 | 11.78985824374389 | 215.0 |
| Tantivy |  9.138888888888888 | 10.966666666666663 | 3.023059524536176 | 3.311595788538611 | -0.8794523824056837 | -1.4565249104105549 | 11.212785715739017 | 11.78985824374389 | 215.0 |

- I also did a comparison using values where the mean is very similar to the variance and Tantivy calculations are more precise (I used Excel as a reference for obtaining the value) because Elastic Search does not implement the Welford Algorithm.


fulmicoton:
@PSeitz can you review?

src/aggregation/metric/stats.rs
```diff
@@ -81,42 +128,253 @@ impl Stats {
...
+    }
+}
+
 /// Intermediate result of the stats aggregation that can be combined with other intermediate
 /// results.
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+
+#[derive(Default, Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct IntermediateStats {
+    stats: IntermediateExtendedStats,
```
PSeitz: I think the memory overhead for e.g. terms -> avg aggregation is too high

Performance may also be impacted
`cargo +nightly bench --features unstable aggregation::agg_bench`

giovannicuccu:
This is just a ping. I'm working on it, I created a dataset and a simple application that uses several gb of memory for computing some statistics. I'm using the main branch as the memory consumption reference and working on a solution for keeping it the same.

PSeitz:
> This is just a ping. I'm working on it, I created a dataset and a simple application that uses several gb of memory for computing some statistics. I'm using the main branch as the memory consumption reference and working on a solution for keeping it the same.

Just as a heads up, please no macros. Generics would be fine, if the complexity stays low. Otherwise we just can keep it seperate

giovannicuccu:
Some notes:
- I used a `Box<ExtendedStats>` because the size of the enum without the box was impacting the memory usage of all the stat searches. In my experiments a term aggregation without h the box wsa consuming about 15 more ram than the actual version.
- I splitted Stats and ExtendedStats, the latter is using the former
* I introduced a new intermediate struct that makes the `SegmentStatsCollector` more flexible making it work with both stats and extendedstats; in future, if needed, you can implement the single stats (min,max,avg) independently with a minimum effort.

giovannicuccu:
Hello, I'd like to know how to proceed with this pull request, it has been pending for a while.

PSeitz:
> Hello, I'd like to know how to proceed with this pull request, it has been pending for a while.

Sorry I forgot about this PR. It's fine to ping occasionally. Can you resolve the merge conflicts?
I consider adding a benchmark that also reports memory consumption for the aggregation benchmarks

giovannicuccu:
Merge is done, do you have an example for benchmark that also reports memory consumption? I ran `cargo bench` but saw no indication of memory consumption.

PSeitz:
> Merge is done, do you have an example for benchmark that also reports memory consumption? I ran `cargo bench` but saw no indication of memory consumption.

Thanks! memory reporting is coming with this PR:
https://github.com/quickwit-oss/tantivy/pull/2378

giovannicuccu:
Hi, quick update: I added an extended stats bench in the corresponding file, my fork is merged with main; I'm waiting for [#2378](https://github.com/quickwit-oss/tantivy/pull/2378) to be merged into main before aligning with with it and then pushing it to this pr.

PSeitz:
https://github.com/quickwit-oss/tantivy/pull/2378 is now merged

giovannicuccu:
new version merged with main plus an update bench for extendedstats

PSeitz:
The memory consumption increased by 10% for those two benchmarks. I don't think you should pay for extended stats if you are not using it
```
terms_many_with_avg_sub_agg                     Memory: 29.0 MB  (+9.01%)    Avg: 131.7313ms  (-3.79%)     Median: 125.7534ms  (+3.04%)    [118.0482ms .. 180.8836ms]
terms_many_json_mixed_type_with_sub_agg_card    Memory: 44.0 MB  (+9.95%)    Avg: 176.1016ms  (+13.39%)    Median: 161.8976ms  (+8.52%)    [151.0041ms .. 251.0137ms]
```

giovannicuccu:
Hi, I dug into the issue and there 2 reason for the additional memory consumption:
1. I created a unified SegmentStatsCollector (both for stats and extendedstats)  that requires more memory. I developed a different solution i.e. a dedicated SegmentExtendedStatsCollector that duplicates part of the code, but does not allocate more memory if you use the regular stats. This changes reduces the memory increase to 5% (i.e. it halves the memory increment from main)
2. I modified the stats to use Kahan summation aglorithm, because it's more precise. The new algorithm requires an extra f64, that is responsible for the remaining 5% memory increase. I can revert the stats  implementation, but you are losing something in terms of precision (Elasticsearch uses Kahan summation).

Should I revert even the modification on stats in order to restore the previous memory usage?

Thanks

src/aggregation/metric/stats.rs
```diff
@@ -80,6 +129,110 @@ impl Stats {
     }
 }
 
+/// Extended stats contains a collection of statistics
+/// they extends stats adding variance, standard deviation
+/// and bound informations
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct ExtendedStats {
```
PSeitz: Can you move the extended stats stuff to `extended_stats.rs`?

src/aggregation/metric/stats.rs
```diff
@@ -143,12 +300,266 @@ impl IntermediateStats {
...
+            sum_of_squares: 0.0,
+            sum_of_squares_elastic: 0.0,
+            delta_sum_for_squares_elastic: 0.0,
+            mean: 0.0,
+            sigma: sigma.unwrap_or(2.0),
+        }
+    }
+    /// Merges the other stats intermediate result into self.
+    pub fn merge_fruits(&mut self, other: IntermediateExtendedStats) {
+        if other.intermediate_stats.count != 0 {
```
PSeitz: this should be an early exit not nesting everything

src/aggregation/metric/stats.rs
```diff
@@ -143,12 +300,266 @@ impl IntermediateStats {
...
+            if self.intermediate_stats.count == 0 {
```
PSeitz: this should be an mem::replace + early exit and not nesting everything

src/aggregation/metric/stats.rs
```diff
@@ -143,12 +300,266 @@ impl IntermediateStats {
...
+                    self.collect(other.intermediate_stats.sum);
+                } else if self.intermediate_stats.count == 1 {
+                    let sum = self.intermediate_stats.sum;
+                    self.intermediate_stats = other.intermediate_stats;
+                    self.sum_of_squares = other.sum_of_squares;
+                    self.sum_of_squares_elastic = other.sum_of_squares_elastic;
+                    self.mean = other.mean;
+                    self.delta_sum_for_squares_elastic = other.delta_sum_for_squares_elastic;
+                    self.collect(sum);
+                } else {
```
PSeitz: these cases are way to special and should be handled by the generic version

src/aggregation/metric/stats.rs
```diff
@@ -143,12 +300,266 @@ impl IntermediateStats {
...
+            None
+        } else {
+            Some(self.mean)
+        };
+        let sum_of_squares = if self.intermediate_stats.count == 0 {
+            None
+        } else {
+            Some(self.sum_of_squares_elastic)
+        };
+        let variance = if self.intermediate_stats.count <= 1 {
```
PSeitz: all these different but same ifs are error prone and should be collapsed into one condition

src/aggregation/metric/stats.rs
giovannicuccu: Yes, but I need to change `IntermediateStats` visibility members and his `collect` method. if it's ok, I can refactor, do you prefer me adding accessor methods o declaring the fields as `pub`?

src/aggregation/metric/stats.rs
giovannicuccu: Ok, my code was inspired by the IntermediateStats one 

https://github.com/quickwit-oss/tantivy/blob/aa26ff50298e8a6a590faf9a66bdd7d2a684ec97/src/aggregation/metric/stats.rs#L119-L133

should I refactor that one as well?

src/aggregation/metric/stats.rs
giovannicuccu: Sorry, but I don't understand it, I try to better clarify the current situation. 

I created a version with a new segment collector for extended stats which reduces the extra memory increase, but I did not pushed yet, because I'm waiting for a response about kahan summation. In the new version the generic component is gone, so I cannot apply the change.

If you were referring to some other modification I ask you for some more details.

src/aggregation/metric/stats.rs
PSeitz: you can add `pub(crate)`

src/aggregation/metric/stats.rs
PSeitz: This is also not nice, but there are all the same check so it's easy to proofread

src/aggregation/metric/stats.rs
PSeitz: yes, we can use kahan summation.

What I mean is that these branches, they complexity they add, instead using the version that can handle all counts, needs to be justified. 

```rust
if other.intermediate_stats.count == 1 {
    self.collect(other.intermediate_stats.sum);
} else if self.intermediate_stats.count == 1 {
    let sum = self.intermediate_stats.sum;
    self.intermediate_stats = other.intermediate_stats;
    self.sum_of_squares = other.sum_of_squares;
    self.sum_of_squares_elastic = other.sum_of_squares_elastic;
    self.mean = other.mean;
    self.delta_sum_for_squares_elastic = other.delta_sum_for_squares_elastic;
    self.collect(sum);
```

giovannicuccu:
Hi, thanks a lot for your review, this submission contains the code modifications.
This version still shows a memory usage increase (halved since the last submission) because of the introduction of Kahan summation in stats.

src/lib.rs
```diff
@@ -412,6 +412,23 @@ pub mod tests {
                 }
             }
         }};
+        ($left:expr, $right:expr, $epsilon:expr) => {{
+            match (&$left, &$right, &$epsilon) {
+                (left_val, right_val, epsilon_val) => {
+                    let diff = (left_val - right_val).abs();
+
+                    if diff > *epsilon_val {
```
PSeitz, line 407: can the code above call this part with epsilon `0.0005`

giovannicuccu: I'm not sure I got it. Are asking for a change of epsilon in extended_stats tests or for a change in the macro using a fixed value?

PSeitz: the macro has two code paths, one with a parameter and one with epsilon `0.0005`. The `0.0005` path can call the  variant with the parameter

giovannicuccu: ok, done, thank you.

src/aggregation/metric/extended_stats.rs
```diff
@@ -0,0 +1,1177 @@
...
+
+impl Default for IntermediateExtendedStats {
+    fn default() -> Self {
+        Self {
+            intermediate_stats: IntermediateStats::default(),
+            sum_of_squares: 0.0,
+            sum_of_squares_elastic: 0.0,
+            delta_sum_for_squares_elastic: 0.0,
+            mean: 0.0,
+            sigma: 2.0,
```
PSeitz, line 196: can you document why this is 2.0 ?

src/aggregation/metric/extended_stats.rs
```diff
@@ -0,0 +1,1177 @@
...
+pub struct IntermediateExtendedStats {
+    intermediate_stats: IntermediateStats,
+    /// The number of extracted values.
+    // The sum of square values, it's referred as M2 in Welford's online algorithm
+    sum_of_squares: f64,
+    // The sum of square values as computed by elastic search
+    sum_of_squares_elastic: f64,
+    /// delta for sum of squares  as computed by elastic search needed for the Kahan algorithm
+    delta_sum_for_squares_elastic: f64,
+    // The mean an intermediate value need for calculating the variance
```
PSeitz: That comment seems broken

src/aggregation/metric/extended_stats.rs
```diff
@@ -0,0 +1,1177 @@
...
+    /// The max value of the fast field values.
+    pub max: Option<f64>,
+    /// The average of the fast field values. `None` if count equals zero.
+    pub avg: Option<f64>,
+    /// The sum of squares of the fast field values. `None` if count equals zero.
+    pub sum_of_squares: Option<f64>,
+    /// The variance of the fast field values. `None` if count is less then 2.
+    pub variance: Option<f64>,
+    /// The variance population of the fast field values, always equal to variance. `None` if count
+    /// is less then 2.
```
PSeitz: less than

`if count is 0 or 1` is maybe easier to read

src/aggregation/metric/extended_stats.rs
```diff
@@ -0,0 +1,1177 @@
...
+            "max" => Ok(self.max),
+            "avg" => Ok(self.avg),
+            "variance" => Ok(self.variance),
+            "variance_sampling" => Ok(self.variance_sampling),
+            "variance_population" => Ok(self.variance_population),
+            "sum_of_squares" => Ok(self.sum_of_squares),
+            "std_deviation" => Ok(self.std_deviation),
+            "std_deviation_sampling" => Ok(self.std_deviation_sampling),
+            "std_deviation_population" => Ok(self.std_deviation_population),
+            "std_deviation_bounds.lower" => Ok(self
```
PSeitz, line 137: this should match the elastic search name, not the internal name of the property

src/aggregation/metric/extended_stats.rs
```diff
@@ -0,0 +1,1177 @@
...
+        };
+        Box::new(ExtendedStats {
+            count: self.intermediate_stats.count,
+            sum: self.intermediate_stats.sum,
+            min,
+            max,
+            avg,
+            sum_of_squares,
+            variance,
+            variance_population: variance,
```
PSeitz, line 289: why are there two fields with the same value?

src/aggregation/metric/extended_stats.rs
```diff
@@ -0,0 +1,1177 @@
...
+            (None, None)
+        } else {
+            (
+                Some(self.sum_of_squares / self.intermediate_stats.count as f64),
+                Some(self.sum_of_squares / (self.intermediate_stats.count - 1) as f64),
+            )
+        };
+        let std_deviation = variance.map(|v| v.sqrt());
+        let std_deviation_sampling = variance_sampling.map(|v| v.sqrt());
+        let std_deviation_bounds = if std_deviation.is_none() {
```
PSeitz: can you use `if let Some(std_deviation, std_deviation_sampling) = ` instead of `is_some` + unwrap?

src/aggregation/metric/extended_stats.rs
giovannicuccu: Perhaps I'm missing something, but the names seem to match the ElasticSearch ones https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-extendedstats-aggregation.html

src/aggregation/metric/extended_stats.rs
giovannicuccu: Basically because of ElasticSearch implementation, the documentation (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-extendedstats-aggregation.html) states: `The std_deviation and variance are calculated as population metrics so they are always the same as std_deviation_population and variance_population respectively.`

giovannicuccu:
This version should reflect the changes you requested with the exception of the stat labels; they seem to  me the same as ElasticSearch

PSeitz (approved):
Thanks! nice job

PR #2249, opened 2023-11-12
docid deltas while indexing

PSeitz:
storing deltas is especially helpful for repetitive data like logs.
In those cases, recording a doc on a term costed 4 bytes instead of 1
byte now.

HDFS Indexing 1.1GB Total memory consumption:
Before:  760 MB
Now:     590 MB

### DIFF
--- a/src/postings/recorder.rs
+++ b/src/postings/recorder.rs
@@ -82,19 +82,10 @@ pub(crate) trait Recorder: Copy + Default + Send + Sync + 'static {
 }
 
 /// Only records the doc ids
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Default)]
 pub struct DocIdRecorder {
     stack: ExpUnrolledLinkedList,
     current_doc: DocId,
-}
-
-impl Default for DocIdRecorder {
-    fn default() -> Self {
-        DocIdRecorder {
-            stack: ExpUnrolledLinkedList::default(),
-            current_doc: u32::MAX,
-        }
-    }
 }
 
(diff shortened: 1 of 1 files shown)


src/postings/recorder.rs
```diff
@@ -125,18 +117,22 @@ impl Recorder for DocIdRecorder {
         let (buffer, doc_ids) = buffer_lender.lend_all();
         self.stack.read_to_end(arena, buffer);
         // TODO avoid reading twice.
+        let mut prev_doc = 0;
         if let Some(doc_id_map) = doc_id_map {
-            doc_ids.extend(
-                VInt32Reader::new(&buffer[..])
-                    .map(|old_doc_id| doc_id_map.get_new_doc_id(old_doc_id)),
-            );
+            doc_ids.extend(VInt32Reader::new(&buffer[..]).map(|delta_doc_id| {
```
adamreichold: Usage of `prev_doc` here looks like [`scan`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.scan) instead of `map` could make this more obvious.

fulmicoton: + 1 in favor of scan of a loop.

Having `prev_doc` outside of the if-statement is not a good idea. Having it in both statement makes the code easier   to (proof)read.

fulmicoton (approved):

src/postings/recorder.rs
PSeitz: switched to scan, but we can't apply the same technique in the other cases, because they contain a mix of delta and non delta encoded values

PR #2253, opened 2023-11-14
Fix bug occuring when merging JSON object indexed with positions.

### LINKED ISSUE #2251: Merge bug for objects with position enabled and including at least one number
minimal reproducing test.

```rust
#[test]
fn test_positions_dynamic_mode() {
    let field = schema_builder.add_json_field("dynamic", TEXT);
    let mut writer: IndexWriter = index.writer_for_tests().unwrap();
    let mut merge_policy = LogMergePolicy::default();
    merge_policy.set_min_num_segments(2);
    writer.set_merge_policy(Box::new(merge_policy));
    // Here a string would work.
    let doc_json = r#"{"tenant_id":75}"#;
    let vals = serde_json::from_str(doc_json).unwrap();
    let mut doc = TantivyDocument::default();
    doc.add_object(field, vals);
    writer.add_document(doc.clone()).unwrap();
    writer.commit().unwrap();
    writer.add_document(doc.clone()).unwrap();
    writer.commit().unwrap();
    writer.wait_merging_threads().unwrap();
    assert_eq!(reader.searcher().segment_readers().len(), 1);
}
```

fulmicoton:
In JSON Object field the presence of term frequencies depend on the field.
Typically, a string with postiions indexed will have positions while numbers won't.

The presence or absence of term freqs for a given term is unfortunately encoded in a very passive way.

It is given by the presence of extra information in the skip info, or the lack of term freqs after decoding vint blocks.

Before, after writing a segment, we would encode the segment correctly (without any term freq for number in json object field). However during merge, we would get the default term freq=1 value. (this is default in the absence of encoded term freqs)

The merger would then proceed and attempt to decode 1 position when there are in fact none.

This PR requires to explictly tell the posting serialize whether term frequencies should be serialized for each new term.

Closes #2251

### DIFF
--- a/src/core/tests.rs
+++ b/src/core/tests.rs
@@ -1,12 +1,13 @@
 use crate::collector::Count;
 use crate::directory::{RamDirectory, WatchCallback};
-use crate::indexer::NoMergePolicy;
+use crate::indexer::{LogMergePolicy, NoMergePolicy};
+use crate::json_utils::JsonTermWriter;
 use crate::query::TermQuery;
-use crate::schema::{Field, IndexRecordOption, Schema, INDEXED, STRING, TEXT};
+use crate::schema::{Field, IndexRecordOption, Schema, Type, INDEXED, STRING, TEXT};
 use crate::tokenizer::TokenizerManager;
 use crate::{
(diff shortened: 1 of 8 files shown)


trinity-1686a (approved):
lgtm

PR #2257, opened 2023-11-15
reduce number of allocations

PSeitz:
Explanation makes up around 75% of all allocations (numbers not perf) when indexing.
It's created during serialization but not called.

- Make Explanation optional in BM25
- Avoid allocations when using Explanation

### DIFF
--- a/src/query/explanation.rs
+++ b/src/query/explanation.rs
@@ -1,6 +1,6 @@
 use std::fmt;
 
-use serde::Serialize;
+use serde::{Serialize, Serializer};
 
 use crate::{DocId, Score, TantivyError};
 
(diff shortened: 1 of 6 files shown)


src/query/explanation.rs
```diff
@@ -16,12 +16,26 @@ pub(crate) fn does_not_match(doc: DocId) -> TantivyError {
...
-    context: Vec<String>,
+    description: StaticOrDynamic,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    details: Option<Vec<Explanation>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    context: Option<Vec<String>>,
+}
+#[derive(Clone)]
+enum StaticOrDynamic {
+    Static(&'static str),
```
fulmicoton: Cow<'static> does not do that?

fulmicoton (approved):

src/query/explanation.rs
PSeitz: yes, that's better

PR #2268, opened 2023-11-22
use fst for sstable index

trinity-1686a:
related to https://github.com/quickwit-oss/quickwit/issues/3964

makes loading an sstable index basically free (previously you'd decode a datastructure, which could take a non-negligible amount of time, in practice ~200ms for a 125m term table, with keys encoding json paths).


sstable/README.md
```diff
@@ -89,33 +89,69 @@ Note: as the SSTable does not support redundant keys, there is no ambiguity betw
...
-+-------+-------+-----+-------------+---------+---------+
-| Block | Block | ... | IndexOffset | NumTerm | Version |
-+-------+-------+-----+-------------+---------+---------+
-|----( # of blocks)---|
++-----+----------------+-------------+-------------+---------+---------+
+| Fst | BlockAddrStore | StoreOffset | IndexOffset | NumTerm | Version |
++-----+----------------+-------------+-------------+---------+---------+
 ```
-- Block(SSTBlock): uses IndexValue for its Values format
+- Fst(Fst): finit state transducer mapping keys to a block number
```
fulmicoton:
```suggestion
- Fst(Fst): finite state transducer mapping keys to a block number
```

(I think I keep mispronouncing finite but I think this is "finite")

sstable/README.md
```diff
@@ -89,33 +89,69 @@ Note: as the SSTable does not support redundant keys, there is no ambiguity betw
...
-- BlockLen(VInt): length of the block
-- FirstOrdinal(VInt): ordinal of the first element in the given block
+### BlockAddrStore
+
++---------+-----------+-----------+-----+-----------+-----------+-----+
+| MetaLen | BlockMeta | BlockMeta | ... | BlockData | BlockData | ... |
++---------+-----------+-----------+-----+-----------+-----------+-----+
+          |---------(N blocks)----------|---------(N blocks)----------|
+
+- MetaLen(u64): lenght of the BlockMeta section
```
fulmicoton:
```suggestion
- MetaLen(u64): length of the BlockMeta section
```

sstable/README.md
```diff
@@ -89,33 +89,69 @@ Note: as the SSTable does not support redundant keys, there is no ambiguity betw
...
+- BlockMeta(BlockAddrBlockMetadata): metadata to seek through BlockData
+- BlockData(CompactedBlockAddr): bitpacked per block metadata
+
+### BlockAddrBlockMetadata
+
++--------+------------+---------------+-----------+-------------+-------------------+-----------------+----------+
+| Offset | RangeStart | FirstOrdianal | RangeSlop | OrdinalSlop | FirstOrdinalNBits | RangeStartNBits | BlockLen |
```
fulmicoton:
```suggestion
| Offset | RangeStart | FirstOrdinal | RangeSlop | OrdinalSlop | FirstOrdinalNBits | RangeStartNBits | BlockLen |
```

sstable/src/dictionary.rs
```diff
@@ -180,24 +183,42 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
...
-        let sstable_index = SSTableIndex::load(sstable_index_bytes)
-            .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "SSTable corruption"))?;
+
+        let sstable_index = match version {
+            2 => SSTableIndex::V2(
+                crate::sstable_index_v2::SSTableIndex::load(sstable_index_bytes).map_err(|_| {
+                    io::Error::new(io::ErrorKind::InvalidData, "SSTable corruption")
+                })?,
+            ),
+            crate::SSTABLE_VERSION => {
```
fulmicoton: I suspect it makes more sense to put the actual value here, no?
```suggestion
            3 => {
```

sstable/src/dictionary.rs
```diff
@@ -180,24 +183,42 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
...
+                let (sstable_index_bytes, mut footerv3_len_bytes) = sstable_index_bytes.rsplit(8);
+                let store_offset = u64::deserialize(&mut footerv3_len_bytes)?;
+                if store_offset != 0 {
```
fulmicoton, line 201: That's good, but a comment about the `store_offset == 0`  would be helpful here.

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+#[derive(Debug)]
+struct BlockAddrBlockMetadata {
+    offset: u64,
+    ref_block_addr: BlockStartAddr,
+    range_start_slop: u32,
+    first_ordinal_slop: u32,
+    range_start_nbits: u8,
+    first_ordinal_nbits: u8,
+    block_len: u16,
+    // these fields are computed on deserialization, and not stored
```
fulmicoton: great comment

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+            + self.range_start_slop as usize * (inner_offset + 2)
+            - self.range_shift as usize;
+
+        Some(BlockAddr {
+            first_ordinal,
+            byte_range: range_start..range_end,
+        })
+    }
+
+    fn bissect_for_ord(&self, data: &[u8], target_ord: TermOrdinal) -> (u64, BlockAddr) {
```
fulmicoton:
```suggestion
    fn bisect_for_ord(&self, data: &[u8], target_ord: TermOrdinal) -> (u64, BlockAddr) {
```
(one s in english)

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+        let inner_target_ord = target_ord - self.ref_block_addr.first_ordinal;
+        let num_bits = self.num_bits() as usize;
+        let range_start_nbits = self.range_start_nbits as usize;
+        let get_ord = |index| {
+            extract_bits(
+                data,
+                num_bits * index as usize + range_start_nbits,
+                self.first_ordinal_nbits,
+            ) + self.first_ordinal_slop as u64 * (index + 1)
```
fulmicoton:
```suggestion
            ) + self.first_ordinal_slope as u64 * (index + 1)
```

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+}
+
+// TODO move this function to tantivy_common?
+fn extract_bits(data: &[u8], addr_bits: usize, num_bits: u8) -> u64 {
+    use byteorder::{ByteOrder, LittleEndian};
+    assert!(num_bits <= 56);
+    let addr_byte = addr_bits / 8;
+    let bit_shift = (addr_bits % 8) as u64;
+    let val_unshifted_unmasked: u64 = if data.len() >= addr_byte + 8 {
+        LittleEndian::read_u64(&data[addr_byte..][..8])
```
fulmicoton: byte order is a bit overkill for this... the std has all you need.
```
 let b = data[addr_byte..addr_byte + 8].try_into().unwrap();
 u64::from_le_bytes(b)
```

generates the optimal code.

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+    let val_shifted_unmasked = val_unshifted_unmasked >> bit_shift;
+    let mask = (1u64 << u64::from(num_bits)) - 1;
+    val_shifted_unmasked & mask
+}
+
+impl BinarySerializable for BlockAddrBlockMetadata {
+    fn serialize<W: Write + ?Sized>(&self, write: &mut W) -> io::Result<()> {
+        self.offset.serialize(write)?;
+        self.ref_block_addr.serialize(write)?;
+        self.range_start_slop.serialize(write)?;
```
fulmicoton: slope

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+    let slop_iterator = elements.clone();
+    let derivation_iterator = elements;
+
+    let mut min_slop_idx = 1;
+    let mut min_slop_val = 0;
+    let mut min_slop = u32::MAX;
+    let mut max_slop_idx = 1;
+    let mut max_slop_val = 0;
+    let mut max_slop = 0;
+    for (index, value) in slop_iterator {
```
fulmicoton: slope_iterator

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+        }
+        if slop >= max_slop {
+            max_slop = slop;
+            max_slop_idx = index;
+            max_slop_val = value;
+        }
+    }
 
-    type ValueWriter = crate::value::index::IndexValueWriter;
+    // above is an heristic giving the "highest" and "lowest" point. It's imperfect in that in that
```
fulmicoton:
```suggestion
    // above is an heuristic giving the "highest" and "lowest" point. It's imperfect in that in that
```

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+    // 0 = min_slop_val - min_slop_idx * A - B
+    // 0 = max_slop_val - max_slop_idx * A + B
+    //
+    // 0 = min_slop_val + max_slop_val - (min_slop_idx + max_slop_idx) * A
+    // (min_slop_val + max_slop_val) / (min_slop_idx + max_slop_idx) = A
+    //
+    // we actually add some correcting factor to have proper rounding, not truncation.
+
+    let denominator = (min_slop_idx + max_slop_idx) as u64;
+    let final_slop = ((min_slop_val + max_slop_val + denominator / 2) / denominator) as u32;
```
fulmicoton:
```suggestion
    let final_slop = ((min_slop_val + max_slop_val + denominator / 2) / denominator) as u32;
```

would using a decimal slope (not a float but just using a 4 bits as decimal for instance) help here?

`let final_slop = ((min_slop_val + max_slop_val + denominator / 2) * 16 / denominator) as u32;`

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+
+    // we don't solve for B because our choice of point is suboptimal, so it's actually a lower
+    // bound and we need to iterate to find the actual worst value.
+
+    let max_derivation = derivation_iterator.fold(0, |max_derivation, (index, value)| {
+        let derivation = (value as i64 - final_slop as i64 * index as i64).unsigned_abs();
+
+        max_derivation.max(derivation)
+    });
```
fulmicoton: why use .fold() instead of `Iterator::map` and `Iterator::max`?

trinity-1686a: a previous iteration of the code computed 2 different max at once, and I didn't think about using Iterator::max after the refactoring

sstable/src/sstable_index.rs
```diff
@@ -1,56 +1,111 @@
-use std::io::{self, Write};
```
fulmicoton, line 1: should we call this file sstable_index_v3.rs?

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+        (
+            inner_offset,
+            self.deserialize_block_addr(data, inner_offset as usize)
+                .unwrap(),
+        )
+    }
+}
+
+// TODO move this function to tantivy_common?
+fn extract_bits(data: &[u8], addr_bits: usize, num_bits: u8) -> u64 {
```
fulmicoton:
```suggestion
#[inline(always)]
fn extract_bits(data: &[u8], addr_bits: usize, num_bits: u8) -> u64 {
```

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+    use std::cmp::Ordering::*;
+    let mut size = max;
+    let mut left = 0;
+    let mut right = size;
+    while left < right {
+        let mid = left + size / 2;
+
+        let cmp = cmp_fn(mid);
+
+        if cmp == Less {
```
fulmicoton: nitpick: match would be nice here.

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+    let derivation_iterator = elements;
+
+    let mut min_slope_idx = 1;
+    let mut min_slope_val = 0;
+    let mut min_slope = u32::MAX;
+    let mut max_slope_idx = 1;
+    let mut max_slope_val = 0;
+    let mut max_slope = 0;
+    for (index, value) in slope_iterator {
+        let slope = (value / index as u64) as u32;
```
fulmicoton: why is index guaranteed to be non zero? can we comment the contract somewhere?

sstable/src/sstable_index.rs
trinity-1686a: that code is enough to make sure we round in the right direction. I *think* using fixed point integers wouldn't really help as we would still need to use a similar trick when collapsing the fixed point integers to whole numbers

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+        }
+        let len = self.buffer_block_metas.len() as u64;
+        len.serialize(wrt)?;
+        wrt.write_all(&self.buffer_block_metas)?;
+        wrt.write_all(&self.buffer_addrs)?;
+        Ok(())
+    }
+}
+
+fn find_best_slope(elements: impl Iterator<Item = (usize, u64)> + Clone) -> (u32, u8) {
```
fulmicoton: is an empty iterator ok?

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+    fn new() -> Self {
+        BlockAddrStoreWriter {
+            buffer_block_metas: Vec::new(),
+            buffer_addrs: Vec::new(),
+            block_addrs: Vec::with_capacity(STORE_BLOCK_LEN),
+        }
+    }
+
+    fn flush_block(&mut self) -> io::Result<()> {
+        let ref_block_addr = self.block_addrs[0].clone();
```
fulmicoton: why is it non empty?

sstable/src/sstable_index.rs
```diff
@@ -139,51 +278,456 @@ impl SSTableIndexBuilder {
...
+        let mut last_block_addr = self.block_addrs.last().unwrap().clone();
+        last_block_addr.byte_range.end -= ref_block_addr.byte_range.start;
+
+        let (range_start_slope, range_start_nbits) = find_best_slope(
+            self.block_addrs
+                .iter()
+                .map(|block| block.byte_range.start as u64)
+                .chain(std::iter::once(last_block_addr.byte_range.end as u64))
+                .enumerate()
+                .skip(1),
```
fulmicoton: how do we know block addrs contains more than one el?

sstable/src/sstable_index.rs
fulmicoton: what does this return?

sstable/src/sstable_index.rs
trinity-1686a: This implementation was in part copied from libcore. It [said the following](https://doc.rust-lang.org/src/core/slice/mod.rs.html#2791-2793) about that comparison:
>  // The reason why we use if/else control flow rather than match
 // is because match reorders comparison operations, which is perf sensitive.
// This is x86 asm for u8: https://rust.godbolt.org/z/8Y8Pra.

sstable/src/sstable_index.rs
trinity-1686a: we don't need to, `find_best_slope` accept empty iterator

sstable/src/sstable_index.rs
trinity-1686a: it is (added a comment)

sstable/src/sstable_index.rs
trinity-1686a: caller make sure of that, but I changed it so it's always safe to call

sstable/README.md
```diff
@@ -89,33 +89,69 @@ Note: as the SSTable does not support redundant keys, there is no ambiguity betw
...
+- RangeStartDelta(var): RangeStartNBits *bits* of little endian number. See below for decoding
+- FirstOrdinalDelta(var): FirstOrdinalNBits *bits* of little endian number. See below for decoding
+- FinalRangeEnd(var): RangeStartNBits *bits* of integer. See below for decoding
+
+converting a BlockData of index Index and a BlockAddrBlockMetadata to an actual block address is done as follow:
+range\_prediction := RangeStart + Index * RangeSlop;
+range\_derivation := RangeStartDelta - (1 << (RangeStartNBits-1));
+range\_start := range\_prediction + range\_derivation
+
+the same computation can be done for ordinal
```
fulmicoton:
```suggestion
the same computation can be done for ordinal.

Note that `range_derivation` can take negative value. `RangeStartDelta` is just its translation to a positive range.
```

sstable/README.md
```diff
@@ -89,33 +89,69 @@ Note: as the SSTable does not support redundant keys, there is no ambiguity betw
...
+
+
+## SingleBlockSStable
+
+The format used for the index is meant to be compact, however it has a constant cost of arround 70
```
fulmicoton:
```suggestion
The format used for the index is meant to be compact, however it has a constant cost of around 70
```

sstable/src/sstable_index.rs
fulmicoton: I meant: wouldn't it make the decimal slope make the prediction for high indexes and eventually help with the compression.

> as we would still need to use a similar trick when collapsing the fixed point integers to whole numbers

I suspect you forgot there is a multiplication in between.

`truncateroundorwhatever(slope*idx)`
vs
`idx * truncateroundorwhatever(slope)`
will actually give you very different precision.

I would not be surprised if you could shave off one or two bits in your encoding.

sstable/src/sstable_index.rs
fulmicoton: I am seeing the same assembly code. Maybe the compiler improved?
anyway no need to fix.

fulmicoton (approved):

sstable/src/sstable_index.rs
trinity-1686a: ftr: I tried and on my sample dataset, there was a negligible gain of about 0.1%. Either I did something wrong and I can't figure out what, or the impact is negligible

PR #2276, opened 2023-12-11
RFC: Transparently fetch multiple documents grouped per block

adamreichold:
My personal take on the discussion in #2252, i.e. expose enough information and control to schedule concurrent reads into independent block caches as to maximize cache effectiveness and then expose this as `Searcher::docs_async` which yields an iterator of futures which user code can then poll concurrently or in parallel as it sees fit.

### DIFF
--- a/src/core/searcher.rs
+++ b/src/core/searcher.rs
@@ -1,4 +1,6 @@
 use std::collections::BTreeMap;
+#[cfg(feature = "quickwit")]
+use std::future::Future;
 use std::sync::Arc;
 use std::{fmt, io};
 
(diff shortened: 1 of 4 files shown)


src/core/searcher.rs
```diff
@@ -112,6 +115,70 @@ impl Searcher {
...
+    ///
+    /// The caller is expected to poll these futures concurrently (e.g. using `FuturesUnordered`)
+    /// or in parallel (e.g. using `JoinSet`) as fits best with the given use case, i.e. whether
+    /// it is predominately I/O-bound or rather CPU-bound.
+    #[cfg(feature = "quickwit")]
+    pub fn docs_async<D: DocumentDeserialize>(
+        &self,
+        doc_addresses: impl IntoIterator<Item = DocAddress>,
+    ) -> crate::Result<
+        impl Iterator<Item = impl Future<Output = crate::Result<Vec<(DocAddress, D)>>>> + '_,
```
adamreichold, line 170: The signature is admittedly somewhat unwieldy even though relatively easy to use in practice as the test shows. I wonder if I should add a `no_run` example showing usage with `FuturesUnordered`? (`no_run` to avoid making the example unreadable due to index setup code. Alternatively the setup code could be hidden.)

adamreichold: Added a `no_run` example on how to use this with `FuturesUnordered` and `block_on`.

PR #2279, opened 2023-12-13
Store List of Fields in Segment

PSeitz:
Fiels may be encoded in the columnar storage or in the inverted index
for JSON fields.
Add a new Segment file that contains the list of fields (schema +
encoded)

### DIFF
--- a/src/core/segment_reader.rs
+++ b/src/core/segment_reader.rs
@@ -3,15 +3,14 @@ use std::ops::BitOrAssign;
 use std::sync::{Arc, RwLock};
 use std::{fmt, io};
 
-use fnv::FnvHashMap;
 use itertools::Itertools;
 
 use crate::core::{InvertedIndexReader, Segment, SegmentComponent, SegmentId};
 use crate::directory::{CompositeFile, FileSlice};
 use crate::error::DataCorruption;
 use crate::fastfield::{intersect_alive_bitsets, AliveBitSet, FacetReader, FastFieldReaders};
+use crate::field_list::read_split_fields;
 use crate::fieldnorm::{FieldNormReader, FieldNormReaders};
-use crate::json_utils::json_path_sep_to_dot;
 use crate::schema::{Field, IndexRecordOption, Schema, Type};
 use crate::space_usage::SegmentSpaceUsage;
 use crate::store::StoreReader;
@@ -44,6 +43,7 @@ pub struct SegmentReader {
     fast_fields_readers: FastFieldReaders,
     fieldnorm_readers: FieldNormReaders,
 
+    list_fields_file: Option<FileSlice>, // Optional field list file for backwards compatibility
     store_file: FileSlice,
     alive_bitset_opt: Option<AliveBitSet>,
     schema: Schema,
@@ -153,6 +153,7 @@ impl SegmentReader {
         let termdict_composite = CompositeFile::open(&termdict_file)?;
 
         let store_file = segment.open_read(SegmentComponent::Store)?;
+        let list_fields_file = segment.open_read(SegmentComponent::FieldList).ok();
 
         crate::fail_point!("SegmentReader::open#middle");
 
@@ -201,6 +202,7 @@ impl SegmentReader {
             segment_id: segment.id(),
             delete_opstamp: segment.meta().delete_opstamp(),
             store_file,
+            list_fields_file,
             alive_bitset_opt,
             positions_composite,
             schema,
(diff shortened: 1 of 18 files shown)


src/core/segment_reader.rs
```diff
@@ -299,87 +301,25 @@ impl SegmentReader {
...
+                .fields()
+                .map(|(_field, entry)| FieldMetadata {
+                    field_name: entry.name().to_string(),
+                    typ: entry.field_type().value_type(),
+                    indexed: entry.is_indexed(),
+                    stored: entry.is_stored(),
+                    fast: entry.is_fast(),
+                })
+                .collect())
         }
```
adamreichold, line 322: It is admittedly code golfing, but I think it makes it a bit less noisier: 

```suggestion
        let fields_metadata = if let Some(list_fields_file) = self.list_fields_file.as_ref() {
            let file = list_fields_file.read_bytes()?;
            read_split_fields(file)?.collect::<io::Result<Vec<FieldMetadata>>>()?
        } else {
            // Schema fallback
            self.schema()
                .fields()
                .map(|(_field, entry)| FieldMetadata {
                    field_name: entry.name().to_string(),
                    typ: entry.field_type().value_type(),
                    indexed: entry.is_indexed(),
                    stored: entry.is_stored(),
                    fast: entry.is_fast(),
                })
                .collect()
        };
        Ok(fields_metadata)
```

src/indexer/path_to_unordered_id.rs
```diff
@@ -24,34 +26,44 @@ impl From<u32> for OrderedPathId {
 
 #[derive(Default)]
 pub(crate) struct PathToUnorderedId {
-    map: FnvHashMap<String, u32>,
+    /// TinySet contains the type codes of the columns in the path.
```
adamreichold, line 29: I think defining a struct with named fields would improve readability here, likely making the comment unnecessary or at least attaching it to the named field directly.

src/indexer/path_to_unordered_id.rs
```diff
@@ -24,34 +26,44 @@ impl From<u32> for OrderedPathId {
...
     ///
     /// The returned vec can be indexed with the unordered id to get the ordered id.
-    pub(crate) fn unordered_id_to_ordered_id(&self) -> Vec<OrderedPathId> {
-        let mut sorted_ids: Vec<(&str, &u32)> =
-            self.map.iter().map(|(k, v)| (k.as_str(), v)).collect();
+    pub(crate) fn unordered_id_to_ordered_id(&self) -> Vec<(OrderedPathId, TinySet)> {
+        let mut sorted_ids: Vec<(&str, (u32, TinySet))> = self
+            .map
+            .iter()
+            .map(|(k, (id, typ_code))| (k.as_str(), (*id, *typ_code)))
```
adamreichold, line 60: I think using the plural `typ_codes` would be good here and below, or maybe `all_codes` or `typ_code_bitvec` as above, to visually indicate that this is not a single code as in e.g. `insert_new_path`.

src/indexer/segment_serializer.rs
```diff
@@ -81,6 +84,11 @@ impl SegmentSerializer {
         &mut self.postings_serializer
     }
 
+    /// Accessor to the ``.
```
adamreichold, line 87: Missing name inside the backticks?

src/postings/postings_writer.rs
```diff
@@ -46,37 +47,38 @@ fn make_field_partition(
...
     per_field_postings_writers: &PerFieldPostingsWriter,
     fieldnorm_readers: FieldNormReaders,
     doc_id_map: Option<&DocIdMapping>,
+    unordered_id_to_ordered_id: &[(OrderedPathId, TinySet)],
     serializer: &mut InvertedIndexSerializer,
 ) -> crate::Result<()> {
     // Replace unordered ids by ordered ids to be able to sort
-    let unordered_id_to_ordered_id: Vec<OrderedPathId> =
-        ctx.path_to_unordered_id.unordered_id_to_ordered_id();
+    let ordered_id_to_path = ctx.path_to_unordered_id.ordered_id_to_path();
```
adamreichold, line 59: Why move the binding up here if it is not used in the loop? I would suggest leaving it below to make that clear to the reader.

src/field_list/mod.rs
```diff
@@ -0,0 +1,369 @@
...
+            io::Error::new(
+                ErrorKind::InvalidData,
+                format!("could not deserialize type {}", data[0]),
+            )
+        })?;
+
+        let data = data[1];
+        let indexed = (data & 0b100) != 0;
+        let stored = (data & 0b010) != 0;
+        let fast = (data & 0b001) != 0;
```
adamreichold, line 41: For extra robustness, one could check that all other bits are indeed zero as if they are not, something fishy must be going on.

src/field_list/mod.rs
```diff
@@ -0,0 +1,369 @@
...
+            let unordered_id =
+                u32::from_be_bytes(key[byte_range_unordered_id.clone()].try_into().unwrap());
+            let (path_id, typ_code_bitvec) = unordered_id_to_ordered_id[unordered_id as usize];
+            if !field_list_set.contains(&(field, path_id, typ_code_bitvec)) {
+                field_list_set.insert((field, path_id, typ_code_bitvec));
+                let mut build_path = |field_name: &str, mut json_path: String| {
+                    // In this case we need to map the potential fast field to the field name
+                    // accepted by the query parser.
+                    let create_canonical =
+                        !field_entry.is_expand_dots_enabled() && json_path.contains('.');
```
adamreichold, line 80: It seems dangerous to pass the name (which is derived from `field_entry`) as an argument but access `field_entry` as a capture. Maybe `build_path` should be a free function taking `field_entry` and `json_path` (and `map_to_canonical`) as arguments to make it obvious what it accesses and how.

src/field_list/mod.rs
```diff
@@ -0,0 +1,369 @@
...
+    let mut payload = Vec::new();
+    // Write Num Fields
+    let length = fields_metadata.len() as u32;
+    payload.extend_from_slice(&length.to_le_bytes());
+
+    for field_metadata in fields_metadata {
+        write_field(field_metadata, &mut payload);
+    }
+    let compression_level = 3;
+    let payload_compressed = zstd::stream::encode_all(&mut &payload[..], compression_level)
```
adamreichold, line 178: Since you are compressing from full buffers to full buffers, Zstd's [bulk API](https://docs.rs/zstd/latest/zstd/bulk/index.html) might yield better results (as @trinity-1686a found elsewhere IIRC).

src/field_list/mod.rs
```diff
@@ -0,0 +1,369 @@
...
+    reader.read_exact(&mut buffer)?;
+    Ok(buffer)
+}
+
+/// Reads the Split fields from a zstd compressed stream of bytes
+pub fn read_split_fields<R: Read>(
+    mut reader: R,
+) -> io::Result<impl Iterator<Item = io::Result<FieldMetadata>>> {
+    let format_version = read_exact_array::<_, 1>(&mut reader)?[0];
+    assert_eq!(format_version, 1);
```
adamreichold, line 217: Shouldn't this be an error instead of a panic?

src/field_list/mod.rs
```diff
@@ -0,0 +1,369 @@
...
+        field_name,
+        typ: field_config.typ,
+        indexed: field_config.indexed,
+        stored: field_config.stored,
+        fast: field_config.fast,
+    })
+}
+
+/// Reads the Split fields from a stream of bytes
+fn read_split_fields_from_zstd<R: Read>(
```
adamreichold, line 252: To be honest, I would suggest merging this into `read_split_fields`. The name was surprising to me as it does nothing specific to Zstd and I think following the logic of constructing the iterator is easier if it is not split over multiple functions.

(If you do go for Zstd's bulk API, this would probably become simpler as well due to working on a full in-memory representation (if that is not consider too memory hungry) which can simply be chunked instead of calling `read_exact` repeatedly.)

src/field_list/mod.rs
```diff
@@ -0,0 +1,369 @@
...
+    mut reader: R,
+) -> io::Result<impl Iterator<Item = io::Result<FieldMetadata>>> {
+    let mut num_fields = u32::from_le_bytes(read_exact_array::<_, 4>(&mut reader)?);
+
+    Ok(std::iter::from_fn(move || {
```
adamreichold, line 257: I am not sure if this break w.r.t. error handling, but wouldn't something like

```rust
Ok((0..num_fields).map(move |_| read_field(&mut reader))
```

work as well?

src/field_list/mod.rs
trinity-1686a: see https://github.com/quickwit-oss/tantivy/pull/1946#issuecomment-1481017149 and https://github.com/quickwit-oss/tantivy/pull/1946#issuecomment-1498195106. This might be more of an issue when decompressing than compressing, but either way the bulk API is faster, so if it's easy to use, we should use it

src/postings/postings_writer.rs
PSeitz: The Vec is reused in another method

adamreichold: Isn't it `unordered_id_to_ordered_id` that is reused in `serialize_segment_fields`? I was thinking about the binding for `ordered_id_to_path` which moved up here from line 79 without any change to how it is used.

PR #2282, opened 2023-12-17
chore: Expose TopDocs::order_by_u64_field again

ditsuke:
The `TopDocs::order_by_u64_field` API was restricted from public use in #2111, but it is useful in contexts where the field type may not be important or known ahead of time.

Note that the unavailability of this API breaks ordering on non-u64 fields in tantity-py. Reference: https://github.com/quickwit-oss/tantivy-py/pull/166


ditsuke:
Bump for review. Open to discussing other ways to solve the issue with tantivy-py, but I think `order_by_u64_field` is a convenient API for general sorting that should remain public.

PSeitz (approved):

PR #2291, opened 2024-01-06
Allow optional whitespace before bound specification in range queries

### LINKED ISSUE #2289: parsing range query with whitespace fails on `0.21` but not `0.20`
Whitespace in a range query causes a `SyntaxError` on `0.21` but was parsed just fine in `0.20`. 

Consider this example:

```rust
use tantivy::schema::{FAST, INDEXED, Schema, STORED};
use tantivy::tokenizer::TokenizerManager;

fn main() {
    let mut schema = Schema::builder();
    schema.add_u64_field("num", INDEXED | STORED | FAST);
    let query_parser = QueryParser::new(schema.build(), vec![], TokenizerManager::default());
    let query = query_parser.parse_query("num: [0 TO 10]").unwrap();
    println!("{query:?}")
}
```
on version `0.21` this results in
```
thread 'main' panicked at src/main.rs:11:60:
called `Result::unwrap()` on an `Err` value: SyntaxError("num: [0 TO 10]")
```

whereas on version `0.20` this prints
```
RangeQuery { field: "num", value_type: U64, lower_bound: Included([0, 0, 0, 0, 0, 0, 0, 0]), upper_bound: Included([0, 0, 0, 0, 0, 0, 0, 10]), limit: None }
```

`0.21.1`

**To Reproduce**

See example above

adamreichold:
Closes #2289

### DIFF
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -469,8 +469,10 @@ fn range(inp: &str) -> IResult<&str, UserInputLeaf> {
         },
     );
 
-    let lower_to_upper =
-        separated_pair(lower_bound, tuple((space1, tag("TO"), space1)), upper_bound);
+    let lower_to_upper = preceded(
+        space0,
+        separated_pair(lower_bound, tuple((space1, tag("TO"), space1)), upper_bound),
+    );
 
     map(
         alt((elastic_unbounded_range, lower_to_upper)),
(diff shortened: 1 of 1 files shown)


query-grammar/src/query_grammar.rs
```diff
@@ -1537,7 +1539,9 @@ mod test {
     fn test_parse_query_with_range() {
         test_parse_query_to_ast_helper("[1 TO 5]", "[\"1\" TO \"5\"]");
         test_parse_query_to_ast_helper("foo:{a TO z}", "\"foo\":{\"a\" TO \"z\"}");
+        test_parse_query_to_ast_helper("foo: \t{a TO z}", "\"foo\":{\"a\" TO \"z\"}");
```
PSeitz, line 1542: not sure we want to support tabs in the query parser

adamreichold: I just added it to highlight this, but since we are using `space0` everywhere already, I think we already support tabs specifically where we support whitespace generally.

PR #2294, opened 2024-01-09
Fix off by one in optional index

### LINKED ISSUE #2293: Range query for optional fast field panics with Index out of bounds in 0.21.1
I tried upgrading to tantivy `0.21.1` and found that some range queries on optional fast field columns were panicking with an index out of bounds error. The error doesn't seem to happen on tantivy `0.21.0`.

---

I was able to reproduce the issue in the `range_query_fast_optional_field_minimum` test (in [fast_field_range_query.rs](https://github.com/quickwit-oss/tantivy/blob/53f2fe1fbee83f22c81e28aaec02fb0628b68966/src/query/range_query/fast_field_range_query.rs#L189-L229)) by increasing the doc count from 1000 to 1000000. Specifically this line:

https://github.com/quickwit-oss/tantivy/blob/53f2fe1fbee83f22c81e28aaec02fb0628b68966/src/query/range_query/fast_field_range_query.rs#L203

Command (after updating the count above):

```
RUST_BACKTRACE=1 cargo test range_query_fast_optional_field_minimum
```

Stack trace:

```
thread 'query::range_query::fast_field_range_query::tests::range_query_fast_optional_field_minimum' panicked at columnar/src/column_index/optional_index/mod.rs:194:26:
index out of bounds: the len is 4 but the index is 4
stack backtrace:
   0: rust_begin_unwind
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/panicking.rs:72:14
   2: core::panicking::panic_bounds_check
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/panicking.rs:190:5
   3: <tantivy_columnar::column_index::optional_index::OptionalIndex as tantivy_columnar::column_index::optional_index::set::Set<u32>>::rank
             at ./columnar/src/column_index/optional_index/mod.rs:194:26
   4: tantivy_columnar::column_index::ColumnIndex::docid_range_to_rowids
             at ./columnar/src/column_index/mod.rs:135:31
   5: tantivy_columnar::column::Column<T>::get_docids_for_value_range
             at ./columnar/src/column/mod.rs:121:27
...
```

PSeitz:
Fixes #2293

Fixes an off by one error in the metadata resize of the optional index
when loading the index.

Merge variables with the same meaning but different names

### DIFF
--- a/columnar/src/column_index/optional_index/mod.rs
+++ b/columnar/src/column_index/optional_index/mod.rs
@@ -21,8 +21,6 @@ const DENSE_BLOCK_THRESHOLD: u32 =
 
 const ELEMENTS_PER_BLOCK: u32 = u16::MAX as u32 + 1;
 
-const BLOCK_SIZE: RowId = 1 << 16;
-
 #[derive(Copy, Clone, Debug)]
 struct BlockMeta {
     non_null_rows_before_block: u32,
@@ -109,8 +107,8 @@ struct RowAddr {
 #[inline(always)]
 fn row_addr_from_row_id(row_id: RowId) -> RowAddr {
     RowAddr {
-        block_id: (row_id / BLOCK_SIZE) as u16,
-        in_block_row_id: (row_id % BLOCK_SIZE) as u16,
+        block_id: (row_id / ELEMENTS_PER_BLOCK) as u16,
+        in_block_row_id: (row_id % ELEMENTS_PER_BLOCK) as u16,
     }
 }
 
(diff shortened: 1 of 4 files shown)


columnar/src/column_index/optional_index/mod.rs
```diff
@@ -490,8 +488,9 @@ fn deserialize_optional_index_block_metadatas(
         start_byte_offset += block_variant.num_bytes_in_block();
         non_null_rows_before_block += num_non_null_rows;
     }
+    let last_block = row_addr_from_row_id(num_rows).block_id;
```
fulmicoton, line 491:
```suggestion
    let last_block = 
       if num_rows == 0 {
          0
       } else {
          row_addr_from_row_id(num_rows - 1).block_id
       }
```

fulmicoton: I think this is equivalent to the previous formula.

PR #2299, opened 2024-01-11
Truncate keys to u16::MAX in term hashmap

PSeitz:
Truncate keys to u16::MAX, instead e.g. storing 0 bytes for keys with length u16::MAX + 1

The term hashmap has a hidden API contract to only accept terms with lenght up u16::MAX.

### DIFF
--- a/stacker/src/shared_arena_hashmap.rs
+++ b/stacker/src/shared_arena_hashmap.rs
@@ -295,6 +295,8 @@ impl SharedArenaHashMap {
     /// will be in charge of returning a default value.
     /// If the key already as an associated value, then it will be passed
     /// `Some(previous_value)`.
+    ///
+    /// The key will be truncated to u16::MAX bytes.
     #[inline]
     pub fn mutate_or_create<V>(
         &mut self,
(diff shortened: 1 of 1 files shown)


stacker/src/shared_arena_hashmap.rs
```diff
@@ -308,6 +312,8 @@ impl SharedArenaHashMap {
         if self.is_saturated() {
             self.resize();
         }
+        // Limit the key size to u16::MAX
+        let key = &key[..std::cmp::min(key.len(), u16::MAX as usize)];
```
adamreichold, line 314: There is [`impl From<u16> for usize`](https://doc.rust-lang.org/stable/std/primitive.usize.html#impl-From%3Cu16%3E-for-usize), i.e. the cast could be replaced by `.into()`.

PSeitz: I prefer to keep it explicit

adamreichold: You can explicitly write `usize::from(u16::MAX)` if you prefer while still avoiding the potentially truncating cast.

fulmicoton (approved):

stacker/src/shared_arena_hashmap.rs
PSeitz: I wish rust would just coerce types that can be safely cast

here I don't think we gain much for proof reading

PR #2304, opened 2024-01-18
feat: add a simple way to chain two tokenizers

ctron:


ctron:
I applied nightly `rustfmt`.

fulmicoton:
@ctron  can you describe your use case?

ctron:
My use case is to have all simple tokens plus all ngrams.

src/tokenizer/chain.rs
```diff
@@ -0,0 +1,185 @@
...
+            }
+            State::Done => false,
+        }
+    }
+
+    fn token(&self) -> &Token {
+        match self.state {
+            State::First => self.first.token(),
+            State::Second => self.second.token(),
+            State::Done => &self.dummy,
```
adamreichold: Do we really need the dummy token? Couldn't the tokenizer just yield `self.second.token()` as it does not really matter what is returned after `advance()` has returned `false`.

src/tokenizer/chain.rs
```diff
@@ -0,0 +1,185 @@
+use tokenizer_api::{Token, TokenStream, Tokenizer};
+
+/// Trait to allow easily chaining two [`Tokenizer`].
+pub trait ChainTokenizerExt: Sized {
+    /// Produce a [`Tokenizer`] which runs through the first tokenizer, and then through the second.
+    fn chain<T: Tokenizer>(self, next: T) -> ChainTokenizer<Self, T>;
```
adamreichold: I wonder if this should be placed in a generic `TokenizerExt` trait as a central place for future tokenizer combinators.

src/tokenizer/chain.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum State {
+    First,
+    Second,
+    Done,
+}
+
+pub struct ChainTokenStream<'a, F, S>
```
adamreichold, line 36: The `Token::position` fields would need updating though, wouldn't they? Meaning the positions of the second stream should be offset by the number of tokens yielded by the first one?

src/tokenizer/chain.rs
```diff
@@ -0,0 +1,185 @@
...
+where
+    F: Tokenizer,
+    S: Tokenizer,
+{
+    first: F::TokenStream<'a>,
```
adamreichold: I wonder if this could be simplified to

```rust
pub struct ChainTokenStream<F, S> {
  first: Option<F>,
  second: S,
}

impl<F,S> TokenStream for ChainTokenStream<F,S> where F: TokenStream, S: TokenStream {
  fn advance(&mut self) -> bool {
    if let Some(first) =  &mut self.first {
       if first.advance() {
         true
       } else {
         self.first = None; 
         false
       }
    } else {
       self.second.advance()
    }
  }
}
```

or does this run afoul of the hidden lifetime `'a`?

src/tokenizer/chain.rs
ctron: True, I was wasn't bold enough to for that. If you like, I can refactor it that way.

ctron:
I was able to incorporate most of the feedback you mentioned. It's less explicit without the `enum`, but works the same way. There was just one call to `second.advance()` missing, as a `TokenStream` seems to start *before* the first item.

I am not sure about the `position` topic, but I also don't fully understand it: If the idea is to enumerate/count all tokens, then yes that should be offset but the length of `first`. If that reflects the position of the token as part of the original input, then that might not be correct.

I'll admit that the whole API around tokenization feels a bit confusing.

adamreichold:
> I was able to incorporate most of the feedback you mentioned. It's less explicit without the enum, but works the same way. There was just one call to second.advance() missing, as a TokenStream seems to start before the first item.

Agreed, it is more implicit. But I was mainly suggesting it for efficiency, i.e. keep the state as small as possible and drop the first tokenizer as soon as we are done with it. If you feel like code-golfing, I think those two calls to `second.advance()` could even be merged:

```rust
fn advance(&mut self) -> bool {
    if let Some(first) = &mut self.first {
        if first.advance() {
            return true;
        } else {
            self.first = None;
        }
    }
    
    self.second.advance()
}
```

ctron:
> If you feel like code-golfing, I think those two calls to second.advance() could even be merged 

I like that, pushed.

So, the remaining thing seems to be the position. I am just not sure what to do with it.

adamreichold:
> So, the remaining thing seems to be the position. I am just not sure what to do with it.

I'd say let's wait for input from @fulmicoton on that. I am myself unsure what downstream consumers expect of the position field. I suspect that this is mainly used for phrase queries with slop which I think would make the current implementation correct, i.e. have `position` still refer to the original input.

fulmicoton:
@PSeitz can you review ?

ctron:
Fixed the test issue.

PSeitz:
There is a hidden contract currently on the tokenizer API which expects positions to be sorted incrementally. This happens in the serialization part in [recorder.rs](https://github.com/quickwit-oss/tantivy/blob/main/src/postings/recorder.rs#L291), when positions are delta encoded.

There are two options:
- Handle unsorted positions in the serialization (allow unsorted positions from the tokenizer)
- Return the tokens in a sorted way

I think handling unsorted positions is not really favorable, since it would carry some performance overhead.

adamreichold:
> Return the tokens in a sorted way

So in this case, we'd need to interleave the output of the two tokenizer dynamically to ensure that one does not outpace the other. Or could just make positions up and offset all positions returned by the second tokenizer?

PR #2308, opened 2024-01-26
Make allocating field names avoidable for range and exists queries.

adamreichold:
If the field names are statically known, `Cow::Borrowed(&'static str)` can handle them without allocations. The general case is still handled by `Cow::Owned(String)`.

### DIFF
--- a/src/query/exist_query.rs
+++ b/src/query/exist_query.rs
@@ -1,4 +1,4 @@
-use core::fmt::Debug;
+use std::borrow::Cow;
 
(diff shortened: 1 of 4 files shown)


PSeitz:
I don't think a single allocation per query won't have a measurable impact on performance, therefore I'd prefer the more readable variant.

adamreichold:
> therefore I'd prefer the more readable variant.

I tried to anticipate this via the trait bound `Into<Cow<'static, str>>` so that call sites do not need to care about the `Cow`s. But I suspect you mean readability of the implementation?

PR #2311, opened 2024-02-02
agg: support to deserialize f64 from string

PSeitz:


src/aggregation/mod.rs
```diff
@@ -154,7 +156,108 @@ use columnar::{ColumnType, MonotonicallyMappableToU64};
...
+    impl<'de> Visitor<'de> for StringOrFloatVisitor {
+        type Value = Option<f64>;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+            formatter.write_str("a string or a float")
+        }
+
+        fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+        where E: de::Error {
+            value.parse::<f64>().map(Some).map_err(|_err| {
```
fulmicoton: we should refuse "nan" and infinity

src/aggregation/mod.rs
```diff
@@ -154,7 +156,108 @@ use columnar::{ColumnType, MonotonicallyMappableToU64};
...
+                de::Error::custom(format!("Failed to parse f64 from string: {:?}", value))
+            })
+        }
+
+        fn visit_string<E>(self, value: String) -> Result<Self::Value, E>
+        where E: de::Error {
+            value.parse::<f64>().map(Some).map_err(|_err| {
+                de::Error::custom(format!("Failed to parse f64 from string: {:?}", value))
+            })
+        }
```
```suggestion
```

fulmicoton: by default visit_str will be called. This is better.

src/aggregation/mod.rs
```diff
@@ -154,7 +156,108 @@ use columnar::{ColumnType, MonotonicallyMappableToU64};
...
+
+        fn visit_string<E>(self, value: String) -> Result<Self::Value, E>
+        where E: de::Error {
+            value.parse::<f64>().map_err(|_err| {
+                de::Error::custom(format!("Failed to parse f64 from string: {:?}", value))
+            })
+        }
+
```
```suggestion
```

src/aggregation/mod.rs
```diff
@@ -154,7 +156,108 @@ use columnar::{ColumnType, MonotonicallyMappableToU64};
...
+    impl<'de> Visitor<'de> for StringOrFloatVisitor {
+        type Value = f64;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+            formatter.write_str("a string or a float")
+        }
+
+        fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+        where E: de::Error {
+            value.parse::<f64>().map_err(|_err| {
```
fulmicoton: same as above. let's refuse nan and infinity.

src/aggregation/mod.rs
PSeitz: elastic search allows NaN and Infinity

In case of missing it could be used to invalidate an aggregation if there are missing elements

fulmicoton: What is the use case for parsing NaN I don't understand?

PSeitz: It could be kind of a safeguard, if you would e.g. expect in your avg aggregation that every doc has at least one value, you could set `missing: NaN`, so the avg aggregation would be set to `NaN` if there are missing values.
There may be other queries like this.

fulmicoton: Not worth the trouble. Let's forbid NaN.

PSeitz: done

fulmicoton: I think this still accepts NaN and inf.

src/aggregation/mod.rs
fulmicoton: same as above.

fulmicoton (approved):

PR #2312, opened 2024-02-05
Add shared search executor

MochiXu:
The approach to integrating `Tantivy` with `ClickHouse` is as follows: 
A `ClickHouse` Table will have multiple `DataParts`, each with its own `Tantivy` index. However, I would like `ClickHouse` to share a thread pool among different `DataParts` during concurrent searches, in order to reduce resource overhead during the search process. I have created a bridging repo([tantivy-search](https://github.com/MochiXu/tantivy-search/tree/tantivy_0.21.1)) between `Tantivy` and `ClickHouse`. 

The code for using a shared thread pool can be referred to at this [link](https://github.com/MochiXu/tantivy-search/blob/c5c741b7abb438bf93da5f43652d72eeabed1e0a/src/search/index_searcher.rs#L104C1-L126C6).

### DIFF
--- a/src/index/index.rs
+++ b/src/index/index.rs
@@ -322,6 +322,12 @@ impl Index {
         Ok(())
     }
 
+    /// Custom thread pool by a outer thread pool.
+    pub fn set_shared_multithread_executor(&mut self, shared_thread_pool: Arc<Executor>) -> crate::Result<()> {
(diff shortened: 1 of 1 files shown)


fulmicoton:
Merged! Just to put it in your radar, Quickwit can be used to search into Clickhouse.

PR #2313, opened 2024-02-07
Fix serde for TopNComputer

PSeitz:
The top hits aggregation changed the TopNComputer to be serializable,
but capacity needs to be carried over, as it contains logic which is
checked against when pushing elements (capacity == 0 is not allowed).

### DIFF
--- a/src/collector/top_score_collector.rs
+++ b/src/collector/top_score_collector.rs
@@ -3,9 +3,8 @@ use std::marker::PhantomData;
 use std::sync::Arc;
 
 use columnar::ColumnValues;
-use serde::de::{self, DeserializeOwned, SeqAccess, Visitor};
-use serde::ser::SerializeSeq;
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
 
 use super::Collector;
 use crate::collector::custom_score_top_collector::CustomScoreTopCollector;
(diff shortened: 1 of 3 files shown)


fulmicoton (changes requested):
I'm surprised we need the TopNComputer to be serializable. Is it for the results or the blueprint part?
It is quite the code smell.

IF it is really required to have it serialize I doubt we need to serialize the capacity.
Isn't it a function of top_n?

PSeitz:
> I'm surprised we need the TopNComputer to be serializable. Is it for the results or the blueprint part? It is quite the code smell.

It's part of the `TopHitsCollector` from the the top hits aggregation, which is part of the `IntermediateMetricResult`. The intermediate results get sent around between nodes.

> 
> IF it is really required to have it serialize I doubt we need to serialize the capacity. Isn't it a function of top_n?

We truncate when we reach the vecs capacity, which is `top_n.max(1) * 2`. I could store it in an extra field instead of using the capacity, but this part of the code is very performance sensitive.

fulmicoton:
No but top_n is part of the serialization... Can't we write deserialization code for the `TopNCollector` that initialize the capacity of the vec depending on the top_n value?

PSeitz:
> No but top_n is part of the serialization... Can't we write deserialization code for the `TopNCollector` that initialize the capacity of the vec depending on the top_n value?

Ah, yes. I did run into some issues with nested generics with a custom deserialization on an intermediate struct. 
But it worked with the serde `from` version now.

src/collector/top_score_collector.rs
```diff
@@ -720,17 +721,43 @@ impl SegmentCollector for TopScoreSegmentCollector {
...
+pub struct TopNComputer<Score, D, const REVERSE_ORDER: bool = true> {
     /// The buffer reverses sort order to get top-semantics instead of bottom-semantics
-    buffer: Vec<ComparableDoc<Score, DocId, REVERSE_ORDER>>,
+    buffer: Vec<ComparableDoc<Score, D, REVERSE_ORDER>>,
     top_n: usize,
     pub(crate) threshold: Option<Score>,
 }
+// Intermediate struct for TopNComputer for deserialization, to fix vec capacity
+#[derive(Deserialize)]
+pub struct TopNComputerDeser<Score, D, const REVERSE_ORDER: bool = true> {
```
fulmicoton:
```suggestion
struct TopNComputerDeser<Score, D, const REVERSE_ORDER: bool> {
```

fulmicoton (approved):
Can we fix the clippy comment too?

PR #2314, opened 2024-02-08
clippy

PSeitz:


src/directory/mmap_directory.rs
```diff
@@ -479,6 +479,7 @@ impl Directory for MmapDirectory {
         let file: File = OpenOptions::new()
             .write(true)
             .create(true) //< if the file does not exist yet, create it.
+            .truncate(false)
```
fulmicoton, line 482: Wouldn't this make more sense?

```suggestion
            .create_new(true) //< if the file already exists, fail
```

PSeitz: This would disable blocking locking. I'm also not sure if we delete the lock

fulmicoton: Ah I totally missed the context. I thought it was for regular writes!

fulmicoton (approved):

PR #2319, opened 2024-02-23
handle ip adresses in term aggregation

PSeitz:
Stores IpAdresses during the segment term aggregation via u64 representation
and convert to u128(IpV6Adress) via downcast when converting to intermediate results.

Enable Downcasting on `ColumnValues`
Expose u64 variant for u128 encoded data via `open_u64_lenient` method.
Remove lifetime in VecColumn, to avoid 'static lifetime requirement coming
from downcast trait.

#1689

### DIFF
--- a/columnar/src/column_values/u128_based/mod.rs
+++ b/columnar/src/column_values/u128_based/mod.rs
@@ -120,7 +120,7 @@ pub fn open_u128_mapped<T: MonotonicallyMappableToU128 + Debug>(
 /// # Notice
 /// In case there are new codecs added, check for usages of `CompactSpaceDecompressorU64` and
 /// also handle the new codecs.
-pub fn open_u128_as_u64(mut bytes: OwnedBytes) -> io::Result<Arc<dyn ColumnValues<u64>>> {
+pub fn open_u128_as_compact_u64(mut bytes: OwnedBytes) -> io::Result<Arc<dyn ColumnValues<u64>>> {
     let header = U128Header::deserialize(&mut bytes)?;
     assert_eq!(header.codec_type, U128FastFieldCodecType::CompactSpace);
     let reader = CompactSpaceU64Accessor::open(bytes)?;
--- a/columnar/src/column_values/u128_based/mod.rs
+++ b/columnar/src/column_values/u128_based/mod.rs
@@ -6,7 +6,9 @@ use std::sync::Arc;
 mod compact_space;
 
 use common::{BinarySerializable, OwnedBytes, VInt};
-use compact_space::{CompactSpaceCompressor, CompactSpaceDecompressor};
+pub use compact_space::{
+    CompactSpaceCompressor, CompactSpaceDecompressor, CompactSpaceU64Accessor,
+};
 
 use crate::column_values::monotonic_map_column;
 use crate::column_values::monotonic_mapping::{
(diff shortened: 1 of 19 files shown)


columnar/src/column_values/u128_based/mod.rs
```diff
@@ -108,6 +110,23 @@ pub fn open_u128_mapped<T: MonotonicallyMappableToU128 + Debug>(
...
+/// Returns the u64 representation of the u128 data.
+/// The internal representation of the data as u64 is useful for faster processing.
+///
+/// In order to convert to u128 back cast to `CompactSpaceU64Accessor` and call
+/// `compact_to_u128`.
+///
+/// # Notice
+/// In case there are new codecs added, check for usages of `CompactSpaceDecompressorU64` and
+/// also handle the new codecs.
+pub fn open_u128_as_u64(mut bytes: OwnedBytes) -> io::Result<Arc<dyn ColumnValues<u64>>> {
```
fulmicoton: can you rename the method? Right now it suggest it is a simple coercion? In reality you are returning the compact space.
Maybe 

```
open_u128_compact_space_as_u64
```

or something like that

columnar/src/column_values/u64_based/blockwise_linear.rs
```diff
@@ -63,7 +63,10 @@ impl BlockwiseLinearEstimator {
         if self.block.is_empty() {
             return;
         }
-        let line = Line::train(&VecColumn::from(&self.block));
+        let column = VecColumn::from(std::mem::take(&mut self.block));
```
fulmicoton, line 66: why was this needed? I don't understand.

columnar/src/column_values/u64_based/blockwise_linear.rs
```diff
@@ -125,7 +128,7 @@ impl ColumnCodecEstimator for BlockwiseLinearEstimator {
                 *buffer_val = gcd_divider.divide(*buffer_val - stats.min_value);
             }
 
-            let line = Line::train(&VecColumn::from(&buffer));
+            let line = Line::train(&VecColumn::from(buffer.to_vec()));
```
fulmicoton, line 131: same question as above

src/aggregation/bucket/term_agg.rs
```diff
@@ -544,6 +548,27 @@ impl SegmentTermCollector {
                 let val = bool::from_u64(val);
                 dict.insert(IntermediateKey::Bool(val), intermediate_entry);
             }
+        } else if self.column_type == ColumnType::IpAddr {
+            let arc_inner = agg_with_accessor
```
fulmicoton: let's put a decent name and a type on this. `arc_inner` is not helpful.
```suggestion
            let arc_inner = agg_with_accessor
```

src/aggregation/bucket/term_agg.rs
```diff
@@ -544,6 +548,27 @@ impl SegmentTermCollector {
...
+                .accessor
+                .values
+                .clone()
+                .downcast_arc::<CompactSpaceU64Accessor>()
```
fulmicoton, line 556: I agree that not doing the segment collection on `u128` is the right way here, but I am surprised downcasting logic is required here. Are there really no better way?

columnar/src/column_values/u64_based/blockwise_linear.rs
PSeitz: Extending the `ColumnValues` trait with another trait (any trait, even an empty one) requires lifetimes to be `'static`. This requirement is no problem when working on the values instead of references, so I changed `VecColumn` to a owned version.

src/aggregation/bucket/term_agg.rs
PSeitz: The main problem here is with regards to generics.
We need a codec specific method on a specific trait variant `ColumnValues<u64>`.

We could return `ColumnValuesU64Ext` instead of `ColumnValues<u64>`, which would include the method

```rust 

trait U64ColumnValuesExt: ColumnValues<u64> {
    compact_to_u128(&self, compact: u32) -> u128 {
        unimplemented!() // unimplemented for all expect compact space
    }
}
```

Having codec specific methods on a common trait is a little bit weird.
Casting is also not great, but having the mechanism may also allow some performance gains by enabling inlining in some cases.

In both cases we will need to know the underlying codec before calling the codec specific method.

columnar/src/column_values/u128_based/mod.rs
PSeitz: renamed to `open_u128_as_compact_u64`

src/aggregation/bucket/term_agg.rs
PSeitz: done

fulmicoton (approved):

PR #2325, opened 2024-03-11
Impl trait `FromStr` for `Language`

MochiXu:
In order to conveniently determine whether the user's Language (in string format) is supported by `Tantivy`, the `FromStr` trait has been implemented for `Language`.

### DIFF
--- a/src/tokenizer/stemmer.rs
+++ b/src/tokenizer/stemmer.rs
@@ -1,12 +1,12 @@
+use std::borrow::Cow;
 use std::mem;
-use std::{borrow::Cow, str::FromStr};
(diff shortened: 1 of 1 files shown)


src/tokenizer/stemmer.rs
```diff
@@ -30,6 +32,36 @@ pub enum Language {
     Turkish,
 }
 
+impl FromStr for Language {
```
PSeitz, line 35: Can't you just try to deserialize?

MochiXu: Okey, I get it.

PR #2327, opened 2024-03-12
Speed up searches by removing repeated memsets coming from vec.resize()

RaphaelMarinier:
Also, reserve exactly the size needed, which is surprisingly needed to get the full speedup of ~5% on a good fraction of the queries.

The memset (accounting for 4% of the CPU cycles) in the range_numeric query is indeed gone with this PR, see flame graphs:
baseline range_numeric
reserve_exact and set_len range_numeric

Overall, it makes things a few percent faster on many queries:
tantivy reserve_exact and set_len vs baseline

Note: I made sure to use the same version of tantivy for the baseline and for this proposed change. The only difference between the two columns is this PR.

It would be possible to have this speedup without using unsafe by never truncating it (as is done in filter_vec_in_place()) and instead passing the number of final elements in the vector back to VecCursor.

### DIFF
--- a/bitpacker/src/bitpacker.rs
+++ b/bitpacker/src/bitpacker.rs
@@ -125,6 +125,8 @@ impl BitUnpacker {
 
     // Decodes the range of bitpacked `u32` values with idx
     // in [start_idx, start_idx + output.len()).
+    // It is guaranteed to completely fill `output` and not read from it, so passing a vector with
+    // un-initialized values is safe.
     //
     // #Panics
     //
@@ -237,7 +239,19 @@ impl BitUnpacker {
         data: &[u8],
         positions: &mut Vec<u32>,
     ) {
-        positions.resize(id_range.len(), 0u32);
+        // We use the code below instead of positions.resize(id_range.len(), 0u32) for performance
+        // reasons: on some queries, the CPU cost of memsetting the array and of using a bigger
+        // vector than necessary is noticeable (~5%).
+        // In particular, searches are a few percent faster when using reserve_exact() as below
+        // instead of reserve().
+        // The un-initialized values are safe as get_batch_u32s() completely fills `positions`
+        // and does not read from it.
+        positions.clear();
+        positions.reserve_exact(id_range.len());
+        #[allow(clippy::uninit_vec)]
+        unsafe {
+            positions.set_len(id_range.len());
+        }
         self.get_batch_u32s(id_range.start, data, positions);
         crate::filter_vec::filter_vec_in_place(value_range, id_range.start, positions)
     }


bitpacker/src/bitpacker.rs
```diff
@@ -237,7 +239,19 @@ impl BitUnpacker {
...
+        // vector than necessary is noticeable (~5%).
+        // In particular, searches are a few percent faster when using reserve_exact() as below
+        // instead of reserve().
+        // The un-initialized values are safe as get_batch_u32s() completely fills `positions`
+        // and does not read from it.
+        positions.clear();
+        positions.reserve_exact(id_range.len());
+        #[allow(clippy::uninit_vec)]
+        unsafe {
+            positions.set_len(id_range.len());
```
adamreichold, line 253: I am pretty sure this is undefined behaviour due to the validity requirements on integers which newly allocated memory does not fulfil.

If you want to apply this optimization, `get_batch_u32s` needs to work with the return value of [`Vec::spare_capacity_mut`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.spare_capacity_mut) to [`write`](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.write-1) into the `MaybeUninit<u32>` values and only call `set_len` after all of these values have been initialized.

RaphaelMarinier: Indeed, thanks for pointing this out, I see that in https://github.com/rust-lang/unsafe-code-guidelines/issues/439 it was decided that the mere existence of an uninitialized integer is undefined behavior. There is also an interesting discussion on that [here](https://github.com/rust-lang/unsafe-code-guidelines/issues/71) where it was controversially concluded that it was UB.

The trick is the same as in https://github.com/quickwit-oss/quickwit/pull/4712, so we should probably decide whether we do neither or both. @fulmicoton 

For that quickwit PR, I see that read_buf (see [rfc](https://rust-lang.github.io/rfcs/2930-read-buf.html), [function doc](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_buf)) could be used, but it's not stable unfortunately.

For this PR, note that this vector [ends up being filled by AVX2 instructions](https://github.com/quickwit-oss/tantivy/blob/f6b0cc1aab50d13ad3e6b2d094a82e53d8060e11/bitpacker/src/filter_vec/avx2.rs#L74), so what you propose does not seem possible: we would either to take a performance hit by storing the output of the AVX2 computation to a temporary variable and write() it into the uninitialized vector, or use MaybeUninit.as_mut_ptr which would lead us back to UB. Or do you have an efficient suggestion to make it work with the AVX instructions?

fulmicoton: > so we should probably decide whether we do neither or both. @fulmicoton

I'd say we should do it for one and not for the other.
For the Read side, the one here is simply not needed. We also own the bitpacking crate. We can just add a method that takes a `MaybeUninit`.

For the Read trait, this is well known problem that has a workaround but it is not stabilized yet.
The problem is that if a Read implementation were to read the buffer we would be in big troubles.

We put that in a place where the Read impl is well defined so I think we are ok.

fulmicoton:
I think there is a problem in the benchmark. That code is not used in most of the queries showing an improvements.

PR #2330, opened 2024-03-15
add method to fetch block of first vals in columnar

PSeitz:
add method to fetch block of first vals in columnar (this is way faster
than single calls for full columns)
add benchmark
fix import warnings

```
test bench_get_block_first_on_full_column                  ... bench:          56 ns/iter (+/- 26)
test bench_get_block_first_on_full_column_single_calls     ... bench:         311 ns/iter (+/- 6)
test bench_get_block_first_on_multi_column                 ... bench:         378 ns/iter (+/- 15)
test bench_get_block_first_on_multi_column_single_calls    ... bench:         546 ns/iter (+/- 13)
test bench_get_block_first_on_optional_column              ... bench:         291 ns/iter (+/- 6)
test bench_get_block_first_on_optional_column_single_calls ... bench:         362 ns/iter (+/- 8)
```


columnar/src/column_values/mod.rs
```diff
@@ -80,6 +80,32 @@ pub trait ColumnValues<T: PartialOrd = u64>: Send + Sync + DowncastSync {
...
+        assert!(indexes.len() == output.len());
+        let out_and_idx_chunks = output.chunks_exact_mut(4).zip(indexes.chunks_exact(4));
+        for (out_x4, idx_x4) in out_and_idx_chunks {
+            out_x4[0] = Some(self.get_val(idx_x4[0]));
+            out_x4[1] = Some(self.get_val(idx_x4[1]));
+            out_x4[2] = Some(self.get_val(idx_x4[2]));
+            out_x4[3] = Some(self.get_val(idx_x4[3]));
+        }
+
+        let step_size = 4;
```
fulmicoton: can we at least make `STEP_SIZE` a const and use it above?

Other possible improvement...
Use the remainder of chunks_exact (in case it is tricky, you can clone the iterator).

fulmicoton (approved):

columnar/src/column_values/mod.rs
PSeitz: I switched to `remainder`

PR #2333, opened 2024-03-21
extend field grouping

### LINKED ISSUE #2332: support more syntax for field grouping
we support field grouping like this `title:(+return +"pink panther")`, but not like `title:(return AND "pink panther")`. That should be added to be as close as possible to lucene when it can reasonably be done

trinity-1686a:
fix #2332 
as field grouping allow basically anything, we simply parse an ast, and set the field of any sub-leaf if it wasn't already. This allow simplifying the code a bit, and makes sure grouping stays on par with the rest of the language

### DIFF
--- a/query-grammar/src/user_input_ast.rs
+++ b/query-grammar/src/user_input_ast.rs
@@ -226,7 +226,7 @@ impl UserInputAst {
         UserInputAst::compose(Occur::Should, asts)
     }
 
-    pub fn set_default_field(&mut self, field: String) {
+    pub(crate) fn set_default_field(&mut self, field: String) {
         match self {
             UserInputAst::Clause(clauses) => clauses
                 .iter_mut()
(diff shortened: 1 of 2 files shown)


query-grammar/src/user_input_ast.rs
```diff
@@ -205,6 +225,16 @@ impl UserInputAst {
     pub fn or(asts: Vec<UserInputAst>) -> UserInputAst {
         UserInputAst::compose(Occur::Should, asts)
     }
+
+    pub fn set_default_field(&mut self, field: String) {
```
adamreichold: Do you really want to commit to this as public API while `UserInputLeaf` is only `pub(crate)`?

trinity-1686a: hum, no I don't

fulmicoton (approved):

Issue #2335, opened 2024-03-28
Panic occured while committing writer with sort by text field.


gy0801151351:
**Which version of tantivy are you using?**
 tantivy v0.21.1

**To Reproduce**

```
use tantivy::schema::{SchemaBuilder, FAST, STORED, STRING, TEXT};
use tantivy::{doc, IndexBuilder, IndexSettings, IndexSortByField, Order};

fn main() -> tantivy::Result<()> {
    let mut schema_builder = SchemaBuilder::new();
    let id = schema_builder.add_text_field("id", STRING | FAST | STORED);
    let name = schema_builder.add_text_field("name", TEXT | STORED);

    let index = IndexBuilder::new()
        .schema(schema_builder.build())
        .settings(IndexSettings {
            sort_by_field: Some(IndexSortByField {
                field: "id".to_string(),
                order: Order::Asc,
            }),
            ..Default::default()
        })
        .create_in_ram()?;

    let mut writer = index.writer(50_000_000)?;
    writer.add_document(doc! {
        id => "0001",
        name => "name1"
    })?;
    writer.add_document(doc! {
        id => "0002",
        name => "name2"
    })?;
    writer.commit()?;
    writer.wait_merging_threads()?;

    Ok(())
}
```

```

index out of bounds: the len is 0 but the index is 0stack backtrace:

   0: std::panicking::begin_panic_handler
             at /rustc/8ace7ea1f7cbba7b4f031e66c54ca237a0d65de6/library\std\src\panicking.rs:647
   1: core::panicking::panic_fmt
             at /rustc/8ace7ea1f7cbba7b4f031e66c54ca237a0d65de6/library\core\src\panicking.rs:72
   2: core::panicking::panic_bounds_check
             at /rustc/8ace7ea1f7cbba7b4f031e66c54ca237a0d65de6/library\core\src\panicking.rs:208
   3: core::slice::index::impl$2::index<u32>
             at /rustc/8ace7ea1f7cbba7b4f031e66c54ca237a0d65de6\library\core\src\slice\index.rs:255
   4: alloc::vec::impl$12::index<u32,usize,alloc::alloc::Global>
             at /rustc/8ace7ea1f7cbba7b4f031e66c54ca237a0d65de6\library\alloc\src\vec\mod.rs:2771
   5: tantivy::indexer::doc_id_mapping::DocIdMapping::get_new_doc_id
[...]

PR #2336, opened 2024-03-29
validate sort by field type

PSeitz:
https://github.com/quickwit-oss/tantivy/issues/2335


src/index/index.rs
```diff
@@ -251,6 +251,15 @@ impl IndexBuilder {
                         sort_by_field.field
                     )));
                 }
+                let supported_field_types = vec![Type::I64, Type::U64, Type::F64, Type::Date];
```
adamreichold:
```suggestion
                let supported_field_types = [Type::I64, Type::U64, Type::F64, Type::Date];
```

fulmicoton (approved):

Issue #2248, opened 2023-11-11
Add cardinality aggregation.


fmassot:
It's already tracked in [Quickwit](https://github.com/quickwit-oss/quickwit/issues/3779) but not in tantivy. 

This is the same aggregation as the one defined in [elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-cardinality-aggregation.html)

saroh:
I gave a quick look, unfortunately no time to implem but here are a few pointers for someone else, or future me :D 

<details><summary>Spoilers ??????</summary>
<p>
The code should be pretty similar to the percentiles aggregation (see implem PR #1984)

`PercentilesCollector` -> `CardinalityCollector` which would hold a hyperloglog implementing add, merge and count.

hashing -> values need to be hashed in order to map them to a uniform distribution. It's possible the hll implem hashes => need to make sure the hash is computent just once

missing values -> it may be faster to simply check if we've seen a missing value and add +1 to the final count rather than blindly adding the same value's hash to the hll each time we encounter one.

</p>
</details>

PR #2337, opened 2024-04-01
feat: cardinality aggregation

raphaelcoeffic:
Implements #2248

### DIFF
--- a/src/aggregation/metric/cardinality.rs
+++ b/src/aggregation/metric/cardinality.rs
@@ -3,6 +3,7 @@ use std::hash::BuildHasher;
 
 use columnar::column_values::CompactSpaceU64Accessor;
 use hyperloglogplus::{HyperLogLog, HyperLogLogPlus};
+use rustc_hash::FxHashSet;
 use serde::{Deserialize, Serialize};
 
 use crate::aggregation::agg_req_with_accessor::AggregationsWithAccessor;
@@ -12,6 +13,7 @@ use crate::aggregation::*;
 
 use crate::TantivyError;
 
+use self::agg_req_with_accessor::AggregationWithAccessor;
 use self::intermediate_agg_result::IntermediateAggregationResult;
 use self::intermediate_agg_result::IntermediateMetricResult;
 
@@ -60,8 +62,7 @@ impl CardinalityAggregationReq {
 
 #[derive(Clone, Debug, PartialEq)]
 pub(crate) struct SegmentCardinalityCollector {
-    // field_type: ColumnType,
-    cardinality: CardinalityCollector,
+    entries: FxHashSet<u64>,
     column_type: ColumnType,
     accessor_idx: usize,
 }
@@ -69,10 +70,55 @@ pub(crate) struct SegmentCardinalityCollector {
 impl SegmentCardinalityCollector {
     pub fn from_req(column_type: ColumnType, accessor_idx: usize) -> Self {
         Self {
-            cardinality: CardinalityCollector::new(),
+            entries: Default::default(),
             column_type,
             accessor_idx,
         }
+    }
+
+    fn into_intermediate_metric_result(
+        self,
+        agg_with_accessor: &AggregationWithAccessor,
+    ) -> crate::Result<IntermediateMetricResult> {
+        let mut collector = CardinalityCollector::new();
+        let entries: Vec<u64> = self.entries.into_iter().collect();
+        if self.column_type == ColumnType::Str {
+            // TODO: better error handling
+            let term_dict = agg_with_accessor.str_dict_column.as_ref().cloned().unwrap();
+            for term_id in entries {
+                let mut buffer = String::new();
+                if !term_dict.ord_to_str(term_id, &mut buffer)? {
+                    return Err(TantivyError::InternalError(format!(
+                        "Couldn't find term_id {term_id} in dict"
+                    )));
+                }
+                collector.sketch.insert_any(&buffer);
+            }
+        } else if self.column_type == ColumnType::IpAddr {
+            let compact_space_accessor = agg_with_accessor
+                .accessor
+                .values
+                .clone()
+                .downcast_arc::<CompactSpaceU64Accessor>()
+                .map_err(|_| {
+                    TantivyError::AggregationError(
+                        crate::aggregation::AggregationError::InternalError(
+                            "Type mismatch: Could not downcast to CompactSpaceU64Accessor"
+                                .to_string(),
+                        ),
+                    )
+                })?;
+            for val in entries {
+                let val: u128 = compact_space_accessor.compact_to_u128(val as u32);
+                collector.sketch.insert_any(&val);
+            }
+        } else {
+            for val in entries {
+                collector.sketch.insert_any(&val);
+            }
+        }
+
+        Ok(IntermediateMetricResult::Cardinality(collector))
     }
 }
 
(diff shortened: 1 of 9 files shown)


src/aggregation/metric/cardinality.rs
```diff
@@ -0,0 +1,381 @@
...
+        &mut self,
+        docs: &[crate::DocId],
+        agg_with_accessor: &mut AggregationsWithAccessor,
+    ) -> crate::Result<()> {
+        let bucket_agg_accessor = &mut agg_with_accessor.aggs.values[self.accessor_idx];
+        self.collect_block_with_field(docs, bucket_agg_accessor);
+
+        let col_block_accessor = &bucket_agg_accessor.column_block_accessor;
+        if self.column_type == ColumnType::Str {
+            for term_id in col_block_accessor.iter_vals() {
```
fulmicoton:
```suggestion
            for term_ord in col_block_accessor.iter_vals() {
```

We call ordered ids ordinals in tantivy.
So here that would be `term_ord`.

src/aggregation/metric/cardinality.rs
```diff
@@ -0,0 +1,381 @@
...
+            let term_dict = agg_with_accessor.str_dict_column.as_ref().cloned().unwrap();
+            for term_id in entries {
+                if term_id == u64::MAX {
+                    let missing_key = self
+                        .missing
+                        .as_ref()
+                        .expect("Found placeholder term_id but `missing` is None");
+                    match missing_key {
+                        Key::Str(missing) => {
+                            self.cardinality.sketch.insert_any(&missing);
```
fulmicoton: It might be worth treating the missing value as special as it can be very frequent, and avoid pushing it in the sketch more than once.

src/aggregation/metric/cardinality.rs
```diff
@@ -0,0 +1,381 @@
...
+                        }
+                        Key::F64(val) => {
+                            let val = f64_to_u64(*val);
+                            self.cardinality.sketch.insert_any(&val);
```
fulmicoton, line 122: I think the result woudl be more accurate if we pushed a type discriminant in the sketch.

src/aggregation/metric/cardinality.rs
```diff
@@ -0,0 +1,381 @@
...
+        }
+    }
+
+    fn into_intermediate_metric_result(
+        mut self,
+        agg_with_accessor: &AggregationWithAccessor,
+    ) -> crate::Result<IntermediateMetricResult> {
+        if self.column_type == ColumnType::Str {
+            let mut buffer = String::new();
+            let entries: Vec<u64> = self.entries.into_iter().collect();
```
fulmicoton: no need to collect it here.

src/aggregation/metric/cardinality.rs
```diff
@@ -0,0 +1,381 @@
...
+                            self.cardinality.sketch.insert_any(&val);
+                        }
+                    }
+                } else {
+                    if !term_dict.ord_to_str(term_id, &mut buffer)? {
+                        return Err(TantivyError::InternalError(format!(
+                            "Couldn't find term_id {term_id} in dict"
+                        )));
+                    }
+                    self.cardinality.sketch.insert_any(&buffer);
```
fulmicoton, line 129: I think the result woudl be more accurate if we pushed a type discriminant in the sketch.

The most efficient way to do this would be to use a salted hasher.

src/aggregation/metric/cardinality.rs
```diff
@@ -0,0 +1,381 @@
...
+    // TODO
+    sketch: HyperLogLogPlus<u64, DefaultBuildHasher>,
+}
+impl Default for CardinalityCollector {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl PartialEq for CardinalityCollector {
```
fulmicoton, line 231: is Eq required? @PSeitz

fulmicoton:
@PSeitz I leave you the rest of the code review. Also can you open a ticket to one or another isolate aggregation? I think it is a bit overkill for most tantivy user to have to depend on hyperloglog.

src/aggregation/metric/cardinality.rs
raphaelcoeffic: fixed

src/aggregation/metric/cardinality.rs
raphaelcoeffic: done

src/aggregation/metric/cardinality.rs
raphaelcoeffic: fixed

src/aggregation/metric/cardinality.rs
raphaelcoeffic: Not sure I understand here. Is that about fields that hold multiple value types?

raphaelcoeffic: done. Test case added as well.

src/aggregation/metric/cardinality.rs
```diff
@@ -0,0 +1,410 @@
...
+        }
+    }
+
+    fn into_intermediate_metric_result(
+        mut self,
+        agg_with_accessor: &AggregationWithAccessor,
+    ) -> crate::Result<IntermediateMetricResult> {
+        if self.column_type == ColumnType::Str {
+            let mut buffer = String::new();
+            let term_dict = agg_with_accessor.str_dict_column.as_ref().cloned().unwrap();
```
PSeitz: .unwrap_or_else(|| {
                    StrColumn::wrap(BytesColumn::empty(agg_with_accessor.accessor.num_docs()))
                });

src/aggregation/metric/cardinality.rs
PSeitz: yes, we store the intermediate results in a hashmap

src/aggregation/metric/cardinality.rs
```diff
@@ -0,0 +1,410 @@
...
+        Self {
+            cardinality: CardinalityCollector::new(column_type as u8),
+            entries: Default::default(),
+            column_type,
+            accessor_idx,
+            missing: missing.clone(),
+        }
+    }
+
+    fn collect_block_with_field(
```
PSeitz: this does not collect, but just fetches the data

src/aggregation/metric/cardinality.rs
PSeitz: it's about different types, we support mixed types on the same column name. Internally they we differentiate them via `(ColumnType, Column)`.
e.g. the bytes of the number `[1u8,2,3,4,5,6,7,8]` are not equal to the bytes of a string `[1u8,2,3,4,5,6,7,8]`. It's not a problem during collection, because we generate two collectors, one for numbers and one for strings. But it may cause collisions during merging.
If you prefix the data with e.g. ColumnType discriminator, you won't get the collision

Edit: Ah it's already done, I just saw the default value which is set to 0

src/aggregation/metric/cardinality.rs
```diff
@@ -0,0 +1,410 @@
...
+
+#[cfg(test)]
+mod tests {
+
+    use std::net::IpAddr;
+    use std::str::FromStr;
+
+    use columnar::MonotonicallyMappableToU64;
+
+    use crate::aggregation::agg_req::Aggregations;
```
PSeitz, line 268: can you add this aggregation to `test_aggregation_flushing`

PSeitz:
Looks good so far, I left some comments.

I don't think `hyperloglog` is really big on its own, but all aggregations together may cost quite a bit if you don't use it.

src/aggregation/metric/cardinality.rs
raphaelcoeffic: done

src/aggregation/metric/cardinality.rs
raphaelcoeffic: done

src/aggregation/metric/cardinality.rs
raphaelcoeffic: done

raphaelcoeffic:
> Looks good so far, I left some comments.

Thx @PSeitz, it should be all fixed now.

PSeitz (approved):

PSeitz:
Thanks @raphaelcoeffic !

PR #2343, opened 2024-04-04
change debug_assert to assert in term length

PSeitz:
change `debug_assert` to `assert` since the bug with the truncated Terms only occurs in prod


fulmicoton:
I agree with the change. Can you check that it does not provoke a regression however?

PSeitz:
I don't see an impact on indexing performance

PR #2344, opened 2024-04-04
uprade dependencies

PSeitz:
The ahash upgrade fixes nightly compilation


fulmicoton:
I'm not sure what is the point of forcing the upgrade of a minor. Shouldn't we just upgrade ahash?

PSeitz:
> I'm not sure what is the point of forcing the upgrade of a minor. Shouldn't we just upgrade ahash?

It forces an upgrade, but I think we can also rely on cargo. Updated the PR

fulmicoton (approved):

PR #2346, opened 2024-04-08
fix postcard compatibility for top_hits, add postcard test

### LINKED ISSUE #2347: Top Hits Agg Performance: Delay fetching `get_document_field_data` in top_hits
`get_document_field_data` is currently called for every document collected instead of the top hits.

### LINKED ISSUE #4840: sending a query with `top_hits` causes a panic
As reported by a user:
While testing quickwit, the following ES query triggers a panic

```json
{
  "query": {
    "bool": {
      "must": [
        {
          "multi_match": {
            "query": "GET POST",
            "fields": [
              "db_statement"
            ]
          }
        },
        {
          "range": {
            "duration": {
              "gte": 0.6,
              "lte": 0.9
            }
          }
        },
        {
          "range": {
            "timestamp": {
              "gte": "2024-03-27T14:00:00Z",
              "lte": "2024-03-27T14:10:00Z"
            }
          }
        }
      ]
    }
  },
  "aggs": {
    "top_hits": {
      "top_hits": {
        "size": 10,
        "_source": {
          "includes": [
            "db_statement",
            "duration",
            "timestamp"
          ]
        },
        "sort": [
          {
            "_score": "desc"
          }
        ]
      }
    }
  },
  "size": 0
}
```

The cause is that the fruit of the top_hits collector are not serializable.

```
thread 'thread 'quickwit-search-3quickwit-search-0' panicked at quickwit-search/src/collector.rsthread ':quickwit-search-1680' panicked at ' panicked at quickwit-search/src/collector.rsquickwit-search/src/collector.rs::68022:::
22680Collector fruit should be serializable.: SerializeSeqLengthUnknown:
:
Collector fruit should be serializable.: SerializeSeqLengthUnknown22
:
Collector fruit should be serializable.: SerializeSeqLengthUnknown
```

top_hits is not considered supported by quickwit at the moment, so in a sense this is ok.

Supporting it is not super simple as it requires coding the fetch docs phase.
We need either to support it, or fix the code to make sure we return an explicit error.

PSeitz:
fixes https://github.com/quickwit-oss/quickwit/issues/4840

Fix intermediate result postcard serialization by removing flatten
Fix intermediate result postcard deserialization by replacing `OwnedValue` (no self-describing formats)

Delay doc value fetchting to end of segment collection, closes https://github.com/quickwit-oss/tantivy/issues/2347
Fix naming and struct organization

### DIFF
--- a/src/aggregation/metric/mod.rs
+++ b/src/aggregation/metric/mod.rs
@@ -25,6 +25,8 @@ mod stats;
 mod sum;
 mod top_hits;
 
+use std::collections::HashMap;
+
 pub use average::*;
 pub use count::*;
 pub use max::*;
@@ -35,6 +37,8 @@ use serde::{Deserialize, Serialize};
 pub use stats::*;
 pub use sum::*;
 pub use top_hits::*;
+
+use crate::schema::OwnedValue;
 
 /// Single-metric aggregations use this common result structure.
 ///
@@ -92,8 +96,9 @@ pub struct TopHitsVecEntry {
 
     /// Search results, for queries that include field retrieval requests
     /// (`docvalue_fields`).
-    #[serde(flatten)]
-    pub search_results: FieldRetrivalResult,
+    #[serde(rename = "docvalue_fields")]
+    #[serde(skip_serializing_if = "HashMap::is_empty")]
+    pub doc_value_fields: HashMap<String, OwnedValue>,
 }
 
 /// The top_hits metric aggregation results a list of top hits by sort criteria.
--- a/src/aggregation/metric/top_hits.rs
+++ b/src/aggregation/metric/top_hits.rs
@@ -1,7 +1,9 @@
 use std::collections::HashMap;
 use std::fmt::Formatter;
+use std::net::Ipv6Addr;
 
 use columnar::{ColumnarReader, DynamicColumn};
+use common::DateTime;
 use regex::Regex;
 use serde::ser::SerializeMap;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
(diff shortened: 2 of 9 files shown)


src/aggregation/metric/mod.rs
```diff
@@ -92,8 +96,9 @@ pub struct TopHitsVecEntry {
 
     /// Search results, for queries that include field retrieval requests
     /// (`docvalue_fields`).
-    #[serde(flatten)]
-    pub search_results: FieldRetrivalResult,
```
fulmicoton, line 96: this should have been `FieldRetrievalResult` too. 

Is it still used? Can we remove or rename it?

src/aggregation/metric/top_hits.rs
```diff
@@ -175,11 +194,24 @@ impl RetrievalFields {
...
+    pub fn value_field_names(&self) -> Vec<&str> {
+        self.doc_value_fields.iter().map(|s| s.as_str()).collect()
+    }
+
     fn get_document_field_data(
         &self,
         accessors: &HashMap<String, Vec<DynamicColumn>>,
         doc_id: DocId,
-    ) -> FieldRetrivalResult {
+    ) -> IntermediateFieldRetrivalResult {
```
fulmicoton: Retrieval

src/aggregation/metric/top_hits.rs
```diff
@@ -92,42 +94,59 @@ pub struct TopHitsAggregation {
...
+        let mut map = serializer.serialize_map(Some(1))?;
+        map.serialize_entry(field, order)?;
+        map.end()
+    }
+}
+
+impl<'de> Deserialize<'de> for KeyOrder {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where D: Deserializer<'de> {
+        let mut k_o = <HashMap<String, Order>>::deserialize(deserializer)?.into_iter();
```
fulmicoton: variable name...

src/aggregation/metric/top_hits.rs
```diff
@@ -92,42 +94,59 @@ pub struct TopHitsAggregation {
...
 
-    fn resolve_field_names(&mut self, reader: &ColumnarReader) -> crate::Result<()> {
+impl TopHitsAggregation {
+    /// Validate and resolve field retrieval parameters
+    pub fn validate_and_resolve_field_names(
+        &mut self,
+        reader: &ColumnarReader,
+    ) -> crate::Result<()> {
         // Tranform a glob (`pattern*`, for example) into a regex::Regex (`^pattern.*$`)
         let globbed_string_to_regex = |glob: &str| {
```
fulmicoton: can we extract this function? (I know this is not from this PR but still)

src/aggregation/metric/top_hits.rs
```diff
@@ -344,7 +366,7 @@ impl PartialEq for ComparableDocFeature {
 impl Eq for ComparableDocFeature {}
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
-struct ComparableDocFeatures(Vec<ComparableDocFeature>, FieldRetrivalResult);
+struct ComparableDocFeatures(Vec<ComparableDocFeature>, IntermediateFieldRetrivalResult);
```
fulmicoton: Can we fix the terrible struct name and remove the 2 positional arguments and use a regular struct?

src/aggregation/metric/top_hits.rs
```diff
@@ -344,7 +366,7 @@ impl PartialEq for ComparableDocFeature {
...
 
 impl Ord for ComparableDocFeatures {
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
```
fulmicoton, line 366: My understanding is that the `TopHitsCollector` needlessly access retrieved fields for every single document it collects. (Not even those who enter the Top K)

This is quite terrible. Can we at least open a ticket to fix this?

src/aggregation/metric/mod.rs
PSeitz: It can't be `FieldRetrievalResult`, since postcard does not support flatten

src/aggregation/metric/top_hits.rs
PSeitz: Yes, that's quite terrible, I opened an issue here:
https://github.com/quickwit-oss/tantivy/issues/2347

PSeitz: I fixed it in this PR, fields will be retrieved after segment collection now

src/aggregation/metric/top_hits.rs
PSeitz: Yes, I did also do a pass on all the names and some comments

src/aggregation/metric/top_hits.rs
fulmicoton: thx

src/aggregation/metric/top_hits.rs
```diff
@@ -92,42 +94,59 @@ pub struct TopHitsAggregation {
...
 }
 
-/// Search query result for each matched document
-/// TODO: move this to a common module
+/// Intermediate Search query result for each matched document
+///
+/// Unlike the final result, this struct is postcard compatible.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
-pub struct FieldRetrivalResult {
+pub struct IntermediateFieldRetrivalResult {
```
fulmicoton: The correct spelling is `Retri*e*val`

src/aggregation/metric/top_hits.rs
fulmicoton: Explicit variable names please

src/aggregation/metric/top_hits.rs
```diff
@@ -1,7 +1,9 @@
 use std::collections::HashMap;
 use std::fmt::Formatter;
```
fulmicoton:
```suggestion
```

fulmicoton (approved):

fulmicoton (changes requested):
ComparableDocFeatures

fulmicoton (approved):

PR #2351, opened 2024-04-09
check unsupported parameters top_hits

PSeitz:


src/aggregation/metric/top_hits.rs
```diff
@@ -148,6 +157,42 @@ impl TopHitsAggregation {
         &mut self,
         reader: &ColumnarReader,
     ) -> crate::Result<()> {
+        let use_doc_value_fields_err = |parameter: &str| {
```
fulmicoton: nitpick

I prefer we avoid closure and have actual functions for those.

fulmicoton (approved):

PR #2360, opened 2024-04-15
Fix trait bound of StoreReader::iter

adamreichold:
Similar to `StoreReader::get`, `StoreReader::iter` should only require `DocumentDeserialize` and not `Document`.


PSeitz (approved):

PSeitz:
Thanks!

PR #2364, opened 2024-04-19
Owned value relying on Vec instead of BTreeMap

fulmicoton:


src/schema/document/owned_value.rs
```diff
@@ -363,7 +364,8 @@ impl From<PreTokenizedString> for OwnedValue {
 
 impl From<BTreeMap<String, OwnedValue>> for OwnedValue {
```
adamreichold, line 369: Should this change to `From<Vec<(String, OwnedValue)>>` as well or do we actually need the conversion (instead of just wrapping as before)?

fulmicoton:
Impact on resident memory on quickwit.
Before on the left, After on the right

image

src/schema/document/owned_value.rs
fulmicoton: it is used due to the serde_json conversion.

adamreichold: And you do not want to apply

```diff
diff --git a/src/schema/document/default_document.rs b/src/schema/document/default_document.rs
index fcf374dfe..077554f13 100644
--- a/src/schema/document/default_document.rs
+++ b/src/schema/document/default_document.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{HashMap, HashSet};
 use std::net::Ipv6Addr;
 
 use common::DateTime;
@@ -150,7 +150,7 @@ impl TantivyDocument {
     }
 
     /// Add a dynamic object field
-    pub fn add_object(&mut self, field: Field, object: BTreeMap<String, OwnedValue>) {
+    pub fn add_object(&mut self, field: Field, object: Vec<(String, OwnedValue)>) {
         self.add_field_value(field, object);
     }
 
diff --git a/src/schema/document/owned_value.rs b/src/schema/document/owned_value.rs
index 738194908..fd1b232a6 100644
--- a/src/schema/document/owned_value.rs
+++ b/src/schema/document/owned_value.rs
@@ -1,4 +1,3 @@
-use std::collections::BTreeMap;
 use std::fmt;
 use std::net::Ipv6Addr;
 
@@ -359,10 +358,9 @@ impl From<PreTokenizedString> for OwnedValue {
     }
 }
 
-impl From<BTreeMap<String, OwnedValue>> for OwnedValue {
-    fn from(object: BTreeMap<String, OwnedValue>) -> OwnedValue {
-        let key_values = object.into_iter().collect();
-        OwnedValue::Object(key_values)
+impl From<Vec<(String, OwnedValue)>> for OwnedValue {
+    fn from(object: Vec<(String, OwnedValue)>) -> OwnedValue {
+        OwnedValue::Object(object)
     }
 }
```

because it is technically breaking and you do need this ASAP as a fix for the performance bug it implies?

Personally, I think the effect so large and the `add_object` API so new that this might be worth a little bit of semver breakage. (Worst case would be yanking 0.20.0 after 0.20.1 is out.)

PSeitz: No, this won't be included in a bugfix release

PSeitz (approved):

Issue #1795, opened 2023-01-17
Refactoring: removing Field and use &str in the public API of tantivy


fulmicoton:
We want to remove `Field` from the public API, or most of the public API.
There are multiple benefits for that refactoring:
- It will eventually make it possible to change schema in tantivy.
- It will make it easier to unify the usage of fastifeld with a strict schema or with a dynamic schema.
- It will remove a bit of a footgun.
- It will improve / make it easier to write better error message

# Pitfalls

We don't want to  hurt (too much):
- Document building performance.
- Document building experience.
- Docstore compression
- Indexing performance.
- Document consumption

# Names

Field names are &str.
A FieldId object still exists, but it is an internal object.

# Composite file

Will be addressed by field names instead of field Id.
The change should be relatively trivial.

# Document

The `DocumentBuilder<'a>` object will be hold a reference to a `&'a Schema`.

User will have to create documents passing field names as &str.

While any &str is accepted, as an optimization if we judge this necessary/helpful-,
we can hold a  `HashMap<(*const u8, usize), InternalFieldId>` in the schema to return more rapidly 
when the user provides a `field_name: &str` borrowed directly from the schema.

The user will then pass `DocumentBuilder` object to the `IndexWriter`, and not document anymore.
The IndexWriter will then just check the schema refered to is correct, and just keep the field
as an `InternalDocument` (not part of the public API).
The `InternalDocument` is very similar to todays `Document`.

# Segment writer

We will keep a notion of `Field` at indexing time, for efficiency reasons.
It will be segment specific.

The segment writer remains uncahnged for the most part.
It still manipulates a `Term` that joins a u32 serialized in a slice.
This Term is however now named `IndexingTerm` and is only private API.

Upon serialization, the writer resolves that u32 field identifier into a str and 
the segment serializer only manipulates fieldnames &str.

# Document Store

Each segment document store comes with its own field id -> field_name maping.
Upon read, the document will rebuild user facing Documents, addressed by strings.

# Query and collectors, fast fields.

The inverted index and fast fields will only offer accessors with &str.
Query and collectors will simply refer to field as `&str` as well.
A Term public object will subsist to be used in some of the public API:
TermQuery building, Query::query_terms, 

It will however be separate from the `IndexingTerm` and may not need to have
this strange serialized contraption.

fulmicoton:
@adamreichold @PSeitz @shikhar your thoughts?

PSeitz:
# Document
[...]

PR #2366, opened 2024-04-22
split Term into Term and IndexingTerm

PSeitz:
- **split term and indexing term**
- **add JsonTermSerializer**
- **cleanup**
- **clippy**

This splits `Term` into `Term` and `IndexingTerm`. `Term` is used for querying, `IndexingTerm` during indexing.
This allows to reduce the complexity of `Term` and a more compact format for `IndexingTerm` by dropping the `type` marker at byte 5. 

Part of: https://github.com/quickwit-oss/tantivy/issues/1795


fulmicoton:
As discussed offline, this looks good but let's not merge not progress on #1795 at the moment.

PSeitz-dd:
https://github.com/quickwit-oss/tantivy/pull/2744

PR #2372, opened 2024-04-24
fix ReferenceValue API flaw

PSeitz:
Remove `Facet` and `TokenizedString` values from the `ReferenceValue` API,
as this requires the trait implementation to have the actual objects stored somewhere, which may not be feasible for some implementations.

Since `TokenizedString` is quite niche, I just copy it into a Box,
instead of designing a reference API around it.

precursor PR to refactor `TantivyDocument` for https://github.com/quickwit-oss/quickwit/issues/4890

### DIFF
--- a/src/schema/document/value.rs
+++ b/src/schema/document/value.rs
@@ -3,7 +3,6 @@ use std::net::Ipv6Addr;
 
 use common::DateTime;
 
-use crate::schema::Facet;
 use crate::tokenizer::PreTokenizedString;
 
 /// A single field value.
@@ -82,7 +81,7 @@ pub trait Value<'a>: Send + Sync + Debug {
     #[inline]
     /// If the Value is a pre-tokenized string, returns the associated string. Returns None
     /// otherwise.
-    fn as_pre_tokenized_text(&self) -> Option<&'a PreTokenizedString> {
+    fn as_pre_tokenized_text(&self) -> Option<Box<PreTokenizedString>> {
         self.as_leaf().and_then(|leaf| leaf.as_pre_tokenized_text())
     }
 
@@ -94,7 +93,7 @@ pub trait Value<'a>: Send + Sync + Debug {
 
     #[inline]
     /// If the Value is a facet, returns the associated facet. Returns None otherwise.
-    fn as_facet(&self) -> Option<&'a Facet> {
+    fn as_facet(&self) -> Option<&'a str> {
         self.as_leaf().and_then(|leaf| leaf.as_facet())
     }
 
@@ -132,7 +131,7 @@ pub trait Value<'a>: Send + Sync + Debug {
 }
 
 /// A enum representing a leaf value for tantivy to index.
-#[derive(Clone, Copy, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub enum ReferenceValueLeaf<'a> {
     /// A null value.
     Null,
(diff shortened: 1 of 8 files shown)


src/schema/document/value.rs
```diff
@@ -146,16 +145,16 @@ pub enum ReferenceValueLeaf<'a> {
...
+    Facet(&'a str),
     /// Arbitrarily sized byte array
     Bytes(&'a [u8]),
     /// IpV6 Address. Internally there is no IpV4, it needs to be converted to `Ipv6Addr`.
     IpAddr(Ipv6Addr),
     /// Bool value
     Bool(bool),
     /// Pre-tokenized str type,
-    PreTokStr(&'a PreTokenizedString),
+    PreTokStr(Box<PreTokenizedString>),
```
fulmicoton, line 158: We have started using PreTokenizedString in the concatenated field implementation in tantivy.
We use it for bool/numbers.

I am bit scared this PR could involve a non-benine regression for concatenate field.
@trinity-1686a has more info about why it is used.

fulmicoton: Maybe `Box<Arc<...>>` !?

fulmicoton:
I don't think you explained the actual thing you are trying to fix.

src/schema/document/value.rs
trinity-1686a: (`Box<Arc<...>>` is just an `Arc<...>` with extra indirection)

src/schema/document/value.rs
```diff
@@ -146,16 +145,16 @@ pub enum ReferenceValueLeaf<'a> {
     F64(f64),
     /// Date/time with nanoseconds precision
     Date(DateTime),
-    /// Facet
-    Facet(&'a Facet),
+    /// Facet, needs to match the format of [Facet].
```
adamreichold: I think this should clarify that the _encoded_ format is meant, not the human-readable slash-separated one.

src/schema/document/value.rs
```diff
@@ -261,9 +260,9 @@ impl<'a> ReferenceValueLeaf<'a> {
     #[inline]
     /// If the Value is a pre-tokenized string, returns the associated string. Returns None
     /// otherwise.
-    pub fn as_pre_tokenized_text(&self) -> Option<&'a PreTokenizedString> {
+    pub fn as_pre_tokenized_text(&self) -> Option<Box<PreTokenizedString>> {
```
adamreichold, line 264: This could still return `Option<&PreTokenizedString>` even if internally, the storage is owned, to avoid the unconditional `clone` in the accessor. (The caller can `clone` or `cloned` this just as well.)

PSeitz: so where would the object live

src/schema/document/value.rs
PSeitz: PreTokenizedString is cloned once already for the tokenization 
`BoxTokenStream::new(PreTokenizedStream::from(tok_str.clone()))`

I think since we can probably remove `PreTokenizedString` from the concat field, we can keep it like this for now and if not revisit it later.

src/schema/document/value.rs
adamreichold: Note, that it cannot return `&'a PreTokenizedString`, but it can still return `&'self PreTokenizedString`, i.e. a short-lived reference into the `Box` bound to the `&self` reference, which the caller can clone if so desired.

src/schema/document/value.rs
fulmicoton: @trinity-1686a Usually you would be right but there is a different, and in this very specific case it is a notable one. mem::size of `Box<Arc<..>>` is larger than that of `Arc<...>`.

You might tell me "who cares": overall the earlier in fact takes more memory.
The difference here is that we are in an enum so this options happens to determine the size of all of the num nodes in our tree.

Overall it actually take LESS here to use `Box<Arc<T>>` instead of `Box<T>`.

Anyway @PSeitz moved concatenate to use JSON field instead so we don't care :)

fulmicoton (approved):
Approving following the work on quickwit#4937.

src/schema/document/value.rs
fulmicoton: I put my foot in my mouth.

PR #2386, opened 2024-05-06
run block decompression from executor

trinity-1686a:
decompressing block of documents is a cpu-bound task, which we currently do in an async context, blocking the executor. Maybe some other places could use the same treatment, this one was identified as very likely to be problematic.

### DIFF
--- a/src/core/executor.rs
+++ b/src/core/executor.rs
@@ -100,20 +100,71 @@ impl Executor {
     #[cfg(feature = "quickwit")]
     pub fn spawn_blocking<T: Send + 'static>(
         &self,
-        task: impl FnOnce() -> T + Send + 'static,
+        cpu_intensive_task: impl FnOnce() -> T + Send + 'static,
     ) -> impl std::future::Future<Output = Result<T, ()>> {
         match self {
-            Executor::SingleThread => Either::Left(std::future::ready(Ok(task()))),
+            Executor::SingleThread => Either::Left(std::future::ready(Ok(cpu_intensive_task()))),
             Executor::ThreadPool(pool) => {
-                let (sender, receiver) = oneshot::channel();
+                let (sender, receiver) = oneshot_with_sentinel::channel();
                 pool.spawn(|| {
-                    // we don't care if the receiver was dropped
-                    let _ = sender.send(task());
+                    if sender.is_closed() {
+                        return;
+                    }
+                    let task_result = cpu_intensive_task();
+                    let _ = sender.send(task_result);
                 });
 
                 let res = receiver.map(|res| res.map_err(|_| ()));
                 Either::Right(res)
             }
+        }
+    }
+}
+
+#[cfg(feature = "quickwit")]
+mod oneshot_with_sentinel {
+    use std::pin::Pin;
+    use std::sync::Arc;
+    use std::task::{Context, Poll};
+    // TODO get ride of this if oneshot ever gains a is_closed()
+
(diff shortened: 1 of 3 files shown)


src/core/executor.rs
```diff
@@ -91,6 +93,29 @@ impl Executor {
             }
         }
     }
+
+    /// Spawn a task on the pool, returning a future completing on task success.
+    ///
+    /// If the task panic, returns `Err(())`.
+    #[cfg(feature = "quickwit")]
+    pub fn spawn_blocking<T: Send + 'static>(
```
fulmicoton, line 101:
```suggestion
    pub(crate) fn spawn_blocking<T: Send + 'static>(
```

fulmicoton (approved):

src/core/executor.rs
```diff
@@ -91,6 +93,29 @@ impl Executor {
...
+        &self,
+        task: impl FnOnce() -> T + Send + 'static,
+    ) -> impl std::future::Future<Output = Result<T, ()>> {
+        match self {
+            Executor::SingleThread => Either::Left(std::future::ready(Ok(task()))),
+            Executor::ThreadPool(pool) => {
+                let (sender, receiver) = oneshot::channel();
+                pool.spawn(|| {
```
fulmicoton, line 109: can you check how we do this in quickwit? (run_cpu_intensive).
One refinement  is that we avoid running the future if the future has been dropped. 
We do that by checking that the oneshot sender is not closed right before executing task.

src/core/executor.rs
```diff
@@ -91,6 +93,29 @@ impl Executor {
...
+                    // we don't care if the receiver was dropped
+                    let _ = sender.send(task());
```
fulmicoton:
```suggestion
                    if sender.is_closed() {
                        return;
                    }
                    let task_result = task();
                    let _ = sender.send(task());
```

src/core/executor.rs
```diff
@@ -91,6 +93,29 @@ impl Executor {
...
         }
     }
+
+    /// Spawn a task on the pool, returning a future completing on task success.
+    ///
+    /// If the task panic, returns `Err(())`.
+    #[cfg(feature = "quickwit")]
+    pub fn spawn_blocking<T: Send + 'static>(
+        &self,
+        task: impl FnOnce() -> T + Send + 'static,
```
fulmicoton: nitpick naming `task` `cpu_intensive_task` might have better here I think

fulmicoton (approved):
See suggestion though

src/core/executor.rs
fulmicoton: we'd also need a test for cancellability. You can borrow the one from tantivy.

trinity-1686a: there doesn't appear to be a `is_closed()` on oneshot's sender. There is on tokio's, but i doubt we want to add that dependency

adamreichold: This sounds like something that could easily be added upstream in the `oneshot` crate? This way everyone would benefit from the improved API and since this the early cancellation check is an optimization, it could land when upstream releases the API addition. (Or a Git dependency on a fork could be used until then.)

adamreichold: (There is also `async-oneshot` which does support this, but I think contributing to an existing dependency is better than vetting a completely new one?)

trinity-1686a: I agree that it could get upstreamed. I'm implementing a wrapper to have it quickly, but will also submit a patch to `oneshot`

PR #2387, opened 2024-05-07
Removed usage of num_cpus

fulmicoton:


PSeitz (approved):

src/index/index.rs
```diff
@@ -340,7 +341,7 @@ impl Index {
     /// Replace the default single thread search executor pool
     /// by a thread pool with as many threads as there are CPUs on the system.
     pub fn set_default_multithread_executor(&mut self) -> crate::Result<()> {
-        let default_num_threads = num_cpus::get();
+        let default_num_threads = available_parallelism().unwrap().get();
```
adamreichold: This function already returns a `Result`, wouldn't it better to propagate the error than to `unwrap` here? Especially since `num_cpus` did handle it differently internally by always falling back to one if the underlying system calls yielded an error, i.e. it basically did `.map_or(1, |num| num.get());`.

PSeitz: Ah yes, failure is not uncommon it seems:

```
This function will, but is not limited to, return errors in the following cases:

    If the amount of parallelism is not known for the target platform.
    If the program lacks permission to query the amount of parallelism made available to it.
```

Issue #2374, opened 2024-04-24
Replace PreTokenizedString


PSeitz:
`PreTokenizedString` is a way to provide pre tokenized text to tantivy.

I think this can be removed and be handled by the user provided tokenizer instead.

E.g. in the example from `pre_tokenized_text.rs` below, the field `title` could contain the serialized JSON instead, which the tokenizer would parse to return the tokens.

```rust
let short_man_json = r#"{
    "title":[{
        "text":"The Old Man",
        "tokens":[
            {"offset_from":0,"offset_to":3,"position":0,"text":"The","position_length":1},
            {"offset_from":4,"offset_to":7,"position":1,"text":"Old","position_length":1},
            {"offset_from":8,"offset_to":11,"position":2,"text":"Man","position_length":1}
        ]
    }]
}"#;
```

PR #2389, opened 2024-05-09
Give allocation control to the caller instead of force a clone

adamreichold:
Achieved by moving the boxes out of the temporary reference wrappers which are cloneable themselves, i.e. if required the caller can clone them already or consume them to reuse existing allocations.

### DIFF
--- a/src/schema/document/value.rs
+++ b/src/schema/document/value.rs
@@ -27,7 +27,7 @@ pub trait Value<'a>: Send + Sync + Debug {
     }
 
     #[inline]
-    /// If the Value is a String, returns the associated str. Returns None otherwise.
+    /// If the Value is a leaf, returns the associated leaf. Returns None otherwise.
     fn as_leaf(&self) -> Option<ReferenceValueLeaf<'a>> {
         if let ReferenceValue::Leaf(val) = self.as_value() {
             Some(val)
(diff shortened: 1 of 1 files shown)


src/schema/document/value.rs
```diff
@@ -259,11 +260,11 @@ impl<'a> ReferenceValueLeaf<'a> {
     }
 
     #[inline]
-    /// If the Value is a pre-tokenized string, returns the associated string. Returns None
-    /// otherwise.
-    pub fn as_pre_tokenized_text(&self) -> Option<Box<PreTokenizedString>> {
+    /// If the Value is a pre-tokenized string, consumes it and returns the string.
+    /// Returns None otherwise.
+    pub fn into_pre_tokenized_text(self) -> Option<Box<PreTokenizedString>> {
```
adamreichold, line 265: Arguably, all methods on the two reference wrapper types could be `into_*(self)` instead of `as_*(&self)` for consistency but since the others are basically `Copy`, it does not make a difference there.

PSeitz:
PreTokenizedStrings is a corner case, which may be removed. Not a candidate for micro optimizations
Related: https://github.com/quickwit-oss/tantivy/issues/2374

adamreichold:
I do not consider this a micro optimization but just good API design: Give control to the caller as you do not know how your API is used. Maybe your corner case is their common case.

Especially since doing this does not cost anybody anything because the overall power of the API is the same due to the cloneability of the reference wrappers.

This also does not increase the maintenance burden in any way. If you do remove `PreTokenizedString`, you can just remove this code as well.

fulmicoton (approved):

PR #2391, opened 2024-05-10
Change in Executor API

fulmicoton:


Cargo.toml
```diff
@@ -15,12 +15,16 @@ rust-version = "1.63"
 exclude = ["benches/*.json", "benches/*.txt"]
 
 [dependencies]
```
adamreichold, line 17: I think this might be a good time to update `rayon` to its latest version 1.10 or at least 1.8 which pulls in the latest `rayon-core` 1.12 (instead of 1.9 as done currently).

src/core/executor.rs
```diff
@@ -11,11 +12,18 @@ use crate::TantivyError;
 /// First dependency hell. It is not a good idea to expose the
 /// API of a dependency, knowing it might conflict with a different version
 /// used by the client. Second, we may stop using rayon in the future.
+#[derive(Clone)]
```
adamreichold, line 10: I think the comment above this line needs to be changed.

src/lib.rs
```diff
@@ -166,7 +166,7 @@ mod future_result;
 
 // Re-exports
 pub use common::DateTime;
-pub use {columnar, query_grammar, time};
+pub use {columnar, query_grammar, rayon, time};
```
adamreichold: It did not miss anything, we actually only need to expose `rayon_core` because we only use `ThreadPool` in our public API and not `ParallelIterator` et al. `rayon_core` is supposedly much more stable exactly to make it easier to share with other crates in the dependency tree.

Cargo.toml
fulmicoton: The way Cargo interpret revision is that it will pull the highest available revision compatible according to semver.
So today, if you don't have a Cargo.lock you are already using the last rayon rev.

adamreichold: Ah yes, sorry about the noise. I was stilling think in 0.x mode...

PR #2396, opened 2024-05-12
Add Function Score Query

### LINKED ISSUE #2395: Adding Function Score Query

Recently I found that tantivy is lacking some common search engine properties like flexible scoring mechanism on retrieving the relevant docs. Currently users are able to tweak the score through `TopDoc`'s `tweak_score` method, the method evaluates the score at the last level which makes customizable scoring based on different search branches (Query) difficult.

For example, when using Disjunction Max with several queries, by passing a flexible closure on defining the score, we can easily fine-grained control the scoring and offsets of each queries. Existing solution relies on only Boosting which can nest several hierarchies and becomes hard to read, and Boosting doesn't allow offsets as well.

---

By introducing the FunctionScoreQuery, users can define their own closure for the score modification algorithm **in the query level**. The score tweak happens before the final `TopDoc`'s `tweak_score` method thus greater flexibility to users.

Introducing the FunctionScoreQuery brings several benefits:
- Rather than nesting several Queries and recursively apply the the `score` method on the scorer **when their requirements are complicated**, users can just define a function that is clean and neat. For simpler usecases, native query types like `BoostQuery` is prefered.
- Other language bindings (like tantivy-py) can hardly inherit the tantivy's `Query` struct due to different language implementations. For example, for tantivy-py, though `pyO3` can turn python objects into rust's struct, we have to first define all the utility classes before tantivy-py can consume.
- Existing Lucene and ElasticSearch (OpenSearch) have similar functions as well, while Lucene's implementation is very verbose, ElasticSearch's implementation restricts the scope inside the server which is not flexible at all. By utilizing rust's closure feature, existing Lucene and ElasticSearch users can easily migrate their workloads to tantivy with simpler syntax than their counterparts.

alex-au-922:
The implementation of FunctionScoreQuery which **accepts an existing query and a closure for applying new scoring**. You may consider this query as a `.map()` function which **does not filter the documents out, but rather change their score**.

Closes #2395

### DIFF
--- a/src/query/function_score_query.rs
+++ b/src/query/function_score_query.rs
@@ -1,11 +1,12 @@
-use crate::query::{EnableScoring, Explanation, Query, Scorer, Weight};
-use crate::{DocId, DocSet, Score, SegmentReader, TantivyError};
 use std::fmt;
 use std::sync::Arc;
 
+use crate::query::{EnableScoring, Explanation, Query, Scorer, Weight};
+use crate::{DocId, DocSet, Score, SegmentReader, TantivyError};
+
 type Function = Arc<dyn Fn(&SegmentReader, Score, DocId) -> Score + Send + Sync + 'static>;
 
-/// A FunctionScoreQuery modifies the score of 
+/// A FunctionScoreQuery modifies the score of
 /// matched documents using a custom function.
 /// ```rust
 /// use tantivy::collector::TopDocs;
@@ -42,7 +43,7 @@ type Function = Arc<dyn Fn(&SegmentReader, Score, DocId) -> Score + Send + Sync 
 ///         let documents = searcher.search(&query, &TopDocs::with_limit(2))?;
 ///         for (score, _) in documents {
 ///            assert_eq!(score, 0.42 * 2.0 + 1.0);
-///         } 
+///         }
 ///     }
 ///
 ///     Ok(())
@@ -56,10 +57,7 @@ pub struct FunctionScoreQuery {
 
 impl FunctionScoreQuery {
     /// Creates a new FunctionScoreQuery.
-    pub fn new(
-        query: Box<dyn Query>,
-        function: Function,
-    ) -> Self {
+    pub fn new(query: Box<dyn Query>, function: Function) -> Self {
         FunctionScoreQuery { query, function }
     }
 }
@@ -97,11 +95,7 @@ struct FunctionScoreWeight {
 }
 
 impl FunctionScoreWeight {
-    pub fn new(
-        weight: Box<dyn Weight>,
-        scoring_enabled: bool,
-        function: Function,
-    ) -> Self {
+    pub fn new(weight: Box<dyn Weight>, scoring_enabled: bool, function: Function) -> Self {
         FunctionScoreWeight {
             weight,
             scoring_enabled,
@@ -190,14 +184,14 @@ impl Scorer for FunctionScorer {
 #[cfg(test)]
 mod tests {
 
-    use super::FunctionScoreQuery;
-    use crate::query::{AllQuery, ConstScoreQuery, Query};
-    use crate::schema::FAST;
-    use crate::schema::{Schema, document::OwnedValue, TEXT, STORED};
-    use crate::{DocAddress, Index, IndexWriter, TantivyDocument};
-    use crate::collector::TopDocs;
-    use crate::{SegmentReader, DocId};
     use std::sync::Arc;
+
+    use super::FunctionScoreQuery;
+    use crate::collector::TopDocs;
+    use crate::query::{AllQuery, ConstScoreQuery, Query};
+    use crate::schema::document::OwnedValue;
+    use crate::schema::{Schema, FAST, STORED, TEXT};
+    use crate::{DocAddress, DocId, Index, IndexWriter, SegmentReader, TantivyDocument};
 
     #[test]
     fn test_function_score_query_explain() -> crate::Result<()> {
@@ -238,7 +232,7 @@ mod tests {
     #[test]
     fn test_function_score_get_fast_field_from_segment_reader() -> crate::Result<()> {
         let mut schema_builder = Schema::builder();
-        let title = schema_builder.add_text_field("title", TEXT | STORED | FAST); 
+        let title = schema_builder.add_text_field("title", TEXT | STORED | FAST);
         let schema = schema_builder.build();
         let index = Index::create_in_ram(schema);
         {
@@ -270,12 +264,12 @@ mod tests {
                 field_reader.ord_to_bytes(ord, &mut bytes).unwrap();
                 let title = String::from_utf8(bytes).unwrap();
                 score * title.len() as f32
-            }
+            },
         );
 
         let query = FunctionScoreQuery::new(
             Box::new(ConstScoreQuery::new(Box::new(AllQuery), 0.42)),
-            title_length_scorer
+            title_length_scorer,
         );
         let results = searcher.search(&query, &TopDocs::with_limit(10))?;
         for (score, doc_address) in results {
@@ -288,5 +282,4 @@ mod tests {
         }
         Ok(())
     }
-
 }
--- a/src/query/function_score_query.rs
+++ b/src/query/function_score_query.rs
@@ -37,8 +37,7 @@ type Function = Arc<dyn Fn(&SegmentReader, Score, DocId) -> Score + Send + Sync 
 ///     {
 ///         let query = FunctionScoreQuery::new(
 ///             Box::new(ConstScoreQuery::new(Box::new(AllQuery), 0.42)),
-///             Arc::new(move 
-/// |_, score, _| score * MULTIPLIER + OFFSET),
+///             Arc::new(move |_, score, _| score * MULTIPLIER + OFFSET),
 ///         );
 ///         let documents = searcher.search(&query, &TopDocs::with_limit(2))?;
 ///         for (score, _) in documents {
@@ -117,10 +116,8 @@ impl Weight for FunctionScoreWeight {
         if !self.scoring_enabled {
             Ok(Box::new(inner_scorer))
         } else {
-            let new_inner_scorer = self.weight.scorer(reader, boost)?;
             Ok(Box::new(FunctionScorer::new(
                 inner_scorer,
-                new_inner_scorer,
                 Arc::new(reader.clone()),
                 self.function.clone(),
             )))
@@ -148,22 +145,19 @@ impl Weight for FunctionScoreWeight {
     }
 }
 
-struct FunctionScorer<TDocSet: DocSet> {
-    docset: TDocSet,
+struct FunctionScorer {
     scorer: Box<dyn Scorer>,
     segment_reader: Arc<SegmentReader>,
     function: Function,
 }
 
-impl<TDocSet: DocSet> FunctionScorer<TDocSet> {
+impl FunctionScorer {
     pub fn new(
-        docset: TDocSet,
         scorer: Box<dyn Scorer>,
         segment_reader: Arc<SegmentReader>,
         function: Function,
     ) -> Self {
         FunctionScorer {
-            docset,
             scorer,
             segment_reader,
             function,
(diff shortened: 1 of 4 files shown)


adamreichold:
How you would say this relates to the existing [`tweak_score`](https://docs.rs/tantivy/latest/tantivy/collector/struct.TopDocs.html#method.tweak_score)? Would that also suffice to handle your use case?

adamreichold:
Ah sorry, you wrote that in the linked issue...

alex-au-922:
> Ah sorry, you wrote that in the linked issue...

Sorry that I'm not quite used to contribute to large OSS projects so I'm not sure the exact practices on linking Issues and PR. Will add the issue link in the PR next time

src/query/function_score_query.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    fn count(&self, reader: &SegmentReader) -> crate::Result<u32> {
+        self.weight.count(reader)
+    }
+}
+
+struct FunctionScorer<TDocSet: DocSet> {
+    docset: TDocSet,
+    scorer: Box<dyn Scorer>,
```
adamreichold: Do we really need both? Or rather, I think these must be one scorer to make sure the docset operations and the scoring operations stay synchronized, otherwise `self.docset.doc()` could refer to a different document than `self.scorer.score()`. (I believe the fast field test case should be able to surface this if it is an issue by adding at least one more document.)

src/query/function_score_query.rs
```diff
@@ -0,0 +1,289 @@
...
+///     let reader = index.reader()?;
+///     let searcher = reader.searcher();
+///     let MULTIPLIER: f32 = 2.0;
+///     let OFFSET: f32 = 1.0;
+///
+///     {
+///         let query = FunctionScoreQuery::new(
+///             Box::new(ConstScoreQuery::new(Box::new(AllQuery), 0.42)),
+///             Arc::new(move 
+/// |_, score, _| score * MULTIPLIER + OFFSET),
```
adamreichold: The formatting seems slightly off here, i.e. the line break is probably unnecessary?

src/query/function_score_query.rs
```diff
@@ -0,0 +1,289 @@
+use crate::query::{EnableScoring, Explanation, Query, Scorer, Weight};
+use crate::{DocId, DocSet, Score, SegmentReader, TantivyError};
+use std::fmt;
+use std::sync::Arc;
+
+type Function = Arc<dyn Fn(&SegmentReader, Score, DocId) -> Score + Send + Sync + 'static>;
```
adamreichold: If this closure is supposed to access the `SegmentReader`, I think it should return `crate::Result<Score>` to be able to forward errors. But then again `Scorer::score` currently does not allow errors so we cannot forward this anywhere.

Not sure how to best handle this, expect for making all scoring fallible, but that might have a measurable performance cost?

alex-au-922: We can handle the fallback case like this:

```rust
type Function = Arc<dyn Fn(&SegmentReader, Score, DocId) -> crate::Result<Score> + Send + Sync + 'static>;

impl Scorer for FunctionScorer {
    fn score(&mut self) -> Score {
        let score = self.scorer.score();
        let doc_id = self.doc();
        match (self.function)(self.segment_reader.as_ref(), score, doc_id) {
            Ok(new_score) => new_score,
            Err(_) => score,
        }
    }
}
```

But then we should have a mechanism to tell the users that some error happens. Do you have any idea on this?

adamreichold: > But then we should have a mechanism to tell the users that some error happens. Do you have any idea on this?

I would refrain from silently not adjusting the score. Either the operations work or they don't. If errors are supported, they should be propagated.

Just falling back to not modifying the score can also be implemented in the given function itself without any support in the scorer if that is indeed the chosen approach in a specific situation.

src/query/function_score_query.rs
alex-au-922: I have found that actually the `docset` is not really needed so I just deleted it. The `FunctionScorer` should look like this:

```rust
struct FunctionScorer {
    scorer: Box<dyn Scorer>,
    segment_reader: Arc<SegmentReader>,
    function: Function,
}
```

src/query/function_score_query.rs
alex-au-922: Should be some bug from my side. Will fix it later with the error handling.

src/query/function_score_query.rs
adamreichold: But could also please extend the fastfield-based test case to actually use this, i.e. have at least two documents which get two different scores so that the earlier issue would have surfaced? Thanks.

adamreichold:
Regarding the formatting errors, note that you'll need to use a nightly toolchain to apply the correct formatting configuration.

src/query/function_score_query.rs
```diff
@@ -0,0 +1,285 @@
...
+    ) -> Self {
+        FunctionScorer {
+            scorer,
+            segment_reader,
+            function,
+        }
+    }
+}
+
+impl DocSet for FunctionScorer {
```
fulmicoton: Can you implement the other methods too? seek for instance, has a default impemnetation but it will be catastrophically slow.

src/query/function_score_query.rs
```diff
@@ -0,0 +1,285 @@
+use std::fmt;
+use std::sync::Arc;
+
+use crate::query::{EnableScoring, Explanation, Query, Scorer, Weight};
+use crate::{DocId, DocSet, Score, SegmentReader, TantivyError};
+
+type Function = Arc<dyn Fn(&SegmentReader, Score, DocId) -> Score + Send + Sync + 'static>;
```
fulmicoton: If your fucntion needs to access a column from a fast field, you don't want to access it at each call. 

The function should be a factory that takes a `SegmentReader` and returns a scoring function

src/query/function_score_query.rs
adamreichold: I think the error handling will be resolved by https://github.com/quickwit-oss/tantivy/pull/2396#discussion_r1598176436 as a side effect.

alex-au-922:
I have removed the use of `SegmentReader` in the current version of `FunctionScoreQuery` as the query could be too versatile that users cannot optimize their query performance with the wrong use of `SegmentReader`.

The use of `SegmentReader` should be thoroughly discussed. We can even just limit them with the use of fastfield only.

The current signature should only be a simple `Fn(Score) -> Score` which can **simplify complicated nested boosting cases** and some **impossible maths like reciprocal, logarithm etc.**.

adamreichold:
> The use of SegmentReader should be thoroughly discussed. We can even just limit them with the use of fastfield only.

I think @fulmicoton's proposed approach is very right on target and has precedence in other Tantivy API, i.e. make the user pass a fallible factory function which given a segment reader will produce an infallible per-document scoring function.

alex-au-922:
> I think @fulmicoton's proposed approach is very right on target and has precedence in other Tantivy API, i.e. make the user pass a fallible factory function which given a segment reader will produce an infallible per-document scoring function.

What comes up in my mind is a new enum like

```rust
pub enum ScoreFunction {
    ScoreOnly(Arc<dyn Fn(Score) -> Score + Send + Sync + 'static>),
    ScoreWithDetails(Arc<dyn Fn(&SegmentReader, Score, DocId) -> Score + Send + Sync + 'static>),
}

pub struct FunctionScoreQuery {
    query: Box<dyn Query>,
    score_function: ScoreFunction,
}
```

so that we can match different types of `ScoreFunction`. However, this cannot prevent users from using the `ScoreWithDetails` when they don't need the `SegmentReader` at all.

If we accept users to dynamically use the segment reader based on different conditions, we have to include it in the closure input parameter and in the end each call will still calling the segment reader. Or do you have any suggestions?

adamreichold:
The idea is not allow the scoring function access to the segment reader but only to the relevant column itself, because after that column has been loaded, access to the values becomes essentially infallible.

```rust
type ScoreFunction = Arc<dyn Fn(DocId, Score) -> Score + Send + Sync + 'static>;

type ScoreFunctionFactory = Arc<dyn Fn(&SegmentReader) -> crate::Result<ScoreFunction> + Send + Sync + 'static>;
```

(The score function will access the column retrieved from the segment reader in the factory by capturing it.)

alex-au-922:
Yes, that makes sense by utilizing the type checker system. However, I'm still thinking how can we be benefited with not accessing the segment reader by every call, as @fulmicoton mentioned:

> If your fucntion needs to access a column from a fast field, you don't want to access it at each call.

I think it's better to separate two structs as `FunctionScoreQuery` (only dealing with the score) and a `FunctionSegmentQuery` (accessing the segment reader)?

fulmicoton:
@alex-au-922 ... Yes that's the approach taken in different part of the code. Collector -> SegmentCollector for instance.

alex-au-922:
Supposed the new signature works like how @adamreichold suggest, for a **new impl**, say `FunctionSegmentQuery`:

```rust
type ScoreFunction = Arc<dyn Fn(DocId, Score) -> Score + Send + Sync + 'static>;

type ScoreFunctionFactory = Arc<dyn Fn(&SegmentReader) -> crate::Result<ScoreFunction> + Send + Sync + 'static>;
```

Then we can create the `score_function` (` Arc<dyn Fn(DocId, Score) -> Score + Send + Sync + 'static>`) when the `Weight` trait first call the `for_each_pruning` in each new encounter of the segment.

```diff
impl Weight for FunctionSegmentWeight {
    fn scorer(&self, reader: &SegmentReader, boost: Score) -> crate::Result<Box<dyn Scorer>> {
        let inner_scorer = self.weight.scorer(reader, boost)?;
+        let score_function = (self.score_function_factory)(reader)?;
        if !self.scoring_enabled {
            Ok(Box::new(inner_scorer))
        } else {
            Ok(Box::new(FunctionSegmentScorer::new(
                inner_scorer,
                score_function,
            )))
        }
    }
    ...
}
```

 As there are many `DocId`s in a segment, we can only access individual column data only when we advance to the corresponding document. That's why our new `ScoreFunction` accepts both `DocId` and `Score`:

```rust
impl Scorer for FunctionSegmentScorer {
    fn score(&mut self) -> Score {
        let doc_id = self.scorer.doc();
        let score = self.scorer.score();
        (self.score_function)(doc_id, score)
    }
}
```

Hope that I understand this correctly. However, I don't think the problem that @adamreichold mentioned has solved:
> (The score function will access the column retrieved from the segment reader in the factory by capturing it.)

To what my understanding, the column data of individual document can only be **retrieved inside the `score()` function** when the scorer advances documents inside the segment one by one. Therefore the best we could do might be 

```rust
FunctionSegmentQuery::new(
    Box::new(ConstScoreQuery::new(Box::new(AllQuery), 0.42)),
    Arc::new(|segment_reader| {
        let title_reader = segment_reader.fast_fields().str("title").unwrap();
        if title_reader.is_none() {
            return Err(TantivyError::SchemaError("title field not found".to_string()));
        }
        let title_reader= title_reader.unwrap();
        Ok(Arc::new(move |doc_id, score| {
            // Calculate the product between the length of title and the original score
            let mut bytes = Vec::new();
            let ord = title_reader.term_ords(doc_id).next().unwrap();
            title_reader.ord_to_bytes(ord, &mut bytes).unwrap();
            let title = String::from_utf8(bytes).unwrap();
            score * title.as_str().len() as f32
        }))
    })
)
```
That means the `field` is loaded but **the actual data is loaded inside the closure**.

---

Or is my example too extreme? Say if we only operates on the fast fields:

```rust
FunctionSegmentQuery::new(
    Box::new(ConstScoreQuery::new(Box::new(AllQuery), 0.42)),
    Arc::new(|segment_reader| {
        let age_reader = segment_reader.fast_fields().u64("age").unwrap();
        if age_reader.is_none() {
            return Err(TantivyError::SchemaError("age field not found".to_string()));
        }
        let age_reader = age_reader.unwrap();
        Ok(Arc::new(move |doc_id, score| {
            let age = age_reader.get_val(doc_id) as f32;
            score * age
        }))
    })
)
```

The `age_reader` will return the infallible age value inside the closure. Is this what you mean?

fulmicoton:
@alex-au-922 This PR is consuming too much maintainer time for little benefit for the project. 
You can implement your custom queries within your own project, outside of tantivy. I will also close the associated issue as wontfix but invite readers to comment on it if they encounter a need for something like that.

I hope for your understanding.

PR #2397, opened 2024-05-12
Implement `std::io::Read` on `FileSlice`

tontinton:
A simple and useful way to read from a file slice.

For example: you can now use `std::io::copy` to copy a file slice into some other buffer, or even better you can now implement `std::io::Read` on a struct holding a file slice using 1 less copy.

### DIFF
--- a/common/src/file_slice.rs
+++ b/common/src/file_slice.rs
@@ -1,4 +1,5 @@
 use std::fs::File;
+use std::io::Read;
 use std::ops::{Deref, Range, RangeBounds};
 use std::sync::Arc;
 use std::{fmt, io};
@@ -368,9 +369,17 @@ impl FileHandle for OwnedBytes {
     }
 }
 
+impl Read for FileSlice {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        let num_read = self.read_bytes_into(buf, self.range.clone())?;
+        self.range.start += num_read;
+        Ok(num_read)
+    }
+}
+
 #[cfg(test)]
 mod tests {
-    use std::io;
+    use std::io::{self, Read};
     use std::ops::Bound;
     use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


PSeitz:
`OwnedBytes` already implements `std::io::Read`, which is returned by `read_bytes`

fulmicoton:
Closing as it does not seem necessary.

tontinton:
`OwnedBytes` allocates a vector, as my code doesn't... why close? @PSeitz @fulmicoton 
I can remove the implementation of `std::io::Read`, but what about the `read_bytes_into`? I need this for my project...

fulmicoton:
@tontinton You don't exactly need it. There are different patterns to keep these methods in your project if you need them.

> why close?

Because we don't want to maintain code that does not seem useful.

tontinton:
@fulmicoton 

To read a file slice when it is really large (to not read it whole into a `OwnedBytes`) I now need something like:

```rust
struct FileSliceReader {
    len: usize,
    slice_left: FileSlice,
    chunk: OwnedBytes,
    bytes_read: usize,
}

impl FileSliceReader {
    fn split_chunk(
        slice: FileSlice,
        left_to_read: usize,
    ) -> std::io::Result<(FileSlice, OwnedBytes)> {
        let chunk_size = left_to_read.min(CHUNK_SIZE);
        let (chunk_slice, next_slice) = slice.split(chunk_size);
        Ok((next_slice, chunk_slice.read_bytes()?))
    }

    fn new(slice: FileSlice) -> std::io::Result<Self> {
        let len = slice.len();
        let (slice_left, chunk) = Self::split_chunk(slice, len)?;
        Ok(Self {
            len,
            slice_left,
            chunk,
            bytes_read: 0,
        })
    }
}

impl Read for FileSliceReader {
    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
        while self.bytes_read < self.len {
            let n = self.chunk.read(buf)?;
            if n == 0 {
                let left = self.len - self.bytes_read;
                (self.slice_left, self.chunk) = Self::split_chunk(self.slice_left.clone(), left)?;
            } else {
                self.bytes_read += n;
                return Ok(n);
            }
        }

        Ok(0)
    }
}
```

Which means I read from the file into a chunk `OwnedBytes` and then copying it again into the `buf` argument, this is a double copy for no reason...

Can you advise me maybe I'm missing something here?

fulmicoton:
Why is the following not working for you?

```
struct FileSliceWithReadImpl(FileSlice);

impl io::Read for FileSliceWithReadImpl {
   //... the code in your PR
}
```

are there internal properties you need to access?

tontinton:
It works, just like in my example, the problem being that I must make a redundant copy because `read_bytes()` allocates me a vector, instead of recieving the already allocated buffer provided in the` read()` function of the `std::io::Read` trait.

This is only an optimization, but I feel like a project aiming for performance like `tantivy` would want something like that.

PR #2402, opened 2024-05-17
compact doc

PSeitz:
Replaces TantivyDocument with a version that is more compact, by storing all data in two vecs. (similar to tape in simd_json)

This new Document has a lower memory footprint, but has a limitation:
* The number of `Field` can't exceed `u16::MAX`

This PR also fixes an issue on the `Value` implementation on `serde_json::Value`, which did not handle date parsing on strings


src/schema/document/default_document.rs
```diff
@@ -1,93 +1,64 @@
...
-/// In this list, one field may appear more than once.
-#[derive(Clone, Debug, serde::Serialize, serde::Deserialize, Default)]
-pub struct TantivyDocument {
-    field_values: Vec<FieldValue>,
+#[repr(packed)]
+#[derive(Debug, Clone)]
+/// A field value pair in the compact tantivy document
+pub struct FieldValueAddr {
+    pub field: u16,
+    pub value: ValueAddr,
```
fulmicoton:
```suggestion
    pub value_addr: ValueAddr,
```

src/schema/document/default_document.rs
```diff
@@ -218,20 +222,457 @@ impl TantivyDocument {
...
+    }
+}
+
+/// The address in the vec
+type Addr = u32;
+
+#[derive(Clone, Copy, Default)]
+#[repr(packed)]
+/// The value type and the address to its payload in the container.
+pub struct ValueAddr {
```
fulmicoton: I suspect it does not have to be public?

src/schema/document/default_document.rs
```diff
@@ -218,20 +222,457 @@ impl TantivyDocument {
...
+}
+impl std::fmt::Debug for ValueAddr {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let val_addr = self.val_addr;
+        f.write_fmt(format_args!("{:?} at {:?}", self.type_id, val_addr))
+    }
+}
+
+impl ValueAddr {
+    pub fn new(type_id: ValueType, val_addr: u32) -> Self {
```
fulmicoton: nitpick: I suspect this object is only built in this file, and dropping the constructor would be an improvement.

src/schema/document/default_document.rs
```diff
@@ -218,20 +222,457 @@ impl TantivyDocument {
...
+            node_addresses_slice,
+        })
+    }
+}
+
+impl<'a> Iterator for CompactDocObjectIter<'a> {
+    type Item = (&'a str, CompactDocValue<'a>);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if !self.node_addresses_slice.is_empty() {
```
fulmicoton:
```suggestion
        if self.node_addresses_slice.is_empty() {
            return None;
        }
        // ...
```

fulmicoton (approved):

src/schema/document/default_document.rs
PSeitz: Yes, I'll probably add an API later for concat fields, so we can just take the ValueAddrs and add them to a Field, instead of operating on the JSON values

PR #2424, opened 2024-06-04
Fix Instances of `if` Being Spelt as `iff` in Documentation

ultrabear:
Hi, I used tantivy for a project and noticed `if` being written as `iff` in the NumericOptions docs as I browsed, I doubt these spellings were intentional and a ripgrep later revealed 16 potential locations where this spelling error had occurred.

This PR changes the spelling in all 16 locations that were found through automated means to say `if` instead of `iff`, it also changes `fieldnorm` to `fieldnorms` in 2 areas (that were on the same line as an `iff`) to make the grammar flow better.

### DIFF
--- a/src/collector/mod.rs
+++ b/src/collector/mod.rs
@@ -153,7 +153,7 @@ pub trait Collector: Sync + Send {
         segment: &SegmentReader,
     ) -> crate::Result<Self::Child>;
 
-    /// Returns true iff the collector requires to compute scores for documents.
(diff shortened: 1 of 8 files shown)


fulmicoton:
https://en.wikipedia.org/wiki/If_and_only_if

ultrabear:
oh

Noratrieb:
"true iff" could be expressed by just "whether", which is a lot simpler

PR #2427, opened 2024-06-07
fix de-escaping too much in query parser

trinity-1686a:
fix broken test in https://github.com/quickwit-oss/quickwit/pull/5093

### DIFF
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -42,16 +42,12 @@ fn field_name(inp: &str) -> IResult<&str, String> {
     )(inp)
 }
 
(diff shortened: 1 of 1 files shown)


PSeitz (approved):
There's a clippy warning

query-grammar/src/query_grammar.rs
```diff
@@ -42,6 +42,36 @@ fn field_name(inp: &str) -> IResult<&str, String> {
     )(inp)
 }
 
+fn interpret_escape(source: &str) -> String {
+    let mut res = String::with_capacity(source.len());
+    let mut in_escape = false;
+    let require_escape = |c: char| {
+        c.is_whitespace()
+            || [
+                '-', '^', '`', ':', '{', '}', '"', '\'', '[', ']', '(', ')', '\\',
```
PSeitz: would probably make sense to move this to a const variable and give it a name

PR #2433, opened 2024-06-13
add columnar format compatibiliy tests

PSeitz:
Add backwards compatibility test for columnar


columnar/src/compat_tests.rs
```diff
@@ -0,0 +1,84 @@
...
+    let mut wrt: Vec<u8> = Vec::new();
+    columnar_writer.serialize(num_docs, None, &mut wrt).unwrap();
+
+    wrt
+}
+
+#[test]
+#[ignore]
+/// Unignore to write columnar for current version to disk
+fn create_format() {
```
fulmicoton, line 32: I'm not fond of the #[ignore] trick.. What about running it all of the time, but skip the generation if the file already exists for instance?

fulmicoton (approved):

PR #2444, opened 2024-06-25
fix out of order bug

PSeitz:


fulmicoton:
we cannot fix like this on serialization, because it can yield collision on serializaiton.
We need to add unit test on serialization, and normalize upstream.

fulmicoton:
See https://github.com/quickwit-oss/tantivy/pull/2445

PR #2446, opened 2024-07-01
fix cardinality aggregation performance

PSeitz:
fix cardinality performance by fetching multiple terms at once. This
avoids decompressing the same block and keeps the buffer state between
terms.

* Fix Cardinatlity performance
* Add cardinality aggregation benchmark
* Bump rust version to 1.66
* Update aggregation docs

Performance comparison to before (AllQuery)
```
full
cardinality_agg                   Memory: 3.5 MB (-0.00%)    Avg: 21.2256ms (-97.78%)    Median: 21.0042ms (-97.82%)    [20.4717ms .. 23.6206ms]
terms_few_with_cardinality_agg    Memory: 10.6 MB            Avg: 81.9293ms (-97.37%)    Median: 81.5526ms (-97.38%)    [79.7564ms .. 88.0374ms]
dense
cardinality_agg                   Memory: 3.6 MB (-0.00%)    Avg: 25.9372ms (-97.24%)    Median: 25.7744ms (-97.25%)    [24.7241ms .. 27.8793ms]
terms_few_with_cardinality_agg    Memory: 10.6 MB            Avg: 93.9897ms (-96.91%)    Median: 92.7821ms (-96.94%)    [90.3312ms .. 117.4076ms]
sparse
cardinality_agg                   Memory: 895.4 KB (-0.00%)    Avg: 22.5113ms (-95.01%)    Median: 22.5629ms (-94.99%)    [22.1628ms .. 22.9436ms]
terms_few_with_cardinality_agg    Memory: 680.2 KB             Avg: 26.4250ms (-94.85%)    Median: 26.4135ms (-94.86%)    [26.3210ms .. 26.6774ms]
```

https://github.com/quickwit-oss/quickwit/issues/3779

### DIFF
--- a/sstable/src/dictionary.rs
+++ b/sstable/src/dictionary.rs
@@ -338,6 +338,44 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
         Ok(true)
     }
 
+    /// Returns the terms for a _sorted_ list of term ordinals.
+    ///
+    /// Returns true if and only if all terms have been found.
+    pub fn ords_to_term_cb<F: FnMut(&[u8])>(
+        &self,
+        ord: impl Iterator<Item = TermOrdinal>,
+        mut cb: F,
+    ) -> io::Result<bool> {
+        let mut bytes = Vec::new();
+        let mut current_block_addr = self.sstable_index.get_block_with_ord(0);
+        let mut current_sstable_delta_reader =
+            self.sstable_delta_reader_block(current_block_addr.clone())?;
+        let mut current_ordinal = 0;
+        for ord in ord {
+            // check if block changed for new term_ord
+            let new_block_addr = self.sstable_index.get_block_with_ord(ord);
+            if new_block_addr != current_block_addr {
+                current_block_addr = new_block_addr;
+                current_ordinal = current_block_addr.first_ordinal;
+                current_sstable_delta_reader =
+                    self.sstable_delta_reader_block(current_block_addr.clone())?;
+                bytes.clear();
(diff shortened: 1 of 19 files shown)


sstable/src/dictionary.rs
```diff
@@ -338,6 +338,44 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
...
+    ) -> io::Result<bool> {
+        let mut bytes = Vec::new();
+        let mut current_block_addr = self.sstable_index.get_block_with_ord(0);
+        let mut current_sstable_delta_reader =
+            self.sstable_delta_reader_block(current_block_addr.clone())?;
+        let mut current_ordinal = 0;
+        for ord in ord {
+            // check if block changed for new term_ord
+            let new_block_addr = self.sstable_index.get_block_with_ord(ord);
+            if new_block_addr != current_block_addr {
```
fulmicoton, line 358: So right now the facts that ord is sorted is only enforced by the comment.

We should either:

We can make sure this function works, but is slow, even for non sorted ord, as follows.

```suggestion
            if new_block_addr != current_block_addr || ord < current_ordinal {
```

Or we could add an assert a rename the function `sorted_ords_to_term_cb`

fulmicoton (approved):
see comments

sstable/src/dictionary.rs
PSeitz: I changed it to `sorted_ords_to_term_cb` + assert, since the performance hit is too big otherwise

PR #2449, opened 2024-07-05
lift clauses in LogicalAst

PSeitz:
(a OR b) OR (c OR d) can be simplified to (a OR b OR c OR d)
(a AND b) AND (c AND d) can be simplified to (a AND b AND c AND d)

This directly affects how queries are executed and the performance

Remove `SumWithCoordsCombiner`, it is effectively the same as `SumCombiner`, but with an unused field

simplify_ast

### DIFF
--- a/src/query/query_parser/logical_ast.rs
+++ b/src/query/query_parser/logical_ast.rs
@@ -39,6 +39,34 @@ impl LogicalAst {
             LogicalAst::Boost(Box::new(self), boost)
         }
     }
+
+    pub fn simplify(self) -> LogicalAst {
+        match self {
+            LogicalAst::Clause(clauses) => {
+                let mut new_clauses: Vec<(Occur, LogicalAst)> = Vec::new();
+
+                for (occur, sub_ast) in clauses {
+                    let simplified_sub_ast = sub_ast.simplify();
+
+                    // If clauses below have the same `Occur`, we can pull them up
+                    match simplified_sub_ast {
+                        LogicalAst::Clause(sub_clauses)
+                            if (occur == Occur::Should || occur == Occur::Must)
+                                && sub_clauses.iter().all(|(o, _)| *o == occur) =>
+                        {
+                            for sub_clause in sub_clauses {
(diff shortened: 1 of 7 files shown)


fulmicoton (approved):

fulmicoton (changes requested):
I don't think it works for all Occur

src/query/query_parser/logical_ast.rs
```diff
@@ -39,6 +39,33 @@ impl LogicalAst {
...
+            LogicalAst::Clause(clauses) => {
+                let mut new_clauses: Vec<(Occur, LogicalAst)> = Vec::new();
+
+                for (occur, sub_ast) in clauses {
+                    let simplified_sub_ast = sub_ast.simplify();
+
+                    // If clauses below have the same `Occur`, we can pull them up
+                    match simplified_sub_ast {
+                        LogicalAst::Clause(sub_clauses)
+                            if sub_clauses.iter().all(|(o, _)| *o == occur) =>
```
fulmicoton: What if Occur is Not?
Can we stay on the safe side and only apply this logic for occur = Must, Should, or Filter?

Also this will only yield the same score if we sum scores of child branches.
(I think this is the case today: we don't rely on dismax do we?)

PSeitz: yes good catch, I excluded it and added a test. There are more optimization we can do here to flatten clauses later on

trinity-1686a (approved):
I believe other simplifications are possible (see below), this one seems correct

- top: Should, inner: single Must => Should
- top: Must, inner: single Should => Must
- ~~top: MustNot, inner: all MustNot => all Should~~ impact scoring (or we need to wrap with a boost of 0)
- top: MustNot, inner: single Must|Should => MustNot
- top: Must, inner: all MustNot => all MustNot

fulmicoton:
@trinity-1686a Can you open a ticket and auto-assign for this?

fulmicoton (approved):

Issue #2023, opened 2023-05-03
Enable str fast field based range queries.


fulmicoton:
This is for quickwit. Which means we at least want a StrFastFieldRange{Query/Weight}

fulmicoton:
@PSeitz Do we have this already I forgot?

PSeitz:
Currently only u64 fast fields are supported.

If we can map the str ranges to term ids as a preprocessing, reusing the existing u64 range queries on fast fields should be straight forward.

Issue #1709, opened 2022-12-05
Range queries on JSON fields


xrl:
**Is your feature request related to a problem? Please describe.**

I cannot perform range queries on JSON fields. For example, the `examples/json_field.rs` has a search like so:

```
    {
        let query = query_parser.parse_query("cart.product_id:103")?;
        let count_docs = searcher.search(&*query, &Count)?;
        assert_eq!(count_docs, 1);
    }
```

But I cannot rewrite the query like this:

```
    {
        let query = query_parser.parse_query("cart.product_id < 110")?;
        let count_docs = searcher.search(&*query, &Count)?;
        assert_eq!(count_docs, 1);
    }
```

the count comes back 0.

**Describe the solution you'd like**

We should be able to search nested JSON documents with the usual `<`, `>`, etc.

adamreichold:
Isn't the syntax for range queries supported by the built-in [`QueryParser`](https://docs.rs/tantivy/latest/tantivy/query/struct.QueryParser.html) different, meaning it would be something like

```
cart.product_id:[0 TO 110}
```

(assuming that 0 is the smallest possible ID).

PSeitz:
The parser doesn't handle this currently, but this should work `cart.product_id:<110` or `cart.product_id:{* TO 110}`

Related quickwit issue: https://github.com/quickwit-oss/quickwit/issues/2431

Nickersoft:
@PSeitz Attempting the `:<` syntax of `:{{* TO 110}}` (Rust is complaining about the unescaped `{}`, thus the double-brackets) returns `Unsupported query: Range query are not supported on json field` for me.

PSeitz:
Indeed, it's disabled. 
I don't think there's a inherent reason, except some code missing to handle that. @fulmicoton?

fulmicoton:
If people want to contribute?

yollotltamayo:
Hi @fulmicoton I would like to contribute, if you can give any entry point to start working on this, thanks.

PSeitz:
@yollotltamayo 

https://github.com/quickwit-oss/tantivy/tree/main/src/query/range_query

The code is sometimes bound to a fixed field in the schema. These would need to be replaced with something that can handle JSON, e.g. "myjson.fielda"
https://github.com/quickwit-oss/tantivy/blob/main/src/query/range_query/range_query.rs#L335

A range query can run on the columnar storage and on the inverted index. I would implement it first for the inverted index, as it should be simpler.

[...]

PR #2452, opened 2024-07-12
refactor fast field query

PSeitz:
As preparation of https://github.com/quickwit-oss/tantivy/issues/2023 and https://github.com/quickwit-oss/tantivy/issues/1709

* Use Term to pass parameters
* merge u64 and ip fast field range query

Side note: I did not rename range_query_u64_fastfield, because then git can't track the changes.


fulmicoton (approved):
nice refactoring.

PR #2453, opened 2024-07-16
add support for str fast field range query

### LINKED ISSUE #2023: Enable str fast field based range queries.

PSeitz:
Add support for range queries on fast fields, by converting term bounds to
term ordinal bounds.

closes https://github.com/quickwit-oss/tantivy/issues/2023

### DIFF
--- a/src/query/range_query/range_query_u64_fastfield.rs
+++ b/src/query/range_query/range_query_u64_fastfield.rs
@@ -5,7 +5,7 @@
 use std::net::Ipv6Addr;
 use std::ops::{Bound, RangeInclusive};
 
-use columnar::{Column, MonotonicallyMappableToU128, MonotonicallyMappableToU64};
+use columnar::{Column, MonotonicallyMappableToU128, MonotonicallyMappableToU64, StrColumn};
 use common::BinarySerializable;
 
 use super::fast_field_range_doc_set::RangeDocSet;
@@ -51,16 +51,22 @@ impl Weight for FastFieldRangeWeight {
         }
         let field_name = reader.schema().get_field_name(self.field);
         let field_type = reader.schema().get_field_entry(self.field).field_type();
+
+        let term = inner_bound(&self.lower_bound)
+            .or(inner_bound(&self.upper_bound))
+            .expect("At least one bound must be set");
+        assert_eq!(
+            term.typ(),
+            field_type.value_type(),
+            "Field is of type {:?}, but got term of type {:?}",
+            field_type,
+            term.typ()
+        );
         if field_type.is_ip_addr() {
             let parse_ip_from_bytes = |term: &Term| {
-                let ip_u128_bytes: [u8; 16] =
-                    term.serialized_value_bytes().try_into().map_err(|_| {
-                        crate::TantivyError::InvalidArgument(
-                            "Expected 8 bytes for ip address".to_string(),
-                        )
-                    })?;
-                let ip_u128 = u128::from_be_bytes(ip_u128_bytes);
-                crate::Result::<Ipv6Addr>::Ok(Ipv6Addr::from_u128(ip_u128))
+                term.value().as_ip_addr().ok_or_else(|| {
+                    crate::TantivyError::InvalidArgument("Expected ip address".to_string())
+                })
             };
             let lower_bound = map_bound_res(&self.lower_bound, parse_ip_from_bytes)?;
             let upper_bound = map_bound_res(&self.upper_bound, parse_ip_from_bytes)?;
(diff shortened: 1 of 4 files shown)


src/schema/field_type.rs
```diff
@@ -201,6 +201,11 @@ impl FieldType {
         matches!(self, FieldType::IpAddr(_))
     }
 
+    /// returns true if this is an str field
+    pub fn is_term(&self) -> bool {
```
trinity-1686a: i don't think that's a good name for this function, in other contexts terms can be arbitrary byte-strings representing any types. `is_str` would be better

src/query/range_query/range_query_u64_fastfield.rs
```diff
@@ -79,33 +85,42 @@ impl Weight for FastFieldRangeWeight {
...
+            let (lower_bound, upper_bound) = if field_type.is_term() {
+                let Some(str_dict_column): Option<StrColumn> =
+                    reader.fast_fields().str(field_name)?
+                else {
+                    return Ok(Box::new(EmptyScorer));
+                };
+                let dict = str_dict_column.dictionary();
+
+                let lower_bound = map_bound(&self.lower_bound, |term| {
+                    term.serialized_value_bytes().to_vec()
```
trinity-1686a, line 97:
```suggestion
                    term.serialized_value_bytes()
```

src/query/range_query/range_query_u64_fastfield.rs
```diff
@@ -79,33 +85,42 @@ impl Weight for FastFieldRangeWeight {
...
+                });
+                let upper_bound = map_bound(&self.upper_bound, |term| {
+                    term.serialized_value_bytes().to_vec()
```
trinity-1686a, line 100:
```suggestion
                    term.serialized_value_bytes()
```

src/query/range_query/range_query_u64_fastfield.rs
```diff
@@ -202,12 +217,52 @@ pub mod tests {
...
+
+        test_query("title:[aaa TO ccc]", 1);
+        test_query("title:[aaa TO bbb]", 1);
+        test_query("title:[bbb TO bbb]", 1);
+        test_query("title:[bbb TO ddd]", 2);
+        test_query("title:[bbb TO eee]", 2);
+        test_query("title:[bb TO eee]", 2);
+        test_query("title:[ccc TO ccc]", 0);
+        test_query("title:[ccc TO ddd]", 1);
+        test_query("title:[ccc TO eee]", 1);
```
trinity-1686a, line 261: this could use tests for exclusive bounds, and half-unbounded ranges

sstable/src/dictionary.rs
```diff
@@ -56,6 +56,51 @@ impl Dictionary<VoidSSTable> {
...
+        Excluded(ref from_val) => Bound::Excluded(transform(from_val)),
+        Included(ref from_val) => Bound::Included(transform(from_val)),
+        Unbounded => Unbounded,
+    }
+}
+
+fn map_bound_res<TFrom, TTo>(
+    bound: &Bound<TFrom>,
+    transform: impl Fn(&TFrom) -> io::Result<Bound<TTo>>,
+) -> io::Result<Bound<TTo>> {
```
trinity-1686a: i honestly don't like this function. The spec is odd. It takes a Bound, but then ignores the variant, and gives no way to the transform to know if the bound was inclusive or not. The places this function is used end up with a closure that captures the `Bound<TFrom>` to properly handle Included vs Excluded.

I'm not sure how to make something that's as flexible, while being clearer.
At the very least there should be a doc-comment warning that its behaviour might be surprising.

sstable/src/dictionary.rs
```diff
@@ -524,6 +637,222 @@ mod tests {
...
+        assert_eq!(dict.term_ord_or_next(b"bbbb").unwrap(), TermOrdHit::Next(1));
+        assert_eq!(dict.term_ord_or_next(b"dd").unwrap(), TermOrdHit::Next(1));
+        assert_eq!(dict.term_ord_or_next(b"ddd").unwrap(), TermOrdHit::Exact(1));
+        assert_eq!(dict.term_ord_or_next(b"dddd").unwrap(), TermOrdHit::Next(2));
+
+        // Shouldn't this be u64::MAX?
+        assert_eq!(
+            dict.term_ord_or_next(b"zzzzzzz").unwrap(),
+            TermOrdHit::Next(2)
+        );
```
trinity-1686a: for very small sstables (only one block), we don't store an index, and the pseudo-index always reply that the answer lies in block number 0. After reading the whole block, `decode_up_to_or_next` conclude that the key is after its own end

sstable/src/dictionary.rs
```diff
@@ -524,6 +637,222 @@ mod tests {
...
+
+        assert_eq!(
+            dict.term_ord_or_next(b"zzzzzzz").unwrap(),
+            TermOrdHit::Next(u64::MAX)
+        );
+    }
+
+    #[test]
+    fn test_term_bounds_to_ord() {
```
trinity-1686a, line 723: this doesn't test the cases where the key is present and lower bound (both Included and Excluded)

src/schema/field_type.rs
PSeitz: yes, good point, renamed it

src/query/range_query/range_query_u64_fastfield.rs
PSeitz: there's a lifetime issue without the vec, because it gets lost in the generic parameter. shouldn't be an issue performance wise

sstable/src/dictionary.rs
PSeitz: I renamed it to `transform_bound_inner` and added a comment

sstable/src/dictionary.rs
PSeitz: Thanks for the clarification

src/query/range_query/range_query_u64_fastfield.rs
PSeitz: I extended the test for half-unbounded and exclusive ranges

sstable/src/dictionary.rs
trinity-1686a: can you update the comment?

trinity-1686a (approved):

sstable/src/dictionary.rs
PSeitz: yep, done

Issue #2454, opened 2024-07-17
Merge policy works unexpectedly for uncommitted documents


simunic-cz:
**Describe the bug**

_Context:_ I'm trying to drastically reduce the number of segments in an index to reduce IO latency (we're running Tantivy in Lambda with EFS attached, so network latency is an issue with 100(0)s of files). I've written a custom merge policy intended to merge segments with less then some threshold (a few thousands) of documents. In the Lambda (test/prod) we `commit` after every change (`delete` + `add` document), however during local testing, I add all _(about 22k)_ documents at once and only then call `commit` once.

What I noticed locally is that with the custom merge policy didn't really work. So I tried to use the `MergeWheneverPossible` merge policy that I found in `tests` here: https://github.com/quickwit-oss/tantivy/blob/main/src/indexer/merge_policy.rs#L48
To my surprise, the `MergeWheneverPossible` merge policy produced even more segments than the default one.

Looking at the implementation at [indexer/segment_updater.rs#L544](https://github.com/quickwit-oss/tantivy/blob/main/src/indexer/segment_updater.rs#L544), I see that committed and uncommitted segments are handled separately. I guess that makes sense. However, I wasn't trying to merge committed and uncommitted segments together - all 22k documents were uncommitted, yet it seems the `MergeWheneverPossible` merge policy didn't apply to all those uncommitted documents.

When I tried to (locally) call `commit` after every single `add_document`, it run as a charm. I mean, true, the indexing took more than an hour, but the number of emitted segments successfully dropped to just a few segments. So I guess the merge policy is correctly applied to committed documents, it's just not applied correctly to all those tens of thousands of uncommitted documents.

When I tried to log what's going on, the logs confirmed an issue:
```rs
fn compute_merge_candidates(&self, segment_metas: &[SegmentMeta]) -> Vec<MergeCandidate> {
    println!("Segment metas (# = {}): {:?}", segment_metas.len(), segment_metas);
    ...
}
```
When `commit` was called after every change, `segment_metas` was indeed a vector full of segments to be considered for merging.
However, when `commit` was called only once, after adding 22k documents, then the `segment_metas` vector never contained more then a single item. Sometimes it was empty, sometimes it contained 1 single item. Therefore no merge could really be considered.

[...]

PR #2457, opened 2024-07-24
fix potential endless loop in merge

PSeitz:
avoid single segments lists without deletes as merge candidates, as they will be moved
to a merge operation and filtered for merging in the next
consider_merge_options call. In rare cases this may end up in a endless
merge loop where only single segments where nothing is to be done are
merged.

#2454

As alternative we could say the merge policy would need to handle that.

### DIFF
--- a/src/indexer/segment_updater.rs
+++ b/src/indexer/segment_updater.rs
@@ -542,7 +542,13 @@ impl SegmentUpdater {
     }
 
     fn consider_merge_options(&self) {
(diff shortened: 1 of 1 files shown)


simunic-cz:
I can confirm that this seems to have solved both the possibly infinite loop and (even rarer case of) finishing successfully yet merge policy not taking effect due to not ever being called with `2` merge candidates.

fulmicoton (approved):

PR #2468, opened 2024-07-31
add Key::I64 and Key::U64 variants in aggregation

### LINKED ISSUE #5254: Aggregations that return field values do not respect output format
**Describe the bug**
The output format is not respected on aggregations that return field values, such as bucket aggregations, For example, u64 fields that are serialized to json will return a number in scientific notation.

**Steps to reproduce (if applicable)**
##### Field Definition
```json5
    , {
        name: '_lid'
      , type: 'u64'
      , coerce: true
      , output_format: 'string'
      , fast: true
      }
```

##### Query

```json5
{
  "size": 0,
  "aggs": {
    "dupes": {
      "terms": { "field": "_lid" }
    }
  }
}'
```

Will return
```json5
  "aggregations": {
    "dupes": {
      "buckets": [
        {
          "doc_count": 1,
          "key": 1.7679365525681357e18
        },
      ]
    }
  }
```

**Expected behavior**
If a field has a defined output format in the index field mapping, it should be used to format the field value before it is serialized

**Build Info**
```json5
{
  "build": {
    "build_date": "2024-07-19T08:56:40Z",
    "build_profile": "release",
    "build_target": "x86_64-unknown-linux-gnu",
    "cargo_pkg_version": "0.8.0",
    "commit_date": "2024-07-19T08:44:01Z",
    "commit_hash": "5638734dda122a87e269758905bf416bee42623c",
    "commit_short_hash": "5638734",
    "commit_tags": [
      "qw-airmail-20240719"
    ],
    "version": "0.8.0-nightly"
  },
  "runtime": {
    "num_cpus": 2,
    "num_threads_blocking": 1,
    "num_threads_non_blocking": 1
  }
}
```

PSeitz:
Currently all `Key` numerical values are returned as f64. This causes problems in some
cases with the precision and the way f64 is serialized.

This PR adds `Key::I64` and `Key::U64` variants and uses them in the term
aggregation.

Fixes https://github.com/quickwit-oss/quickwit/issues/5254

### DIFF
--- a/src/aggregation/agg_req_with_accessor.rs
+++ b/src/aggregation/agg_req_with_accessor.rs
@@ -186,6 +186,8 @@ impl AggregationWithAccessor {
                     .map(|missing| match missing {
                         Key::Str(_) => ColumnType::Str,
                         Key::F64(_) => ColumnType::F64,
+                        Key::I64(_) => ColumnType::I64,
+                        Key::U64(_) => ColumnType::U64,
                     })
                     .unwrap_or(ColumnType::U64);
                 let column_and_types = get_all_ff_reader_or_empty(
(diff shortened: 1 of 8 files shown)


src/aggregation/agg_req_with_accessor.rs
```diff
@@ -339,9 +341,18 @@ fn get_missing_val(
         Key::Str(_) if column_type == ColumnType::Str => Some(u64::MAX),
         // Allow fallback to number on text fields
         Key::F64(_) if column_type == ColumnType::Str => Some(u64::MAX),
+        Key::U64(_) if column_type == ColumnType::Str => Some(u64::MAX),
```
fulmicoton, line 354: I don't understand what `get_missing_val` is about. 

It is clearly doing something strange. Do you remember?
Can you comment/rename to make it make sense?

src/aggregation/metric/cardinality.rs
```diff
@@ -191,6 +191,12 @@ impl SegmentCardinalityCollector {
                         let val = f64_to_u64(*val);
                         self.cardinality.sketch.insert_any(&val);
                     }
+                    Key::U64(val) => {
```
fulmicoton, line 195: sorry for using this PR.. 

`.expect("Found placeholder term_ord but `missing` is None");`

I think we are just supposed to ignore the value.

fulmicoton (approved):

src/aggregation/agg_req_with_accessor.rs
PSeitz: yes, there is definitely a comment missing to explain it

```
/// Get the missing value as internal u64 representation
///
/// For terms we use u64::MAX as sentinel value
/// For numerical data we convert the value into the representation
/// we would get from the fast field, when we open it as u64_lenient_for_type.
///
/// That way we can use it the same way as if it would come from the fastfield.
```

PR #2476, opened 2024-08-13
make find_field_with_default return json fields without path

trinity-1686a:
this is useful for concatenate fields (they are json under the hood), and possibly json fields if we ever allow value at the root of a json field. We still forbid non-json field with a path


fmassot (approved):
Can we add a unit test for this edge case maybe?

PR #2477, opened 2024-08-16
add FastFieldRangeQuery, rename

PSeitz:
add FastFieldRangeQuery
add reexport
rename module

preparation for support on string range queries in quickwit

### DIFF
--- a/src/query/range_query/mod.rs
+++ b/src/query/range_query/mod.rs
@@ -2,10 +2,12 @@ use crate::schema::Type;
 
 mod fast_field_range_doc_set;
 mod range_query;
-mod range_query_u64_fastfield;
+mod range_query_fastfield;
+
+pub use common::bounds::BoundsRange;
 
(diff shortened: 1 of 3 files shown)


trinity-1686a (approved):
`FastFieldRangeWeight`  implements `Query`, imo that should be removed if we now have a proper Query type for that

PSeitz:
> `FastFieldRangeWeight`  implements `Query`, imo that should be removed if we now have a proper Query type for that

The RangeWeight is still a valid export, which can be used in a custom Query

trinity-1686a:
i don't mean to remove `FastFieldRangeWeight`, but to remove `impl Query for FastFieldRangeWeight`

PSeitz:
> i don't mean to remove `FastFieldRangeWeight`, but to remove `impl Query for FastFieldRangeWeight`

Ah yes that makes sense

Issue #2466, opened 2024-07-30
Improve Date precision truncation handling


PSeitz:
The way we handle precision truncation of `DateTime` when querying is opaque and error-prone. When using a date in the inverted index we want to truncate it to seconds, since they are all stored with second precision. In the columnar storage the precision is configurable.

Currently `DateTime` is stored truncated in `Term`, but this is not great since we don't know if `Term` is used in the inverted index or columnar. 

Ideally we would store `DateTime` with full precision and get a truncated version before using it in a query. E.g range queries may use the inverted index or the columnar storage.

Related: https://github.com/quickwit-oss/tantivy/pull/2456
Depends on splitting term into indexing term and term: https://github.com/quickwit-oss/tantivy/pull/2366

PR #2486, opened 2024-09-04
store DateTime as nanoseconds in doc store

### LINKED ISSUE #2380: Store DateTime as nanos in docstore
we use `BinarySerializable` to encode datetimes into the doc store. Strangely, while DateTime are precise up to ns, we only store µs there. We should store the value we got with full precision instead 
https://github.com/quickwit-oss/tantivy/blob/99a59ad37e833e1129f7e1abd25738a0adfae73a/common/src/datetime.rs#L166-L176

PSeitz:
The doc store DateTime was truncated to microseconds previously. This
removes this truncation, while still keeping backwards compatibility.

This is done by passing the doc store version to the DocumentBinarySerializer and having different deserialization logic depending on the version for dates.

bump version format to 7.
add compat test to check the date time truncation.

closes https://github.com/quickwit-oss/tantivy/issues/2380


fmassot:
@fulmicoton can you review it?

fulmicoton:
Closes #2466 I assume

src/store/footer.rs
```diff
@@ -26,15 +27,18 @@ impl BinarySerializable for DocStoreFooter {
 
     fn deserialize<R: io::Read>(reader: &mut R) -> io::Result<Self> {
         let doc_store_version = u32::deserialize(reader)?;
-        if doc_store_version != DOC_STORE_VERSION {
-            panic!("actual doc store version: {doc_store_version}, expected: {DOC_STORE_VERSION}");
+        if doc_store_version > DOC_STORE_VERSION {
```
fulmicoton, line 30: can we introduce a `DocStoreVersion` enum?

src/store/reader.rs
```diff
@@ -203,8 +205,9 @@ impl StoreReader {
     pub fn get<D: DocumentDeserialize>(&self, doc_id: DocId) -> crate::Result<D> {
         let mut doc_bytes = self.get_document_bytes(doc_id)?;
 
-        let deserializer = BinaryDocumentDeserializer::from_reader(&mut doc_bytes)
-            .map_err(crate::TantivyError::from)?;
+        let deserializer =
+            BinaryDocumentDeserializer::from_reader(&mut doc_bytes, self.doc_store_version)
```
fulmicoton, line 243: yeah it is a very clean, future proof solution.

common/src/serialize.rs
```diff
@@ -25,6 +25,63 @@ impl io::Write for Counter {
...
 
+/// The configuration for binary serialization.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct BinarySerializableConfig {
+    /// The precision with which DateTime is de/serialized
+    pub date_time_precision: DateTimePrecision,
+}
+
+/// Trait for configurable binary serialization
+pub trait ConfigurableBinarySerializable: fmt::Debug + Sized {
```
fulmicoton: The introduction of  ConfigurableBinarySerializable seems quite overkill.

I understand this is an attempt to be future proof, but let's wait for the need to resurface 2 or 3 times before going there.

Also, I don't think common is the right place for this trait. We could have it closer to the BinaryDocumentSerializer

src/schema/document/se.rs
```diff
@@ -94,56 +112,36 @@ where W: Write
...
                 }
                 ReferenceValueLeaf::Date(val) => {
-                    self.write_type_code(type_codes::DATE_CODE)?;
-                    val.serialize(self.writer)
-                }
-                ReferenceValueLeaf::Facet(val) => {
-                    self.write_type_code(type_codes::HIERARCHICAL_FACET_CODE)?;
-
-                    Cow::Borrowed(val).serialize(self.writer)
+                    self.serialize_with_type_code(type_codes::DATE_CODE, &val)
```
fulmicoton: I think we should just have some short date specific code here.

src/schema/document/de.rs
```diff
@@ -416,57 +430,94 @@ where R: Read
...
+            &doc_store_version_to_serialize_config(self.doc_store_version)?,
+        )
+        .map(u64_to_f64)
+        .map_err(DeserializeError::from)
     }
 
     fn deserialize_datetime(self) -> Result<DateTime, DeserializeError> {
         self.validate_type(ValueType::DateTime)?;
-        <DateTime as BinarySerializable>::deserialize(self.reader).map_err(DeserializeError::from)
+        <DateTime as ConfigurableBinarySerializable>::deserialize(
```
fulmicoton: I think we should just ahve some short date specific code here.

fulmicoton (commented):
Working on the DocumentBinarySerializer is a great idea. Introducing a ConfigurableBinarySerializer is overkill at this point however. Can we just push the logic on the date specific method in `de.rs`, `se.rs`.

src/store/footer.rs
PSeitz: yep, I converted it to an enum

common/src/serialize.rs
PSeitz: Yes, it's kinda overkill currently, I removed it

src/schema/document/se.rs
PSeitz: yes, I moved the version handling here

fulmicoton (approved):

PR #2499, opened 2024-09-23
feat: make SegmentMeta.with_max_doc public

triandco:
I was working with an implementation of tantivy where one segment can be shared between multiple participants. This allows us to support many distributed indexing use cases.

The current API already support this 99% of this use case. Exposing Segment.with_max_doc publicly is the last bit.


PSeitz:
Formatting on nightly seems to have changed

triandco:
@PSeitz,  Can I do anything here? like matching the style in the format check output?

PSeitz:
`cargo +nightly fmt --all`

PSeitz (approved):

PR #2512, opened 2024-10-09
fix `OwnedBytes` debug panic

b41sh:
If the data length is 9, `OwnedBytes` debug may panic when truncate the slice.


PSeitz (approved):

PSeitz:
Thanks!

PR #2513, opened 2024-10-10
feat(QueryParser): add support for patterns like "field:*".

LebranceBW:
# ChangeLog
* Add entries in `LogicalLiteral` and `LogicalAst` to support `ExistsQuery` pattern.

### DIFF
--- a/src/query/query_parser/query_parser.rs
+++ b/src/query/query_parser/query_parser.rs
@@ -849,33 +849,36 @@ impl QueryParser {
                 (Some(logical_ast), errors)
             }
             UserInputLeaf::Exists { field: full_path } => {
-                let mut errors = Vec::<QueryParserError>::new();
                 let (field, json_path) = try_tuple!(self
                     .split_full_path(&full_path)
                     .ok_or_else(|| QueryParserError::FieldDoesNotExist(full_path.clone())));
                 let field_entry = self.schema.get_field_entry(field);
                 let field_type = field_entry.field_type();
-                // `ExistsQuery` currently supports only FAST fields.
                 if !field_type.is_fast() {
                     // FIXME: FieldNotIndexed error is not accurate.
-                    errors.push(QueryParserError::FieldNotIndexed(
-                        field_entry.name().to_string(),
-                    ));
-                    return (None, errors);
+                    return (
+                        None,
+                        vec![QueryParserError::FieldNotIndexed(
+                            field_entry.name().to_string(),
+                        )],
+                    );
                 }
-                // Though JSON fields cannot be FAST fields, `split_full_path`
-                // may still return a FAST field with a non-empty JSON path.
-                if !json_path.is_empty() {
-                    errors.push(QueryParserError::UnsupportedQuery(format!(
-                        "Json path is not supported for field {:?}",
-                        field_entry.name()
-                    )));
-                    return (None, errors);
+                if json_path.is_empty() == field_type.is_json() {
+                    let error = if field_type.is_json() {
+                        QueryParserError::UnsupportedQuery(format!(
+                            "JSON path must be provided when querying the existence of a JSON \
+                             field: {}",
+                            field_entry.name()
+                        ))
+                    } else {
+                        QueryParserError::FieldDoesNotExist(full_path.clone())
+                    };
+                    return (None, vec![error]);
                 }
                 let logical_ast = LogicalAst::Leaf(Box::new(LogicalLiteral::Exist {
-                    field_name: field_entry.name().to_string(),
+                    field_name: full_path,
                 }));
-                (Some(logical_ast), errors)
+                (Some(logical_ast), Vec::new())
             }
         }
     }
@@ -1090,6 +1093,7 @@ mod test {
         schema_builder.add_bool_field("bool", INDEXED);
         schema_builder.add_bool_field("notindexed_bool", STORED);
         schema_builder.add_u64_field("u64_ff", FAST);
+        schema_builder.add_json_field("json_ff", TEXT | FAST);
         schema_builder.build()
     }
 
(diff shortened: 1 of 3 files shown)


src/query/query_parser/query_parser.rs
```diff
@@ -848,12 +848,32 @@ impl QueryParser {
...
+                let field_entry = self.schema.get_field_entry(field);
+                let field_type = field_entry.field_type();
+                if !field_type.is_fast() {
+                    // FIXME: FieldNotIndexed error is not accurate.
+                    errors.push(QueryParserError::FieldNotIndexed(
+                        field_entry.name().to_string(),
+                    ));
+                    return (None, errors);
+                }
+                if !json_path.is_empty() && field_type.is_json() {
```
trinity-1686a: i think this condition is wrong. it generates an error when the json path contains something and the field is json. It should error when there is a path, but the field isn't json ~~(and maybe when there is no path and we're on a json field, i don't remember off hand if tantivy allows things at the root of a json field. I'll edit this comment after verifying)~~
edit: we don't support putting values at the root of a json field currently, so either it's a json field and there must be a path, or it isn't and there mustn't (though the error messages should likely differ)

LebranceBW: you are right. I'll fix it later.

LebranceBW: `ExistsQuery` does not support JSON fields, so checking them is redundant.  Just make sure fat field field + JSON path combination is banned.

Comment closed, if there still are some issue. feel free to reopen the it.

trinity-1686a: I'm not sure what leads you to believe json fields are not supported in `ExistsQuery`, i'm pretty sure we use them inside Quickwit, and it seems to work when providing the full path and not just the field to the `ExistsQuery` (the changes in `examples/json_fields.rs` are just a poor (wo)man's test, not something that should be committed)

```diff
diff --git a/examples/json_field.rs b/examples/json_field.rs
index 334ea10e4..a9e1286a4 100644
--- a/examples/json_field.rs
+++ b/examples/json_field.rs
@@ -14,7 +14,7 @@ fn main() -> tantivy::Result<()> {
     let mut schema_builder = Schema::builder();
     schema_builder.add_date_field("timestamp", FAST | STORED);
     let event_type = schema_builder.add_text_field("event_type", STRING | STORED);
-    let attributes = schema_builder.add_json_field("attributes", STORED | TEXT);
+    let attributes = schema_builder.add_json_field("attributes", STORED | TEXT | FAST);
     let schema = schema_builder.build();
 
     // # Indexing documents
@@ -99,7 +99,7 @@ fn main() -> tantivy::Result<()> {
     }
     {
         // Default json fields can still be accessed by full path
-        let query = query_parser.parse_query("attributes.event_type:holiday-sale")?;
+        let query = query_parser.parse_query("attributes.event_type:*")?;
         let hits = searcher.search(&*query, &TopDocs::with_limit(2))?;
         assert_eq!(hits.len(), 1);
     }
diff --git a/src/query/query_parser/query_parser.rs b/src/query/query_parser/query_parser.rs
index a7433df97..36248a164 100644
--- a/src/query/query_parser/query_parser.rs
+++ b/src/query/query_parser/query_parser.rs
@@ -862,7 +862,7 @@ impl QueryParser {
                     ));
                     return (None, errors);
                 }
-                if !json_path.is_empty() && field_type.is_json() {
+                if json_path.is_empty() == field_type.is_json() {
                     errors.push(QueryParserError::UnsupportedQuery(format!(
                         "Json path is not supported for field {:?}",
                         field_entry.name()
@@ -870,7 +870,7 @@ impl QueryParser {
                     return (None, errors);
                 }
                 let logical_ast = LogicalAst::Leaf(Box::new(LogicalLiteral::Exist {
-                    field_name: field_entry.name().to_string(),
+                    field_name: full_path,
                 }));
                 (Some(logical_ast), errors)
             }
```

LebranceBW: > I'm not sure what leads you to believe json fields are not supported in `ExistsQuery`, i'm pretty sure we use them inside Quickwit, and it seems to work when providing the full path and not just the field to the `ExistsQuery` (the changes in `examples/json_fields.rs` are just a poor (wo)man's test, not something that should be committed)
> 
> ```diff
> diff --git a/examples/json_field.rs b/examples/json_field.rs
> index 334ea10e4..a9e1286a4 100644
> --- a/examples/json_field.rs
> +++ b/examples/json_field.rs
> @@ -14,7 +14,7 @@ fn main() -> tantivy::Result<()> {
>      let mut schema_builder = Schema::builder();
>      schema_builder.add_date_field("timestamp", FAST | STORED);
>      let event_type = schema_builder.add_text_field("event_type", STRING | STORED);
> -    let attributes = schema_builder.add_json_field("attributes", STORED | TEXT);
> +    let attributes = schema_builder.add_json_field("attributes", STORED | TEXT | FAST);
>      let schema = schema_builder.build();
>  
>      // # Indexing documents
> @@ -99,7 +99,7 @@ fn main() -> tantivy::Result<()> {
>      }
>      {
>          // Default json fields can still be accessed by full path
> -        let query = query_parser.parse_query("attributes.event_type:holiday-sale")?;
> +        let query = query_parser.parse_query("attributes.event_type:*")?;
>          let hits = searcher.search(&*query, &TopDocs::with_limit(2))?;
>          assert_eq!(hits.len(), 1);
>      }
> diff --git a/src/query/query_parser/query_parser.rs b/src/query/query_parser/query_parser.rs
> index a7433df97..36248a164 100644
> --- a/src/query/query_parser/query_parser.rs
> +++ b/src/query/query_parser/query_parser.rs
> @@ -862,7 +862,7 @@ impl QueryParser {
>                      ));
>                      return (None, errors);
>                  }
> -                if !json_path.is_empty() && field_type.is_json() {
> +                if json_path.is_empty() == field_type.is_json() {
>                      errors.push(QueryParserError::UnsupportedQuery(format!(
>                          "Json path is not supported for field {:?}",
>                          field_entry.name()
> @@ -870,7 +870,7 @@ impl QueryParser {
>                      return (None, errors);
>                  }
>                  let logical_ast = LogicalAst::Leaf(Box::new(LogicalLiteral::Exist {
> -                    field_name: field_entry.name().to_string(),
> +                    field_name: full_path,
>                  }));
>                  (Some(logical_ast), errors)
>              }
> ```

On my. I'm sorry that I reached this arbitrary conclusion without actually verifying it through coding. In fact, this idea came from here.
[src/fastfield/mod.rs](https://github.com/quickwit-oss/tantivy/blob/c71ea7b2effe6494c1a2b9e234db71f89255c06a/src/fastfield/mod.rs#L15) which says basic types are supported.  It appears that works plain fields within JSON leaves as well.
Thank you for correcting my mistaken understanding; I will rewrite the code again and add the corresponding unit tests.

PR #2516, opened 2024-10-10
add RegexPhraseQuery

PSeitz:
RegexPhraseQuery supports phrase queries with regex. It supports regex
and wildcards. E.g. a query with wildcards:
"b* b* wolf" matches "big bad wolf"
Slop is supported as well:
"b* wolf"~2 matches "big bad wolf"

Regex queries may match a lot of terms where we still need to
keep track which term hit to load the positions.
The phrase query algorithm groups terms by their frequency
together in the union to prefilter groups early.

This PR comes with some new datastructures:

SimpleUnion - A union docset for a list of docsets. It doesn't do any
caching and is therefore well suited for datasets with lots of skipping.
(phrase search, but intersections in general)

LoadedPostings - Like SegmentPostings, but all docs and positions are loaded in
memory. SegmentPostings uses 1840 bytes per instance with its caches,
which is equivalent to 460 docids.
LoadedPostings is used for terms which have less than 100 docs.
LoadedPostings is only used to reduce memory consumption.

BitSetPostingUnion - Creates a `Posting` that uses the bitset for docid
hits and the docsets for positions. The BitSet is the precalculated
union of the docsets
In the RegexPhraseQuery there is a size limit of 512 docsets per PreAggregatedUnion,
before creating a new one.

Renamed `Union` to `BufferedUnionScorer` to better differentiate from other unions.
Added proptests to test different union types.

## Comparison With Lucene
https://github.com/quickwit-oss/search-benchmark-game/tree/regex_phrase_query

Below is a benchmark from the search benchmark game on the wikipedia dataset.
Lucene does not seem to support Regex Phrase Queries with Slop, so it's not part of the benchmark.

Interestingly the slower "grad* ma*ent admission test" is due to the regex evaluation on the FST. 

regex_perf

### DIFF
--- a/src/core/tests.rs
+++ b/src/core/tests.rs
@@ -421,7 +421,7 @@ fn test_non_text_json_term_freq() {
     let mut term = Term::from_field_json_path(field, "tenant_id", false);
     term.append_type_and_fast_value(75i64);
 
-    let mut postings = inv_idx
+    let postings = inv_idx
         .read_postings(&term, IndexRecordOption::WithFreqsAndPositions)
         .unwrap()
         .unwrap();
--- a/src/core/tests.rs
+++ b/src/core/tests.rs
@@ -421,7 +421,7 @@ fn test_non_text_json_term_freq() {
     let mut term = Term::from_field_json_path(field, "tenant_id", false);
     term.append_type_and_fast_value(75i64);
 
-    let postings = inv_idx
+    let mut postings = inv_idx
         .read_postings(&term, IndexRecordOption::WithFreqsAndPositions)
         .unwrap()
         .unwrap();
--- a/src/index/inverted_index_reader.rs
+++ b/src/index/inverted_index_reader.rs
@@ -66,6 +66,11 @@ impl InvertedIndexReader {
         self.termdict.get(term.serialized_value_bytes())
     }
 
+    /// Returns the term info associated with the str term.
+    pub fn get_term_info_str(&self, term: &str) -> io::Result<Option<TermInfo>> {
+        self.termdict.get(term.as_bytes())
+    }
+
     /// Return the term dictionary datastructure.
     pub fn terms(&self) -> &TermDictionary {
         &self.termdict
--- a/src/index/inverted_index_reader.rs
+++ b/src/index/inverted_index_reader.rs
@@ -66,11 +66,6 @@ impl InvertedIndexReader {
         self.termdict.get(term.serialized_value_bytes())
     }
 
-    /// Returns the term info associated with the str term.
-    pub fn get_term_info_str(&self, term: &str) -> io::Result<Option<TermInfo>> {
-        self.termdict.get(term.as_bytes())
-    }
-
     /// Return the term dictionary datastructure.
     pub fn terms(&self) -> &TermDictionary {
         &self.termdict
--- a/src/postings/loaded_postings.rs
+++ b/src/postings/loaded_postings.rs
@@ -8,7 +8,7 @@ use crate::DocId;
 ///
 /// It exists mainly to reduce memory usage.
 /// `SegmentPostings` uses 1840 bytes per instance due to its caches.
-/// It you need to keep many terms around with few docs, it's cheaper to load all the
+/// If you need to keep many terms around with few docs, it's cheaper to load all the
 /// postings in memory.
 ///
 /// This is relevant for `RegexPhraseQuery`, which may have a lot of
@@ -50,15 +50,16 @@ impl LoadedPostings {
 impl From<(Vec<DocId>, Vec<Vec<u32>>)> for LoadedPostings {
     fn from(doc_ids_and_positions: (Vec<DocId>, Vec<Vec<u32>>)) -> LoadedPostings {
         let mut position_offsets = Vec::new();
-        let mut positions = Vec::new();
-        for pos in &doc_ids_and_positions.1 {
-            position_offsets.push(positions.len() as u32);
-            positions.extend_from_slice(pos);
+        let mut all_positions = Vec::new();
+        let (doc_ids, docid_positions) = doc_ids_and_positions;
+        for positions in docid_positions {
+            position_offsets.push(all_positions.len() as u32);
+            all_positions.extend_from_slice(&positions);
         }
-        position_offsets.push(positions.len() as u32);
+        position_offsets.push(all_positions.len() as u32);
         LoadedPostings {
-            doc_ids: doc_ids_and_positions.0.into_boxed_slice(),
-            positions: positions.into_boxed_slice(),
+            doc_ids: doc_ids.into_boxed_slice(),
+            positions: all_positions.into_boxed_slice(),
             position_offsets: position_offsets.into_boxed_slice(),
             cursor: 0,
         }
@@ -76,7 +77,7 @@ impl DocSet for LoadedPostings {
     }
 
     fn doc(&self) -> DocId {
-        if self.cursor == self.doc_ids.len() {
+        if self.cursor >= self.doc_ids.len() {
             return TERMINATED;
         }
         self.doc_ids[self.cursor]
@@ -89,8 +90,8 @@ impl DocSet for LoadedPostings {
 impl Postings for LoadedPostings {
     fn term_freq(&self) -> u32 {
         let start = self.position_offsets[self.cursor] as usize;
-        let stop = self.position_offsets[self.cursor + 1] as usize;
-        (stop - start) as u32
+        let end = self.position_offsets[self.cursor + 1] as usize;
+        (end - start) as u32
     }
 
     fn append_positions_with_offset(&mut self, offset: u32, output: &mut Vec<u32>) {
--- a/src/postings/loaded_postings.rs
+++ b/src/postings/loaded_postings.rs
@@ -87,7 +87,7 @@ impl DocSet for LoadedPostings {
     }
 }
 impl Postings for LoadedPostings {
-    fn term_freq(&mut self) -> u32 {
+    fn term_freq(&self) -> u32 {
         let start = self.position_offsets[self.cursor] as usize;
         let stop = self.position_offsets[self.cursor + 1] as usize;
         (stop - start) as u32
--- a/src/postings/loaded_postings.rs
+++ b/src/postings/loaded_postings.rs
@@ -46,17 +46,18 @@ impl LoadedPostings {
     }
 }
 
+#[cfg(test)]
 impl From<(Vec<DocId>, Vec<Vec<u32>>)> for LoadedPostings {
-    fn from(doc_ids: (Vec<DocId>, Vec<Vec<u32>>)) -> LoadedPostings {
+    fn from(doc_ids_and_positions: (Vec<DocId>, Vec<Vec<u32>>)) -> LoadedPostings {
         let mut position_offsets = Vec::new();
         let mut positions = Vec::new();
-        for pos in &doc_ids.1 {
+        for pos in &doc_ids_and_positions.1 {
             position_offsets.push(positions.len() as u32);
             positions.extend_from_slice(pos);
         }
         position_offsets.push(positions.len() as u32);
         LoadedPostings {
-            doc_ids: doc_ids.0,
+            doc_ids: doc_ids_and_positions.0,
             positions,
             position_offsets,
             cursor: 0,
@@ -94,8 +95,8 @@ impl Postings for LoadedPostings {
 
     fn append_positions_with_offset(&mut self, offset: u32, output: &mut Vec<u32>) {
         let start = self.position_offsets[self.cursor] as usize;
-        let stop = self.position_offsets[self.cursor + 1] as usize;
-        for pos in &self.positions[start..stop] {
+        let end = self.position_offsets[self.cursor + 1] as usize;
+        for pos in &self.positions[start..end] {
             output.push(*pos + offset);
         }
     }
--- a/src/postings/loaded_postings.rs
+++ b/src/postings/loaded_postings.rs
@@ -1,0 +1,153 @@
+use crate::docset::{DocSet, TERMINATED};
+use crate::postings::{Postings, SegmentPostings};
+use crate::DocId;
+
+/// `LoadedPostings` is a `DocSet` and `Postings` implementation.
+/// It is used to represent the postings of a term in memory.
+/// It is suitable if there are few documents for a term.
+///
+/// It exists mainly to reduce memory usage.
+/// `SegmentPostings` uses 1840 bytes per instance due to its caches.
+/// It you need to keep many terms around with few docs, it's cheaper to load all the
+/// postings in memory.
+///
+/// This is relevant for `RegexPhraseQuery`, which may have a lot of
+/// terms.
+/// E.g. 100_000 terms would need 184MB due to SegmentPostings.
+pub struct LoadedPostings {
+    doc_ids: Vec<DocId>,
+    position_offsets: Vec<u32>,
+    positions: Vec<u32>,
+    cursor: usize,
+}
+
+impl LoadedPostings {
+    /// Creates a new `LoadedPostings` from a `SegmentPostings`.
+    ///
+    /// It will also preload positions, if positions are available in the SegmentPostings.
+    pub fn load(segment_postings: &mut SegmentPostings) -> LoadedPostings {
+        let num_docs = segment_postings.doc_freq() as usize;
+        let mut doc_ids = Vec::with_capacity(num_docs);
+        let mut positions = Vec::with_capacity(num_docs);
+        let mut position_offsets = Vec::with_capacity(num_docs);
+        while segment_postings.doc() != TERMINATED {
+            position_offsets.push(positions.len() as u32);
+            doc_ids.push(segment_postings.doc());
+            segment_postings.append_positions_with_offset(0, &mut positions);
+            segment_postings.advance();
+        }
+        position_offsets.push(positions.len() as u32);
+        LoadedPostings {
+            doc_ids,
+            positions,
+            position_offsets,
+            cursor: 0,
+        }
+    }
+}
+
+impl From<(Vec<DocId>, Vec<Vec<u32>>)> for LoadedPostings {
+    fn from(doc_ids: (Vec<DocId>, Vec<Vec<u32>>)) -> LoadedPostings {
+        let mut position_offsets = Vec::new();
+        let mut positions = Vec::new();
+        for pos in &doc_ids.1 {
+            position_offsets.push(positions.len() as u32);
+            positions.extend_from_slice(pos);
+        }
+        position_offsets.push(positions.len() as u32);
+        LoadedPostings {
+            doc_ids: doc_ids.0,
+            positions,
+            position_offsets,
+            cursor: 0,
+        }
+    }
+}
+
+impl DocSet for LoadedPostings {
+    fn advance(&mut self) -> DocId {
+        self.cursor += 1;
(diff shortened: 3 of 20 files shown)


src/query/union/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+    fn simple_union_from_docs_list(docs_list: &[Vec<DocId>]) -> Box<dyn DocSet> {
+        Box::new(SimpleUnion::build(
+            vec_doc_set_from_docs_list(docs_list).collect::<Vec<VecDocSet>>(),
+        ))
+    }
+    fn bitset_from_docs_list(docs_list: &[Vec<DocId>]) -> BitSetDocSet {
+        let max_doc = docs_list
+            .iter()
+            .flat_map(|docs| docs.iter())
+            .cloned()
```
fulmicoton: are we sure the compiler does a good job on this?

```suggestion
            .flat_map(|docs| docs.iter().copied())
```

also are the docs_list sorted? If they are, why are we computing the max this way?

```
doc_lists
   .iter()
   .flat_map(|docs| docs.last().copied())
   .max()
   .unwrap_or(0)
```

src/query/union/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+            .max()
+            .unwrap_or(0);
+        let mut doc_bitset = BitSet::with_max_value(max_doc + 1);
```
fulmicoton, line 55:
```suggestion
        let mut doc_bitset = BitSet::with_max_value(max_doc + 1);
```

why `max_doc  + 1`

src/query/union/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    use super::{SimpleUnion, *};
+    use crate::docset::{DocSet, TERMINATED};
+    use crate::postings::tests::test_skip_against_unoptimized;
+    use crate::query::score_combiner::DoNothingCombiner;
+    use crate::query::union::bitset_union::BitSetPostingUnion;
+    use crate::query::{BitSetDocSet, ConstScorer, VecDocSet};
+    use crate::{tests, DocId};
+
+    pub fn vec_doc_set_from_docs_list(
```
fulmicoton:
```suggestion
   fn vec_doc_set_from_docs_list(
```

src/query/union/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+        docs_list: &[Vec<DocId>],
+    ) -> impl Iterator<Item = VecDocSet> + '_ {
+        docs_list.iter().cloned().map(VecDocSet::from)
+    }
+    pub fn union_from_docs_list(docs_list: &[Vec<DocId>]) -> Box<dyn DocSet> {
```
fulmicoton:
```suggestion
   fn union_from_docs_list(docs_list: &[Vec<DocId>]) -> Box<dyn DocSet> {
```

src/postings/postings.rs
```diff
@@ -12,16 +12,33 @@ use crate::docset::DocSet;
 /// for merging segments or for testing.
 pub trait Postings: DocSet + 'static {
     /// The number of times the term appears in the document.
-    fn term_freq(&self) -> u32;
+    fn term_freq(&mut self) -> u32;
```
fulmicoton: What requires this?

src/index/inverted_index_reader.rs
```diff
@@ -66,6 +66,11 @@ impl InvertedIndexReader {
         self.termdict.get(term.serialized_value_bytes())
     }
 
+    /// Returns the term info associated with the str term.
+    pub fn get_term_info_str(&self, term: &str) -> io::Result<Option<TermInfo>> {
+        self.termdict.get(term.as_bytes())
+    }
```
fulmicoton:
```suggestion
```

I don't think it helps.

src/postings/loaded_postings.rs
```diff
@@ -0,0 +1,153 @@
...
+        LoadedPostings {
+            doc_ids,
+            positions,
+            position_offsets,
+            cursor: 0,
+        }
+    }
+}
+
+impl From<(Vec<DocId>, Vec<Vec<u32>>)> for LoadedPostings {
```
fulmicoton, line 50:
```suggestion
#[cfg(test)]
impl From<(Vec<DocId>, Vec<Vec<u32>>)> for LoadedPostings {
```

src/postings/loaded_postings.rs
```diff
@@ -0,0 +1,153 @@
...
+    fn from(doc_ids: (Vec<DocId>, Vec<Vec<u32>>)) -> LoadedPostings {
```
fulmicoton:
```suggestion
    fn from(doc_ids_and_positions: (Vec<DocId>, Vec<Vec<u32>>)) -> LoadedPostings {
```

src/postings/loaded_postings.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    fn size_hint(&self) -> u32 {
+        self.doc_ids.len() as u32
+    }
+}
+impl Postings for LoadedPostings {
+    fn term_freq(&mut self) -> u32 {
+        let start = self.position_offsets[self.cursor] as usize;
+        let stop = self.position_offsets[self.cursor + 1] as usize;
```
fulmicoton:
```suggestion
        let end = self.position_offsets[self.cursor + 1] as usize;
```

rust uses start..end in its naming.

src/query/phrase_query/mod.rs
```diff
@@ -19,15 +21,15 @@ pub mod tests {
     use crate::schema::{Schema, Term, TEXT};
     use crate::{assert_nearly_equals, DocAddress, DocId, IndexWriter, TERMINATED};
 
-    pub fn create_index(texts: &[&'static str]) -> crate::Result<Index> {
+    pub fn create_index<S: AsRef<str>>(texts: &[S]) -> crate::Result<Index> {
```
fulmicoton, line 24: do we need this?

src/query/phrase_query/phrase_weight.rs
```diff
@@ -50,27 +50,14 @@ impl PhraseWeight {
             .map(|similarity_weight| similarity_weight.boost_by(boost));
         let fieldnorm_reader = self.fieldnorm_reader(reader)?;
         let mut term_postings_list = Vec::new();
-        if reader.has_deletes() {
```
fulmicoton, line 53: I think the point of this was to avoid paying for the position check if we have deletes.

src/query/union/mod.rs
PSeitz: Yes, they are sorted. It's test code, so I think the simplest version should be fine

src/query/union/mod.rs
PSeitz: BitSet excludes max_doc. We may want to change that, it could be a source of bugs

src/postings/postings.rs
PSeitz: In the BitSetPostingUnion, we need to `seek` in the docset list, since we arrive at a hit from the bitset, the docsets are then scanned for the positions.

```rust
impl<TDocSet: Postings> Postings for BitSetPostingUnion<TDocSet> {
    fn term_freq(&mut self) -> u32 {
        let curr_doc = self.bitset.doc();
        let mut term_freq = 0;
        for docset in &mut self.docsets {
            if docset.doc() < curr_doc {
                docset.seek(curr_doc);
            }
            if docset.doc() == curr_doc {
                term_freq += docset.term_freq();
            }
        }
        term_freq
    }
```

src/index/inverted_index_reader.rs
PSeitz: yes, that's a residue

src/query/phrase_query/mod.rs
PSeitz: yes for the proptest

src/query/phrase_query/phrase_weight.rs
PSeitz: I saw that `read_postings` and `read_postings_no_deletes` is the same. The bitset is passed as parameter, so this seems to be an outdated API. I'll remove it in a follow-up.


```rust
    pub fn read_postings(
        &self,
        term: &Term,
        option: IndexRecordOption,
    ) -> io::Result<Option<SegmentPostings>> {
        self.get_term_info(term)?
            .map(move |term_info| self.read_postings_from_terminfo(&term_info, option))
            .transpose()
    }

    pub(crate) fn read_postings_no_deletes(
        &self,
        term: &Term,
        option: IndexRecordOption,
    ) -> io::Result<Option<SegmentPostings>> {
        self.get_term_info(term)?
            .map(|term_info| self.read_postings_from_terminfo(&term_info, option))
            .transpose()
    }
```

src/postings/postings.rs
fulmicoton: And you don't want to make that update on `.advance()` calls, because this computation is somewhat expensive, and you want it to be lazy (e.g. `term_query and regexquery`), or is there a different reason?

PSeitz: Yes, advancing all the docsets is quite expensive, so it's only done once we have a hit. In `advance` we only forward the bitset, which is an aggregation of the doc sets in the union.

src/core/tests.rs
```diff
@@ -421,7 +421,7 @@ fn test_non_text_json_term_freq() {
     let mut term = Term::from_field_json_path(field, "tenant_id", false);
     term.append_type_and_fast_value(75i64);
 
-    let postings = inv_idx
+    let mut postings = inv_idx
```
fulmicoton:
```suggestion
    let postings = inv_idx
```

src/postings/loaded_postings.rs
```diff
@@ -0,0 +1,154 @@
...
+use crate::postings::{Postings, SegmentPostings};
+use crate::DocId;
+
+/// `LoadedPostings` is a `DocSet` and `Postings` implementation.
+/// It is used to represent the postings of a term in memory.
+/// It is suitable if there are few documents for a term.
+///
+/// It exists mainly to reduce memory usage.
+/// `SegmentPostings` uses 1840 bytes per instance due to its caches.
+/// It you need to keep many terms around with few docs, it's cheaper to load all the
```
fulmicoton: this comment was super helpful by the way

src/postings/loaded_postings.rs
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+impl From<(Vec<DocId>, Vec<Vec<u32>>)> for LoadedPostings {
+    fn from(doc_ids_and_positions: (Vec<DocId>, Vec<Vec<u32>>)) -> LoadedPostings {
+        let mut position_offsets = Vec::new();
+        let mut positions = Vec::new();
+        for pos in &doc_ids_and_positions.1 {
```
fulmicoton: nitpick: destructuring at the beginning of the function makes the code a tad more readable.
```suggestion
     let (doc_ids, positions) = doc_ids_and_positions;
```

src/postings/loaded_postings.rs
```diff
@@ -0,0 +1,154 @@
...
+        self.cursor += 1;
+        if self.cursor >= self.doc_ids.len() {
+            self.cursor = self.doc_ids.len();
+            return TERMINATED;
+        }
+        self.doc()
+    }
+
+    fn doc(&self) -> DocId {
+        if self.cursor == self.doc_ids.len() {
```
fulmicoton: it makes one want to add an extra SENTINEL value at the end of the doc_ids array, to remove the if statement

src/postings/loaded_postings.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    fn size_hint(&self) -> u32 {
+        self.doc_ids.len() as u32
+    }
+}
+impl Postings for LoadedPostings {
+    fn term_freq(&self) -> u32 {
+        let start = self.position_offsets[self.cursor] as usize;
+        let stop = self.position_offsets[self.cursor + 1] as usize;
+        (stop - start) as u32
```
fulmicoton:
```suggestion
        let end = self.position_offsets[self.cursor + 1] as usize;
        (end - start) as u32
```

src/postings/segment_postings.rs
```diff
@@ -249,15 +250,13 @@ impl Postings for SegmentPostings {
                     .iter()
                     .cloned()
                     .sum::<u32>() as u64);
-            output.resize(term_freq as usize, 0u32);
-            position_reader.read(read_offset, &mut output[..]);
+            output.resize(prev_len + term_freq as usize, 0u32);
```
fulmicoton, line 254: I had totally forbidden about this zeroing. It might be non-benign perf wise in phrase queries.

src/query/phrase_query/phrase_weight.rs
fulmicoton: can we open a ticket to clean this up?

src/query/phrase_query/regex_phrase_weight.rs
```diff
@@ -0,0 +1,487 @@
...
+/// See RegexPhraseWeight::get_union_from_term_infos for some design decisions.
+pub struct RegexPhraseWeight {
+    field: Field,
+    phrase_terms: Vec<(usize, String)>,
+    similarity_weight_opt: Option<Bm25Weight>,
+    slop: u32,
+    max_expansions: u32,
+    /// wildcard_mode is true if the query is interpeted as wildcard query instead of regex.
+    /// e.g. wol*
+    wildcard_mode: bool,
```
fulmicoton: nitpick: I think wildcard mode is an anti feature.

src/query/automaton_weight.rs
```diff
@@ -64,6 +65,18 @@ where
 
         term_stream_builder.into_stream()
     }
+
+    /// Returns the term infos that match the automaton
+    pub fn get_term_infos(&self, reader: &SegmentReader) -> crate::Result<Vec<TermInfo>> {
```
fulmicoton:
```suggestion
    pub fn get_match_term_infos(&self, reader: &SegmentReader) -> crate::Result<Vec<TermInfo>> {
```

src/query/term_query/term_scorer.rs
```diff
@@ -82,7 +82,7 @@ impl TermScorer {
         self.fieldnorm_reader.fieldnorm_id(self.doc())
     }
 
-    pub fn explain(&self) -> Explanation {
+    pub fn explain(&mut self) -> Explanation {
```
fulmicoton:
```suggestion
    pub fn explain(&self) -> Explanation {
```

src/query/union/bitset_union.rs
```diff
@@ -0,0 +1,87 @@
...
+use crate::query::BitSetDocSet;
+use crate::DocId;
+
+/// Creates a `Posting` that uses the bitset for hits and the docsets for PostingLists.
+///
+/// It is used for the regex phrase query, where we need the union of a large amount of
+/// terms, but need to keep the docsets for the postings.
+pub struct BitSetPostingUnion<TDocSet> {
+    /// The docsets are required to load positions
+    docsets: Vec<RefCell<TDocSet>>,
```
fulmicoton: Why not 

```suggestion
    docsets: RefCell<Vec<TDocSet>>,
```
?

src/query/union/simple_union.rs
```diff
@@ -0,0 +1,113 @@
+use crate::docset::{DocSet, TERMINATED};
+use crate::postings::Postings;
+use crate::DocId;
+
+/// A `SimpleUnion` is a `DocSet` that is the union of multiple `DocSet`.
+/// Unlike `BufferedUnion`, it doesn't do any horizon precomputation.
+///
+/// For that reason SimpleUnion is a good choice for queries that skip a lot.
+pub struct SimpleUnion<TDocSet> {
```
fulmicoton, line 9: BTW PISA does that for union and outperforms tantivy and lucene

src/query/union/simple_union.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    fn update_current(&mut self) {
+        let mut next_doc = TERMINATED;
+
+        for docset in &mut self.docsets {
+            next_doc = next_doc.min(docset.doc());
+        }
+        self.doc = next_doc;
+    }
```
fulmicoton: I don't think the name is good, and the `&mut docsets` is useful,

```suggestion
    fn initialize_first_doc_id(&mut self) -> DocId {
        let mut next_doc = TERMINATED;

        for docset in &self.docsets {
            next_doc = next_doc.min(docset.doc());
        }
        self.doc = next_doc;
    }
```

src/query/union/simple_union.rs
```diff
@@ -0,0 +1,113 @@
...
+            .iter()
+            .map(|docset| docset.size_hint())
+            .max()
+            .unwrap_or(0u32)
+    }
+
+    fn count_including_deleted(&mut self) -> u32 {
+        let mut count = 0;
+        if self.doc != TERMINATED {
+            count += 1;
```
fulmicoton, line 105:
```

File: src/query/union/simple_union.rs [hunk 25]
```diff
@@ -0,0 +1,113 @@
...
+        }
```
fulmicoton:
```suggestion
        ```suggestion
        if self.doc == TERMINATED {
            return 0u32;
        }
        let mut count = 1u32;
```

src/query/union/simple_union.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn advance_to_next(&mut self) -> DocId {
+        let mut next_doc = TERMINATED;
+
+        for docset in &mut self.docsets {
+            if docset.doc() <= self.doc {
+                docset.advance();
+            }
+            next_doc = next_doc.min(docset.doc());
+        }
+        self.doc = next_doc;
```
fulmicoton, line 42: If you have a lot of docsets, I think a BinaryHeap might be faster.
Well... you can forget it I think.

src/query/phrase_query/phrase_weight.rs
PSeitz: https://github.com/quickwit-oss/tantivy/issues/2525

src/postings/loaded_postings.rs
PSeitz: I changed it to `>=`, so it takes over the job of the bounds check below
(btw. some branch predictors have performance problems with very dense if statements, e.g. 3 branches in 16bytes of instructions)

src/postings/segment_postings.rs
PSeitz: I'll leave a TODO here, we could work on initialized data, but it's a little bit tricky with regards to soundness

src/query/phrase_query/regex_phrase_weight.rs
PSeitz: true, I'll move the wildcard conversion to an utility function

src/query/union/bitset_union.rs
PSeitz: ah that makes much more sense. The performance difference should be gone with this.

src/query/union/simple_union.rs
PSeitz: I did some benchmarks some time ago and the intersections + union are faster, but the pure unions are slower with this variant

src/query/union/simple_union.rs
PSeitz: I did some tests initially with `BinaryHeap`, but it was quite slow when there were a lot of docsets. It might have been related to the size of `SegmentPostings` which I wasn't aware of yet.

I think there is a case for `BinaryHeap` here, but we would need a custom one to fully utilize its structure, e.g. some early exit when iterating the leaf nodes

fulmicoton (approved):

PR #2524, opened 2024-10-17
make casing in docs more consistent

marvin-j97:


fulmicoton:
I kind of prefer tantivy in lowercase. That's how we printed it on t-shirts :)

marvin-j97:
> I kind of prefer tantivy in lowercase. That's how we printed it on t-shirts :)

I thought so too, it was mostly capitalized at the start of a sentence, so I changed it to match that.

Maybe it should be lowercased everywhere?

fulmicoton (approved):

fulmicoton:
yes thank you

PR #2538, opened 2024-11-07
add seek_into_the_danger_zone on DocSet

### LINKED ISSUE #2531: Intersection Performance
The current intersection code looks like this (simplified)

```rust
fn advance(&mut self) -> DocId {
    let (left, right) = (&mut self.left, &mut self.right);
    let mut candidate = left.advance();

    loop {
        let right_doc = right.seek(candidate);
        candidate = left.seek(right_doc);
        if candidate == right_doc {
            return candidate;
        }
    }
}
```

`left` is the doc_set with the smallest `size_hint` in the group. It is advanced first then we seek on `right`.
`seek` is not only going to the passed candidate, but searches for the next hit in that docset.

The problem with this approach is that we have Docset types with vastly different cost for seek which is not reflected here.

E.g. fast field range queries may do a full scan, if there is no next hit. phrase queries load the positions to check if a we have a hit.

Ideally we would use the "cheapest" docset as the driver for the intersection, and all others just check if they have a hit for the passed candidate.
Some docsets, like a precomputed bitset are excellent drivers, but some, like fast field based range, queries aren't.
Cheapness is typically related to number of docs in the docset only for _same type_ of docsets. Between different docsets, we may have something like cost category (can be covered by e.g. using the high bits in u64).

To reflect that a cost based metric would make sense and an api that doesn't advance past `target`.

### Proposal: Seek Exact + Cost Based

```rust
pub trait DocSet: Send {
    ...
    /// Returns a best-effort hint of the
    /// length of the docset.
    fn size_hint(&self) -> u32;

    /// Returns a best-effort hint of the
    /// cost to drive the docset.
    fn cost(&self) -> u64{
        self.size_hint()
    }

    /// Seeks to the target if necessary and checks if the target is an exact match.
    ///
    /// Some implementations may choose to advance past the target if beneficial for performance.
    /// The return value is `true` if the target is in the docset.
    fn seek_exact(&mut self, target: DocId) -> bool {
        let current_doc = self.doc();
        if current_doc < target {
            self.seek(target);
        }
        self.doc() == target
    }
   ...
}
```

Related: https://github.com/quickwit-oss/tantivy/issues/2266

PSeitz:
## Add seek_into_the_danger_zone to DocSet

Adds `seek_into_the_danger_zone` to `DocSet` for a more efficient intersection.

Unlike `seek`, `seek_exact` does not require the DocSet to advance to the next hit, if the target does not exist.

Follow-up of https://github.com/quickwit-oss/tantivy/pull/2707
Closes #2531

## Future Work

Implement `seek_into_the_dangerzone` for RangeDocSet (fast field range queries)

Evaluate replacing `seek` with `seek_exact` to reduce code complexity

## Performance

Performance looks very good:

seek_exact perf

Especially the two phrase critic query is now properly executed:

seek_exact perf2


src/docset.rs
```diff
@@ -49,6 +49,25 @@ pub trait DocSet: Send {
         doc
     }
 
+    /// Seeks to the target if possible and returns true if the target is in the DocSet.
+    ///
+    /// Implementations may choose to advance past the target if target does not exist.
+    ///
+    /// DocSets that already have an efficient `seek` method don't need to implement `seek_exact`.
+    /// All wapper DocSets should forward `seek_exact` to the underlying DocSet.
```
fulmicoton:
```suggestion
    /// All wrapper DocSets should forward `seek_exact` to the underlying DocSet.
```

src/docset.rs
```diff
@@ -49,6 +49,25 @@ pub trait DocSet: Send {
...
+    ///
+    /// ## API Behaviour
+    /// If `seek_exact` is returning true, a call to `doc()` has to return target.
+    /// If `seek_exact` is returning false, a call to `doc()` may return the previous doc,
```
fulmicoton: can you clarify that the result should still be a document in the posting list or TERMINATED, and that it could be something before or after the target.

fulmicoton: We also need to clarify what happens if the target is lower than the current doc.

src/docset.rs
```diff
@@ -49,6 +49,25 @@ pub trait DocSet: Send {
...
+    /// which may be lower than target.
+    fn seek_exact(&mut self, target: DocId) -> bool {
+        let current_doc = self.doc();
+        if current_doc < target {
+            self.seek(target);
+        }
+        self.doc() == target
```
fulmicoton, line 78: why not:
```suggestion
        let doc = self.seek(target);
        doc == target
```

src/docset.rs
```diff
@@ -87,6 +106,23 @@ pub trait DocSet: Send {
...
+    /// By default this returns `size_hint()`.
+    ///
+    /// DocSets may have vastly different cost depending on their type,
+    /// e.g. an intersection with 10 hits is much cheaper than
+    /// a phrase search with 10 hits, since it needs to load positions.
+    ///
+    /// ### Future Work
+    /// We may want to differentiate `DocSet` costs more more granular, e.g.
+    /// creation_cost, advance_cost, seek_cost on to get a good estimation
+    /// what query types to choose.
```
fulmicoton, line 121: of course this will always be handwavy, but we might want to give one absolute frame of reference as a target.

for instance 1 unit of cost could be the amortized cost of advancing a term scorer by 1.

src/query/all_query.rs
```diff
@@ -58,6 +58,15 @@ impl DocSet for AllScorer {
         self.doc
     }
 
+    fn seek(&mut self, target: DocId) -> DocId {
+        debug_assert!(target >= self.doc);
+        self.doc = target;
+        if self.doc >= self.max_doc {
+            self.doc = TERMINATED;
+        }
+        self.doc
```
fulmicoton, line 71:
```suggestion
        self.doc = if target_doc >= self.max_doc {
            TERMINATED
        } else {
            target
        };
        self.doc
```

src/query/intersection.rs
```diff
@@ -95,32 +100,39 @@ impl<TDocSet: DocSet, TOtherDocSet: DocSet> DocSet for Intersection<TDocSet, TOt
...
             loop {
-                let right_doc = right.seek(candidate);
-                candidate = left.seek(right_doc);
-                if candidate == right_doc {
+                if right.seek_exact(candidate) {
                     break;
                 }
+                // `left.advance().max(right.doc())` yielded a regression in the search game
+                // benchmark It may make sense in certain scenarios though.
+                candidate = left.advance();
```
fulmicoton: if the regression is small I would sleep better with the `max(right.doc())`.
(There are a bunch of adversarial cases.)

src/query/intersection.rs
```diff
@@ -136,12 +148,37 @@ impl<TDocSet: DocSet, TOtherDocSet: DocSet> DocSet for Intersection<TDocSet, TOt
         doc
     }
 
+    /// Seeks to the target if necessary and checks if the target is an exact match.
+    ///
+    /// Some implementations may choose to advance past the target if beneficial for performance.
+    /// The return value is `true` if the target is in the docset, and `false` otherwise.
+    fn seek_exact(&mut self, target: DocId) -> bool {
+        self.left.seek_exact(target)
```
fulmicoton: so with after calling seek_exact, we do not necessarily end up with `self.doc()` returning a valid doc. Is this already taking in account today?

src/docset.rs
fulmicoton: actually it seem like your current implementation might return `self.doc()` that is not in the docset at all.

This is tricky, because right now we are saying that self.doc could return "anything".

This is not exactly true though.
We need to have an accurate definition of what can we done with docset after having called seek_exact once.

src/query/intersection.rs
```diff
@@ -11,14 +12,14 @@ use crate::{DocId, Score};
 /// For better performance, the function uses a
 /// specialized implementation if the two
 /// shortest scorers are `TermScorer`s.
-pub fn intersect_scorers(mut scorers: Vec<Box<dyn Scorer>>) -> Box<dyn Scorer> {
+pub fn intersect_scorers(mut scorers: Vec<Box<dyn Scorer>>, num_docs: u32) -> Box<dyn Scorer> {
```
fulmicoton: What is num_docs here?
This is unclear.

Is it an estimation of the number of docs in the intersection? max_doc?

can you document?

src/query/range_query/fast_field_range_doc_set.rs
```diff
@@ -136,9 +139,6 @@ impl<T: Send + Sync + PartialOrd + Copy + Debug + 'static> DocSet for RangeDocSe
         if let Some(docid) = self.loaded_docs.next() {
             return docid;
         }
-        if self.next_fetch_start >= self.column.num_docs() {
-            return TERMINATED;
```
fulmicoton, line 151: is this not required? can you explain when & why?

src/query/range_query/fast_field_range_doc_set.rs
```diff
@@ -174,7 +174,18 @@ impl<T: Send + Sync + PartialOrd + Copy + Debug + 'static> DocSet for RangeDocSe
...
+    }
+
+    /// Returns a best-effort hint of the
+    /// cost to drive the docset.
+    fn cost(&self) -> u64 {
+        // Advancing the docset is pretty expensive since it scans the whole column, there is no
+        // index currently (will change with an kd-tree)
+        // Since we use SIMD to scan the fast field range query we lower the cost a little bit.
+        // Ideally this would take the fast field codec into account
+        (self.column.num_docs() as f64 * 0.8) as u64
```
fulmicoton, line 206: I would have expected something higher than `self.column.num_docs()`

src/query/size_hint.rs
```diff
@@ -0,0 +1,138 @@
...
+///
+/// # Arguments
+/// * `docset_sizes` - An iterator over the sizes of the docsets (number of documents in each set).
+/// * `max_docs` - The maximum number of docs that can hit, usually number of documents in the
+///   segment.
+///
+/// # Returns
+/// The estimated number of documents in the intersection.
+pub fn estimate_intersection<I>(mut docset_sizes: I, max_docs: u32) -> u32
+where I: Iterator<Item = u32> {
```
fulmicoton, line 12:
```suggestion
where I: Iterator<Item = u32> {
     if max_doc == 0u32 { return 0u32; }
```

src/query/size_hint.rs
```diff
@@ -0,0 +1,138 @@
...
+    // Terms tend to be not really randomly distributed.
+    // This factor is used to adjust the estimate.
+    let mut co_loc_factor: f64 = 1.3;
+
+    let mut intersection_estimate = match docset_sizes.next() {
```
fulmicoton, line 20: nitpick
`let Some else ` is ok here.

src/query/size_hint.rs
```diff
@@ -0,0 +1,138 @@
...
+    // Assuming random distribution of terms, the probability of a document being in the
+    // union is the complement of the probability of it not being in any of the sets.
+    for size in docset_sizes {
+        let prob_in_set = (size as f64 / max_docs as f64) * co_loc_factor;
+        not_in_any_set_prob *= 1.0 - prob_in_set;
+    }
+
+    let union_estimate = (max_docs as f64 * (1.0 - not_in_any_set_prob)).round();
+
+    union_estimate.min(u32::MAX as f64) as u32
```
fulmicoton:
```suggestion
    union_estimate.min(max_docs as f64) as u32
```

src/query/intersection.rs
PSeitz: The regression was not small, so it might be better to identify the cases in which `max(right.doc())` is useful and have a separate loop for them

src/docset.rs
```diff
@@ -87,6 +106,23 @@ pub trait DocSet: Send {
     /// length of the docset.
     fn size_hint(&self) -> u32;
 
+    /// Returns a best-effort hint of the
```
fulmicoton: This is not accurate enough to be usable, and this is actually important here.

Is it the cost of calling advance? Is it the cost of consuming the entire docset? Or is it something else?

src/query/intersection.rs
PSeitz: number of docs in the segment, renamed and added a comment

src/query/range_query/fast_field_range_doc_set.rs
PSeitz: I moved it to `fetch_block`, so it is easier to use

src/query/range_query/fast_field_range_doc_set.rs
PSeitz: It depends on the number of actual hits, if the column has many hits, the cost would be higher, otherwise it would be lower. 

If it is higher than num_docs, we would never choose ff range query as the driver in a DocSet, when intersecting a term query with a fast field. But it's the faster choice when the term query has a lot of docids and the range query has not.

src/docset.rs
PSeitz: I added come comments to clarify

src/query/all_query.rs
PSeitz: I think the slightly less nested variant is a little bit more readable

src/docset.rs
PSeitz: I added some comments to clarify the contract

src/query/size_hint.rs
PSeitz: With the conversion to f64 I think `match` is a little more readable here

src/query/intersection.rs
fulmicoton: i can give you one adversarial case. 

intersection between two terms. 

The first term matches 0..1_000_000.
The second term is matching docs 1_000_000..10_000_000.

a more like scenario could be 
first term matches 0..1_000_000
second term is random.


This kind of problem actually happen in reality when document are ordered in a certain way.
For instance, eBay and Indeed are sorting by location.
We often sort by time, which is correlated in strange ways (for instance you can think of status:ERROR).

src/query/intersection.rs
PSeitz: Yes, the intersection code does not use self.doc if seek_exact returns true

src/query/intersection.rs
PSeitz: Generally I think it would make sense for all queries that don't implement a specialized `seek_exact` to use the `max(right.doc())` loop. 
I don't think there's a way in rust to detect this?

For now we could just detect if the right docset is a `TermScorer` and then use the `max(right.doc())` variant. 

I created a issue here to add some missing benchmarks:
https://github.com/quickwit-oss/tantivy/issues/2544

fulmicoton: How large was the regression here? Do you have benchmarks?

PSeitz: Actually I think the benchmark was bogus
1. We are not allowed to call `max(right.doc)` in the general case. 
2. If the candidate is coming from right, we would need to seek_exact on left.

It wasn't covered by a test then but it is now via proptest.

src/docset.rs
```diff
@@ -49,6 +49,25 @@ pub trait DocSet: Send {
...
+    /// Implementations may choose to advance past the target if target does not exist.
+    ///
+    /// DocSets that already have an efficient `seek` method don't need to implement `seek_exact`.
+    /// All wapper DocSets should forward `seek_exact` to the underlying DocSet.
+    ///
+    /// ## API Behaviour
+    /// If `seek_exact` is returning true, a call to `doc()` has to return target.
+    /// If `seek_exact` is returning false, a call to `doc()` may return the previous doc,
+    /// which may be lower than target.
+    fn seek_exact(&mut self, target: DocId) -> bool {
```
fulmicoton: let's find a better name

src/query/union/buffered_union.rs
```diff
@@ -197,6 +197,18 @@ where
         }
     }
 
+    fn seek_into_the_danger_zone(&mut self, target: DocId) -> bool {
+        let is_hit = self
+            .docsets
+            .iter_mut()
+            .all(|docset| docset.seek_into_the_danger_zone(target));
```
fulmicoton: why all? isn't that a union?

also are we ok with the lazy evaluation here?

src/query/intersection.rs
```diff
@@ -146,6 +161,19 @@ impl<TDocSet: DocSet, TOtherDocSet: DocSet> DocSet for Intersection<TDocSet, TOt
         doc
     }
 
+    /// Seeks to the target if necessary and checks if the target is an exact match.
+    ///
+    /// Some implementations may choose to advance past the target if beneficial for performance.
+    /// The return value is `true` if the target is in the docset, and `false` otherwise.
```
fulmicoton-dd, line 167: After this call, what is the result returned by `.doc()`?
This is not necessarily an actual document of the DocSet is it?

fulmicoton-dd: my bad this is documented on the trait.

fulmicoton-dd (approved):

PR #2555, opened 2024-12-06
accept * as field name

PSeitz:
accept * as field name in the query parser

### DIFF
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -6,7 +6,7 @@ use nom::bytes::complete::tag;
 use nom::character::complete::{
     anychar, char, digit1, multispace0, multispace1, none_of, one_of, satisfy, u32,
 };
-use nom::combinator::{eof, map, map_res, opt, peek, recognize, value, verify};
+use nom::combinator::{eof, map, map_res, not, opt, peek, recognize, value, verify};
 use nom::error::{Error, ErrorKind};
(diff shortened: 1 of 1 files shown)


trinity-1686a (approved):

fulmicoton:
@PSeitz shouldn't we merge this?

fulmicoton (changes requested):
This PR lacks a lot of explanation.
The motivation to begin with.
I assume you want to add "*:" because quickwit relies on the tantivy grammar to build its AST and DDG includes `*:` in its syntax.

That info needs to be in the code (and if you want a longer background explanation in the commit message & PR description).


Also, this implementation raises a lot of question:
- With this PR what happens for regular tantivy users with a `*:hello` query?
- Is this behavior unit tested? Does it make sense?
- What is happens for queries like,  `field*withastar:hello`. Is this behavior good, is it unit tested?
- What is happens for queries like,  `*fieldstartingwithastar:hello`. Is this behavior good, is it unit tested?
- What is happens for queries like,  `*fieldendingwithastar*:hello`. Is this behavior good, is it unit tested?

A thing we could consider, is to have a different query grammar for quickwit/pomsky,  or to make it possible for tantivy-grammar users to plug their our own definition of what a valid field name is.

trinity-1686a:
i'm not actually sure we need this anyway, in the situation where this seemed useful, the field name actually gets remapped to something else entirely, and we don't go through this query parser

PR #2556, opened 2024-12-09
update measure_time

### LINKED ISSUE #2554: RUSTSEC-2024-0384: Unmaintained transitive dependency
**Describe the bug**

Hey! I use cargo-deny and noticed I was getting an error for one of tantivys transitive dependencies.

[measure_time](https://github.com/PSeitz/rust_measure_time) depends on [instant](https://github.com/sebcrozet/instant) which has been marked as unmaintained. There is no security vuln associated or anything requiring immediate action as best as I could tell.

There is only one place in tantivy I could see that uses `measure_time`, [src/indexer/merger.rs](https://github.com/quickwit-oss/tantivy/blob/876a579e5da1223b40df044be396faac8151ff91/src/indexer/merger.rs#L8).

It is importing the `debug_time` macro and using it to instrument a few methods.
```rust
        debug_time!("write-fast-fields");
```

Given the small usage here, could this replaced with a `log::debug!` and remove the dependency to `measure_time`? If so, I'm happy to open a PR doing that. :grinning: 

Advisory:
https://rustsec.org/advisories/RUSTSEC-2024-0384

0.22

**To Reproduce**
```shell
cargo deny check
```

PSeitz:
closes https://github.com/quickwit-oss/tantivy/issues/2554

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ rayon = "1.5.2"
 lru = "0.12.0"
 fastdivide = "0.4.0"
 itertools = "0.13.0"
-measure_time = "0.8.2"
+measure_time = "0.9.0"
 arc-swap = "1.5.0"
 
 columnar = { version = "0.3", path = "./columnar", package = "tantivy-columnar" }


fulmicoton (approved):

PSeitz:
Superseeded by https://github.com/quickwit-oss/tantivy/pull/2557

PR #2558, opened 2024-12-11
Exist queries match subpath fields

rdettai:
Assuming the following document where `myfield` is a JSON fast field:
 ```json
 {
   "myfield": {
     "mysubfield": "hello"
   }
 }
 ```

"exists" queries on  `myfield` now also match the document


rdettai:
This is a prerequisite to handle the fast field case in https://github.com/quickwit-oss/quickwit/issues/5560

fulmicoton (approved):

PR #2559, opened 2024-12-18
allow warming partially an sstable for an automaton

trinity-1686a:


sstable/src/streamer.rs
```diff
@@ -129,8 +137,11 @@ where
     }
 
     /// See `into_stream(..)`
-    pub async fn into_stream_async(self) -> io::Result<Streamer<'a, TSSTable, A>> {
-        let delta_reader = self.delta_reader_async().await?;
+    pub async fn into_stream_async(
```
fulmicoton: - typo on undex.
- this methods needs comments to explain what merge_holes_under is about
- it is worth keeping the original method, and just have it call the new method with an argument of 0.

sstable/src/block_match_automaton.rs
```diff
@@ -0,0 +1,208 @@
+use tantivy_fst::Automaton;
+
+/// Returns whether a block whose starting key (exclusive) and final key (inclusive) can match the
+/// provided automaton, without actually looking at the block content.
+pub(crate) fn block_match_automaton(
```
fulmicoton:
```suggestion
pub(crate) fn can_block_match_automaton(
```

sstable/src/block_match_automaton.rs
```diff
@@ -0,0 +1,208 @@
...
+    start_key: Option<&[u8]>,
+    end_key: &[u8],
+    automaton: &impl Automaton,
+) -> bool {
+    let initial_block = start_key.is_none();
+    let start_key = start_key.unwrap_or(&[]);
+
+    debug_assert!(start_key <= end_key);
+
+    let prefix_len = start_key
```
fulmicoton: there is a function called `common_prefix_len`

sstable/src/block_match_automaton.rs
```diff
@@ -0,0 +1,208 @@
...
+        .iter()
+        .zip(end_key)
+        .take_while(|(c1, c2)| c1 == c2)
+        .count();
+
+    let mut base_state = automaton.start();
+    for c in &start_key[0..prefix_len] {
```
fulmicoton:
```suggestion
    for b in &start_key[0..prefix_len] {
```

sstable/src/block_match_automaton.rs
```diff
@@ -0,0 +1,208 @@
+use tantivy_fst::Automaton;
+
+/// Returns whether a block whose starting key (exclusive) and final key (inclusive) can match the
+/// provided automaton, without actually looking at the block content.
+pub(crate) fn block_match_automaton(
+    start_key: Option<&[u8]>,
```
fulmicoton:
```suggestion
    start_key_opt: Option<&[u8]>,
```
no shadowing please

sstable/src/block_match_automaton.rs
```diff
@@ -0,0 +1,208 @@
...
+    // )
+    //
+    // these are all written as regex, but can be converted to operations we can do:
+    // - [x-y] is a for c in x..y
+    // - .* is a can_match()
+    // - .+ is a for c in 0..=255 { accept(c).can_match() }
+    // - ? is a the thing before can_match(), or current state.is_match()
+    // - | means test both side
+
+    let mut start_range = *start_key.get(prefix_len).unwrap_or(&0);
```
fulmicoton: that's a bug isn't it? `prefix_len` could be `start_key.len()`

sstable/src/block_match_automaton.rs
```diff
@@ -0,0 +1,208 @@
...
+        start_range += 1;
+    }
+    for c in start_range..end_range {
+        let new_state = automaton.accept(&base_state, c);
+        if automaton.can_match(&new_state) {
+            return true;
+        }
+    }
+    if start_key.len() > prefix_len {
+        if start_key.len() <= prefix_len {
```
fulmicoton: this cannot happen... see line above.

sstable/src/block_match_automaton.rs
```diff
@@ -0,0 +1,208 @@
...
+            // case [abcd, abcde], we need to handle \0 which wasn't processed
+            let new_state = automaton.accept(&base_state, start_range);
+            if automaton.can_match(&new_state) {
+                eprintln!("ho");
```
fulmicoton:
```suggestion
```

sstable/src/block_match_automaton.rs
trinity-1686a: this was working properly, but is harder to understand than it needs. The way we handle when `start_key` is a prefix of `end_key` as been changed to hopefully make it clearer

src/index/inverted_index_reader.rs
```diff
@@ -315,6 +330,60 @@ impl InvertedIndexReader {
...
+        // there is a caching layer below, which might not always be true, but is in Quickwit.
+        let term_info = self.get_term_range_async(.., automaton, None, 0).await?;
+
+        // TODO this operation is often cheap for "friendly" automatons, but can be very costly for
+        // "unfriendly" ones such as ".*a{50}" (very few terms if any  match this pattern, but we
+        // can't know early). In this case, we decompress and iterate over the entire sstable, while
+        // still being in async context. Ideally we should spawn this on a threadpool.
+        let range_to_load = term_info
+            .map(|term_info| term_info.postings_range)
+            .coalesce(|range1, range2| {
```
fulmicoton: TIL about coalesce

trinity-1686a: it's from itertools, and it's very convenient for doing that kind of things

src/index/inverted_index_reader.rs
```diff
@@ -315,6 +330,60 @@ impl InvertedIndexReader {
...
+        automaton: A,
+        // with_positions: bool, at the moment we have no use for it, and supporting it would add
+        // complexity to the coalesce
+    ) -> io::Result<bool>
+    where
+        A::State: Clone,
+    {
+        // merge holes under 4MiB, that's how many bytes we can hope to receive during a TTFB from
+        // S3 (~80MiB/s, and 50ms latency)
+        let merge_holes_under = (80 * 1024 * 1024 * 50) / 1000;
```
fulmicoton:
```suggestion
        let merge_holes_under_bytes = (80 * 1024 * 1024 * 50) / 1000;
```

fulmicoton (approved):

sstable/src/block_match_automaton.rs
```diff
@@ -132,7 +132,7 @@ fn match_range_start<S, A: Automaton<State = S>>(
     automaton: &A,
     mut state: S,
 ) -> bool {
-    // case [abcdgj, abcpqr], `abcd` is already consumed, we need to handle:
+    // case ]abcdgj, abcpqr], `abcd` is already consumed, we need to handle:
```
fulmicoton: I prefer your notation, unfortunately in my experience, non-french people seem more familiar with `(left, right]`.

src/index/inverted_index_reader.rs
```diff
@@ -351,11 +351,16 @@ impl InvertedIndexReader {
         let mut _term_info = self
             .get_term_range_async(.., automaton.clone(), None, merge_holes_under)
             .await?;
+
         // we build a 2nd iterator, this one with no holes, so we don't go through blocks we can't
         // match, and just download them to reduce our query count. This makes the assumption
         // there is a caching layer below, which might not always be true, but is in Quickwit.
         let term_info = self.get_term_range_async(.., automaton, None, 0).await?;
```
fulmicoton: term_info for this thing is poor naming. This is an iterator over term_infos.

5 lines later you also introduce term_info as a closure argument.

src/index/inverted_index_reader.rs
```diff
@@ -351,11 +351,16 @@ impl InvertedIndexReader {
...
 
+        // TODO this operation is often cheap for "friendly" automatons, but can be very costly for
+        // "unfriendly" ones such as ".*a{50}" (very few terms if any  match this pattern, but we
+        // can't know early). In this case, we decompress and iterate over the entire sstable, while
+        // still being in async context. Ideally we should spawn this on a threadpool.
         let range_to_load = term_info
```
fulmicoton: Same as above, it is difficult for readers to know what is the type of range_to_load.

src/index/inverted_index_reader.rs
```diff
@@ -351,11 +351,16 @@ impl InvertedIndexReader {
...
+
         // we build a 2nd iterator, this one with no holes, so we don't go through blocks we can't
         // match, and just download them to reduce our query count. This makes the assumption
         // there is a caching layer below, which might not always be true, but is in Quickwit.
         let term_info = self.get_term_range_async(.., automaton, None, 0).await?;
 
+        // TODO this operation is often cheap for "friendly" automatons, but can be very costly for
+        // "unfriendly" ones such as ".*a{50}" (very few terms if any  match this pattern, but we
+        // can't know early). In this case, we decompress and iterate over the entire sstable, while
+        // still being in async context. Ideally we should spawn this on a threadpool.
```
fulmicoton: I see. So if I understand correctly, the trouble is that the computation is done within an iterator. 

We eventually yield, but that's only when we effectively get an element.

Is decompressing a single block on the runtime is fast enough, we could probably write ugly code and have our iterator return `Option<Range<usize>>` to offer the runtime to yield more often.

If we do it this way, we might also want to actively yield.
I believe tokio only force yield on await, after await has been reach N times. (to be checked). That N might be way too high for us.

The second is probably not of a problem

src/index/inverted_index_reader.rs
```diff
@@ -348,20 +348,20 @@ impl InvertedIndexReader {
...
-        let mut _term_info = self
+        let mut _term_infos = self
             .get_term_range_async(.., automaton.clone(), None, merge_holes_under_bytes)
             .await?;
 
         // we build a 2nd iterator, this one with no holes, so we don't go through blocks we can't
         // match, and just download them to reduce our query count. This makes the assumption
         // there is a caching layer below, which might not always be true, but is in Quickwit.
-        let term_info = self.get_term_range_async(.., automaton, None, 0).await?;
+        let term_infos = self.get_term_range_async(.., automaton, None, 0).await?;
```
fulmicoton:
```suggestion
        let term_info_stream = self.get_term_range_async(.., automaton, None, 0).await?;
```

fulmicoton (approved):

fulmicoton:
(approved but we need to fix the runtime footgun you identified of course)

src/index/inverted_index_reader.rs
```diff
@@ -315,6 +330,60 @@ impl InvertedIndexReader {
...
+    ) -> io::Result<bool>
+    where
+        A::State: Clone,
+    {
+        // merge holes under 4MiB, that's how many bytes we can hope to receive during a TTFB from
+        // S3 (~80MiB/s, and 50ms latency)
+        let merge_holes_under_bytes = (80 * 1024 * 1024 * 50) / 1000;
+        // we build a first iterator to download everything. Simply calling the function already
+        // loads everything, but doesn't start iterating over the sstable.
+        let mut _term_infos = self
```
fulmicoton:
```suggestion
        let _term_infos = self
```

src/termdict/mod.rs
```diff
@@ -74,6 +74,7 @@ const CURRENT_TYPE: DictionaryType = DictionaryType::SSTable;
 
 // TODO in the future this should become an enum of supported dictionaries
 /// A TermDictionary wrapping either an FST based dictionary or a SSTable based one.
+#[cfg_attr(feature = "quickwit", derive(Clone))]
```
trinity-1686a: i needed this to be able to clone the dictionary we pass something valid for 'static to the worker thread, though i don't really like that. The alternative would be to add an Arc inside InvertedIndexReader

src/index/inverted_index_reader.rs
```diff
@@ -334,44 +334,65 @@ impl InvertedIndexReader {
...
+            // Quickwit.
+            // We build things from this closure otherwise we get into lifetime issues that can only
+            // be solved with self referential strucs. Returning an io::Result from here is a bit
+            // more leaky abstraction-wise, but a lot better than the alternative
+            let mut stream = termdict.search(automaton).into_stream()?;
+
+            // we could do without an iterator, but this allows us access to coalesce which simplify
+            // things
+            let posting_ranges_iter =
+                std::iter::from_fn(move || stream.next().map(|(_k, v)| v.postings_range.clone()))
```
fulmicoton: can we break it into several statement.

For instance:
- Create the iterator of postings ranges.
- doe the coalesce thing after.

src/termdict/mod.rs
fulmicoton: The fst dictionary cannot be made cloneable?
I prefer either the Arc solution then.

fulmicoton (approved):
minor comments

src/termdict/mod.rs
trinity-1686a: > The fst dictionary cannot be made cloneable?

it contains an fst Map. I don't see an obvious reason why that can't be Clone, but it currently isn't, and would require updating tantivy_fst. We could Arc that Map instead or arcing inside the InvertedIndexReader, and make the Dictionary be clone independently of feature flags

PR #2535, opened 2024-11-05
Make `NUM_MERGE_THREADS` configurable

Barre:
This PR addresses #2533

It makes the amount of merge threads configurable.

PR #2562, opened 2024-12-26
Improve `IndexWriter` customisation via builder

ChillFish8:
Recently been hacking the index writer apart to do some mildly questionable thing, as part of that I found myself wanting to customise some more parts of the writer but adding another method with _more_ config options was starting to get unwieldy. Especially since https://github.com/quickwit-oss/tantivy/pull/2535 

This PR adds a builder notibly `IndexWriterOptions` which improves the ergonomics of configuring the writer.

However, I did also make the options _another_ method to avoid making it a huge breaking change, but maybe a breaking change of the create API fine. 

Also sorry, I accidentally wrapped some of the work I did around deferring indexer threads being spawned until required, will remove at a later point.

### DIFF
--- a/src/index/index.rs
+++ b/src/index/index.rs
@@ -15,7 +15,9 @@ use crate::directory::MmapDirectory;
 use crate::directory::{Directory, ManagedDirectory, RamDirectory, INDEX_WRITER_LOCK};
 use crate::error::{DataCorruption, TantivyError};
 use crate::index::{IndexMeta, SegmentId, SegmentMeta, SegmentMetaInventory};
-use crate::indexer::index_writer::{MAX_NUM_THREAD, MEMORY_BUDGET_NUM_BYTES_MIN};
+use crate::indexer::index_writer::{
+    IndexWriterOptions, MAX_NUM_THREAD, MEMORY_BUDGET_NUM_BYTES_MIN,
+};
 use crate::indexer::segment_updater::save_metas;
 use crate::indexer::{IndexWriter, SingleSegmentIndexWriter};
 use crate::reader::{IndexReader, IndexReaderBuilder};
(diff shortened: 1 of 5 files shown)


src/indexer/index_writer.rs
```diff
@@ -266,19 +278,17 @@ impl<D: Document> IndexWriter<D> {
     /// `TantivyError::InvalidArgument`
     pub(crate) fn new(
         index: &Index,
-        num_threads: usize,
-        memory_budget_in_bytes_per_thread: usize,
+        options: IndexWriterOptions,
```
fulmicoton, line 281: I think a reference is a tiny bit better here.

```suggestion
        options: &IndexWriterOptions,
```

src/index/index.rs
```diff
@@ -560,32 +554,40 @@ impl Index {
...
         &self,
         num_threads: usize,
         overall_memory_budget_in_bytes: usize,
     ) -> crate::Result<IndexWriter<D>> {
-        self.writer_with_num_threads_and_num_merge_threads(
-            num_threads,
-            overall_memory_budget_in_bytes,
-            DEFAULT_NUM_MERGE_THREADS,
-        )
+        let memory_arena_in_bytes_per_thread = overall_memory_budget_in_bytes / num_threads;
```
fulmicoton, line 585: can we enforce one way or another that num_threads is > 0.

(Just an assert with an error message is ok...)

src/index/index.rs
```diff
@@ -521,30 +521,24 @@ impl Index {
...
-    /// Each thread will receive a budget of `overall_memory_budget_in_bytes / num_threads`.
+    /// - `options` defines the writer configuration which includes things like buffer sizes,
+    ///   indexer threads, etc...
     ///
     /// # Errors
-    /// If the lockfile already exists, returns `Error::DirectoryLockBusy` or an `Error::IoError`.
+    /// If the lockfile already exists, returns `TantivyError::LockFailure`.
     /// If the memory arena per thread is too small or too big, returns
     /// `TantivyError::InvalidArgument`
-    pub fn writer_with_num_threads_and_num_merge_threads<D: Document>(
```
fulmicoton, line 543: is this method published in a version? If so can we keep an implem of it and call writer_with_options?

fulmicoton (changes requested):
see comment

src/index/index.rs
ChillFish8: Yes will add

src/index/index.rs
ChillFish8: IIRC it isn't, this method was added in https://github.com/quickwit-oss/tantivy/pull/2535 ~3 weeks ago, I think the last published version was April. Although I think @Barre is probably using the mentioned method already (pinging so they're aware of this change.)

fulmicoton: that's ok then. As long as it isn't published!

src/index/index.rs
fulmicoton: last remaining point here.

src/indexer/index_writer.rs
```diff
@@ -45,6 +45,22 @@ fn error_in_index_worker_thread(context: &str) -> TantivyError {
...
+/// A builder for creating a new [IndexWriter] for an index.
+pub struct IndexWriterOptions {
+    #[builder(default = MEMORY_BUDGET_NUM_BYTES_MIN)]
+    /// The memory budget per indexer thread.
+    ///
+    /// When an indexer thread has buffered this much data in memory
+    /// it will flush the segment to disk (although this is not searchable until commit is called.)
+    memory_budget_per_thread: usize,
+    /// The number of indexer worker threads to use.
+    num_worker_threads: usize,
```
fulmicoton, line 59: should we default that to 1?

fulmicoton (approved):

fulmicoton (approved):

PR #2568, opened 2025-01-15
allow term starting with wildcard in query parser

### LINKED ISSUE #5306: support wildcard in non-suffix position
currently we allow `field:value*`, but not `field:val*ue`

trinity-1686a:
fix https://github.com/quickwit-oss/quickwit/issues/5306#issuecomment-2589428099

### DIFF
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -321,7 +321,17 @@ fn exists(inp: &str) -> IResult<&str, UserInputLeaf> {
         UserInputLeaf::Exists {
             field: String::new(),
         },
-        tuple((multispace0, char('*'), peek(alt((multispace1, eof))))),
+        tuple((
+            multispace0,
+            char('*'),
+            peek(alt((
+                value(
+                    "",
+                    satisfy(|c: char| c.is_whitespace() || ESCAPE_IN_WORD.contains(&c)),
+                ),
+                eof,
+            ))),
+        )),
     )(inp)
 }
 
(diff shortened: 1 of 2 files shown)


tandr:
Is there any chance for this fix to be reviewed and pulled in any time soon, sorry?

query-grammar/src/query_grammar.rs
```diff
@@ -331,7 +331,8 @@ fn exists_precond(inp: &str) -> IResult<&str, (), ()> {
         peek(tuple((
             field_name,
             multispace0,
-            char('*'), // when we are here, we know it can't be anything but a exists
+            char('*'),
+            peek(alt((multispace1, eof))), // we need to check this isn't a wildcard query
```
fulmicoton: will that work with something like 

`(toto:*)`

or in a more legitimate case

`(hello AND toto:*) OR happy`

query-grammar/src/query_grammar.rs
```diff
@@ -1626,11 +1627,11 @@ mod test {
     fn test_exist_query() {
         test_parse_query_to_ast_helper("a:*", "\"a\":*");
         test_parse_query_to_ast_helper("a: *", "\"a\":*");
-        // an exist followed by default term being b
-        test_is_parse_err("a:*b", "(*\"a\":* *b)");
 
-        // this is a term query (not a phrase prefix)
+        // these are term/wildcard query (not a phrase prefix)
         test_parse_query_to_ast_helper("a:b*", "\"a\":b*");
```
fulmicoton, line 1654: we should use a different way to serialize field exists query.

for instance `$field_exists("fieldname")`

right now these unit test are ambiguous for humans, and possibly for machines.

query-grammar/src/query_grammar.rs
trinity-1686a: it did not, and wouldn't have been possible to test without fixing comment#2

guilload (approved):

rtpg:
https://quickwit.io/docs/reference/query-language I think the query language docs might not be up to date with this (very helpful!) change

trinity-1686a:
there are a few things that needs to be updated in that documentation. quickwit main should support /regexes/ now for instance. I don't know when one of us will work on updating the documentation, but feel free to do it yourself and cc me on the PR if you do.

PR #2582, opened 2025-02-20
Catch panics during merges

### LINKED ISSUE #2411: Rayon thread pool abort on panic
As spotted by @adamreichold on #2409.

It might be good to catch panic with `catch_unwind`.

rdettai:
Closes https://github.com/quickwit-oss/tantivy/issues/2411

We recently encountered a bug in merges that was triggering a panic, which in turn was made into an abort by rayon.

This PR proposes a more graceful handling of the panic.

### DIFF
--- a/src/indexer/segment_updater.rs
+++ b/src/indexer/segment_updater.rs
@@ -1,3 +1,4 @@
+use std::any::Any;
 use std::borrow::BorrowMut;
 use std::collections::HashSet;
 use std::io::Write;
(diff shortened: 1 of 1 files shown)


rdettai (commented):
@fulmicoton should we also do it here? 
https://github.com/rdettai/tantivy/blob/6dadfa85758d207a9c634a9ac4cb630977d65a14/src/indexer/segment_updater.rs#L341-L344

src/indexer/segment_updater.rs
```diff
@@ -506,11 +518,34 @@ impl SegmentUpdater {
             // Its lifetime is used to track how many merging thread are currently running,
             // as well as which segment is currently in merge and therefore should not be
             // candidate for another merge.
-            match merge(
-                &segment_updater.index,
-                segment_entries,
-                merge_operation.target_opstamp(),
-            ) {
+            let merge_panic_res = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
```
fulmicoton-dd, line 521: thx for the explicit naming!

fulmicoton (approved):

PR #2586, opened 2025-02-24
follow up on the fix of multiply with overflow

fulmicoton:


PSeitz (approved):
LGTM, `Bitpacker` already uses 16 bytes anyways

bitpacker/src/bitpacker.rs
```diff
@@ -161,23 +162,23 @@ impl BitUnpacker {
...
+        let num_blocks: usize = (end_idx - highway_start) / BitPacker1x::BLOCK_LEN;
 
         // Entrance ramp
-        get_batch_ramp(start_idx, &mut output[..entrance_ramp_len as usize]);
+        get_batch_ramp(start_idx, &mut output[..entrance_ramp_len]);
 
         // Highway
-        let mut offset = (highway_start * self.num_bits) as usize / 8;
-        let mut output_cursor = (highway_start - start_idx) as usize;
+        let mut offset = (highway_start * self.num_bits) / 8;
```
rdettai: nit: this function might be a bit easier to read if the widening from u32 to usize was performed only when it is actually necessary:
```rust
// keep indexes as u32 and do
let mut offset = (highway_start as usize * self.num_bits) / 8;
```

fulmicoton-dd (approved):

PR #2588, opened 2025-02-25
fix clippy

PSeitz:


bitpacker/src/bitpacker.rs
```diff
@@ -48,7 +48,7 @@ impl BitPacker {
 
     pub fn flush<TWrite: io::Write + ?Sized>(&mut self, output: &mut TWrite) -> io::Result<()> {
         if self.mini_buffer_written > 0 {
-            let num_bytes = (self.mini_buffer_written + 7) / 8;
+            let num_bytes = self.mini_buffer_written.div_ceil(8);
```
fulmicoton, line 51: Here it does not matter, but this has a small micro impact on performance.

bitpacker/src/bitpacker.rs
```diff
@@ -65,7 +65,7 @@ impl BitPacker {
 
 #[derive(Clone, Debug, Default, Copy)]
 pub struct BitUnpacker {
-    num_bits: usize,
+    num_bits: u32,
```
fulmicoton: why?

bitpacker/src/blocked_bitpacker.rs
```diff
@@ -34,7 +34,7 @@ struct BlockedBitpackerEntryMetaData {
 
 impl BlockedBitpackerEntryMetaData {
     fn new(offset: u64, num_bits: u8, base_value: u64) -> Self {
-        let encoded = offset | (u64::from(num_bits) << (64 - 8));
+        let encoded = offset | ((num_bits as u64) << (64 - 8));
```
fulmicoton: ?

fulmicoton (changes requested):
I think part of this PR is obsolete. Probably a failed rebase?

bitpacker/src/bitpacker.rs
PSeitz: ah good to know, I thought it would lead to the same instruction

PSeitz:
> I think part of this PR is obsolete. Probably a failed rebase?

Yes, a failed rebase. I restarted from scratch instead

PR #2589, opened 2025-02-27
Update stemmers library

fat-fellow:
It seems that the original repository is abandoned, but in recent years there has been an update to the available stemmer languages.


fulmicoton:
I'm sure you do not mean any wrong, but merging that kind of PR is a big security hazard.

fat-fellow:
may I ask you a question, how it supposed to be implemented then? the original repo is not being updated.

PR #2600, opened 2025-03-18
Replace the default score combiner for boolean queries by a boosted variant

umweltinfo:
Based on user feedback, and using the default `QueryParser` interpretation, we found this modification of plain BM25 to yield significantly more relevant results and considerably reducing the need for explicit `Must` queries.


fulmicoton:
@umweltinfo We cannot merge that kind of PR.
From your point of view, it is an obvious improvement, for which you have first hand experience.
From the rest of the world pov, it is changing the default behavior of tantivy for every users silently based on one user's undocumented experience.

umweltinfo:
> From the rest of the world pov, it is changing the default behavior of tantivy for every users silently based on one user's undocumented experience.

Indeed it is. In our experience, it worked much better as default though.

But in any case, we can certainly continue to use it by rewriting queries to use a custom variant of `BooleanQuery`. I'll open a different PR to make that rewriting easier/more efficient. (We currently need to `box_clone` all sub queries to repackage them with the custom combiner.)

PR #2606, opened 2025-03-30
Add serde json serialize to UserInputAst

katlim-br:
We have two goals with this PR:

1. Allow us to reuse this grammar parser and validate the query input from our users.
  - We use Node, so we need to reuse this logic.
  - Serializing the output as json makes it much more interoperable since we likely need to go Rust > C++ > Node.
2. Use the AST to also validate an object against it.

UPDATE: I found that tests are failing with a fresh clone. I added [this issue](https://github.com/quickwit-oss/tantivy/issues/2607). FIXED.


query-grammar/Cargo.toml
```diff
@@ -13,3 +13,5 @@ edition = "2021"
 
 [dependencies]
 nom = "7"
+serde = { version = "1.0.219", features = ["derive"] }
```
PSeitz, line 16: Can you update the versions in the root Cargo to use the same versions?

katlim-br: done

katlim-br:
I saw the unit tests workflow failing due to formatting, so I applied the fixes for query-grammar only (haha although many files from other parts of the repo need to be fixed)

query-grammar/Cargo.toml
katlim-br: Also, btw, there are other places where the serde versions differ, but I only changed the root.

Do we want to update these too?

katlim-br:
I have this other option:

```rust
use tantivy_query_grammar::{UserInputAst, UserInputBound, UserInputLiteral, UserInputLeaf, Delimiter, Occur};
use serde::{Serialize, Deserialize};

#[derive(Serialize, Deserialize)]
pub enum OccurRemote {
    /// For a given document to be considered for scoring,
    /// at least one of the queries with the Should or the Must
    /// Occur constraint must be within the document.
    Should,
    /// Document without the queries are excluded from the search.
    Must,
    /// Document that contain the query are excluded from the
    /// search.
    MustNot,
}

#[derive(Serialize, Deserialize)]
pub enum DelimiterRemote {
    SingleQuotes,
    DoubleQuotes,
    None,
}

#[derive(Serialize, Deserialize)]
pub enum UserInputAstRemote {
    Clause(Vec<(Option<OccurRemote>, UserInputAstRemote)>),
    Leaf(Box<UserInputLeafRemote>),
    Boost(Box<UserInputAstRemote>, f64),
}

#[derive(Serialize, Deserialize)]
pub enum UserInputBoundRemote {
    Inclusive(String),
    Exclusive(String),
    Unbounded,
}

#[derive(Serialize, Deserialize)]
pub struct UserInputLiteralRemote {
    pub field_name: Option<String>,
    pub phrase: String,
    pub delimiter: DelimiterRemote,
    pub slop: u32,
    pub prefix: bool,
}

#[derive(Serialize, Deserialize)]
pub enum UserInputLeafRemote {
    Literal(UserInputLiteralRemote),
    All,
    Range {
        field: Option<String>,
        lower: UserInputBoundRemote,
        upper: UserInputBoundRemote,
    },
    Set {
        field: Option<String>,
        elements: Vec<String>,
    },
    Exists {
        field: String,
    },
}

#[derive(Serialize, Deserialize)]
pub struct LenientErrorRemote {
    pub pos: usize,
    pub message: String,
}

impl From<Occur> for OccurRemote {
    fn from(occur: Occur) -> Self {
        match occur {
            Occur::Should => OccurRemote::Should,
            Occur::Must => OccurRemote::Must,
            Occur::MustNot => OccurRemote::MustNot,
        }
    }
}

impl From<Delimiter> for DelimiterRemote {
    fn from(delimiter: Delimiter) -> Self {
        match delimiter {
            Delimiter::SingleQuotes => DelimiterRemote::SingleQuotes,
            Delimiter::DoubleQuotes => DelimiterRemote::DoubleQuotes,
            Delimiter::None => DelimiterRemote::None,
        }
    }
}

impl From<UserInputBound> for UserInputBoundRemote {
    fn from(bound: UserInputBound) -> Self {
        match bound {
            UserInputBound::Inclusive(s) => UserInputBoundRemote::Inclusive(s),
            UserInputBound::Exclusive(s) => UserInputBoundRemote::Exclusive(s),
            UserInputBound::Unbounded => UserInputBoundRemote::Unbounded,
        }
    }
}

impl From<UserInputLiteral> for UserInputLiteralRemote {
    fn from(literal: UserInputLiteral) -> Self {
        UserInputLiteralRemote {
            field_name: literal.field_name,
            phrase: literal.phrase,
            delimiter: literal.delimiter.into(),
            slop: literal.slop,
            prefix: literal.prefix,
        }
    }
}

impl From<UserInputLeaf> for UserInputLeafRemote {
    fn from(leaf: UserInputLeaf) -> Self {
        match leaf {
            UserInputLeaf::Literal(l) => UserInputLeafRemote::Literal(l.into()),
            UserInputLeaf::All => UserInputLeafRemote::All,
            UserInputLeaf::Range { field, lower, upper } => UserInputLeafRemote::Range {
                field,
                lower: lower.into(),
                upper: upper.into(),
            },
            UserInputLeaf::Set { field, elements } => UserInputLeafRemote::Set {
                field,
                elements,
            },
            UserInputLeaf::Exists { field } => UserInputLeafRemote::Exists { field },
        }
    }
}

impl From<UserInputAst> for UserInputAstRemote {
    fn from(ast: UserInputAst) -> Self {
        match ast {
            UserInputAst::Clause(clauses) => UserInputAstRemote::Clause(
                clauses.into_iter()
                    .map(|(occur, ast)| (occur.map(Into::into), ast.into()))
                    .collect()
            ),
            UserInputAst::Leaf(leaf) => UserInputAstRemote::Leaf(Box::new((*leaf).into())),
            UserInputAst::Boost(ast, boost) => UserInputAstRemote::Boost(Box::new((*ast).into()), boost),
        }
    }
}
```

This uses [this pattern](https://serde.rs/remote-derive.html) (however, I had compile errors by using the `#[serde:(remote)]` syntax).

It works with the first tests.

I would still prefer the library itself to provide the JSON serialization (so it is future proof).

katlim-br:
@PSeitz once approved and merged, should I update the tantivy-py repo?

query-grammar/src/user_input_ast.rs
```diff
@@ -1,9 +1,11 @@
...
 
+use serde::Serialize;
+
 use crate::Occur;
 
-#[derive(PartialEq, Clone)]
+#[derive(PartialEq, Clone, Serialize)]
 pub enum UserInputLeaf {
     Literal(UserInputLiteral),
     All,
```
rdettai: It would look more idiomatic to have either:
- `#[serde(rename_all_fields = "snake_case")]`
- an [internal tagging](https://serde.rs/enum-representations.html#internally-tagged)

fulmicoton-dd: agreed! Can you make the change [katlim-br](https://github.com/katlim-br) on all of the enum of the tree?

fulmicoton (changes requested):
Please apply snake case + internal tagging for all enums?

query-grammar/src/user_input_ast.rs
katlim-br: ok about snake_case

About internal tagging, do you have any example or preference? I see is barely used across the whole repo, so I don't see enough examples to see a pattern.

katlim-br: I added tag=type to UserInputBound and UserInputLeaf. 
I think it will allow to read the data based on the "type" field easier (instead of using the exported name of the field).

katlim-br: About snake_case:
I saw all fields to be already in snake case, and I think enum values should remain the same across languages like `Inclusive`, `Unbounded`, `Should`, or `Must`.

katlim-br:
> Please apply snake case + internal tagging for all enums?

Fixed, please let me know if it's ok.

I also added more unit tests so it is more visible how the objects end up in json

query-grammar/src/user_input_ast.rs
fulmicoton-dd: Snake case and internal tagging everywhere please. 

> I think enum values should remain the same across languages like Inclusive, Unbounded, Should, or Must.

The format should be language agnostic yes. I don't see how it is an argument for camel case however.

Quickwit, elasticsearch, and tantivy's schema type are using snake case for enum types.
It seems generally more natural to me ™.

query-grammar/src/user_input_ast.rs
```diff
@@ -187,7 +191,7 @@ impl UserInputBound {
     }
 }
 
-#[derive(PartialEq, Clone)]
+#[derive(PartialEq, Clone, Serialize)]
 pub enum UserInputAst {
```
rdettai: tagging missing here (with snake case values)

query-grammar/src/user_input_ast.rs
```diff
@@ -152,7 +155,8 @@ impl fmt::Debug for UserInputLiteral {
     }
 }
 
-#[derive(PartialEq, Debug, Clone)]
+#[derive(PartialEq, Debug, Clone, Serialize)]
+#[serde(tag = "type", content = "value")]
 pub enum UserInputBound {
```
rdettai, line 164: make the tag values snake case (here and all other serialized enums)

query-grammar/src/user_input_ast.rs
katlim-br: By adding
```
#[serde(rename_all = "snake_case")]
pub enum UserInputLeaf {
```

I get
```
BEFORE
 right: "{\"Leaf\":{\"type\":\"Literal\",\"field_name\":\"title\",\"phrase\":\"hello\",\"delimiter\":\"None\",\"slop\":0,\"prefix\":false}}"

AFTER
  left: "{\"leaf\":{\"type\":\"literal\",\"field_name\":\"title\",\"phrase\":\"hello\",\"delimiter\":\"none\",\"slop\":0,\"prefix\":false}}"

```
Is this the intention? (type: literal, delimeter: none)

fulmicoton-dd: yes.

katlim-br: done

query-grammar/src/user_input_ast.rs
katlim-br: fixed

query-grammar/src/user_input_ast.rs
katlim-br: done

fulmicoton-dd (approved):

query-grammar/src/user_input_ast.rs
katlim-br: Last update.

I agreed with Remi to have these:

```
Clause map to {"clause":[...]}
Leaf be flattened out to the UserInputLeaf, e.g {"type":"literal","field_name":"title","phrase":"hello","delimiter":"none","slop":0,"prefix":false}
```

About Boost, he wanted to do:
```
{type: literal, ...., boost: 3}
```
But the boost is defined as a tuple containing the AST, and the value. I tried many ways but it is not possible. The only way is to create a custom serializer (but I don't think is worth it), or change the Rust definition but that is out of scope.

Please let me know if this is ok now.

@rdettai @fulmicoton @fulmicoton-dd

rdettai:
A proposed variation that makes `UserInputAst` more consistent with the rest and the Quickwit QueryAst https://github.com/katlim-br/tantivy/pull/1

query-grammar/Cargo.toml
rdettai: dependencies should be `{ workspace = true }` which is now the norm but didn't exist back then. I'll do it in a separate PR

rdettai (approved):
LGTM

query-grammar/src/user_input_ast.rs
rdettai: we made a last tweak to have a consistent `type` tag across all nodes

PR #2614, opened 2025-04-03
set root version to be 0.24.0

katlim-br:
I missed updating the cargo root version to 0.24.0 in this [PR](https://github.com/quickwit-oss/tantivy/pull/2606) so it can be published.

I aim to publish the `tantivy` package along with the `tantivy-query-grammar` to `crates.io`, so I can then ask the Python bindings to be updated (tantivy-py).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tantivy"
-version = "0.23.0"
+version = "0.24.0"
 authors = ["Paul Masurel <paul.masurel@gmail.com>"]
 license = "MIT"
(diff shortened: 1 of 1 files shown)


fulmicoton:
0.23 has not been published yet. @PSeitz is working on this. Why do you want to bump to 0.24?

katlim-br:
> 0.23 has not been published yet. @PSeitz is working on this. Why do you want to bump to 0.24?

I wanted to bump, so I can publish into tantivy-py, but let's just close this for now. I was able to import the dependency using the gh commit.

thanks

PR #2620, opened 2025-04-09
update edition to 2024

PSeitz:
- **update common to edition 2024**
- **update bitpacker to edition 2024**
- **update stacker to edition 2024**
- **update query-grammar to edition 2024**
- **update sstable to edition 2024 + fmt**

Note: relative drop order changed in Rust 2024. This should be fine except for locks, where it may cause issues (or fix them since the behavior is more consistent)
```
warning: relative drop order changing in Rust 2024
   --> src/directory/managed_directory.rs:138:19
    |
127 |             let meta_informations_rlock = self
    |                 -----------------------
    |                 |
    |                 `meta_informations_rlock` calls a custom destructor
    |                 `meta_informations_rlock` will be dropped later as of Edition 2024
...
138 |             match self.acquire_lock(&META_LOCK) {
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                   |
    |                   this value will be stored in a temporary; let us call it `#1`
    |                   up until Edition 2021 `#1` is dropped last but will be dropped earlier in Edition 2024
...
152 |         }
    |         - now the temporary value is dropped here, before the local variables in the block or statement
    |
    = warning: this changes meaning in Rust 2024
    = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/temporary-tail-expr-scope.html>
    = note: `#1` may invoke a custom destructor because it contains a trait object
```


columnar/src/columnar/merge/mod.rs
```diff
@@ -144,16 +144,17 @@ fn merge_column(
...
-                    column_indexes.push(ColumnIndex::Empty {
-                        num_docs: num_docs_per_column[i],
-                    });
-                    column_values.push(None);
+                match dynamic_column_opt.and_then(dynamic_column_to_u64_monotonic) {
+                    Some(Column { index: idx, values }) => {
+                        column_indexes.push(idx);
+                        column_values.push(Some(values));
+                    }
+                    _ => {
```
fulmicoton-dd:
```suggestion
                   None => {
```

fulmicoton-dd (approved):

Issue #2577, opened 2025-02-04
Attempt to multiply with overflow


Barre:
While searching my index of around 3B documents which looks like this for the biggest segments:

```json

{
  "index_settings": {
    "docstore_compression": "lz4",
    "docstore_blocksize": 16384
  },
  "segments": [
    {
      "segment_id": "66e5b4e1-b1d5-4aed-aba6-51565f3d6acc",
      "max_doc": 1054732330,
      "deletes": null
    },
    {
      "segment_id": "61483fc5-0bd4-4e85-b9d5-1723b6d90173",
      "max_doc": 105936249,
      "deletes": null
    },
    {
      "segment_id": "0646c5b8-8bf6-4a95-be54-27724cf3f0e4",
      "max_doc": 53501794,
      "deletes": null
    },
    {
      "segment_id": "965d3f1c-d4a7-4545-946e-382fc469885b",
      "max_doc": 53411250,
      "deletes": null
    },

```

Performing a search such as:

```rust

        let (count, top_docs) = searcher.search(
            &query,
            &(
                Count,
                TopDocs::with_limit(items_per_page)
                    .and_offset(offset)
                    .order_by_fast_field::<DateTime>(
                        "date",
                        tantivy::Order::Desc,
                    ),
            ),
        )?;
```

Produces the following in debug mode 

```
thread 'tantivy-search-2' panicked at .cargo/git/checkouts/tantivy-f70b7ea03dadae9a/4aa8cd2/bitpacker/src/bitpacker.rs:97:28:
attempt to multiply with overflow
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/core/src/panicking.rs:76:14
   2: core::panicking::panic_const::panic_const_mul_overflow
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/core/src/panicking.rs:182:21
   3: tantivy_bitpacker::bitpacker::BitUnpacker::get
             at .cargo/git/checkouts/tantivy-f70b7ea03dadae9a/4aa8cd2/bitpacker/src/bitpacker.rs:97:28
[...]

PR #2627, opened 2025-04-18
Update `LogMergePolicy` to skip to a target number of documents

kryesh:
From discord conversation:
https://discord.com/channels/908281611840282624/915785344396439552/1341705839668625468

This updated logic makes `LogMergePolicy` aim for a specific target number of documents, and opportunistically skip merge operations to reach that target document count.
Pros:
* Reduced IO/CPU usage from skipping intermediate merge operations
* No longer susceptible to creating huge merge operations by merging many large segments into a single segment many times larger than the target size (previously `max_docs_before_merge`)
  *  Oversized merge operations can lead to other issues such as #2577
    * Limits applied to input segment size aren't effective if many segments are being merged
  *  The theoretical maximum size of a segment with this updated logic is `(target_segment_size * 2) - 2`

Cons:
*  If an index has a little over `target_segment_size` total docs then it may get merged to a single segment and thus not parallelize well when searching


src/indexer/log_merge_policy.rs
```diff
@@ -76,62 +76,97 @@ impl LogMergePolicy {
...
+                batch.push(seg);
+
+                // If the current batch has enough documents to be merged, create a merge
+                // candidate and push it to candidates
+                if batch_docs >= self.target_segment_size {
+                    unmerged_docs -= batch_docs;
+                    batch_docs = 0;
+                    candidates.push(MergeCandidate(
+                        // drain to reuse the buffer
+                        batch.drain(..).map(|seg| seg.id()).collect(),
```
fulmicoton, line 120: Nothing prevents this merge to have 1000 segments.

src/indexer/log_merge_policy.rs
```diff
@@ -76,62 +76,97 @@ impl LogMergePolicy {
...
+                    ));
+                    // If there aren't enough documents to create another segment of the target size
+                    // then break
+                    if unmerged_docs <= self.target_segment_size {
+                        break;
```
fulmicoton, line 125: Ok we break here. Which means we keep going through the rest of the function.

But the levels are now empty, because we popped them... This is terrible logic.

src/indexer/log_merge_policy.rs
```diff
@@ -76,62 +76,97 @@ impl LogMergePolicy {
...
+                    // update current_max_log_size to create a new group
+                    current_max_log_size = segment_log_size;
+                }
+                current_max_log_size
             })
-            .map(|segments| MergeCandidate(segments.iter().map(|&seg| seg.id()).collect()))
-            .collect()
+            .into_iter()
+            .for_each(|(_, group)| {
+                let mut hit_delete_threshold = false;
```
fulmicoton: for loop would be nicer than for_each here.

src/indexer/log_merge_policy.rs
```diff
@@ -76,62 +76,97 @@ impl LogMergePolicy {
...
+                group.for_each(|(_, seg)| {
```
fulmicoton: same. Iterator are great. Abusing them is just making code less readable.

src/indexer/log_merge_policy.rs
```diff
@@ -76,62 +76,97 @@ impl LogMergePolicy {
...
-    segment.num_deleted_docs() as f32 / segment.max_doc() as f32
 }
 
 impl MergePolicy for LogMergePolicy {
     fn compute_merge_candidates(&self, segments: &[SegmentMeta]) -> Vec<MergeCandidate> {
-        let size_sorted_segments = segments
+        // Filter for segments that have less than the target number of docs, count total unmerged
+        // docs, and sort in descending order
+        let mut unmerged_docs = 0;
+        let mut levels = segments
```
fulmicoton: this vec does not contains "levels" anymore. Why is it named levels?

src/indexer/log_merge_policy.rs
fulmicoton: aaaaaaaaaaaarg the original merge policy has the same flaw!

src/indexer/log_merge_policy.rs
kryesh: Yeah it shouldn't be called that anymore, I'll rename it

src/indexer/log_merge_policy.rs
kryesh: Should that even be prevented? The logic here only runs when there are enough unmerged documents to create a segment above the `target_segment_size` so it will keep adding segments to merge until this op hits `target_segment_size`. This is an optimisation to prevent documents in those small segments from being merged over and over again.

src/indexer/log_merge_policy.rs
kryesh: We don't pop all the segments, this loop only pops segments while there are enough unmerged docs to create a segment above `target_segment_size`. This is the exit condition for if we need to go around again after creating a segment at the target size. Breaking here means that the remaining segments smaller than `target_segment_size` can go through the log merge logic rather than this skip logic.

src/indexer/log_merge_policy.rs
kryesh: Sure, will update

src/indexer/log_merge_policy.rs
kryesh: as above, will update

kryesh:
Have added a commit addressing the feedback @fulmicoton, it looks like there might need to be some clarification on how the logic works though so I'll run through an example input starting from here: https://github.com/quickwit-oss/tantivy/blob/d516fc52672de7844835e66a437ea433e350b5f4/src/indexer/log_merge_policy.rs#L105
```
target_segment_size = 1000
min_num_segments = 2
sorted_segments =  vec![250, 250, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100]
unmerged_docs = 1500
```

First - the condition at https://github.com/quickwit-oss/tantivy/blob/d516fc52672de7844835e66a437ea433e350b5f4/src/indexer/log_merge_policy.rs#L106 will be met so it starts building a batch.

Next - the while loop will pop the smallest segments off the end to make the merge candidate, since it needs 1000 events it will pop the last 10 segments with 100 docs, so on the 10th iteration of the loop we'll be in this state when we hit this check https://github.com/quickwit-oss/tantivy/blob/d516fc52672de7844835e66a437ea433e350b5f4/src/indexer/log_merge_policy.rs#L116:
```
sorted_segments =  vec![250, 250]
unmerged_docs = 1500
batch = vec![100, 100, 100, 100, 100, 100, 100, 100, 100, 100]]
batch_docs = 1000
```

Next - the skip batch is created and a new merge candidate with all the segments in the batch will be added to candidates, unmerged_docs will be reduced by the amount in the batch which will make it 500, causing the check on https://github.com/quickwit-oss/tantivy/blob/d516fc52672de7844835e66a437ea433e350b5f4/src/indexer/log_merge_policy.rs#L125 to break out of the while loop. Then the remaining 2 segments will be merged by the log merge logic.

src/indexer/log_merge_policy.rs
```diff
@@ -76,62 +76,97 @@ impl LogMergePolicy {
...
-fn deletes_ratio(segment: &SegmentMeta) -> f32 {
-    if segment.max_doc() == 0 {
-        return 0f32;
+    fn segment_above_deletes_threshold(&self, segment: &SegmentMeta) -> bool {
+        match segment.max_doc() {
+            0 => false,
+            _ => {
+                (segment.num_deleted_docs() as f32 / segment.max_doc() as f32)
+                    > self.del_docs_ratio_before_merge
+            }
```
fulmicoton: A match statement is overkill.

You can either use a if statement and early return, or remove the division.

segment.num_deleted_docs() as f32 > segment.max_doc() as f32 * self.del_docs_ratio_before_merge

fulmicoton:
@stuhood Can you review? Datadog/Quickwit does not use this merge policy.

I think the PR code is not great, but the existing log merge policy in tantivy is pretty terrible.
We can merge if you are ok to merge.

stuhood:
> @stuhood Can you review? Datadog/Quickwit does not use this merge policy.

Mm, yea: we don't either.

----

It looks like the intent of your change is to avoid repeatedly re-merging very small segments. Unlike an LSM-tree, Tantivy does not have a "memtable", and so there is no minimum segment size, and very small segments might be created. `LogMergePolicy::clip_min_size` can cause these very small segments to be merged repeatedly, because it effectively lies about the actual size of a small segment. If my math is right, due to `clip_min_size`, a segment containing 1 document might be merged 5 times before it reaches the `DEFAULT_MIN_LAYER_SIZE`.

As @fulmicoton pointed out, the effect of your change might be to merge an arbitrary number of segments (1000 etc). But probably more importantly, it might merge segments of drastically different sizes: imagine that deletes are arriving which shrink segments which were _larger_ than your `target_segment_size` to _smaller_ than your `target_segment_size` (i.e. from 1.01GB to 0.99GB for a target of 1GB). Now that much larger segment is eligible to be merged with potentially super tiny segments: if some brand new segments arrive which are only 10MB, they might end up being compacted with your (now) 0.99GB segment. And this might happen repeatedly.

This is why the log scale exists: it causes extra [write amplification](https://www.scylladb.com/2018/01/31/compaction-series-leveled-compaction/) to merge tiny segments with larger segments unnecessarily. A `target_segment_size` which is too high, _or_ the `clip_min_size` behavior can cause lots of redundant merging.

I like the general thrust of this PR, but I think that what I would propose would be to:
* remove the `min_layer_size` and the `clip_min_size` behavior
  * as discussed above: they lead to write amplification
* significantly lower your new `target_segment_size` (to e.g. 100 or 500 documents), and rename it to `target_min_segment_size` to replace the previous intent of the `min_layer_size` settings
  * this value would act like "the memtable size" in an LSM tree: it would be the initial value in the lowest level of the log-scale behavior
  * when this value is lower, the consequences of the case I described above (deletes causing a segment to wobble back and forth around the target) are less severe
* remove the `max_docs_before_merge` setting, and instead allow the log scale to be applied to all segments which are larger than `target_min_segment_size`
  * AFAICT, the `max_docs_before_merge` setting used to represent a ceiling on which segments would be merged. but I don't think that this is a good idea, for a few reasons:
    1. term dictionaries must be consulted and decompressed in every segment, which has overhead. doing this fewer times is preferable
    2. metadata has a cost: having thousands of tiny segments increases the cost of storing and loading metadata

stuhood:
> remove the `min_layer_size` and the `clip_min_size` behavior ... significantly lower your new `target_segment_size` (to e.g. 100 or 500 documents), and rename it to `target_min_segment_size` to replace the previous intent of the `min_layer_size` settings 

...or, alternatively: don't add `target_segment(_min)_size`, and instead lower the `min_layer_size` to 100 or 500 docs to make the write amplification there less severe. Would still recommend removing `max_docs_before_merge`.

kryesh:
The main issue I was encountered wasn't the very small segments, but repeated merges of intermediate/large segments which take a lot longer and block other operations - but yes the intent was to prevent repeated merges for the same documents.

Regarding deletes and having a segment flapping between just above and just below the target size, would changing the default delete ratio to something like 30% then updating the filter logic to look at a combination of max_doc and the delete ratio for the segment make more sense? Possibly merge segments with either of this criteria instead of num_docs below the target?:
- max_doc below the target
- max_doc above the target + hit the delete ratio threshold

The min size/clipping behaviour is a holdover from the previous policy, I think it'd be worth testing with it removed and see how it reduces write amplification. I can look into this.

Removing the segment limit/target is nice in theory like you mention with the term dictionaries etc - but there are practical issues that prevent that from working such as the 32 bit doc address value imposing a hard limit, #2577 is one example of this. There needs to be some form of maximum size enforced since not having a limit leads to other problems with huge merge operations which was another goal of this PR. If anything making the default target size bigger to approach the practical limit might make more sense, 200m or so?

stuhood:
> Removing the segment limit/target is nice in theory like you mention with the term dictionaries etc - but there are practical issues that prevent that from working such as the 32 bit doc address value imposing a hard limit, #2577 is one example of this. There needs to be some form of maximum size enforced since not having a limit leads to other problems with huge merge operations which was another goal of this PR.

Yea, true. But `const DEFAULT_MAX_DOCS_BEFORE_MERGE: usize = 10_000_000;` is overkill for accomplishing that. Could massively increase it.

kryesh:
@stuhood  @fulmicoton Pushed some more updates to address the feedback so far.


>As @fulmicoton pointed out, the effect of your change might be to merge an arbitrary number of segments (1000 etc). But probably more importantly, it might merge segments of drastically different sizes: imagine that deletes are arriving which shrink segments which were larger than your target_segment_size to smaller than your target_segment_size (i.e. from 1.01GB to 0.99GB for a target of 1GB). Now that much larger segment is eligible to be merged with potentially super tiny segments: if some brand new segments arrive which are only 10MB, they might end up being compacted with your (now) 0.99GB segment. And this might happen repeatedly.

To address this I updated the filter logic at the start to be a bit smarter than naively using `num_docs`, with this change borderline segments that have some documents deleted won't immediately trigger a merge until a certain ratio is reached. I also updated the default delete ratio - but I'm not sure what the "backward compatibility" comment on the current version is referring to.

Another leftover behaviour from the current policy is this chunk which forces a merge if any segment has hit the delete threshold.
```rust
let mut hit_delete_threshold = false;
for (_, segment) in group {
    batch.push(segment.id());
    hit_delete_threshold |= self.segment_above_deletes_threshold(segment);
}

if batch.len() >= self.min_num_segments || hit_delete_threshold {
```
should this still be a thing? or just use `num_docs` like all the other segments and remove this check? Not sure if there's a historic reason for this or if deletes are really bad for performance or something so figured I'd ask.

fulmicoton:
Not the scope of this PR, but here is my thoughts on what a simple rewrite of the log merge policy could look like.
https://github.com/quickwit-oss/tantivy/issues/2794

stuhood (commented):
Changing the `DEFAULT_MIN_LAYER_SIZE` gets this much closer I think.

My primary concern though is that it adds a _lot_ of code that effectively diverges from the name of the policy: about 100 lines dedicated to something that sidesteps the "Log" part of LogMerge.

> The main issue I was encountered wasn't the very small segments, but repeated merges of intermediate/large segments which take a lot longer and block other operations - but yes the intent was to prevent repeated merges for the same documents.

@kryesh : If that's the case, and it's actually medium/large segments that you're worried about, then perhaps what you want to do is significantly increase the `min_num_segments` value (e.g. to 32), so that merges happen much less frequently (and more segments are alive at a time)?

With the new `DEFAULT_TARGET_SEGMENT_SIZE = 200_000_000` value that you've set, my concern is that your "target" size is now just a strange sort of "alternate level" which isn't actually computed from the log scale. It is _not_ doing what we discussed in https://github.com/quickwit-oss/tantivy/pull/2627#issuecomment-3708798652 (acting as a cap on max segment size): that would be something else.

kryesh:
Fair, but these options are exposed to users anyway so in my case I'd just use my current target setting of ~33m which I've found works well. The value of 200m means there is an effective hard cap of 400m (399_999_998) docs per segment.
This PR was always adding an alternate level that bypasses the log scale since the current log algorithm isn't capable of enforcing a maximum size.

PR #2632, opened 2025-04-27
refactoring

adamnemecek:


bitpacker/benches/bench.rs
```diff
@@ -12,7 +12,7 @@ mod tests {
     #[inline(never)]
     fn create_bitpacked_data(bit_width: u8, num_els: u32) -> Vec<u8> {
         let mut bitpacker = BitPacker::new();
-        let mut buffer = Vec::new();
+        let mut buffer = vec![];
```
PSeitz, line 15: Vec::new() is preferred for empty Vecs

Issue #2628, opened 2025-04-18
iOS compilation failure due to zstd


Parth:
I'm working on [lockbook, a cross platform, secure note taking app](https://github.com/lockbook/lockbook). We recently found our build broken while updating our rust version and traced this behavior to zstd. This failure is described here: https://github.com/gyscos/zstd-rs/issues/337

I do see, however that zstd is an optional feature here 

https://github.com/quickwit-oss/tantivy/blob/5379c99ea2a4639b659f640d9fd2c43030929161/Cargo.toml#L31

but is used normally here https://github.com/quickwit-oss/tantivy/blob/5379c99ea2a4639b659f640d9fd2c43030929161/sstable/Cargo.toml#L19

Is it the intention for zstd to be truly optional? If so, may I send a PR to make `sstable` reflect this? This would allow us to remove zstd from our dependency tree and proceed with our upgrade.

Or is it the case that this feature is truly not optional and does need to be solved with zstd itself?

Relevant `cargo tree` output copied from the zstd issue:

```
zstd v0.13.3
└── tantivy-sstable v0.3.0
    └── tantivy-columnar v0.3.0
        └── tantivy v0.22.0
            └── lb-rs v0.9.22 (/Users/parth/Documents/lockbook/lockbook/libs/lb/lb-rs)
                ├── lb-c v0.9.22 (/Users/parth/Documents/lockbook/lockbook/libs/lb/lb-c)
                │   └── workspace-ffi v0.9.22 (/Users/parth/Documents/lockbook/lockbook/libs/content/workspace-ffi)
                └── workspace v0.9.22 (/Users/parth/Documents/lockbook/lockbook/libs/content/workspace)
                    └── workspace-ffi v0.9.22 (/Users/parth/Documents/lockbook/lockbook/libs/content/workspace-ffi)
```

PSeitz:
I think zstd could be optional in `sstable` and should be straightforward to add in a PR. cc @trinity-1686a

trinity-1686a:
sstable's codec is written in such a way that compression is already optional (enabled/disabled based on if it's helping or not with size). It should be rather easy to disable zstd entirely from sstables. Obviously sstables written with zstd enabled would not always be readable by a software built without zstd, but a software with zstd support would be able to read both and wouldn't notice a thing.

[...]

PR #2633, opened 2025-05-03
make zstd optional in sstable

Parth:
- zstd-compression is optional in tantivy, it was (presumably) inadvertently mandatory in `sstable`. See more info here:
  - https://github.com/quickwit-oss/tantivy/issues/2628
  - https://github.com/gyscos/zstd-rs/issues/337
- This PR makes columnar-zstd-compression optional, but enabled by default due to non-rust components of it's build process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,6 @@ Tantivy 0.23 - Unreleased
 ================================
 Tantivy 0.23 will be backwards compatible with indices created with v0.22 and v0.21. The new minimum rust version will be 1.75.
 
-> [!WARNING]
-> Users of `sstable` may need to enable the `zstd-compression` to read blocks written by prior versions of `sstable`
-
 #### Bugfixes
 - fix potential endless loop in merge [#2457](https://github.com/quickwit-oss/tantivy/pull/2457)(@PSeitz)
 - fix bug that causes out-of-order sstable key. [#2445](https://github.com/quickwit-oss/tantivy/pull/2445)(@fulmicoton)
@@ -15,7 +12,6 @@ Tantivy 0.23 will be backwards compatible with indices created with v0.22 and v0
 
 #### Breaking API Changes
 - remove index sorting [#2434](https://github.com/quickwit-oss/tantivy/pull/2434)(@PSeitz)
-- make zstd optional in sstable [#2633](https://github.com/quickwit-oss/tantivy/pull/2633)(@parth)
 
 #### Features/Improvements
 - **Aggregation**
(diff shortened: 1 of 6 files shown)


PSeitz:
I would prefer a separate feature flag `columnar-zstd-compression` with some comments in Cargo.toml

Parth:
@PSeitz Do you mean `sstable-zstd-compression`? 

Also do you want `zstd-compression` at the tantivy / columnar level to enable this automatically? 

I'll fix the `fmt` as well

Parth:
I think I see what you wanted so I pushed a commit. Happy to iterate if that's not exactly right just lmk. Also pushed a `cargo +nightly fmt`.

PSeitz:
> I think I see what you wanted so I pushed a commit. Happy to iterate if that's not exactly right just lmk. Also pushed a `cargo +nightly fmt`.

Yes, for a tantivy user the sstable is more like an implementation detail, unlike the columnar storage.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@ Tantivy 0.23 - Unreleased
 ================================
 Tantivy 0.23 will be backwards compatible with indices created with v0.22 and v0.21. The new minimum rust version will be 1.75.
 
+> [!WARNING]
```
PSeitz: that's already outdated with the last release

https://github.com/quickwit-oss/tantivy/pull/2621/files

Parth: I'm not sure I understand: I didn't update the lines linked in that pr so even if we do 0.24 I'm not sure I see the problem. 

I could rebase but I don't think that's what you mean because that pr isn't merged.

PSeitz: 0.24 is already released, but the CHANGELOG pr is not yet merged

columnar/Cargo.toml
```diff
@@ -33,6 +33,6 @@ harness = false
 name = "bench_access"
 harness = false
 
-
 [features]
 unstable = []
+zstd-compression = ["sstable/zstd-compression"]
```
PSeitz, line 38: that should be a default feature flag

Parth: Are you saying I should make it a default flag and then eliminate the changelog entry so we can merge this PR?

PSeitz: yes

Parth: understood, pushed

Parth:
@trinity-1686a / @PSeitz lmk if there's anything else I can do to bring this closer to being merged and ultimately released

PSeitz-dd (approved):

PSeitz:
LGTM, Thanks for the PR!

PR #2641, opened 2025-05-29
Add concurrency limiting to Executor to prevent thread pool exhaustion

Barre:
**Problem**

When using the multithreaded executor with Tantivy, a single search query can exhaust all threads in the pool, preventing other queries from running concurrently. This happens because `Executor::map()` uses `rayon::scope()` which spawns all tasks at once, potentially using all available threads for a single operation.

This causes issues in production scenarios where:

- A slow/expensive query blocks fast queries from executing
- Multiple concurrent searches cannot effectively share the thread pool
- The only workaround is creating multiple `IndexReader` instances, which is inefficient

**Solution**

This PR adds an optional concurrency limit to the `Executor` that restricts how many tasks a single `map()` operation can spawn concurrently. This ensures that no single query can monopolize the entire thread pool.

### DIFF
--- a/src/core/executor.rs
+++ b/src/core/executor.rs
@@ -1,9 +1,43 @@
-use std::sync::Arc;
+use crate::TantivyError;
+use std::sync::{Arc, Condvar, Mutex};
 
 #[cfg(feature = "quickwit")]
 use futures_util::{future::Either, FutureExt};
 
-use crate::TantivyError;
+struct TaskLimiter {
+    available: Mutex<usize>,
+    condvar: Condvar,
+}
+
+impl TaskLimiter {
+    fn new(max_tasks: usize) -> Self {
+        TaskLimiter {
+            available: Mutex::new(max_tasks),
+            condvar: Condvar::new(),
+        }
+    }
+
+    fn acquire(&self) -> TaskPermit<'_> {
+        let mut available = self.available.lock().unwrap();
+        while *available == 0 {
+            available = self.condvar.wait(available).unwrap();
+        }
+        *available -= 1;
+        TaskPermit { limiter: self }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


src/core/executor.rs
```diff
@@ -34,33 +73,80 @@ impl Executor {
...
+                    Some(limit) => {
+                        if *limit == 0 {
+                            return Err(TantivyError::InternalError(
+                                "Concurrency limit cannot be zero.".to_string(),
+                            ));
+                        }
+                        Some(Arc::new(TaskLimiter::new(*limit)))
+                    }
+                    _ => None,
+                };
```
ChillFish8, line 131: The error for creating the threadpool with a concurrency limit should not be deferred until it is used.

src/core/executor.rs
```diff
@@ -34,33 +73,80 @@ impl Executor {
...
+                            let limiter = limiter.clone();
+
                             // We name references for f and fruit_sender_ref because we do not
                             // want these two to be moved into the closure.
                             let f_ref = &f;
                             let fruit_sender_ref = &fruit_sender;
                             scope.spawn(move |_| {
+                                if let Some(limiter) = limiter {
+                                    let _permit = limiter.acquire();
+                                }
```
ChillFish8, line 148: This doesn't prevent the task from causing issues with other searchers, the semaphore will park the thread it is being executed on, so if you have a pool with `8` threads, and a concurrency limit `2` and do a search with `8` segments to map, one search will lock out all others until complete, this effectively only limits the maximum number of CPU cores the index reader can max out globally, rather than allow multiple searchers to share the pool without standing on each others toes.

Also the actual limiting bit doesn't work because the permit is dropped as soon as it leaves the `if` block.

PR #2642, opened 2025-06-02
Add string fast field support to `TopDocs`.

stuhood:
Creating as a draft for discussion, as there is a fair chance that I've missed an existing way to do this.

### DIFF
--- a/src/collector/top_score_collector.rs
+++ b/src/collector/top_score_collector.rs
@@ -2,7 +2,7 @@ use std::fmt;
 use std::marker::PhantomData;
 use std::sync::Arc;
 
-use columnar::ColumnValues;
+use columnar::{ColumnValues, StrColumn};
 use serde::{Deserialize, Serialize};
 
 use super::Collector;
@@ -14,6 +14,7 @@ use crate::collector::{
 };
 use crate::fastfield::{FastFieldNotAvailableError, FastValue};
 use crate::query::Weight;
+use crate::termdict::TermOrdinal;
 use crate::{DocAddress, DocId, Order, Score, SegmentOrdinal, SegmentReader, TantivyError};
 
 struct FastFieldConvertCollector<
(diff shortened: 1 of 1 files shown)


src/collector/top_score_collector.rs
```diff
@@ -83,6 +84,175 @@ where
...
+                },
+            )
+        } else {
+            self.ff.dictionary().sorted_ords_to_term_cb(
+                fruit.iter().rev().map(|(term_ord, _)| *term_ord),
+                |term| {
+                    terms.push(
+                        std::str::from_utf8(term)
+                            .expect("Failed to decode term as unicode")
+                            .to_owned(),
```
fulmicoton-dd, line 216: nitpick. same thing but more explicit.
```suggestion
                            .to_string(),
```

src/collector/top_score_collector.rs
```diff
@@ -83,6 +84,175 @@ where
...
+        // Collect terms.
+        let mut terms = Vec::with_capacity(fruit.len());
+        let result = if self.order.is_asc() {
+            self.ff.dictionary().sorted_ords_to_term_cb(
+                fruit.iter().map(|(term_ord, _)| u64::MAX - term_ord),
+                |term| {
+                    terms.push(
+                        std::str::from_utf8(term)
+                            .expect("Failed to decode term as unicode")
+                            .to_owned(),
```
fulmicoton-dd, line 204:
```suggestion
                            .to_string(),
```

src/collector/top_score_collector.rs
```diff
@@ -83,6 +84,175 @@ where
...
+
+    fn collect(&mut self, doc: DocId, score: Score) {
+        self.collector.collect(doc, score);
+    }
+
+    fn harvest(self) -> Vec<(String, DocAddress)> {
+        let fruit = self.collector.harvest();
+
+        // Collect terms.
+        let mut terms = Vec::with_capacity(fruit.len());
```
fulmicoton-dd:
```suggestion
        let mut terms: Vec<String> = Vec::with_capacity(fruit.len());
```

src/collector/top_score_collector.rs
```diff
@@ -83,6 +84,175 @@ where
...
+where TCollector: SegmentCollector<Fruit = Vec<(TermOrdinal, DocAddress)>> + 'static
+{
+    type Fruit = Vec<(String, DocAddress)>;
+
+    fn collect(&mut self, doc: DocId, score: Score) {
+        self.collector.collect(doc, score);
+    }
+
+    fn harvest(self) -> Vec<(String, DocAddress)> {
+        let fruit = self.collector.harvest();
```
fulmicoton-dd:
```suggestion
        let fruit: Vec<(TermOrdinal, DocAddress)> = self.collector.harvest();
```

Also, I think renaming fruit into something more explicit: top_ordinals or something like that

src/collector/top_score_collector.rs
```diff
@@ -83,6 +84,175 @@ where
...
+    ) -> crate::Result<Self::Fruit> {
+        if self.limit == 0 {
+            return Ok(Vec::new());
+        }
+        if self.order.is_desc() {
+            let mut top_collector: TopNComputer<_, _, true> =
+                TopNComputer::new(self.limit + self.offset);
+            for child_fruit in child_fruits {
+                for (feature, doc) in child_fruit {
+                    top_collector.push(feature, doc);
```
fulmicoton-dd, line 151: We could probably do a tiny better by make the conversion to str lazy.
The benefit would be to make sure we only decode `K + num_segments - 1`, instead of `num_segments x K`

fulmicoton: No need to change that in this PR. Can you open an issue however?

fulmicoton:
@stuhood Can you have a look at the comments? I think this is more or less good to merge

src/collector/top_score_collector.rs
stuhood: I find that `to_owned` is more explicit.

The reason is that `to_string` is provided by `Display`, which is implemented by many, many things: you might accidentally call it on a `u32`, for example (and that could happen silently during a refactoring). On the other hand, `to_owned` is much more narrowly scoped, and operates on very specific pairs of types.

src/collector/top_score_collector.rs
stuhood: https://github.com/quickwit-oss/tantivy/issues/2652

stuhood (commented):
Thanks! See also #2651, which relates to this issue, because this issue attempts to use `TopNComputer` in ascending order.

src/collector/top_score_collector.rs
fulmicoton: Well, let's agree to disagree. Here to be honest it is non-consequential so I don't care at all.

src/collector/top_score_collector.rs
```diff
@@ -1214,6 +1398,73 @@ mod tests {
...
+        assert_eq!(
+            &query(&index, Order::Desc, 3, 0)?,
+            &[
+                ("tokyo".to_owned(), DocAddress::new(0, 2)),
+                ("georgetown".to_owned(), DocAddress::new(0, 1)),
+                ("alberta".to_owned(), DocAddress::new(0, 0)),
+            ]
+        );
+
+        assert_eq!(
```
fulmicoton, line 1451: please add a test with offset 0, and limit 2.

fulmicoton: and offset 3 limit 3

fulmicoton (approved):

fulmicoton (approved):

PR #2646, opened 2025-06-07
Reduce sorting in TopDocs

stuhood:
Reduce sorting in `TopDocs` by not sorting individual segments -- they will be merged and top-n'd anyway.

```
full
top_docs_small_shallow    Memory: 97.3 KB (-0.58%)     Avg: 4.5728ms (-2.23%)      Median: 4.5707ms (-2.05%)      [4.5540ms .. 4.6014ms]
top_docs_small_deep       Memory: 6.2 MB (-0.01%)      Avg: 14.2786ms (-33.59%)    Median: 14.2741ms (-33.70%)    [14.1030ms .. 14.5956ms]
top_docs_large_shallow    Memory: 698.7 KB (-0.05%)    Avg: 6.8426ms (-8.48%)      Median: 6.8434ms (-8.26%)      [6.7612ms .. 6.9138ms]
top_docs_large_deep       Memory: 6.8 MB (-0.01%)      Avg: 14.2691ms (-37.26%)    Median: 14.2846ms (-37.28%)    [14.0335ms .. 14.3867ms]
dense
top_docs_small_shallow    Memory: 92.9 KB (-0.35%)    Avg: 4.6792ms (-2.13%)      Median: 4.6785ms (-2.01%)      [4.6559ms .. 4.7226ms]
top_docs_small_deep       Memory: 5.9 MB              Avg: 15.0375ms (-29.50%)    Median: 15.0420ms (-29.46%)    [14.8413ms .. 15.1579ms]
top_docs_large_shallow    Memory: 662.9 KB            Avg: 7.0038ms (-6.82%)      Median: 6.9954ms (-6.79%)      [6.9232ms .. 7.0746ms]
top_docs_large_deep       Memory: 6.4 MB              Avg: 15.2383ms (-32.39%)    Median: 15.2466ms (-32.32%)    [15.1046ms .. 15.3693ms]
sparse
top_docs_small_shallow    Memory: 40.7 KB (+0.59%)    Avg: 17.1796ms (+1.15%)    Median: 17.1670ms (+1.01%)    [17.1109ms .. 17.3223ms]
top_docs_small_deep       Memory: 2.9 MB              Avg: 22.8191ms (-6.98%)    Median: 22.8086ms (-6.93%)    [22.6527ms .. 23.3086ms]
top_docs_large_shallow    Memory: 324.8 KB            Avg: 18.6777ms (-0.57%)    Median: 18.6723ms (-0.64%)    [18.5654ms .. 18.8087ms]
[...]

PR #2651, opened 2025-06-17
Fixes for `TopDocs::order_by_string_fast_field` and `TopNComputer`

stuhood:
#2642 added `TopDocs::order_by_string_fast_field`, but the test coverage in that PR did not catch two bugs:

1.  The `TopNComputer` was refactored in https://github.com/quickwit-oss/tantivy/pull/2198 to support both ascending and descending order. But that refactor did not adjust the `TopNComputer::threshold` to include the `ComparableDoc`, which is what provides the relevant `Ord` implementation.
    * The effect of this was that the threshold comparison for `TopNComputer` was still hardcoded for descending order. Unit tests which tested with fewer than `limit * 2` items OR which didn't have an item less than the threshold arrive _after_ the first `limit * 2` items (when the threshold is computed for the first time) would not expose the bug.
2. `sorted_ords_to_term_cb` only supported unique `ords`, likely because its previous consumers in aggregates would always de-dupe ords before lookup.
    * It is much easier for consumers who would like to look up non-unique term ordinals to allow them to be duplicated (as it introduces only two comparisons in a path that is always IO bound).

### DIFF
--- a/src/collector/top_score_collector.rs
+++ b/src/collector/top_score_collector.rs
@@ -1061,7 +1061,7 @@ mod tests {
 
     use super::{TopDocs, TopNComputer};
     use crate::collector::top_collector::ComparableDoc;
-    use crate::collector::Collector;
+    use crate::collector::{Collector, DocSetCollector};
     use crate::query::{AllQuery, Query, QueryParser};
     use crate::schema::{Field, Schema, FAST, STORED, TEXT};
     use crate::time::format_description::well_known::Rfc3339;
@@ -1559,6 +1559,72 @@ mod tests {
         assert_eq!(&query(&index, Order::Asc, 3, 3)?, &[]);
 
         Ok(())
+    }
+
+    proptest! {
+        #[test]
+        fn test_top_field_collect_string_prop(
+          order in prop_oneof!(Just(Order::Desc), Just(Order::Asc)),
+          limit in 1..256_usize,
+          offset in 0..256_usize,
+          segments_terms in
+            proptest::collection::vec(
+                proptest::collection::vec(0..32_u8, 1..32_usize),
+                0..8_usize,
+            )
+        ) {
+            let mut schema_builder = Schema::builder();
+            let city = schema_builder.add_text_field("city", TEXT | FAST);
+            let schema = schema_builder.build();
+            let index = Index::create_in_ram(schema);
+            let mut index_writer = index.writer_for_tests()?;
+
+            // A Vec<Vec<u8>>, where the outer Vec represents segments, and the inner Vec
+            // represents terms.
+            for segment_terms in segments_terms.into_iter() {
+                for term in segment_terms.into_iter() {
+                    let term = format!("{term:0>3}");
+                    index_writer.add_document(doc!(
+                        city => term,
+                    ))?;
+                }
+                index_writer.commit()?;
+            }
+
+            let searcher = index.reader()?.searcher();
+            let top_n_results = searcher.search(&AllQuery, &TopDocs::with_limit(limit)
+                .and_offset(offset)
+                .order_by_string_fast_field("city", order.clone()))?;
+            let all_results = searcher.search(&AllQuery, &DocSetCollector)?.into_iter().map(|doc_address| {
+                // Get the term for this address.
+                // NOTE: We can't determine the SegmentIds that will be generated for Segments
+                // ahead of time, so we can't pre-compute the expected `DocAddress`es.
+                let column = searcher.segment_readers()[doc_address.segment_ord as usize].fast_fields().str("city").unwrap().unwrap();
+                let term_ord = column.term_ords(doc_address.doc_id).next().unwrap();
+                let mut city = Vec::new();
+                column.dictionary().ord_to_term(term_ord, &mut city).unwrap();
+                (String::try_from(city).unwrap(), doc_address)
+            });
+
+            // Using the TopDocs collector should always be equivalent to sorting, skipping the
+            // offset, and then taking the limit.
+            let sorted_docs: Vec<_> = if order.is_desc() {
+                let mut comparable_docs: Vec<ComparableDoc<_, _, true>> =
+                    all_results.into_iter().map(|(feature, doc)| ComparableDoc { feature, doc}).collect();
+                comparable_docs.sort();
+                comparable_docs.into_iter().map(|cd| (cd.feature, cd.doc)).collect()
+            } else {
+                let mut comparable_docs: Vec<ComparableDoc<_, _, false>> =
+                    all_results.into_iter().map(|(feature, doc)| ComparableDoc { feature, doc}).collect();
+                comparable_docs.sort();
+                comparable_docs.into_iter().map(|cd| (cd.feature, cd.doc)).collect()
+            };
+            let expected_docs = sorted_docs.into_iter().skip(offset).take(limit).collect::<Vec<_>>();
+            prop_assert_eq!(
+                expected_docs,
+                top_n_results
+            );
+        }
     }
 
     #[test]
--- a/src/collector/top_score_collector.rs
+++ b/src/collector/top_score_collector.rs
@@ -1057,6 +1057,8 @@ where
 
 #[cfg(test)]
 mod tests {
+    use proptest::prelude::*;
+
     use super::{TopDocs, TopNComputer};
     use crate::collector::top_collector::ComparableDoc;
     use crate::collector::Collector;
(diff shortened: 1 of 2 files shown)


src/collector/top_score_collector.rs
```diff
@@ -806,14 +816,14 @@ where
     /// If the document is below the current threshold, it will be ignored.
     #[inline]
     pub fn push(&mut self, feature: Score, doc: D) {
-        if let Some(last_median) = self.threshold.clone() {
-            if feature < last_median {
+        let comparable_doc = ComparableDoc { doc, feature };
+        if let Some(last_median) = self.threshold.as_ref() {
+            if &comparable_doc > last_median {
```
stuhood, line 1005: Calling this out in particular, as we discussed it downstream:

The reason this swapped is that `TopNComputer` used to be hardcoded for descending order: when support was added for configuring the order via `ComparableDoc`, this codepath wasn't updated to use that wrapper's `Ord` implementation, and so it continued to only support descending order on the feature.

fulmicoton-dd:
@stuhood I don't understand what this is about. Can you improve the description? In particular, the title suggests this is a bugfix... Is this really a bugfix? It then points to commits that are not in tantivy.

stuhood:
> @stuhood I don't understand what this is about. Can you improve the description? In particular, the title suggests this is a bugfix... Is this really a bugfix? It then points to commits that are not in tantivy.

Yes, it is a bugfix. Have improved the description.

I don't quite understand why that commit isn't rendering as part of Tantivy, but: it is https://github.com/quickwit-oss/tantivy/pull/2198 ... which claims to have been merged as https://github.com/quickwit-oss/tantivy/commit/dd7e745392904a3fc236e5508c7d2374e1010732 ... which the github UI claims is not in Tantivy? Strange.

fulmicoton:
@PSeitz can you review?

src/collector/top_score_collector.rs
```diff
@@ -915,48 +927,41 @@ mod tests {
         );
     }
 
-    #[test]
-    fn test_empty_topn_computer() {
```
PSeitz-dd: don't remove tests, the proptest is no replacement for simple test cases

src/collector/top_score_collector.rs
```diff
@@ -668,19 +668,26 @@ impl Collector for TopDocs {
 
         if let Some(alive_bitset) = reader.alive_bitset() {
             let mut threshold = Score::MIN;
-            top_n.threshold = Some(threshold);
             weight.for_each_pruning(Score::MIN, reader, &mut |doc, score| {
                 if alive_bitset.is_deleted(doc) {
                     return threshold;
                 }
                 top_n.push(score, doc);
-                threshold = top_n.threshold.unwrap_or(Score::MIN);
```
PSeitz-dd, line 861: I think`TopDocs` used as `Collector` never supported desc ordering as there is not an API for that. 
It's a little bit weird, since the constructors on `TopDocs` don't always return `TopDocs`.

Also in this PR treshold is still set to `Score::MIN` initially.

src/collector/top_score_collector.rs
```diff
@@ -915,48 +927,41 @@ mod tests {
...
-                }
-            ]
-        );
-    }
-
-    #[test]
-    fn test_topn_computer_no_panic() {
-        for top_n in 0..10 {
-            let mut computer: TopNComputer<u32, u32> = TopNComputer::new(top_n);
+    proptest! {
```
PSeitz-dd, line 1184: Can you add a simple test case for the reverse ordering bugfix?

PSeitz-dd:
Can you add a high level test, that would have failed?

PSeitz:
I fixed the failing CI

src/collector/top_score_collector.rs
stuhood: That was added in the first commit, but I replaced it with the proptests because they provide more coverage: https://github.com/quickwit-oss/tantivy/pull/2651/commits/cfe04ab6ac9dedfa45ebe45d384940603a0a8e15

It sounds like you would like to have both though?

PSeitz-dd: Yes, in my experience proptests are great, but not a replacement for regular tests. 
* They don't document border cases well and how they are supposed to work
* When refactoring they don't show as well as a simple test cases what exactly breaks
* Since they are very generic, they are easy to get wrong and miss something

src/collector/top_score_collector.rs
stuhood: > I think`TopDocs` used as `Collector` never supported desc ordering as there is not an API for that. It's a little bit weird, since the constructors on `TopDocs` don't always return `TopDocs`.

Yea. I think that it was always accomplished for fast field ordering by negating (`-`) the value when scoring, and then negating it again when collecting.

But that doesn't work for strings, which is where I first encountered it: #2642.

> Also in this PR threshold is still set to `Score::MIN` initially.

Right... so I think that this line is correct. Or are you saying otherwise?

PSeitz-dd: Not sure we are talking about the same thing. `TopDocs` is a `Collector`, but also an entry point to construct other collectors.
When used as an collector I don't think it supports reverse ordering. Can you show otherwise, e.g. a test that would fail without that change

PSeitz-dd: Maybe `TopDocs` should not be an entry point to create other collectors, these could be instantiated on their own.

stuhood: > Not sure we are talking about the same thing. `TopDocs` is a `Collector`, but also an entry point to construct other collectors.
> When used as an collector I don't think it supports reverse ordering. Can you show otherwise, e.g. a test that would fail without that change

`order_by_u64_field` and `order_by_fast_field` support specifying an `Order`, but as you said: they construct separate `Collector` instances.

Those `Collector` instances negate their values to get the inverted ordering: for example, `FastFieldConvertCollector`: https://github.com/quickwit-oss/tantivy/blob/988c2b35e7f44c08dfc620b3926661931e54614e/src/collector/top_score_collector.rs#L77-L83

The new `StringConvertCollector` added in #2642 could not (trivially) invert the sort order for strings (would have needed a wrapper type and more boilerplate), so it uses `TopNComputer`: https://github.com/quickwit-oss/tantivy/blob/988c2b35e7f44c08dfc620b3926661931e54614e/src/collector/top_score_collector.rs#L146-L175

I can add a test that shows `StringConvertCollector` failing to produce sorted strings without this PR, but it feels like the test on `TopNComputer` is sufficient?

> Maybe `TopDocs` should not be an entry point to create other collectors, these could be instantiated on their own.

Possibly... but having the `TopDocs::order_by_*` methods namespaced is a fairly nice API for consumers... needing to know the names of the various underlying `Collector` impls would be more challenging IMO, even if they were all renamed.

PSeitz: This comment thread is about the change in the `impl Collector for TopDocs`. Can you show a test that would fail without this change, because I don't see a way `impl Collector for TopDocs` would be instantiated with reverse sorting.

> Possibly... but having the TopDocs::order_by_* methods namespaced is a fairly nice API for consumers... needing to know the names of the various underlying Collector impls would be more challenging IMO, even if they were all renamed.

Namespacing in a module should also be nice to use, and make things less confusing.

stuhood: > This comment thread is about the change in the `impl Collector for TopDocs`. Can you show a test that would fail without this change, because I don't see a way `impl Collector for TopDocs` would be instantiated with reverse sorting.

`StringConvertCollector` uses `TopNComputer` in ASC order. Have pushed a new proptest to validate that `StringConvertCollector` is correct after these changes.

stuhood:
Have rebased and pushed two more commits:

* Allowing `sorted_ords_to_term_cb` to handle duplicate (but still sorted) ords. We were using this patch downstream to allow for string sorting, but it hadn't made its way upstream, and caused the `TopDocs::order_by_string_fast_field` method added in #2642 to be broken for duplicate matches.
* Adding a proptest for `order_by_string_fast_field`, which caught the above issue, as well as reproducing the issue with `TopNComputer`.

src/collector/top_score_collector.rs
PSeitz: This thread is about the change in **`impl Collector for TopDocs`**, not the fix in `TopNComputer` or `StringConvertCollector`. Neither uses the `impl Collector for TopDocs`.

`impl Collector for TopDocs` can't be instantiated with reverse ordering, so why is there a change for it.

Even conceptually I don't think `TopDocs` supports reverse ordering due to how WAND is implemented.

Please have a look at the code where the comment is and what you changed there.

stuhood: > `impl Collector for TopDocs` can't be instantiated with reverse ordering, so why is there a change for it.
> 
> Even conceptually I don't think `TopDocs` supports reverse ordering due to how WAND is implemented.
> 
> Please have a look at the code where the comment is and what you changed there.

This line of code changed because the type of the `threshold` field changed in order to make `TopNComputer` correct for its current public interface.

PSeitz: Ah okay, so that's not impacted by the bug

sstable/src/dictionary.rs
```diff
@@ -515,16 +515,25 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
...
-                current_block_addr = new_block_addr;
-                current_ordinal = current_block_addr.first_ordinal;
-                current_sstable_delta_reader =
-                    self.sstable_delta_reader_block(current_block_addr.clone())?;
-                bytes.clear();
+            // only advance forward if the new ord is different than the one we just processed
+            //
+            // this allows the input TermOrdinal iterator to contain duplicates, so long as it's
+            // still sorted
+            if Some(ord) != prev_ord {
```
PSeitz, line 524: This changes the logic and may cause some unexpected surprises

I think it would be better for the caller to filter duplicates in the incoming iterator, e.g. via [dedup](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.dedup)

src/collector/top_score_collector.rs
```diff
@@ -913,7 +920,7 @@ pub struct TopNComputer<Score, D, const REVERSE_ORDER: bool = true> {
     /// The buffer reverses sort order to get top-semantics instead of bottom-semantics
     buffer: Vec<ComparableDoc<Score, D, REVERSE_ORDER>>,
     top_n: usize,
-    pub(crate) threshold: Option<Score>,
+    pub(crate) threshold: Option<ComparableDoc<Score, D, REVERSE_ORDER>>,
```
PSeitz, line 923: `TopNComputer` is quite performance sensitive. I think this changes the memory layout of the struct. Did you test the performance of the change?

It may cause a performance regression in `TopDocs`.

sstable/src/dictionary.rs
stuhood: > This changes the logic and may cause some unexpected surprises

Duplicates were not supported before (you'd get an assertion failure), so it should only make a case work that didn't before? I can add additional tests if that would help. 

> I think it would be better for the caller to filter duplicates in the incoming iterator, e.g. via dedup

That makes it more difficult for consumers to then zip the results back to a collection: they need a mapping table or `HashMap` to look up offsets, rather than being able to literally `zip` them, as we do here: https://github.com/quickwit-oss/tantivy/blob/988c2b35e7f44c08dfc620b3926661931e54614e/src/collector/top_score_collector.rs#L230-L232

src/collector/top_score_collector.rs
stuhood: I've added some benchmarks in #2646: are you able to help me get that one landed?

sstable/src/dictionary.rs
PSeitz: Which assertion? So this one is not strict enough? `assert!(ord >= current_ordinal);`

cc @trinity-1686a

src/collector/top_score_collector.rs
PSeitz: We usually bench using https://github.com/quickwit-oss/search-benchmark-game. These benchmarks are highlevel enough to give a good idea of the performance.

sstable/src/dictionary.rs
trinity-1686a: `assert!(ord >= current_ordinal);` would be the assertion that failed before. At the end of the main loop, current_ordinal is set to `ord + 1`, not ord. This is an artifact of how the sub loop works with `current_ordinal..=ord`, a range which isn't empty when `current_ordinal==ord`, and/or that current_ordinal is initialized to 0 without calling advance (so the current position is more like minus 1).

I think `current_ordinal` is a miss-nommer, it isn't the ordinal of the entry we have in `bytes`, it's the ordinal of the entry we will have after calling `current_sstable_delta_reader.advance()`.

this change *does* fix the issue, but in a non-obvious way: if we pull for 4 (chosen by a fair dice roll) two times, the first iteration will bring `current_ordinal` to 5, the second will skip the assertion due to the newly added if, and do a for over `5..=4` which is an empty range, so everything works out. The fix has nothing to do with not calling `get_block_with_ord()` though, that function is stateless, it just skip the assertion in the specific case where it's too restrictive.

I think a better fix is to rename `current_ordinal` to `next_ordinal`, because that's what it is, and change the assertion to `assert!(ord+1 >= next_ordinal);`. There is probably an alternative fix where current_ordinal can keep its name, and `current_sstable_delta_reader.advance()` is called once just after initializing the delta reader, but i have a hard time wrapping my head around whether it's sound (not in the rust unsafety sense) when loading a new block

src/collector/top_score_collector.rs
PSeitz-dd: I ran the benchmark and this causes a performance regression

Screenshot 2025-07-11 at 15 30 09

fulmicoton-dd:
@PSeitz @stuhood 
If this is a bug in main we need to fix it right away.
If the PR is not ready for merge, let's revert the PR introducing the bug

PSeitz:
> @PSeitz @stuhood If this is a bug in main we need to fix it right away. If the PR is not ready for merge, let's revert the PR introducing the bug

As I understand it, this is a bug that exists since tantivy 0.22 and ordering by a fast field in ASC order is affected

stuhood:
@PSeitz : Thanks for opening #2672. Have broken out #2673 for the rest of the fix here.

PR #2655, opened 2025-06-22
QueryParser: make empty-query behaviour configurable (set_empty_query_match_all)

MagellaX:
an empty user query ("", whitespace-only, or a string that tokenises to nothing) is always translated into EmptyQuery, which matches no documents.
Issue #386 requests making this behaviour configurable because analytics dashboards and “show everything” listings often prefer an empty search box to return all documents.
What’s in this PR
1. New field empty_query_match_all: bool (default false) added to QueryParser.
API additions
set_empty_query_match_all(should_match_all: bool) – opt-in/out.
get_empty_query_match_all() -> bool – read current setting.
Centralised conversion – new helper logical_ast_to_query ensures the flag is respected by all parse_* and build_* entry points.
If the parsed AST is empty -> returns
AllQuery when the flag is true.
EmptyQuery (unchanged legacy behaviour) when false.
No breaking changes – the default remains “match nothing”; all existing tests continue to pass.
Docs – inline rustdoc comments explain the new behaviour.

Usage example
-------------```rust
let mut query_parser = QueryParser::for_index(&index, default_fields);
// Opt-in so that an empty search matches everything.
query_parser.set_empty_query_match_all(true);
// "" now yields AllQuery instead of EmptyQuery
let query = query_parser.parse_query("").unwrap();
assert_eq!(format!("{query:?}"), "AllQuery");

Impact
------  
• Allows downstream applications to implement a “show all” behaviour without post-processing.  
• Does not affect existing users unless they explicitly enable it.

Closes [#386]


src/query/query_parser/query_parser.rs
```diff
@@ -856,6 +873,20 @@ impl QueryParser {
             ),
         }
     }
+
+    /// Convert a logical AST into a runnable `Query`, taking into account the parser configuration.
+    fn logical_ast_to_query(&self, logical_ast: LogicalAst) -> Box<dyn Query> {
+        match trim_ast(logical_ast) {
```
fulmicoton, line 987: that logic is possibly broken.

We cannot have the check on "is an empty query" done after complex processing like simplification or trim_ast or whatever.

For this feature, an empty query should really be defined off the user input ast.

MagellaX: Hmm...seems right, the emptiness check must be based on the user input AST. I reworked it so we detect UserInputAst::Clause([ ]) directly and short-circuit before any simplification/trim. All parse/build entry points now apply this early check; the default stays EmptyQuery unless set_empty_query_match_all(true) is set.

src/query/query_parser/query_parser.rs
```diff
@@ -320,13 +324,49 @@ impl QueryParser {
...
+    #[inline]
+    fn empty_query_result(&self) -> Box<dyn Query> {
+        if self.empty_query_match_all {
+            Box::new(AllQuery)
+        } else {
+            Box::new(EmptyQuery)
+        }
+    }
+
+    #[inline]
```
fulmicoton:
```suggestion
```

src/query/query_parser/query_parser.rs
```diff
@@ -320,13 +324,49 @@ impl QueryParser {
...
+    pub fn set_empty_query_match_all(&mut self, should_match_all: bool) {
+        self.empty_query_match_all = should_match_all;
+    }
+
+    /// Returns the current behaviour for empty queries.
+    pub fn get_empty_query_match_all(&self) -> bool {
+        self.empty_query_match_all
+    }
+
+    #[inline]
```
fulmicoton:
```suggestion
```

fulmicoton (changes requested):
LGTM but can you add unit tests?

PR #2661, opened 2025-06-30
clippy

PSeitz:
mostly string formatting and io error handling


src/store/reader.rs
```diff
@@ -320,7 +320,7 @@ impl StoreReader {
                     doc_pos = 0;
                 }
 
-                let alive = alive_bitset.map_or(true, |bitset| bitset.is_alive(doc_id));
+                let alive = alive_bitset.is_none_or(|bitset| bitset.is_alive(doc_id));
```
trinity-1686a: do we want to use these functions?

trinity-1686a (approved):

src/store/reader.rs
PSeitz: I prefer
```rust
alive_bitset.map(|bitset| bitset.is_alive(doc_id)).unwrap_or(true)
```

PR #2662, opened 2025-06-30
Utilise tracing and reduce log verbosity

cetra3:
Another potentially breaking/unwelcome change!

This updates tantivy to utilise `tracing` and turns down some log verbosity

### DIFF
--- a/src/indexer/index_writer.rs
+++ b/src/indexer/index_writer.rs
@@ -193,7 +193,7 @@ fn index_documents<D: Document>(
         }
         let mem_usage = segment_writer.mem_usage();
         if mem_usage >= memory_budget - MARGIN_IN_BYTES {
-            info!(
+            debug!(
                 "Buffer limit reached, flushing segment with maxdoc={}.",
                 segment_writer.max_doc()
             );
@@ -562,7 +562,7 @@ impl<D: Document> IndexWriter<D> {
     ///
     /// The opstamp at the last commit is returned.
     pub fn rollback(&mut self) -> crate::Result<Opstamp> {
-        info!("Rolling back to opstamp {}", self.committed_opstamp);
+        debug!("Rolling back to opstamp {}", self.committed_opstamp);
         // marks the segment updater as killed. From now on, all
         // segment updates will be ignored.
         self.segment_updater.kill();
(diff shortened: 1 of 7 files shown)


AdamGS:
my 2c as a person that played around with tantivy - I think the real change this could power in the future is propagating `tracing`-based traces across threads and sync/async code, making everything work with application level traces

src/indexer/index_writer.rs
```diff
@@ -562,7 +562,7 @@ impl<D: Document> IndexWriter<D> {
     ///
     /// The opstamp at the last commit is returned.
     pub fn rollback(&mut self) -> crate::Result<Opstamp> {
-        info!("Rolling back to opstamp {}", self.committed_opstamp);
+        debug!("Rolling back to opstamp {}", self.committed_opstamp);
```
fulmicoton, line 565: that seems like a bad idea.

src/indexer/prepared_commit.rs
```diff
@@ -45,7 +45,7 @@ impl<'a, D: Document> PreparedCommit<'a, D> {
     /// this operation is not at all really light.
     /// At this point deletes have not been flushed yet.
     pub fn commit_future(self) -> FutureResult<Opstamp> {
-        info!("committing {}", self.opstamp);
+        debug!("committing {}", self.opstamp);
```
fulmicoton, line 48: same. this seems like a bad idea.
I don't understand what this is about. If we put everything as debug, won't it make difficult for people to discriminate what they want to see?

fulmicoton:
Is tracing compatible with the env_logger? I can't tell if this PR will break existing code or not.

AdamGS:
it can, just requires a couple of lines of setup, see the first example [here](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#examples).

src/indexer/prepared_commit.rs
cetra3: It feels like this is an internal detail to tantivy and not something someone using tantivy as a library should be concerned with?  That's sort of the guiding reason behind this decision

PSeitz: What would they be concerned with? I think committing to disk is one the most important events during indexing

fulmicoton:
@cetra3 closing this.

cetra3:
@fulmicoton let me know if there is any way we could get this integrated or if it's just not something you are keen on at the moment?

PR #2664, opened 2025-07-01
attempt to optimize sorted_ords_to_term_cb

trinity-1686a:
make it so that operation doesn't need to query the index nearly as much (that operation isn't free on index v3)

### DIFF
--- a/sstable/src/sstable_index_v3.rs
+++ b/sstable/src/sstable_index_v3.rs
@@ -58,10 +58,13 @@ impl SSTableIndex {
     }
 
     /// Get the [`BlockAddr`] of the block containing the `ord`-th term.
-    pub(crate) fn get_block_with_ord(&self, ord: TermOrdinal) -> BlockAddr {
+    pub(crate) fn get_block_with_ord<const FETCH_NEXT_ORD: bool>(
+        &self,
+        ord: TermOrdinal,
+    ) -> (BlockAddr, u64) {
         match self {
             SSTableIndex::V2(v2_index) => v2_index.get_block_with_ord(ord),
-            SSTableIndex::V3(v3_index) => v3_index.get_block_with_ord(ord),
+            SSTableIndex::V3(v3_index) => v3_index.get_block_with_ord::<FETCH_NEXT_ORD>(ord),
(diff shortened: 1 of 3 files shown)


trinity-1686a:
we could get away with always returning the next block 1st ordinal, that looked like a regression in some bench, but it's also possible that was only noise

sstable/src/sstable_index_v3.rs
```diff
@@ -58,10 +58,13 @@ impl SSTableIndex {
     }
 
     /// Get the [`BlockAddr`] of the block containing the `ord`-th term.
-    pub(crate) fn get_block_with_ord(&self, ord: TermOrdinal) -> BlockAddr {
+    pub(crate) fn get_block_with_ord<const FETCH_NEXT_ORD: bool>(
```
PSeitz, line 61: I think `FETCH_NEXT_ORD` could be just a regular parameter. With inlining it will be the same code, and without I think it won't make a performance difference.

fulmicoton: we could make this method inline always maybe.

fulmicoton: There are quite a bunch of fatty calls in between. I wouldn't bet on inlining happening here.

trinity-1686a: also the goal of having this as a const generics was to not impact perf when we don't care about the upper bound ord, which isn't very expensive, but isn't free to fetch. But it looks like most benchs having regressed might not be using sstables at all (noisy bench?), so it's possible this added complexity brings nothing

sstable/src/sstable_index_v3.rs
```diff
@@ -591,7 +610,10 @@ impl BlockAddrStore {
         )
     }
 
-    fn binary_search_ord(&self, ord: TermOrdinal) -> (u64, BlockAddr) {
+    fn binary_search_ord<const FETCH_NEXT_ORD: bool>(
```
fulmicoton, line 613: A comment to explain this signature please. This is not obvious at all.

trinity-1686a:
superseded by https://github.com/quickwit-oss/tantivy/pull/2933

PR #2672, opened 2025-07-16
Fix TopNComputer for reverse order

PSeitz:
Similar fix to #2651, but more lightweight (no change of `Score`)

thx @stuhood

### DIFF
--- a/src/collector/top_score_collector.rs
+++ b/src/collector/top_score_collector.rs
@@ -970,7 +970,7 @@ impl<Score, D, const R: bool> From<TopNComputerDeser<Score, D, R>> for TopNCompu
     }
 }
 
-impl<Score, D, const R: bool> TopNComputer<Score, D, R>
+impl<Score, D, const REVERSE_ORDER: bool> TopNComputer<Score, D, REVERSE_ORDER>
 where
     Score: PartialOrd + Clone,
     D: Ord,
@@ -991,7 +991,10 @@ where
     #[inline]
     pub fn push(&mut self, feature: Score, doc: D) {
         if let Some(last_median) = self.threshold.clone() {
-            if feature < last_median {
+            if !REVERSE_ORDER && feature > last_median {
+                return;
+            }
+            if REVERSE_ORDER && feature < last_median {
                 return;
             }
         }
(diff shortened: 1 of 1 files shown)


fulmicoton-dd (approved):

src/collector/top_score_collector.rs
```diff
@@ -991,7 +991,10 @@ where
...
     pub fn push(&mut self, feature: Score, doc: D) {
         if let Some(last_median) = self.threshold.clone() {
-            if feature < last_median {
+            if !REVERSE_ORDER && feature > last_median {
+                return;
+            }
+            if REVERSE_ORDER && feature < last_median {
                 return;
             }
         }
```
stuhood, line 1000: Unfortunately, after further consideration, I think that the previous version was slightly incorrect, and so this one is too.

There are probably multiple factors, but one reason why https://github.com/quickwit-oss/tantivy/pull/2651 would be slower is that it is comparing both the `Score` _and_ the `DocId`. Before and after this PR, `main` is only comparing the `Score` to the threshold, which means that it will not tiebreak using the `DocId`.

Tiebreaking with the `DocId` is necessary for deterministic ordering of results: this will eliminate docs with equal `Score`s even if they have higher `DocId`s, which mismatches the behavior of `ComparableDoc` (and thus `truncate_top_n`/`into_sorted_vec`).

stuhood: > Tiebreaking with the `DocId` is necessary for deterministic ordering of results...

And if it isn't, then we should remove it / make it optional, because it's not free! 😅

stuhood: > > Tiebreaking with the `DocId` is necessary for deterministic ordering of results...
> 
> And if it isn't, then we should remove it / make it optional, because it's not free! 😅

See https://github.com/quickwit-oss/tantivy/pull/2681#issue-3301620721 on this topic: I think that we could introduce optional `DocId`/`DocAddress` tiebreaking with that API, and then remove it from `TopN` consumers who don't need it.

PSeitz: #2651 compares Docid for the threshold, but that's not necessary. It should be a very fast pre-filtering. We still compare Docids via `ComparableDoc` during the sorts

The check does not filter docs with equal scores only with smaller ones

stuhood: Oh, true.  And the `ComparableDoc` comparator _should_ be doing the same thing and exiting early without comparing the docid.

Sorry for the noise.

PR #2673, opened 2025-07-16
Fix `TopDocs::order_by_string_fast_field` for duplicates

stuhood:
`sorted_ords_to_term_cb` currently only supports unique `ords`, likely because its previous consumers in aggregates would always de-dupe ords before lookup. That means that `TopDocs::order_by_string_fast_field` is broken for duplicates.

It is much easier for consumers who would like to look up non-unique term ordinals to allow them to be duplicated (as it introduces only two comparisons in a path that should always be IO bound): it allows a consumer to `zip` the results directly back to the inputs without a mapping table, as we do here: https://github.com/quickwit-oss/tantivy/blob/4e84c70387d14a46baf609c087498b670023192c/src/collector/top_score_collector.rs#L228-L242

### DIFF
--- a/sstable/src/dictionary.rs
+++ b/sstable/src/dictionary.rs
@@ -507,47 +507,58 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
     /// Returns true if and only if all terms have been found.
     pub fn sorted_ords_to_term_cb<F: FnMut(&[u8]) -> io::Result<()>>(
         &self,
-        ord: impl Iterator<Item = TermOrdinal>,
+        mut ords: impl Iterator<Item = TermOrdinal>,
         mut cb: F,
     ) -> io::Result<bool> {
+        let Some(mut ord) = ords.next() else {
+            return Ok(true);
+        };
+
+        // Open the block for the first ordinal.
         let mut bytes = Vec::new();
-        let mut current_block_addr = self.sstable_index.get_block_with_ord(0);
+        let mut current_block_addr = self.sstable_index.get_block_with_ord(ord);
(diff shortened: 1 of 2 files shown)


sstable/src/dictionary.rs
```diff
@@ -515,16 +515,25 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
...
-            assert!(ord >= current_ordinal);
-            // check if block changed for new term_ord
-            let new_block_addr = self.sstable_index.get_block_with_ord(ord);
-            if new_block_addr != current_block_addr {
-                current_block_addr = new_block_addr;
-                current_ordinal = current_block_addr.first_ordinal;
-                current_sstable_delta_reader =
-                    self.sstable_delta_reader_block(current_block_addr.clone())?;
-                bytes.clear();
+            // only advance forward if the new ord is different than the one we just processed
```
trinity-1686a: i already reviewed this bit in https://github.com/quickwit-oss/tantivy/pull/2651#discussion_r2176873213
tldr: current_ordinal is a misnomer, and that causes the comment to be wrong (advancing is done in the loop below), and the fix to work "by accident"

trinity-1686a (changes requested):

sstable/src/dictionary.rs
stuhood: Have refactored this into effectively a do-while (which is hopefully a bit clearer), and made the maintenance of this variable a little more obvious.

Please take another look!

trinity-1686a (approved):

trinity-1686a:
thanks!

PR #2676, opened 2025-07-17
ignore failure to parse query when other default field suceeded

### LINKED ISSUE #2649: Invalid query ParseIntError when the default fields contains u64 field and the query is str
**Describe the bug**

- What did you do?

In `QueryParser::for_index`, I set fields `title` (text) and `year` (u64) as default fields, and I query for "foo" (not parseable as u64).

- What happened?

This error is raised:

```
Error: InvalidArgument("Query is invalid. ExpectedInt(ParseIntError { kind: InvalidDigit })")
```

- What was expected?

The error should be silently ignored, and search in field `title` only.

commit 4a6123d3ffb552be2a97fc080360b64e7f7865bb and `0.24.1` (https://crates.io/crates/tantivy/0.24.1).

**To Reproduce**

If your bug is deterministic, can you give a minimal reproducing code?
Some bugs are not deterministic. Can you describe with precision in which context it happened?
If this is possible, can you share your code?

The bug is deterministic. Here is a minimal reproducible example:

```rust
use tantivy::{Index, TantivyDocument};
use tantivy::schema::{Schema, INDEXED, STORED, TEXT};

#[test]
fn test_queryparser_parseinterror() -> tantivy::Result<()> {
    let title = schema_builder.add_text_field("title", TEXT | STORED);
    let year = schema_builder.add_u64_field("year", INDEXED);

    let mut index_writer = index.writer(100_000_000)?;
    let mut doc1 = TantivyDocument::default();
    doc1.add_text(title, "foo bar");
    doc1.add_u64(year, 1999);
    index_writer.add_document(doc1)?;
    let mut doc2 = TantivyDocument::default();
    doc2.add_text(title, "qux");
    doc2.add_u64(year, 2022);
    index_writer.add_document(doc2)?;
    index_writer.commit()?;

    let query_parser = QueryParser::for_index(&index, vec![title, year]);
    let _query = query_parser.parse_query("foo")?;  // <-- error occurs here

    Ok(())
}

### PR DESCRIPTION
**Pull request by trinity-1686a**
fix #2649

### DIFF
--- a/src/query/query_parser/query_parser.rs
+++ b/src/query/query_parser/query_parser.rs
@@ -778,6 +778,12 @@ impl QueryParser {
                         asts.push(LogicalAst::Leaf(Box::new(ast)).boost(boost));
                     }
                 }
+                if !asts.is_empty() {
+                    // if some fields failed but other succeeded, we consider this a success, it
+                    // probably means the default_fields contains
+                    // text and non-text fields, and the non-text ones failed
+                    errors.clear();
+                }
                 let result_ast: LogicalAst = if asts.len() == 1 {
                     asts.into_iter().next().unwrap()
                 } else {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by mineshp-mecha — APPROVED** [review]

File: src/query/query_parser/query_parser.rs [hunk 1]
```diff
@@ -778,6 +778,12 @@ impl QueryParser {
                         asts.push(LogicalAst::Leaf(Box::new(ast)).boost(boost));
                     }
                 }
+                if !asts.is_empty() {
+                    // if some fields failed but other succeeded, we consider this a success, it
+                    // probably means the default_fields contains
+                    // text and non-text fields, and the non-text ones failed
+                    errors.clear();
```
PSeitz-dd, line 785: Should we still return the errors? The API seems to allow to return both

trinity-1686a: higher up, we either return both the query and the vec of errors (lenient mode), or we fail if there is any error (strict/default mode). Returning an error here is fatal for the strict parser (we share as much as possible between both to limit maintenance burden and make sure both behave the same overall)

PSeitz-dd (approved):

PR #2677, opened 2025-07-21
feat: Regex query grammar

Darkheir:
Add support for regexes in the query grammar.

Regexes are not widely documented in Lucene documentation but are supported inside queries ([Source](https://github.com/apache/lucene/blob/main/lucene/queryparser/src/java/org/apache/lucene/queryparser/classic/QueryParser.jj#L212)).

We can find a bit more documentation about them inside the Elasticsearch [query string documentation](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-query-string-query#_regular_expressions).

### DIFF
--- a/query-grammar/src/infallible.rs
+++ b/query-grammar/src/infallible.rs
@@ -46,7 +46,9 @@ fn unwrap_infallible<T>(res: Result<T, nom::Err<Infallible>>) -> T {
 ///
 /// It's less generic than the original to ease type resolution in the rest of the code.
 pub(crate) fn opt_i<I: Clone, O, F>(mut f: F) -> impl FnMut(I) -> JResult<I, Option<O>>
-where F: nom::Parser<I, O, nom::error::Error<I>> {
+where
+    F: nom::Parser<I, O, nom::error::Error<I>>,
+{
     move |input: I| {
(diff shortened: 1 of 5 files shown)


fulmicoton:
I am a bit scared of adding support for regex. I have seen companies not knowing their search engine supported regex (it was lucene 2.4) by default in the query parser, and had their search engine exposed to the internet. 
Attacker could exploit that for all kinds of usage.

It is definitely useful though.

Could we make it somehow optional and off by default? Maybe have a QueryParserBuilder that makes it possible to set  different params.

Darkheir:
I understand your concerns, I'm going to rework the PR to have something more customizable

Darkheir:
@fulmicoton I pushed a new commit to disable regexes by default.

I didn't use a `QueryParserBuilder` because we currently have a lot of methods in the `QueryParser` struct that would belong inside this builder (`set_conjunction_by_default`, `set_field_boost`, `set_field_fuzzy`, ...) but moving them into a builder struct would be a breaking change and duplicating those methods felt a bit overkill.

Feel free to ask for changes if you feel there is a better way to achieve it.

fulmicoton (approved):

PR #2678, opened 2025-07-22
feat: Support spaces between field name and value

Darkheir:
Allow optional spaces between the field name and the value the same way Lucene allows them

### DIFF
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -36,7 +36,7 @@ fn field_name(inp: &str) -> IResult<&str, String> {
                 alt((first_char, escape_sequence())),
                 many0(alt((simple_char, escape_sequence(), char('\\')))),
             )),
-            char(':'),
+            tuple((multispace0, char(':'), multispace0)),
         ),
         |(first_char, next)| once(first_char).chain(next).collect(),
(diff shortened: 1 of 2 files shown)


fulmicoton:
@trinity-1686a can you review? Does it looks sound?

query-grammar/src/query_grammar.rs
```diff
@@ -1283,6 +1283,10 @@ mod test {
             super::field_name("~my~field:a"),
             Ok(("a", "~my~field".to_string()))
         );
+        assert_eq!(
+            super::field_name(".my.field.name : a"),
```
trinity-1686a, line 1287: can you also add a `test_parse_query_to_ast_helper` test somewhere to make sure this works (and stays working) for both the strict and the lenient parser

trinity-1686a (approved):
lgtm, afaict it doesn't introduce any problematic ambiguity
it makes redundant some of the multisplace handling in the lenient parser, but i'll clean that up myself eventually

Darkheir:
Thank you for the review :)

I added few tests, but don't hesitate to ask for more if you see some edge cases that could not be covered by the existing ones.

trinity-1686a:
thanks!

PR #2679, opened 2025-07-29
per field size details

fulmicoton:
Added per-field size details.

This also does a bunch of refactoring.

merging field metadata does not silently asserts that arguments should be sorted.
merging does not set `stored`.

We do not rely on a hashmap to group fields, but instead rely on the fact that
the term dictionary is sorted.

The inverted level method that exposes field metadata is not exposed
as public anymore.

### DIFF
--- a/src/index/inverted_index_reader.rs
+++ b/src/index/inverted_index_reader.rs
@@ -44,23 +44,17 @@ pub(crate) struct InvertedIndexFieldSpace {
     pub num_terms: u64,
 }
 
-impl InvertedIndexFieldSpace {
-    // Creates an empty FieldSpace object for a given term.
-    fn try_from(term: &[u8]) -> Option<InvertedIndexFieldSpace> {
-        let index = term.iter().position(|&byte| byte == JSON_END_OF_PATH)?;
-        let field_type_code = term.get(index + 1).copied()?;
-        let field_type = Type::from_code(field_type_code)?;
-        // Let's flush the current field.
-        let field_name = String::from_utf8_lossy(&term[..index]).to_string();
-        Some(InvertedIndexFieldSpace {
-            field_name,
-            field_type,
-            postings_size: ByteCount::default(),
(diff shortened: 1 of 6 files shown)


src/index/segment_reader.rs
```diff
@@ -507,190 +540,191 @@ fn intersect_alive_bitset(
...
+
+//         {
+//             let mut index_writer: IndexWriter = index.writer_for_tests()?;
+//             index_writer.add_document(doc!(name => "tantivy"))?;
+//             index_writer.add_document(doc!(name => "horse"))?;
+//             index_writer.add_document(doc!(name => "jockey"))?;
+//             index_writer.add_document(doc!(name => "cap"))?;
+//             // we should now have one segment with two docs
+//             index_writer.delete_term(Term::from_field_text(name, "horse"));
+//             index_writer.delete_term(Term::from_field_text(name, "cap"));
```
fulmicoton: wip

src/index/inverted_index_reader.rs
```diff
@@ -36,6 +35,39 @@ pub struct InvertedIndexReader {
...
+    pub field_type: Type,
+    pub postings_size: ByteCount,
+    pub positions_size: ByteCount,
+    pub num_terms: u64,
+}
+
+impl InvertedIndexFieldSpace {
+    // Creates an empty FieldSpace object for a given term.
+    fn try_from(term: &[u8]) -> Option<InvertedIndexFieldSpace> {
+        let index = term.iter().position(|&byte| byte == JSON_END_OF_PATH)?;
```
Abdul-Andha: nit: I think its worth clarifying this function only accepts json

Abdul-Andha (approved):

src/index/inverted_index_reader.rs
fulmicoton: you are absolutely right.

PR #2680, opened 2025-08-01
Clippy and cleanups

fulmicoton-dd:


stacker/src/lib.rs
```diff
@@ -1,5 +1,3 @@
-#![cfg_attr(all(feature = "unstable", test), feature(test))]
```
fulmicoton-dd, line 1: this is now forbidden in stable. Can you investigate what you want to do with this?

PSeitz: I think we can just move all benches to `binggan`

Issue #494, opened 2019-02-07
Near Real Time indexing.


fulmicoton:
# Use case

A lot of users or potential users of tantivy want to use tantivy in client applications. (e.g. indexing the web page they see.). 

They really do not care much about data ingestion, as they handle live-human generated data, but they do care about finding their data quickly after ingesting it. 

Actually `.commit()` after every single document inserted is not much of a problem, but the creation of a new segment and the associated file might be a bit too aggressive.

# Proposed solution

We introduce the notion of a soft commit. soft commits acts similarly to commits, except the `meta.json` file is not written. Soft commits is implemented as another `SegmentRegister` in the `SegmentManager`.

We then introduce a `.persist()` method in directories as well. The role of the method is to ensure that the data is persisted and blocks until this is done. (returning a future is ok too)

`.prepare commit(bool)` would then get an extra boolean that says whether this is a soft commit or not. 
On non-soft prepare commit, the directory `.persist()` method is called.
On `commit()` , `meta.json` is written.

We also need to introduce a NRTDirectory that consists in a RAMDirectory masking another directory.  ( in most case an MMapDirectory).
All writes (except committed segment merges?) go to the RAMDirectory.

On `.persist()` this in RAM data gets written to the `MMapDirectory`.

In the future, we will probably want to start automatically writing file when they exceed a given size.

petr-tik:
Hey @diggsey, 
IIRC, you want to make newly indexed documents searchable immediately. 

Does this ticket capture and address the use case you were asking me about at the Rust London meetup after my presentation about tantivy? 

If so, can you please share more detail and numbers of your use case? 

## Constraints and use cases? 

Your current infrastructure takes too much time to make new documents available and I was wondering:  
What is the average and variance in those times? High std dev or a low std dev? How predictable is the performance from load, peak times or client?
How big are your corpora of documents? From 1GB to 100s of GB? How many corpora? What is the potential scale of your product – bigger corpora or more corpora of the same size, or both?   
Are you read-heavy, write-heavy or roughly similar? How much does it change with time of day/year? 
Do you run on-prem or in public cloud? 
How complex are the queries that users run? 
Outside the 1ms delay, what are your other pain points with current indexing and search infrastructure? 

## Overall setup:
Each client loads, indexes and searchers only their own documents. 

You partition different clients’ documents (I am guessing that ES chooses a partition by client_id as key?) to separate instances. 


[...]

PR #2682, opened 2025-08-11
Add soft-commit and NRT overlay: IndexWriter::soft_commit, IndexWriter::nrt_searcher, and NrtDirectory

MagellaX:
# PR description

This change introduces Near Real-Time indexing support:

## Soft commit
- Adds `PreparedCommit::soft_commit()` and `IndexWriter::soft_commit()` to publish current uncommitted segments to the in-memory committed set without persisting `meta.json`.
- Deletes up to the soft-commit opstamp are applied; no GC or `meta.json` write happens.

## NRT searcher
- Adds `IndexWriter::nrt_searcher()` to construct a Searcher directly from the in-memory committed segments for instant visibility after soft commits.

## NRTDirectory overlay
- Adds `NrtDirectory`, an overlay that writes to an in-memory `RamDirectory` and reads from overlay first, then base directory.
- `sync_directory()` persists overlay files to the base directory to make data durable when desired.

---

## Compatibility
- Default commit semantics are unchanged. Existing users are unaffected unless the new APIs are used.
- Normal commit still persists `meta.json` and triggers garbage collection as before.

---

## Why
- Enables low-latency searchability for recently indexed documents in client applications without writing `meta.json` on every change.  
- Aligns with the design discussed in tantivy issue #494.

---

## How to use
1. Wrap your base directory with `NrtDirectory` when creating/opening the index.
2. Call `index_writer.soft_commit()` after indexing to publish for NRT search.
3. Use `index_writer.nrt_searcher()` to obtain a Searcher that sees soft-committed data immediately.
4. Call `index_writer.commit()` whenever you are ready to persist `meta.json` and make changes durable on disk.

---


fulmicoton:
@MagellaX can you fix the PR. I think the first two commits are unrelated.

src/directory/nrt_directory.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    /// Persist overlay files into the base directory if missing there.
+    fn persist_overlay_into_base(&self) -> crate::Result<()> {
+        let snapshot_paths: Vec<PathBuf> = {
+            let guard = self.overlay_paths.read().unwrap();
+            guard.iter().cloned().collect()
+        };
+        for path in snapshot_paths {
```
fulmicoton, line 43: ok, but if we update the meta json file before a segment file, and this function has an io Error in the middle, the index will be corrupted right now?

fulmicoton:
This is a cool effort! ... But I stopped reviewing in the middle as too many problems have been overlooked.

MagellaX:
> @MagellaX can you fix the PR. I think the first two commits are unrelated.

just give me a sec, i am checking

src/directory/nrt_directory.rs
MagellaX: I’ve moved the meta publication to the end of sync_directory() and removed the early base write in atomic_write. We now snapshot overlay paths, persist all data files first, and atomically write meta.json last to avoid partial states. Build is green; please have another look.

PR #2689, opened 2025-08-28
Added unit test for aggregation

fulmicoton:


src/aggregation/agg_tests.rs
```diff
@@ -155,7 +155,7 @@ fn test_aggregation_flushing(
         searcher.search(&AllQuery, &collector).unwrap()
     };
 
-    let res: Value = serde_json::from_str(&serde_json::to_string(&agg_res)?)?;
```
PSeitz, line 158: This is on purpose to test JSON serialization roundtrips

PR #2693, opened 2025-09-11
Add fast field fallback for term query if not indexed

PSeitz-dd:
Currently we fail if we try to search (via `TermQuery`) on a field which is not indexed, but has a fast field (columnar storage).

This will add a path to search on the fast field for `TermQuery`, by reusing the range query implementation.

## Caveats
This comes with 2 caveats:

* This is considerable slower since it requires to scan the entire fast field. In combination with `fast` this is a real footgun. This should probably come with renaming `FAST` to `COLUMN`.
* In the range query case we do not support scoring, so this the fallback will only be executed if the query is without scoring


fulmicoton-dd (approved):
thank you!

PR #2695, opened 2025-09-18
Refactor aggregation accessors

rdettai-sk:
This PR is a refactoring of AggregationWithAccessor to streamline the way accessors are defined and missing values are configured. Some comments were already present in the code about the necessity of this refactoring.

The approach is to make the coupling between AggregationWithAccessor and the collectors more explicit without changing the existing logic.

This refactoring is the first step of a broader effort aiming at adding a new bucket aggregation called composite aggregations (https://www.elastic.co/docs/reference/aggregations/search-aggregations-bucket-composite-aggregation)

### DIFF
--- a/src/aggregation/agg_req_with_accessor.rs
+++ b/src/aggregation/agg_req_with_accessor.rs
@@ -34,21 +34,34 @@ impl AggregationsWithAccessor {
     }
 }
 
-pub(crate) enum MissingTermCollection {
-    Inline { missing_value_for_accessor: u64 },
-    TermMissingCollector,
+/// Configures how the aggregation should behave when the field is missing from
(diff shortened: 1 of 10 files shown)


rdettai-sk:
The failing CI has nothing to do with this PR, it's because of rust 1.90 (see https://github.com/quickwit-oss/tantivy/pull/2696)

rdettai-sk:
After having gone through with an implem of composite aggregations (https://github.com/quickwit-oss/tantivy/pull/2714), I realize that this refacto needs to go even further to be effective. I'm closing it for now.

PSeitz:
@rdettai-sk I'm currently refactoring aggregations and will replace `AggregationWithAccessor` with a better solution, that allows per request caching and datastructures

PR #2698, opened 2025-09-19
perf: deduplicate queries

PSeitz-dd:
Currently we may waste performance by executing the same query multiple times. This PR deduplicates duplicate queries in `UserInputAst` after parsing. 

e.g. 
`a a a ` is equal to `a`
`to be or not to be` is equal to `to be or not`

Screenshot 2025-09-19 at 11 09 33

### DIFF
--- a/query-grammar/src/lib.rs
+++ b/query-grammar/src/lib.rs
@@ -31,11 +31,11 @@ pub fn parse_query_lenient(query: &str) -> (UserInputAst, Vec<LenientError>) {
 
 #[cfg(test)]
 mod tests {
-    use crate::{parse_query, parse_query_lenient};
+    use crate::{UserInputAst, parse_query, parse_query_lenient};
 
     #[test]
     fn test_deduplication() {
-        let ast = parse_query("a a").unwrap();
+        let ast: UserInputAst = parse_query("a a").unwrap();
         let json = serde_json::to_string(&ast).unwrap();
         assert_eq!(
             json,
--- a/query-grammar/src/lib.rs
+++ b/query-grammar/src/lib.rs
@@ -34,6 +34,16 @@ mod tests {
     use crate::{parse_query, parse_query_lenient};
 
     #[test]
+    fn test_deduplication() {
+        let ast = parse_query("a a").unwrap();
+        let json = serde_json::to_string(&ast).unwrap();
+        assert_eq!(
+            json,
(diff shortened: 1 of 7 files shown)


query-grammar/src/lib.rs
```diff
@@ -33,6 +33,16 @@ pub fn parse_query_lenient(query: &str) -> (UserInputAst, Vec<LenientError>) {
 mod tests {
     use crate::{parse_query, parse_query_lenient};
 
+    #[test]
+    fn test_deduplication() {
+        let ast = parse_query("a a").unwrap();
```
fulmicoton-dd: which ast is that? can you add the type?

query-grammar/src/query_grammar.rs
```diff
@@ -1078,12 +1079,25 @@ pub fn parse_to_ast_lenient(query_str: &str) -> (UserInputAst, Vec<LenientError>
     (rewrite_ast(res), errors)
 }
 
-/// Removes unnecessary children clauses in AST
-///
-/// Motivated by [issue #1433](https://github.com/quickwit-oss/tantivy/issues/1433)
 fn rewrite_ast(mut input: UserInputAst) -> UserInputAst {
-    if let UserInputAst::Clause(terms) = &mut input {
-        for term in terms {
+    if let UserInputAst::Clause(sub_clauses) = &mut input {
```
fulmicoton-dd, line 1083: :| I don't think this is a good idea to do manipulation over the user input ast.
Can this be done over the logic ast? Or the conversion from one to the other?

PSeitz-dd: We already do manipulate the user input ast, I just extended the logic. 

`LogicAst` is not suitable currently, with the `Regex` we can't have `Eq` or `Hash`. `Regex` creation could be delayed though and just store a `String` instead.

* The main advantage doing it here is that we get the optimization directly available in Quickwit.
* The downside I guess is that you already get an optimized result, but for that I'd rather add a new method that returns the parsed AST without optimization e.g. `parse_without_rewrite_ast`.

query-grammar/src/lib.rs
PSeitz-dd: added the type

query-grammar/src/query_grammar.rs
fulmicoton-dd: The downside:
- the user input ast gets further away from the user input.
- we lose the ability to have several language map into the logical ast.

fulmicoton-dd (approved):

PR #2699, opened 2025-09-19
fix compiler warnings

PSeitz-dd:


columnar/src/column_index/optional_index/mod.rs
```diff
@@ -1,4 +1,4 @@
-use std::io::{self, Write};
+use std::io::{self};
```
trinity-1686a:
```suggestion
use std::io;
```

src/query/bm25.rs
```diff
@@ -68,12 +66,6 @@ fn compute_tf_cache(average_fieldnorm: Score) -> [Score; 256] {
     cache
 }
 
-#[derive(Clone, PartialEq, Debug, Serialize, Deserialize)]
-pub struct Bm25Params {
```
trinity-1686a, line 72: is there something else people should use/be pointed too (this is public)?

trinity-1686a (approved):

src/query/bm25.rs
PSeitz: it's pub but not exported outside the mod

PR #2707, opened 2025-09-25
Add DocSet::cost()

PSeitz:
- Add DocSet::cost()
- Use cost() instead of size_hint() to order scorers in intersect_scorers

This pulls out the safe changes without the more risky seek APIs changes from PR #2538

### DIFF
--- a/src/query/size_hint.rs
+++ b/src/query/size_hint.rs
@@ -1,0 +1,141 @@
+/// Computes the estimated number of documents in the intersection of multiple docsets
+/// given their sizes.
+///
+/// # Arguments
+/// * `docset_sizes` - An iterator over the sizes of the docsets (number of documents in each set).
+/// * `max_docs` - The maximum number of docs that can hit, usually number of documents in the
+///   segment.
+///
+/// # Returns
+/// The estimated number of documents in the intersection.
+pub fn estimate_intersection<I>(mut docset_sizes: I, max_docs: u32) -> u32
+where I: Iterator<Item = u32> {
+    if max_docs == 0u32 {
+        return 0u32;
+    }
+    // Terms tend to be not really randomly distributed.
+    // This factor is used to adjust the estimate.
+    let mut co_loc_factor: f64 = 1.3;
+
+    let mut intersection_estimate = match docset_sizes.next() {
+        Some(first_size) => first_size as f64,
+        None => return 0, // No docsets provided, so return 0.
+    };
+
+    let mut smallest_docset_size = intersection_estimate;
+    // Assuming random distribution of terms, the probability of a document being in the
+    // intersection
+    for size in docset_sizes {
+        // Diminish the co-location factor for each additional set, or we will overestimate.
+        co_loc_factor = (co_loc_factor - 0.1).max(1.0);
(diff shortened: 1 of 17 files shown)


src/query/size_hint.rs
```diff
@@ -0,0 +1,141 @@
...
+    // Terms tend to be not really randomly distributed.
+    // This factor is used to adjust the estimate.
+    let mut co_loc_factor: f64 = 1.3;
+
+    let mut intersection_estimate = match docset_sizes.next() {
+        Some(first_size) => first_size as f64,
+        None => return 0, // No docsets provided, so return 0.
+    };
+
+    let mut smallest_docset_size = intersection_estimate;
```
fulmicoton, line 25: tbh I would have prefered just the min as an estimate.

fulmicoton, line 25: i think there was no reason to have smallest_docset_size be a f64

src/query/union/buffered_union.rs
```diff
@@ -50,6 +51,7 @@ pub struct BufferedUnionScorer<TScorer, TScoreCombiner = DoNothingCombiner> {
     doc: DocId,
     /// Combined score for current `doc` as produced by `TScoreCombiner`.
     score: Score,
+    num_docs: u32,
```
fulmicoton, line 55: max_doc?

src/query/union/buffered_union.rs
```diff
@@ -81,6 +83,7 @@ impl<TScorer: Scorer, TScoreCombiner: ScoreCombiner> BufferedUnionScorer<TScorer
     pub(crate) fn build(
         docsets: Vec<TScorer>,
         score_combiner_fn: impl FnOnce() -> TScoreCombiner,
+        num_docs: u32,
```
fulmicoton, line 88: Can you add a comment explaining what num_docs is about? I really did not understand what we are supposed to pass here.

src/query/union/mod.rs
```diff
@@ -27,11 +27,17 @@ mod tests {
         docs_list.iter().cloned().map(VecDocSet::from)
     }
     fn union_from_docs_list(docs_list: &[Vec<DocId>]) -> Box<dyn DocSet> {
+        let max_doc = docs_list
+            .iter()
+            .flat_map(|docs| docs.iter().copied())
```
fulmicoton, line 32: wait... docs_lists are not sorted here?
Can't we just take the last?

fulmicoton (approved):

src/query/size_hint.rs
PSeitz-dd: the min estimate was quite off from the actual number of hits

src/query/union/mod.rs
PSeitz-dd: I think there is one test that generates not sorted lists
```rust
        aux_test_union(&[
            tests::sample_with_seed(100_000, 0.01, 1),
            tests::sample_with_seed(100_000, 0.05, 2),
            tests::sample_with_seed(100_000, 0.001, 3),
        ]);
```

src/query/union/buffered_union.rs
PSeitz-dd: added docs for the parameter, it's the number of docs in the segment

PR #2710, opened 2025-10-02
feat: add range (= max - min) into stats

cometkim:
I found this is useful when querying something like duration or timespan.

### DIFF
--- a/src/aggregation/bucket/histogram/histogram.rs
+++ b/src/aggregation/bucket/histogram/histogram.rs
@@ -1199,6 +1199,7 @@ mod tests {
                     "count": 0,
                     "min": Value::Null,
                     "max": Value::Null,
+                    "range": Value::Null,
                     "avg": Value::Null,
                 }
             })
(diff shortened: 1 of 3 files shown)


cometkim:
`range` might be confused with `ranges`. Maybe something like `duration` could be used instead.

fulmicoton:
That does not seem really useful. This can be computed from min, max on the application, isn't it? Also, aggregations are trying to replicate elasticsearch api.

cometkim:
Fair enough.

However, it isn't enough for complex queries that uses the result in sorting or filtering.

To be compatible with ElasticSearch and compete fairly with other analytics engines, at least the `bucket_script` aggregation should be supported. Even if it doesn't support the full ES [Painless](https://www.elastic.co/docs/explore-analyze/scripting/modules-scripting-painless) syntax, it should at least [Lucene's expression language](https://lucene.apache.org/core/9_1_0/expressions/org/apache/lucene/expressions/js/package-summary.html).

One certain thing is scripting will be an inefficient way to use values ​​that are already computed by stats aggregation.

cometkim:
Dynamic scripting will be powerful, but it's difficult to optimize and can potentially make operations more difficult. Wouldn't it be more feasible to support more of extended aggregations, even it's not the part of ES?

PR #2711, opened 2025-10-03
feat: added filter aggregation

### LINKED ISSUE #2706: Add Elasticsearch Filter Aggregation Support
**Problem Definition**

We need filter aggregations for our analytics queries - stuff like "get me the average price overall, plus the average price for just t-shirts, plus the count of electronics items" in a single query.

Right now there's no way to do this in Tantivy. We tried using `MultiCollector` but it doesn't work - all collectors get the same documents, so you can't filter differently for each one. The collectors get confused when you try to filter at that level.

So we're stuck running separate queries for each filter, which is slow and defeats the purpose.

For example, this common Elasticsearch query pattern is not supported:

```json
{
  "aggs": {
    "avg_price": { "avg": { "field": "price" } },
    "t_shirts": {
      "filter": { "term": { "type": "t-shirt" } },
      "aggs": {
        "avg_price": { "avg": { "field": "price" } }
      }
    }
  }
}
```

**Proposed Solution**

Add filter aggregation support to Tantivy. Basically implement the same thing ES has - a bucket aggregation that creates a single bucket with documents matching a query, and you can nest other aggregations inside it.

Would be great if it:
- Uses the same JSON format as Elasticsearch (for compatibility)
- Works with sub-aggregations like the other bucket types
- Handles basic queries like term, range, bool
- Returns the same JSON structure as Elasticsearch

This would provide Elasticsearch compatibility for filter aggregations.

**Alternatives Considered**

We looked at a few other approaches:

1. **Custom multi-collector**: Build something that routes docs to different collectors based on filters. Seems overly complex and would need big changes to the collector system.

2. **Add filters to existing aggregations**: Like making every aggregation accept an optional filter param. Would work but you'd have to modify every single aggregation type.

3. **Just live with separate queries**: What we're doing now. Works but it's slow and feels hacky.

The filter aggregation approach seems cleanest since it fits into the existing bucket aggregation pattern and gives us ES compatibility.

**Related Issues:**
- #1690 - Multi-collector support for filtered aggregations

mdashti:
# Ticket(s) Closed

- Closes https://github.com/quickwit-oss/tantivy/issues/2706

## What

Implements filter aggregation support in Tantivy, enabling multiple filtered aggregations in a single query.

## Why

Currently, there's no way to compute aggregations on different filtered subsets of documents in a single query. Users must run separate queries for each filter, which is slow and inefficient. For example, computing "average price overall + average price for t-shirts + count of electronics" requires three separate queries.

Elasticsearch's filter aggregation solves this by creating a single bucket containing documents matching a query, with support for nested sub-aggregations. This is a common analytics pattern that Tantivy now supports!

## How

Added a new `FilterAggregation` bucket aggregation type that:

1. Accepts both query strings (parsed via `QueryParser`) and direct `Query` objects for custom query types
2. Uses `DocumentQueryEvaluator` to evaluate filter queries per-document during aggregation collection, avoiding separate query executions
3. Extended aggregation collectors to receive `SegmentReader` references, enabling filter aggregations to create query weights and scorers per segment

Some Implementation Details:
- `FilterAggregation` supports two modes:
  - `FilterQuery::QueryString`: Parsed using Tantivy's standard `QueryParser`
  - `FilterQuery::Direct`: Accepts `Box<dyn Query>` for custom query extensions
- `FilterSegmentCollector` evaluates the filter query on each document collected by the main query
- Documents matching the filter are counted and passed to sub-aggregation collectors
- Results include `doc_count` and flattened sub-aggregation results
- 
## Tests

Test suite with 20 tests covering:

- Basic Filtering: Single filters, no matches, multiple independent filters
- Query Types: Term queries, range queries, boolean queries, bool field queries  
- Nested Filters: 2-level nesting, deep nesting (4+ levels), multiple branches at each level
- Sub-Aggregations: Terms aggregations, multiple metric aggregations
- Edge Cases: Empty indexes, malformed queries, base query interaction
- Custom Queries: Direct Query objects, serialization behavior, equality checks
- Correctness: Validation against equivalent separate query execution

All tests use the `assert_agg_results!` macro for clean, consistent result validation with floating-point tolerance.


examples/filter_aggregation.rs
```diff
@@ -0,0 +1,212 @@
...
+        }
+    });
+    assert_eq!(serde_json::to_value(&result)?, expected);
+    println!("{}\n", serde_json::to_string_pretty(&result)?);
+
+    // Example 3: Nested filters
+    println!("=== Example 3: Nested filters ===");
+    let agg_req = json!({
+        "all_products": {
+            "filter": "*",
```
PSeitz-dd: As I understand it `*` is a no-op as filter, so it's not super suitable as an example I think (It may people give the impression it's the same as the `global` aggregation)

src/aggregation/agg_result.rs
```diff
@@ -172,6 +174,10 @@ impl BucketResult {
                 sum_other_doc_count: _,
                 doc_count_error_upper_bound: _,
             } => buckets.iter().map(|bucket| bucket.get_bucket_count()).sum(),
+            BucketResult::Filter(filter_result) => {
+                // add one for the filter bucket itself
+                1 + filter_result.sub_aggregations.get_bucket_count()
```
PSeitz-dd:
```suggestion
                filter_result.sub_aggregations.get_bucket_count()
```
The bucket count is the user facing bucket count, which filter does not alter

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,380 @@
...
+
+/// Represents different ways to specify a filter query
+#[derive(Debug)]
+pub enum FilterQuery {
+    /// Query string that will be parsed using Tantivy's standard parsing facilities
+    /// Accepts query strings that can be parsed by QueryParser::parse_query()
+    QueryString(String),
+
+    /// Direct Query object for extension query types
+    /// This bypasses JSON parsing entirely for maximum performance
```
PSeitz-dd: What JSON parsing?
Edit: I see, you mean this will not be accessible when deserializing from JSON?

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,380 @@
...
+    Direct(Box<dyn Query>),
```
PSeitz-dd:
```suggestion
    CustomQuery(Box<dyn Query>),
```

For me both variants are direct, but this one allows to pass in custom queries

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,380 @@
...
+                // Return a clone of the direct query, ignoring the parser
+                Ok(query.box_clone())
+            }
+        }
+    }
+
+    /// Get the fast field names used by this aggregation (none for filter aggregation)
+    pub fn get_fast_field_names(&self) -> Vec<&str> {
+        // Filter aggregation doesn't use fast fields directly
+        vec![]
```
PSeitz-dd, line 336: This is a recently added feature https://github.com/quickwit-oss/tantivy/pull/2693
When a field is not indexed, but fast, we search on it via its fast field.

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,380 @@
...
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where S: Serializer {
+        match &self.query {
+            FilterQuery::QueryString(query_string) => {
+                // Serialize the query string directly
+                query_string.serialize(serializer)
+            }
+            FilterQuery::Direct(_) => {
+                // Direct queries cannot be serialized
+                Err(serde::ser::Error::custom(
```
PSeitz-dd: This limitation should be mentioned on the enum variant

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,380 @@
...
+}
+
+// PartialEq is required because AggregationVariants derives it
+// We implement it manually to handle Box<dyn Query> which doesn't impl PartialEq
+impl PartialEq for FilterAggregation {
+    fn eq(&self, other: &Self) -> bool {
+        match (&self.query, &other.query) {
+            (FilterQuery::QueryString(a), FilterQuery::QueryString(b)) => a == b,
+            // Direct queries cannot be compared for equality
+            _ => false,
```
PSeitz-dd, line 388: I'm not sure why we have PartialEq on `AggregationVariants`, maybe we can remove it

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,380 @@
...
+        })?;
+
+        let segment_reader = self.segment_reader.as_ref().ok_or_else(|| {
+            TantivyError::InvalidArgument(
+                "DocumentQueryEvaluator not initialized for segment".to_string(),
+            )
+        })?;
+
+        // This already handles all optimizations (fast fields, bitsets, etc.)
+        let mut scorer = weight.scorer(segment_reader, 1.0)?;
```
PSeitz: I think this would be way too slow for aggregations that have many hits. I think a better way would be to execute the search and collect it in a BitSet upfront (to be able to handle out of order docs due to caching), same as we do for range queries. It's not ideal though in cases we skip a lot, since we may fill the Bitset and never read it.

Range queries on the inverted index would very likely overload the system https://github.com/quickwit-oss/tantivy/blob/main/src/query/range_query/range_query.rs#L215

I'm currently refactoring aggregations to be able to store precomputed data in a better way

src/aggregation/bucket/filter.rs
mdashti: I tried to remove `PartialEq` from `AggregationVariants`, but it seems that's used in several tests. Let's do this change in a separate PR.

src/aggregation/bucket/filter.rs
mdashti: Thanks for the feedback! You're right that the initial implementation would've been too slow, and we observed it in our benchmarks. Then, in [f144df4](https://github.com/quickwit-oss/tantivy/pull/2711/commits/f144df44da3365222a17ef20ff33d6d6e9b7183c), improved it. But, your suggestion to use BitSet also makes a lot of sense. Applied it in [a570edb](https://github.com/quickwit-oss/tantivy/pull/2711/commits/a570edb048674b665f2a93896238a8e87d1b49ba)

There's a clear tradeoff between [f144df4](https://github.com/quickwit-oss/tantivy/pull/2711/commits/f144df44da3365222a17ef20ff33d6d6e9b7183c) and [a570edb](https://github.com/quickwit-oss/tantivy/pull/2711/commits/a570edb048674b665f2a93896238a8e87d1b49ba), but I couldn't make a case for BitSet to works much more inefficiently, but easy to make cases that BitSet approach works 2x better.

examples/filter_aggregation.rs
mdashti: Good suggestion. Made the example more interesting.

src/aggregation/agg_result.rs
mdashti: Oh! I didn't know that. Thanks. Fixed.

src/aggregation/bucket/filter.rs
mdashti: Changed it to `CustomQuery` based on your other suggestion and fixed this comment, too.

src/aggregation/bucket/filter.rs
mdashti: That's a better name. Changed it to `CustomQuery`

src/aggregation/bucket/filter.rs
mdashti: I see. Fixed the comment here to be more clear.

src/aggregation/bucket/filter.rs
mdashti: Done.

mdashti (commented):
@PSeitz @PSeitz-dd Thanks for the comments.

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,447 @@
...
+    /// This executes the query upfront and collects results into a BitSet,
+    /// unless the query is AllQuery in which case we skip BitSet creation.
+    fn new(
+        query: Box<dyn Query>,
+        schema: Schema,
+        segment_reader: &SegmentReader,
+    ) -> crate::Result<Self> {
+        // Optimization: Detect AllQuery and skip BitSet creation
+        // AllQuery matches all documents, so we can just accept everything
+        if query.as_any().downcast_ref::<AllQuery>().is_some() {
```
PSeitz, line 439: I don't think we should optimize this case, since all query here doesn't make sense

src/aggregation/bucket/filter.rs
PSeitz: > There's a clear tradeoff between [f144df4](https://github.com/quickwit-oss/tantivy/pull/2711/commits/f144df44da3365222a17ef20ff33d6d6e9b7183c) and [a570edb](https://github.com/quickwit-oss/tantivy/pull/2711/commits/a570edb048674b665f2a93896238a8e87d1b49ba), but I couldn't make a case for BitSet to works much more inefficiently, but easy to make cases that BitSet approach works 2x better.

I don't understand this, what is inefficient, what is 2x better?

Caching on the segment collector level is dangerous. In some queries you may have 100k instances of it, due to nested buckets. I'm working on a refactor to move caching in a per request data structure. I think it would make sense to rebase this PR after I merge it.

src/aggregation/bucket/filter.rs
mdashti: I see your point-of-view. In our experience, for machine-generated queries, it simplifies the code/query generators to use `FilteredAggregation` with `*`. Also, it helps to keep the same query output, which simplifies the output processing. IMO, this optimization is harmless, but if you insist I can remove it.

PSeitz-dd: Do you have some benchmarks on this? Is there some cost for queries for the `Option` wrapping?

PSeitz-dd:
I refactored aggregations in this PR: https://github.com/quickwit-oss/tantivy/pull/2715

You can now store the Bitset outside of the collector and reuse it between instances of collectors.

src/aggregation/bucket/filter.rs
mdashti: Added some benchmarks to [17d5b3f](https://github.com/quickwit-oss/tantivy/pull/2711/commits/17d5b3f7f2ac5420e1ae3c1dcb2ca2f56cc55d85), but also used this specialized bechmark:
```
// Benchmark to measure the performance impact of the AllQuery optimization in filter aggregation
//
// This benchmark specifically tests:
// 1. Filter aggregation with AllQuery (optimized - no BitSet)
// 2. Filter aggregation with AllQuery (unoptimized - full BitSet)
// 3. Filter aggregation with TermQuery (baseline)
//
// The AllQuery optimization avoids creating a BitSet when the filter matches all documents,
// which saves both memory and CPU time.

use binggan::plugins::PeakMemAllocPlugin;
use binggan::{black_box, InputGroup, PeakMemAlloc, INSTRUMENTED_SYSTEM};
use serde_json::json;
use tantivy::aggregation::agg_req::Aggregations;
use tantivy::aggregation::AggregationCollector;
use tantivy::query::AllQuery;
use tantivy::schema::{Schema, FAST, TEXT};
use tantivy::{doc, Index, IndexWriter};

#[global_allocator]
pub static GLOBAL: &PeakMemAlloc<std::alloc::System> = &INSTRUMENTED_SYSTEM;

fn main() {
    // Create test indices with different sizes
    let inputs = vec![
        ("10k_docs", create_test_index(10_000).unwrap()),
        ("100k_docs", create_test_index(100_000).unwrap()),
        ("1m_docs", create_test_index(1_000_000).unwrap()),
    ];

    let mut group = InputGroup::new_with_inputs(inputs);
    group.add_plugin(PeakMemAllocPlugin::new(GLOBAL));

    // Register benchmarks
    group.register("filter_allquery_optimized", |index| {
        filter_allquery_optimized(index);
    });

    group.register("filter_allquery_with_sub_aggs", |index| {
        filter_allquery_with_sub_aggs(index);
    });

    group.register("filter_termquery_baseline", |index| {
        filter_termquery_baseline(index);
    });

    group.register("no_filter_baseline", |index| {
        no_filter_baseline(index);
    });

    group.run();
}

fn create_test_index(num_docs: usize) -> tantivy::Result<Index> {
    let text_field = schema_builder.add_text_field("text", TEXT);
    let score_field = schema_builder.add_u64_field("score", FAST);
    let category_field = schema_builder.add_text_field("category", TEXT | FAST);

    let index = Index::create_in_ram(schema);
    let mut index_writer: IndexWriter = index.writer(50_000_000)?;

    // Add documents
    for i in 0..num_docs {
        index_writer.add_document(doc!(
            text_field => "test document",
            score_field => (i % 1000) as u64,
            category_field => format!("category_{}", i % 10),
        ))?;
    }

    index_writer.commit()?;
    Ok(index)
}

// Benchmark: Filter aggregation with AllQuery (optimized path - no BitSet)
fn filter_allquery_optimized(index: &Index) {
    let agg_req: Aggregations = serde_json::from_value(json!({
        "filtered": {
            "filter": "*",
            "aggs": {
                "avg_score": { "avg": { "field": "score" } }
            }
        }
    }))
    .unwrap();

    let collector = AggregationCollector::from_aggs(agg_req, Default::default());
    black_box(searcher.search(&AllQuery, &collector).unwrap());
}

// Benchmark: Filter aggregation with AllQuery and multiple sub-aggregations
fn filter_allquery_with_sub_aggs(index: &Index) {
    let agg_req: Aggregations = serde_json::from_value(json!({
        "filtered": {
            "filter": "*",
            "aggs": {
                "avg_score": { "avg": { "field": "score" } },
                "max_score": { "max": { "field": "score" } },
                "min_score": { "min": { "field": "score" } },
                "sum_score": { "sum": { "field": "score" } },
                "count_score": { "value_count": { "field": "score" } }
            }
        }
    }))
    .unwrap();

    let collector = AggregationCollector::from_aggs(agg_req, Default::default());
    black_box(searcher.search(&AllQuery, &collector).unwrap());
}

// Benchmark: Filter aggregation with TermQuery (baseline - always uses BitSet)
fn filter_termquery_baseline(index: &Index) {
    let agg_req: Aggregations = serde_json::from_value(json!({
        "filtered": {
            "filter": "text:test",
            "aggs": {
                "avg_score": { "avg": { "field": "score" } }
            }
        }
    }))
    .unwrap();

    let collector = AggregationCollector::from_aggs(agg_req, Default::default());
    black_box(searcher.search(&AllQuery, &collector).unwrap());
}

// Benchmark: No filter aggregation (baseline - direct aggregation)
fn no_filter_baseline(index: &Index) {
    let agg_req: Aggregations = serde_json::from_value(json!({
        "avg_score": { "avg": { "field": "score" } }
    }))
    .unwrap();

    let collector = AggregationCollector::from_aggs(agg_req, Default::default());
    black_box(searcher.search(&AllQuery, &collector).unwrap());
}

```

Then, ran the benchmark twice, once with AllQuery optimization (i.e., [17d5b3f](https://github.com/quickwit-oss/tantivy/pull/2711/commits/17d5b3f7f2ac5420e1ae3c1dcb2ca2f56cc55d85)):
```
1m_docs
filter_allquery_optimized                  Memory: 53.1 KB (+17.24%)    Avg: 2.8051ms (+5.52%)     Median: 2.7943ms (+7.18%)     [2.7116ms .. 3.1025ms]      
filter_allquery_optimized_with_sub_aggs    Memory: 90.1 KB (+53.12%)    Avg: 13.7210ms (+5.73%)    Median: 13.4924ms (+5.98%)    [13.3372ms .. 18.7915ms]    
no_filter_baseline                         Memory: 45.9 KB (+33.29%)    Avg: 2.7527ms (+4.04%)     Median: 2.7369ms (+6.73%)     [2.6691ms .. 3.3107ms]   
```

And once without the AllQuery optimization (via commenting out [these 3 lines](https://github.com/paradedb/tantivy/blob/17d5b3f7f2ac5420e1ae3c1dcb2ca2f56cc55d85/src/aggregation/bucket/filter.rs#L221-L223)):
```
1m_docs
filter_allquery_NOT_optimized                  Memory: 119.1 KB (+124.28%)    Avg: 5.8934ms (+110.09%)    Median: 5.8463ms (+109.22%)    [5.7593ms .. 7.4810ms]      
filter_allquery_NOT_optimized_with_sub_aggs    Memory: 126.2 KB (+40.06%)     Avg: 16.5306ms (+20.48%)    Median: 16.4271ms (+21.75%)    [16.2381ms .. 19.4332ms]    
no_filter_baseline                             Memory: 39.8 KB (-13.38%)      Avg: 2.7175ms (-1.28%)      Median: 2.6844ms (-1.92%)      [2.6135ms .. 3.1884ms] 
```

As you see, it's twice as slow with `AllQuery`, as it does one more pass over the data.

mdashti:
@PSeitz @PSeitz-dd Merged the changes from `main` into this branch. Can you please take another look?

src/aggregation/bucket/filter.rs
mdashti: Re-did the benchmark after [24ab4cb](https://github.com/quickwit-oss/tantivy/pull/2711/commits/24ab4cb1d2c75019e6ffaab8b4ae5c788df97fcc) (i.e., removing the Option wrapper):

```
1m_docs
filter_allquery_optimized        Memory: 115.7 KB (-2.84%)    Avg: 2.9247ms (-50.37%)     Median: 2.7583ms (-52.82%)     [2.6998ms .. 5.0336ms]      
filter_allquery_with_sub_aggs    Memory: 122.8 KB (-2.68%)    Avg: 14.0793ms (-14.83%)    Median: 13.4256ms (-18.27%)    [13.2402ms .. 20.2725ms]    
no_filter_baseline               Memory: 34.5 KB (-13.38%)    Avg: 2.7908ms (+2.69%)      Median: 2.7107ms (+0.98%)      [2.6237ms .. 4.1228ms]     


1m_docs
filter_allquery_NOT_optimized        Memory: 115.7 KB (-0.01%)    Avg: 5.8795ms (+101.03%)    Median: 5.8717ms (+112.87%)    [5.7623ms .. 6.3808ms]      
filter_allquery_NOT_with_sub_aggs    Memory: 122.8 KB (-0.01%)    Avg: 16.4571ms (+16.89%)    Median: 16.4363ms (+22.43%)    [16.1566ms .. 16.9539ms]    
no_filter_baseline                   Memory: 34.5 KB              Avg: 2.7032ms (-3.14%)      Median: 2.6775ms (-1.23%)      [2.6390ms .. 3.0310ms] 
```

mdashti: Conclusion:
 - Let's avoid the Option wrapper to save memory, even though there's no noticeable performance impact.
 - Having the AllQuery optimization doesn't have any noticeable performance overhead, but helps with the performance of auto-generated queries that require the same output structure (~2x).

PSeitz:
The failing CI is a flaky test from me

PSeitz:
> The failing CI is a flaky test from me

Should be fixed now (after rebase)

mdashti:
Thanks @PSeitz. I've also noticed and fixed it in https://github.com/quickwit-oss/tantivy/pull/2723. Closed that PR.
Now, this PR is good for final review.

benches/agg_bench.rs
```diff
@@ -472,3 +478,61 @@ fn get_test_index_bench(cardinality: Cardinality) -> tantivy::Result<Index> {
 
     Ok(index)
 }
+
+// Filter aggregation benchmarks
+
+fn filter_agg_all_query(index: &Index) {
```
PSeitz-dd:
```suggestion
fn filter_agg_all_query_count_agg(index: &Index) {
```

benches/agg_bench.rs
```diff
@@ -472,3 +478,61 @@ fn get_test_index_bench(cardinality: Cardinality) -> tantivy::Result<Index> {
...
+            "filter": "*",
+            "aggs": {
+                "count": { "value_count": { "field": "score" } }
+            }
+        }
+    });
+    execute_agg(index, agg_req);
+}
+
+fn filter_agg_term_query(index: &Index) {
```
PSeitz-dd:
```suggestion
fn filter_agg_term_query_count_agg(index: &Index) {
```

benches/agg_bench.rs
```diff
@@ -472,3 +478,61 @@ fn get_test_index_bench(cardinality: Cardinality) -> tantivy::Result<Index> {
...
+            "filter": "text:cool",
+            "aggs": {
+                "count": { "value_count": { "field": "score" } }
+            }
+        }
+    });
+    execute_agg(index, agg_req);
+}
+
+fn filter_agg_all_query_with_sub_agg(index: &Index) {
```
PSeitz-dd:
```suggestion
fn filter_agg_all_query_with_sub_aggs(index: &Index) {
```

benches/agg_bench.rs
```diff
@@ -472,3 +478,61 @@ fn get_test_index_bench(cardinality: Cardinality) -> tantivy::Result<Index> {
...
+                "terms_text": {
+                    "terms": { "field": "text_few_terms" }
+                }
+            }
+        }
+    });
+    execute_agg(index, agg_req);
+}
+
+fn filter_agg_term_query_with_sub_agg(index: &Index) {
```
PSeitz-dd:
```suggestion
fn filter_agg_term_query_with_sub_aggs(index: &Index) {
```

benches/agg_bench.rs
mdashti: Fixed.

benches/agg_bench.rs
mdashti: Fixed.

benches/agg_bench.rs
mdashti: Fixed.

benches/agg_bench.rs
mdashti: Fixed.

mdashti (commented):
@PSeitz-dd Thanks for the comments. I also noticed there was a bug and fixed it in [94bdd5d](https://github.com/quickwit-oss/tantivy/pull/2711/commits/94bdd5db74a33d5df8a2f3f6eb2240964bbba251)

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,446 @@
...
+    }
+
+    /// Parse the query into a Tantivy Query object
+    ///
+    /// For query strings, this uses the QueryParser::parse_query() method.
+    /// For direct Query objects, returns a clone.
+    fn parse_query(&self, schema: &Schema) -> crate::Result<Box<dyn Query>> {
+        match &self.query {
+            FilterQuery::QueryString(query_str) => {
+                let tokenizer_manager = TokenizerManager::default();
```
PSeitz: The default tokenizer manager will fail for any fields with custom tokenizers. We'll need a mechanism to pass the TokenizerManager in there. 
Probably the same way we pass the aggregations limits, we could put them both in a AggContextParams struct or similar.
```rust
    fn for_segment(
        &self,
        segment_local_id: crate::SegmentOrdinal,
        reader: &crate::SegmentReader,
    ) -> crate::Result<Self::Child> {
        AggregationSegmentCollector::from_agg_req_and_reader(
            &self.agg,
            reader,
            segment_local_id,
            &self.limits,
        )
    }
```

src/aggregation/bucket/filter.rs
PSeitz: The information is required for prefetching, as it is now this would need to be passed through the stack. But there is no mechanism for queries currently to get that info.

So the caller would need to do the collection for filter queries themselves.

src/aggregation/bucket/filter.rs
PSeitz: okay let's keep it

src/query/query_parser/query_parser.rs
```diff
@@ -2026,89 +2026,4 @@ mod test {
             );
         }
     }
-
-    #[test]
-    pub fn test_set_default_field_integer() {
```
PSeitz, line 2031: I think this was removed by accident

tests/filter_aggregation.rs
```diff
@@ -0,0 +1,1013 @@
+//! Test suite for Filter Aggregation
```
PSeitz: can you move this to the filter aggregation implementation?

src/query/query_parser/query_parser.rs
mdashti: Oops. Yeah. Fixed it.

tests/filter_aggregation.rs
mdashti: Done.

src/aggregation/bucket/filter.rs
mdashti: Thanks for catching this. Used the default and forgot to pipe it through. Fixed it.

src/aggregation/bucket/filter.rs
mdashti: Added further comments. IMO, it should be fixed with a broader change in a follow-up PR.

mdashti (commented):
@PSeitz Thanks for the comments. Please take another look.

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,1758 @@
...
+    /// Custom Query object for programmatic query construction
+    ///
+    /// This variant allows passing pre-constructed Query objects directly,
+    /// which is useful for:
+    /// - Custom query types not expressible as query strings
+    /// - Programmatic query construction
+    /// - Extension query types
+    ///
+    /// Note: This variant cannot be serialized to JSON (only QueryString can be serialized)
+    CustomQuery(Box<dyn SerializableQuery>),
```
PSeitz: Why do we use `SerializableQuery` when the query cannot be serialized?

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,1758 @@
...
+        //
+        // For prefetching optimization, callers must analyze the query themselves to
+        // determine fast field usage. This requires:
+        // 1. Parsing the query string to extract field references
+        // 2. Checking the schema to see if those fields are indexed or fast-only
+        // 3. Collecting fast field names for non-indexed fields
+        //
+        // This limitation exists because:
+        // - Query::weight() is called during execution, not during planning
+        // - The fallback decision is made per-segment based on field configuration
```
PSeitz: I think the decision depends on the schema, which is not segment specific

src/aggregation/bucket/filter.rs
PSeitz: I think a query constructor would be more suitable here, than de/serializing runtime objects, which may carry state.

src/aggregation/bucket/filter.rs
mdashti: Fixed the comment in a4df7df8d506843489bc4e7f14ac40f7945f5dec

src/aggregation/bucket/filter.rs
mdashti: Good suggestion to use a query constructor/builder. Fixed it in [4b94636](https://github.com/quickwit-oss/tantivy/pull/2711/commits/4b94636dc23733b5e1cf7ac19fde7a10a014761d)

mdashti (commented):
@PSeitz @PSeitz-dd Thanks for the comments.

mdashti:
@PSeitz @PSeitz-dd can you please take a look this PR again?

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,1922 @@
...
+            }
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for FilterAggregation {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where D: Deserializer<'de> {
+        // Deserialize as query string
+        let query_string = String::deserialize(deserializer)?;
```
PSeitz: So we never deserialize custom queries?

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,1922 @@
...
+            .field("num_matches", &self.bitset.len())
+            .finish()
+    }
+}
+
+/// Segment collector for filter aggregation
+pub struct SegmentFilterCollector {
+    /// Document evaluator for the filter query
+    /// The evaluator internally stores a SegmentReader
+    evaluator: DocumentQueryEvaluator,
```
PSeitz: this should be moved to FilterAggReqData

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,1922 @@
...
+    /// Document count in this bucket
+    doc_count: u64,
+    /// Sub-aggregation collectors
+    sub_aggregations: Option<Box<dyn SegmentAggregationCollector>>,
+    /// Accessor index for this filter aggregation
+    accessor_idx: usize,
+    /// Reusable buffer for matching documents to minimize allocations
+    matching_docs_buffer: Vec<DocId>,
```
PSeitz: this should be moved to FilterAggReqData

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,1922 @@
...
+    /// - Programmatic query construction based on schema
+    /// - Extension query types
+    ///
+    /// **Note**: The builder is serializable, but the cached query is not serialized
+    /// (it's rebuilt on deserialization).
+    CustomBuilder {
+        /// Builder that constructs the query
+        builder: Box<dyn QueryBuilder>,
+        /// Cached query instance (built on first use, not serialized)
+        cached_query: Arc<Mutex<Option<Box<dyn Query>>>>,
```
PSeitz: I don't think we should put cached objects here.
I also don't understand why we would need that, we can just execute the query directly and put the result into `FilterAggReqData` (the precomputed bitset of valid docs)

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,1922 @@
...
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        });
+
+        let aggregations: Aggregations = serde_json::from_value(agg)?;
```
PSeitz, line 1080: All the tests should include a deser roundtrip
```suggestion
        let aggregations: Aggregations = serde_json::from_str(& serde_json::to_string(serde_json::from_value(agg)?).unwrap()).unwrap();
```

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,1894 @@
...
+///     }
+/// }
+///
+/// // Create an instance
+/// let builder = TermQueryBuilder {
+///     field_name: "category".to_string(),
+///     term_text: "electronics".to_string(),
+/// };
+/// ```
+pub trait QueryBuilder: Debug + Send + Sync + erased_serde::Serialize {
```
PSeitz: If we never deserialize this, there is no reason to serialize it

src/aggregation/bucket/filter.rs
```diff
@@ -0,0 +1,1894 @@
...
+        // Verify separate query result matches
+        assert_agg_results!(&separate_result, separate_expected);
+
+        // This test demonstrates that filter aggregation produces the same results
+        // as running a separate query with the same condition
+        Ok(())
+    }
+
+    #[test]
+    fn test_filter_aggregation_performance() -> crate::Result<()> {
```
PSeitz: benchmark in a test doesn't make sense

src/aggregation/bucket/filter.rs
mdashti: It was here to cache a custom query once it's built. But, now with moving the result to `FilterAggReqData`, not needed anymore. Fixed.

src/aggregation/bucket/filter.rs
mdashti: Thanks for the suggestion. Moved.

src/aggregation/bucket/filter.rs
mdashti: Thanks. Moved.

src/aggregation/bucket/filter.rs
mdashti: Done.

src/aggregation/bucket/filter.rs
mdashti: Good point. But I think there's value supporting serialization/deserialization for it. Added it using typetag.

src/aggregation/bucket/filter.rs
mdashti: You're right. We already have a separate benchmark. Removed this.

src/aggregation/bucket/filter.rs
mdashti: As mentioned in a separate comment, added serialization/deserialization using typetag.

src/aggregation/bucket/filter.rs
mdashti: Initially, I thought it's better to cache if we're going to repeatedly build on this builder. But, it's a pre-mature optimization. Removed it.

mdashti (commented):
@PSeitz @PSeitz-dd Thanks for the comments.

mdashti:
@PSeitz @PSeitz-dd Can you please take another look?

PSeitz (approved):

PSeitz:
Thanks!

PR #2717, opened 2025-10-14
Add Filtering for Term Aggregations

### LINKED ISSUE #2702: Add Filtering for Term Aggregations
Currently term aggregation returns all terms in a field. 

For filtered facets `include` (and maybe `exclude`) filters would be necessary.
https://www.elastic.co/docs/reference/aggregations/search-aggregations-bucket-terms-aggregation#_filtering_values_4

## Implementation
One way to achieve this would be to search the term dictionary in the columnar storage upfront for the `include` filter and build a Hashmap for the term ids. This Hashmap can then be used to filter the term ids when reading them.

This would only work for terms, for numbers we don't have a dictionary.

### Problems

The SSTable in the columnar storage only supports exact or starts_with queries effectively, contains queries require a full scan of the dictionary.

PSeitz:
Closes #2702

If a `include` or `exclude` is set, ignore results from column types that are not strings.

### DIFF
--- a/src/aggregation/agg_data.rs
+++ b/src/aggregation/agg_data.rs
@@ -209,6 +209,45 @@ pub struct PerRequestAggSegCtx {
 }
 
 impl PerRequestAggSegCtx {
+    /// Estimate the memory consumption of this struct in bytes.
+    fn get_memory_consumption(&self) -> usize {
+        self.term_req_data
+            .iter()
+            .map(|b| b.as_ref().unwrap().get_memory_consumption())
+            .sum::<usize>()
+            + self
+                .histogram_req_data
+                .iter()
+                .map(|b| b.as_ref().unwrap().get_memory_consumption())
+                .sum::<usize>()
+            + self
+                .range_req_data
+                .iter()
+                .map(|b| b.as_ref().unwrap().get_memory_consumption())
+                .sum::<usize>()
+            + self
+                .stats_metric_req_data
+                .iter()
+                .map(|t| t.get_memory_consumption())
+                .sum::<usize>()
+            + self
+                .cardinality_req_data
+                .iter()
+                .map(|t| t.get_memory_consumption())
+                .sum::<usize>()
+            + self
+                .top_hits_req_data
+                .iter()
+                .map(|t| t.get_memory_consumption())
+                .sum::<usize>()
+            + self
+                .missing_term_req_data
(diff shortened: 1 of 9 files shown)


src/aggregation/bucket/term_agg.rs
```diff
@@ -48,6 +47,8 @@ pub struct TermsAggReqData {
     pub name: String,
     /// The normalized term aggregation request.
     pub req: TermsAggregationInternal,
+    /// Preloaded allowed term ords (string columns only). If set, only ords present are collected.
+    pub allowed_term_ids: Option<BitSet>,
```
rdettai-sk, line 51: a bitset has a moderate memory footprint, but in general should we maybe track the ReqData size as part of the request memory consumption?

PSeitz: yes good point, I'll update the tracking

src/aggregation/agg_data.rs
```diff
@@ -819,6 +876,66 @@ fn build_terms_or_cardinality_nodes(
...
+        let allowed = allowed.as_mut().unwrap();
+        for_each_matching_term_ord(str_col, exclude, |ord| allowed.remove(ord))?;
+    }
+
+    Ok(allowed)
+}
+
+/// Apply a callback to each matching term ordinal for the given include/exclude parameter.
+fn for_each_matching_term_ord(
+    str_col: &StrColumn,
```
fulmicoton, line 909: I think this argument should be the dictionary, not the column.

src/aggregation/agg_data.rs
```diff
@@ -819,6 +876,66 @@ fn build_terms_or_cardinality_nodes(
     Ok(nodes)
 }
 
+/// Builds a single BitSet of allowed term ordinals for a string dictionary column according to
+/// include/exclude parameters.
+fn build_allowed_term_ids_for_str(
+    str_col: &StrColumn,
```
fulmicoton, line 882: I think this argument should be the dictionary, not the column.

fulmicoton (approved):

src/aggregation/agg_data.rs
PSeitz: For that we would need to make `sstable` non-optional or export it from `columnar`. It seems like a potential high impact change and therefore won't include it in this PR.

PR #2718, opened 2025-10-15
Add a fast-field variant of TermSet

stuhood:
The `TermSet` `Query` currently produces one `Scorer`/`DocSet` per matched term by scanning the term dictionary and then consuming posting lists. For very large sets of terms and a fast field, it is faster to scan the fast field column while intersecting with a `HashSet` of (encoded) term values.

Following the pattern set by the two execution modes of `RangeQuery`, this PR introduces a variant of `TermSet` which uses fast fields, and then uses it when there are more than 1024 input terms (an arbitrary threshold!).

Performance is significantly improved for large `TermSet`s of primitives.

### DIFF
--- a/src/query/mod.rs
+++ b/src/query/mod.rs
@@ -22,9 +22,9 @@ mod range_query;
 mod regex_query;
 mod reqopt_scorer;
 mod scorer;
-mod set_query;
 mod size_hint;
 mod term_query;
+mod term_set_query;
 mod union;
 mod weight;
 
@@ -61,8 +61,8 @@ pub use self::regex_query::RegexQuery;
 pub use self::reqopt_scorer::RequiredOptionalScorer;
 pub use self::score_combiner::{DisjunctionMaxCombiner, ScoreCombiner, SumCombiner};
 pub use self::scorer::Scorer;
-pub use self::set_query::TermSetQuery;
 pub use self::term_query::TermQuery;
+pub use self::term_set_query::*;
 pub use self::union::BufferedUnionScorer;
 #[cfg(test)]
 pub use self::vec_docset::VecDocSet;
--- a/src/query/range_query/mod.rs
+++ b/src/query/range_query/mod.rs
@@ -9,7 +9,7 @@ pub use common::bounds::BoundsRange;
 pub use self::range_query::*;
 pub use self::range_query_fastfield::*;
 
-// TODO is this correct?
+// NOTE: Keep in sync with `FastFieldRangeWeight::scorer`.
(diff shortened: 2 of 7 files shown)


PSeitz:
I think a precomputed `BitSet` should also work well, similar to what I did in [regex_phrase_weight](https://github.com/quickwit-oss/tantivy/blob/main/src/query/phrase_query/regex_phrase_weight.rs#L101)

stuhood:
> I think a precomputed `BitSet` should also work well, similar to what I did in [regex_phrase_weight](https://github.com/quickwit-oss/tantivy/blob/main/src/query/phrase_query/regex_phrase_weight.rs#L101)

Hm, interesting.

That could be implemented using either the posting lists or fast fields, I assume? So it seems like it's independent of which index we use.

For very high term counts, I think that you'll still want to scan the fast field, since it is so dense? Requires benchmarking probably.

PSeitz:
> > I think a precomputed `BitSet` should also work well, similar to what I did in [regex_phrase_weight](https://github.com/quickwit-oss/tantivy/blob/main/src/query/phrase_query/regex_phrase_weight.rs#L101)
> 
> Hm, interesting.
> 
> That could be implemented using either the posting lists or fast fields, I assume? So it seems like it's independent of which index we use.
> 
> For very high term counts, I think that you'll still want to scan the fast field, since it is so dense? Requires benchmarking probably.

The BitSet would be precomputed on the inverted index. I think we would need a benchmark to justify the additional code. (also good to know :)

stuhood:
> I think we would need a benchmark to justify the additional code. 

FWIW: The benchmarks that I did in ParadeDB for this PR are over here: https://github.com/paradedb/paradedb/pull/3412 ... for large term sets, consuming the posting lists uses a very large amount of memory, and does a lot of seeking: this implementation is about 7 times faster in one worker process. Can ignore the "multiple workers as pessimization" aspect: that's specific to having a huge query in ParadeDB.

PSeitz:
> > I think we would need a benchmark to justify the additional code.
> 
> FWIW: The benchmarks that I did in ParadeDB for this PR are over here: [paradedb/paradedb#3412](https://github.com/paradedb/paradedb/pull/3412) ... for large term sets, consuming the posting lists uses a very large amount of memory, and does a lot of seeking: this implementation is about 7 times faster in one worker process. Can ignore the "multiple workers as pessimization" aspect: that's specific to having a huge query in ParadeDB.

I mean comparing the BitSet variant (via `BitSetDocSet`) with the fastfield variant.
In the BitSet variant we would fill the BitSet one by one from the postinglists, which would fix the memory usage to `num_docs/8` bytes and remove all seeking.

PR #2719, opened 2025-10-17
fix merge intermediate aggregation results

PSeitz:
Previously the merging relied on the order of the results, which is invalid since https://github.com/quickwit-oss/tantivy/pull/2035. 
This bug is only hit in specific scenarios, when the aggregation collectors are built in a different order on different segments.

### DIFF
--- a/src/aggregation/bucket/term_agg.rs
+++ b/src/aggregation/bucket/term_agg.rs
@@ -781,13 +781,20 @@ mod tests {
     use time::{Date, Month};
 
     use crate::aggregation::agg_req::Aggregations;
+    use crate::aggregation::intermediate_agg_result::{
+        IntermediateAggregationResult, IntermediateAggregationResults,
+    };
     use crate::aggregation::tests::{
         exec_request, exec_request_with_query, exec_request_with_query_and_memory_limit,
         get_test_index_from_terms, get_test_index_from_values_and_terms,
     };
-    use crate::aggregation::AggregationLimitsGuard;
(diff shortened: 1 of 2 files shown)


fulmicoton-dd (approved):

src/aggregation/bucket/term_agg.rs
```diff
@@ -2274,4 +2281,127 @@ mod tests {
...
+        agg_req2.insert(
+            "tags".to_string(),
+            serde_json::from_value(json!({ "terms": { "field": "tags" } }))?,
+        );
+        agg_req2.insert(
+            "hosts".to_string(),
+            serde_json::from_value(json!({ "terms": { "field": "host" } }))?,
+        );
+        // make sure the order of the aggregation request is different
+        assert_ne!(agg_req.keys().next(), agg_req2.keys().next());
```
rdettai-sk, line 2375: @PSeitz this fails on my ARM machine. I don't understand why comparing key ordering on HashMap makes sense here.

PSeitz: I just pushed a fix for this, does it still fail?

rdettai-sk: The test doesn't fail anymore, but I still feel it makes to many assumptions on HashMap's internals.

PSeitz: I think you are right. It should just try instead and if the order is not like expected just quit

PR #2723, opened 2025-10-21
Fixed a flaky test: `terms_aggs_hosts_and_tags_merge_on_mixed_order_request`

mdashti:
## What

Removes a flaky assertion in the `terms_aggs_hosts_and_tags_merge_on_mixed_order_request` test that was causing non-deterministic failures.

## Why

The test attempted to verify that two `HashMap`-based aggregation requests had different iteration orders by asserting `assert_ne!(agg_req.keys().next(), agg_req2.keys().next())`. However, `HashMap` does not guarantee any particular iteration order, making this assertion non-deterministic and prone to failure even when items are inserted in different sequences.

The test was originally added as a regression test for PR #2719, which fixed a bug where `merge_fruits` relied on iteration order (using `.zip()`) instead of key-based lookup. The flaky assertion is not necessary to validate this fix.

## Tests

- Verified the test now passes consistently: `cargo test --lib aggregation::bucket::term_agg::tests::terms_aggs_hosts_and_tags_merge_on_mixed_order_request`
- All other term_agg tests continue to pass

### DIFF
--- a/src/aggregation/bucket/term_agg.rs
+++ b/src/aggregation/bucket/term_agg.rs
@@ -2285,8 +2285,10 @@ mod tests {
     #[test]
     fn terms_aggs_hosts_and_tags_merge_on_mixed_order_request() -> crate::Result<()> {
         // This test ensures that merging of aggregation results works correctly
-        // even if the order of the aggregation requests is different and
-        // running on different indexes with the same data.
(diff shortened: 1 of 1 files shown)


mdashti:
Closed this in favour of https://github.com/quickwit-oss/tantivy/pull/2722

PR #2726, opened 2025-10-30
Lazy scorers

fulmicoton:
Follow up of https://github.com/quickwit-oss/tantivy/pull/2681

- Allow lazy evaluation of score. As soon as we identified that a doc won't
reach the topK threshold, we can stop the evaluation.
- Adds a new type of ordering. Asc, Desc, Ascending but none at the end.
The latter is more natural for most search application, but is not what is common in SQL (which I assume is @stuhood  / paradedb use case)
- Allow for a different segment level score, segment level score and their conversion.
- Rationalization of part of the code / API

This PR breaks public API, but fixing code should be straightforward.


src/collector/top_score_collector.rs
```diff
@@ -1048,6 +1045,46 @@ where
...
+            let feature = score_tweaker.score(doc, score);
+            self.append_within_capacity(ComparableDoc { feature, doc });
+        }
+    }
+}
+
+// Push an element provided there is enough capacity to do so.
+// TODO replace me when push_within_capacity is stabilized.
+#[inline(always)]
+fn push_within_capacity<T>(el: T, buf: &mut Vec<T>) {
```
stuhood: Nit: the upstream method returns `Result<(), T>`... could do that here, and then debug-assert on the result in the caller?

src/collector/tweak_score_top_collector.rs
```diff
@@ -76,49 +128,229 @@ where
...
+    type Fruit = Vec<(TSegmentScoreTweaker::Score, DocAddress)>;
 
     fn collect(&mut self, doc: DocId, score: Score) {
-        let score = self.segment_scorer.score(doc, score);
-        self.segment_collector.collect(doc, score);
+        // Thanks to generics, this if-statement is free.
+        if TSegmentScoreTweaker::is_lazy() {
+            self.segment_collector
+                .topn_computer
+                .push_lazy(doc, score, &mut self.segment_scorer);
```
stuhood: This assumes that the body of `TopSegmentCollector::collect` is empty other than this call, which is a bit fragile.

Does that suggest that `(TopSegment)Collector::collect` should have a lazy variant too? I can't imagine what that would look like.

src/collector/tweak_score_top_collector.rs
```diff
@@ -76,49 +128,229 @@ where
...
+    type SegmentScore = TScore;
+
     fn score(&mut self, doc: DocId, score: Score) -> TScore {
         (self)(doc, score)
     }
+
+    /// Convert a segment level score into the global level score.
+    fn convert_score(&self, score: Self::SegmentScore) -> Self::Score {
+        score
+    }
```
stuhood: Nit: could have a name which better describes what it does.

stuhood (commented):
I think that this interface works! Thanks a lot for exploring it.

My only suggestion would be to break interfaces a little bit harder (naming wise), if possible: the capabilities exposed by this API go way beyond "tweaking a score", and into sorting on arbitrary features.

src/collector/tweak_score_top_collector.rs
fulmicoton-dd: agreed

src/collector/top_score_collector.rs
```diff
@@ -1040,14 +748,65 @@ where
...
+        score_tweaker: &mut TSegmentSortKeyComputer,
+    ) {
+        if TSegmentSortKeyComputer::is_lazy() {
+            if let Some(threshold) = self.threshold.as_ref() {
+                let Some((_cmp, feature)) =
+                    score_tweaker.accept_sort_key_lazy::<REVERSE_ORDER>(doc, score, threshold)
+                else {
+                    return;
+                };
+                push_assuming_capacity(
```
stuhood: This method needs a call to `self.truncate_top_n()` before this call.

stuhood: I think that right now this isn't covered by a test because there are no tests using:
```rust
impl<HeadSortKeyComputer, TailSortKeyComputer> SortKeyComputer
    for (HeadSortKeyComputer, TailSortKeyComputer)
```
... so there are no implementations which are `is_lazy = true`.

src/collector/sort_key/sort_key_computer.rs
```diff
@@ -0,0 +1,361 @@
...
+        &mut self,
+        doc_id: DocId,
+        score: Score,
+        threshold: &Self::SegmentSortKey,
+    ) -> Option<(std::cmp::Ordering, Self::SegmentSortKey)> {
+        let excluded_ordering = if REVERSE_ORDER {
+            Ordering::Greater
+        } else {
+            Ordering::Less
+        };
```
stuhood: Based on some testing (hardcoding `is_lazy = true` and then increasing the testcase size of `test_top_field_collect_string_prop`), these cases are reversed.

Lifting over some of the tests from #2681 might be a good way to get some coverage of the multi-sort key cases quickly.

stuhood:
I made a sketch of how this design could support implementing `Collector::collect_block` over here: https://github.com/quickwit-oss/tantivy/pull/2728 ... seems promising, thanks!

fulmicoton:
@stuhood Can you review for good now? I did not integrate your work on batching yet. I think this can be done in a separate PR. 

The PR ended up larger, because I added the flexibility to deal with ordering equivalent to 
`ORDER BY col DESC NULLS LAST` as well as `ORDER BY col DESC` to be compatible with both Quickwit and ParadeDB.

src/collector/sort_key/sort_key_computer.rs
```diff
@@ -0,0 +1,631 @@
...
+    }
+
+    /// Indicates whether the sort key actually uses the similarity score (by default BM25).
+    /// If set to false, the similary score might not be computed (as an optimization),
+    /// and the score fed in the segment sort key computer could take any value.
+    fn requires_scoring(&self) -> bool {
+        false
+    }
+
+    /// Sorting by score is special in that it allows for the Block-Wand optimization.
```
stuhood: This comment is specific to the overridden method for scores.

src/collector/sort_key/sort_by_static_fast_value.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use crate::collector::sort_key::NaturalComparator;
+use crate::collector::{SegmentSortKeyComputer, SortKeyComputer};
+use crate::fastfield::{FastFieldNotAvailableError, FastValue};
+use crate::{DocId, Score, SegmentReader};
+
+/// Sorts by a fast value (u64, i64, f64, bool).
+///
+/// The field must appear explicitly in the schema, with the right type, and declared as
+/// a fast field..
```
stuhood, line 13:
```suggestion
/// a fast field.
```

src/collector/sort_key/sort_by_static_fast_value.rs
```diff
@@ -0,0 +1,98 @@
...
+///
+/// If the field is multivalued, only the first value is considered.
+///
+/// Document that do not have this value are still considered.
```
stuhood:
```suggestion
/// Documents that do not have this value are still considered.
```

src/collector/sort_key/sort_by_string.rs
```diff
@@ -0,0 +1,72 @@
...
+use crate::{DocId, Score};
+
+/// Sort by the first value of a string column.
+///
+/// The string can be dynamic (coming from a json field)
+/// or static (being specificaly defined in the configuration).
+///
+/// If the field is multivalued, only the first value is considered.
+///
+/// Document that do not have this value are still considered.
```
stuhood:
```suggestion
/// Documents that do not have this value are still considered.
```

src/collector/sort_key/sort_by_score.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    // Sorting by score is special in that it allows for the Block-Wand optimization.
+    fn collect_segment_top_k(
+        &self,
+        k: usize,
+        weight: &dyn crate::query::Weight,
+        reader: &crate::SegmentReader,
+        segment_ord: u32,
+    ) -> crate::Result<Vec<(Self::SortKey, DocAddress)>> {
```
stuhood, line 34: I completely agree with not looping the batching into this change.

One thing that I wonder is whether batched, lazy collection could be used to implement block-wand, which might make scores less of a special case! Something to experiment with in the future.

src/collector/sort_key/sort_by_score.rs
```diff
@@ -0,0 +1,77 @@
+use crate::collector::sort_key::NaturalComparator;
+use crate::collector::{SegmentSortKeyComputer, SortKeyComputer, TopNComputer};
+use crate::{DocAddress, DocId, Score};
+
+/// Sort by similarity score.
+#[derive(Clone, Debug, Copy)]
+pub struct SortBySimilarityScore;
```
stuhood, line 7: Naming nit: here we spell out "similarity score", whereas with `TopDocs::order_by_score` and the other method removals/changes, I think that we recognize rightly that a "score" should probably _only_ mean a "similarity score". So I think that you could probably drop `Similarity` from the name here, which would make this more symmetrical with `order_by_score`.

stuhood (approved):
Thanks so much for tackling this! The reduction in API surface area is really wonderful, and it looks like this enables the boxed/erased features in a cleaner way than #2681 did too.

Also, based on another experiment I was doing: I like removing `impl Collector for TopDocs` (despite the API change), because it cleans up the builder interface to not have to carry the generic type around.

Thanks again!

stuhood:
There is about a 60% regression on TweakScore and about 15% on string sorts; other than that our benchmarks look good.

Will triage those today.

stuhood:
> There is about a 60% regression on TweakScore and about 15% on string sorts; other than that our benchmarks look good.

@fulmicoton, @PSeitz: Both of these appear to be https://github.com/quickwit-oss/tantivy/issues/2775. Please let me know your thoughts there!

PR #2727, opened 2025-11-01
Implement `collect_block` for `Collector`s which wrap other `Collector`s

stuhood:
The:
* tuple `Collector`s
* `MultiCollector`
* `FilterCollector`s

...do not have explicit overrides of `collect_block`, meaning that wrapping a `Collector` which _does_ implement that method loses its advantage.

Noticed in a profile while using a tuple `Collector` of `TopDocs` and `Aggregations` (because `Aggregations` take advantage of `collect_block`).


fulmicoton (approved):
Thank you!

PR #2728, opened 2025-11-03
Implement `collect_block` for lazy scorers.

stuhood:
The new API proposed in #2726 enables implementing `Collector::collect_block`, which allows for slightly more efficient bulk pushes to the `TopNComputer` when a `SegmentSortKeyComputer::is_lazy`, and when scores are not required.

We can:
* Check capacity and truncate once for the entire block.
* Lift the check for the presence of a threshold out of the loop over the block.

It would additionally allow for bulk fetching (using `ColumnValues::first_vals`) the first sort key column, as future work. 

I haven't benchmarked this yet! But `collect_block` is quite beneficial for aggregations.


fulmicoton-dd:
@stuhood @philippemnoel In case there was an ambiguity here. I did not close this branch. I think this was closed automatically after I merged/deleted lazy-scorers. This is a good idea, and we should ship it.

If you can reopen a PR I'd be happy to review it. (I cannot reopen that one)

stuhood:
Great, thanks! I'll revisit this this month. We also have an idea about threshold sharing that we might pursue first.

stuhood:
> @stuhood @philippemnoel In case there was an ambiguity here. I did not close this branch. I think this was closed automatically after I merged/deleted lazy-scorers. This is a good idea, and we should ship it.

Spent some time over the holiday expanding this patch to experiment with predicate-pushdown into `Column`: https://github.com/quickwit-oss/tantivy/pull/2793

Issue #44, opened 2016-10-17
Geo-search


fulmicoton:
This would require a longlat field type... (and coords if we want simple 2D as well

Then the user would want to perform queries such as 
- return document at a distance of less than X km
- return document within this polygon

Fiedzia:
Two more practical requirements that I would love to have:

1. Store multiple values
2. Return *which* value was found

those two things have a range of usecases that are very practical for me

hwchen:
Curious about the status of this issue. Is this still a desired feature?

I have a need for geo queries in a future project, and would be willing to help implement, although I might need some guidance.

fulmicoton:
@hwchen Sorry about the late answer

This is a desired feature but this is also a lot of work.
I'm thinking of following Lucene idea of using Bkd Tree as it fits well the search indexing scheme.

You can work on it if you are confident, but be careful... This is one of the hardest ticket around and
I won't be able to give much guidance.

hwchen:
Thanks, I'll keep this in mind. Maybe in a year or so I'll pick this up if nobody else has :)

natew:
Plus one on this feature, very relevant to any local app - Uber Eats, Yelp, Maps, Delivery.

shuoli84:
FYI: did a quick search, found one nice bkd tree impl https://github.com/peermaps/eyros

PSeitz:
Nice, rucene also has a bkd implementation https://github.com/zhihu/rucene/blob/master/src/core/util/bkd/bkd_writer.rs

shuoli84:
> FYI: did a quick search, found one nice bkd tree impl https://github.com/peermaps/eyros

Hmm, seems the license of eyros is not compatible with MIT.

shuoli84:
Another FYI.

From https://www.elastic.co/blog/lucene-points-6.0, the blog which announced lucene spatial support, there is a quote on future work:

> Currently, dimensional values must be single points, but some geo-spatial use cases require indexing shapes as well, which can be done with a generalization of the k-d tree called an R-Tree, for example, and I expect that will be a future improvement to Lucene (patches welcome again!).

And there is a rtree crate called rstar, which provides a good  in mem rtree. https://crates.io/crates/rstar

Jure-BB:
Point search specific methods are easier to implement since they don't require specialized trees and they also perform better. It would make sense to implement point search first as it is probably most commonly used, and implement indexing of shapes at a latter time. 

[...]

PR #2729, opened 2025-11-04
Geo-search

flatheadmill:
* Regarding #44.
---
This pull request adds spatial indexing to Tantivy using block kd-trees. The implementation tessellates polygons into triangles and indexes the triangles in the block kd-tree.

Search supports three query types: "intersects" finds documents whose geometries overlap a query rectangle, "within" finds documents whose geometries fall entirely inside the query bounds, and "contains" finds documents whose geometries completely enclose the query region. I've attempted to follow Tantivy's established patterns. Spatial fields use a new `FieldType::Spatial` variant. I used `PreTokenizedString` as a go-by.

The user provided `f64` geographic coordinates are serialized using XOR compression that exploits spatial locality. When consecutive vertices are close together, their bit representations share high-order bits, and XOR reveals this redundancy as zeros that varint encoding can compress. The compression falls back to uncompressed when data is incompressible. Looked at alternatives, `bitshuffle+zstd` might be nice for large polygons, but small polygons would be smaller than a `zstd` block. The XOR is simple and a good place to start.

The block kd-tree implementation uses bulk-loaded immutable construction with recursive median partitioning, creating a somewhat balanced block kd-tree with 512 triangles per leaf. The tree stores triangles. Polygons are tessellated into triangles prior to indexing. `f64` dimensions are converted to `i32` prior to tessellation and the block kd-tree stores the dimensions as `i32`. Triangles are stored in an encoded format that places a bounding box for the triangle in the first four words, followed by the spare x/y, boundary flags and finally the doc_id. The boundary flags indicate if a side of the triangle shares a boundary with the tessellated polygon and is used for "within" and "contains" testing.

The tree construction uses Rust's standard library `select_nth_unstable_by_key` for median partitioning, operating directly on `&mut [Triangle]` slices. Tree construction is a simple recursive descent based around this standard library function. The recursive descent receives slices of progressively smaller sub-ranges until reaching leaf size. This approach works identically whether the triangles come from a `Vec` during initial indexing or from a memory-mapped file during merge.

Merge is implemented by writing these `Triangle` structures into a temporary file, memory mapping the temporary file, and then performing an unsafe cast of the memory to `&mut [Traingle]`. Because the file is memory mapped, the `select_nth_unstable_by_key` tree construction can rely on the operating system to manage the memory when merging large segments.

If you are zealously opposed to unsafe code then `&mut [Triangle]` and `select_nth_unstable_by_key` have to be rewritten, but I don't see how it is a problem. The `Vec<Triangle>` is Rust-safe. The temporary file is entirely implemented as a transient step in the block kd-tree merge. Ought to be able to add a compile switch for endianess. It is not a persistent file. It is temporary only for the merge. It doesn't not have to have the same endianess on different architectures.

This implementation is certainly Lucene inspired. The triangle encoding is a 1:1 port with the exception of replacing orient with the robust orient2d implementation of Shewchuk's 2d orient. The rest is more inspiration than port. Porting from Lucene 1:1 not sensible. Lucene needs to operate on mapped memory via Java arrays, it goes to great lengths to avoid the creation of POJOs to keep the garbage collector out of the picture. The Rust implementation can read more like Rust.

`f64` to `i32` is Lucene inspired, roughly centimeter precision. The bulk-loaded partition construction is Lucene inspired, but the in place sorting and reuse of the standard library is a feature of Rust. Delta encoding for `i32`/`u32` is Lucene inspired. Currently using barycentric point-in-triangle instead of Lucene's orientation tests.

Dependences added are [`robust`](https://github.com/georust/robust) for `orient2d` and [`i_triangle`](https://github.com/iShape-Rust/iTriangle) for its integer-based Delaunay triangulation that tracks boundary edges.

I have avoiding making anything "pluggable" and bonded tightly to Tantivy and the chosen 3rd-party libraries for simplicity and easy reading.

---

In the above you'll see a discussion of how the use of `select_nth_unstable_by_key` with `&mut [Triangle]` against a memory mapped file that has been cast to `&mut [Triangle]`, please consider it.

Please consider the minimalist Geometry enum. Rather than create yet another object hierarchy of `Point`, `Polygon`, `BoundingBox`, et. al. I used underlying representations that can be easily handed off to GeoRust or similar. I didn't want to make GeoRust or GeoJSON parsing a Tantivy dependency. The block kd-tree is indented for bounding box and point queries, not for the full range of queries you would find in PostGIS, like polygon in polygon. Such a test could be performed while iterating the search results.

Note that the block kd-tree is not going work with geometries that cross the antimeridian, but this is a known issue, and data sets will probably already split a polygon that crosses the antimeridian into a multi-polygon with a polygon for each side of the antimeridian. This is even in the GeoJSON spec.

Note that there is a Spatial field and a Geometry type stored in that field. Would it be better to have a Geographic or Geo field since the implementation is geo specific?

As I write this, "contains" and "within" are implemented in the block kd-tree but not exposed through the interfaces. I will implement "disjoint." There are `todo!()`s to implement in the `Field::Spatial(_)` match arms. There are some tree balancing improvements that I'd like to add. I'd like to put a lot of Geofabrik data through it and see how well it performs. However, it is ready for consideration by the Tantivy maintainers.

---

Run `cargo run --example geo_json`, a minimal round-trip through the index.


fulmicoton:
@flatheadmill This is a super interesting PR. A bit of contributions has accumulated, so it might take a bit of time to get it reviewed. To make things smoother, would you be ok for a quick call to walk me through the code?

flatheadmill:
Of course. Doesn't have to be quick for my sake. Check your email.

flatheadmill:
For all the talk of k-dimensions, the end result of writing the tree is a run-of-the-mill binary tree whose search functions are very easy to understand. In the end, we have a binary tree of inner nodes that have a left and right pointer to child nodes and a bounding box. During a search the bounding box allows us to eliminate sub-trees that do not intersect. The binary tree terminates in leaf notes that have a bounding box and a pointer to where triangles are stored. If the path you follow intersects your sought bounding box, you scan the triangles referenced by the leaf node.

---

Triangles? Yes, the polygons you add to the index are tessellated into triangles. The i_triangle crate used for tessellation in this pull request has an animation of triangulation in its [GitHub README](https://github.com/iShape-Rust/iTriangle). The animation is all you need to understand what we're trying to accomplish. We are tokenizing a polygon, turning it into triangles. We know we've matched a document if we match a triangle in the polygon. Triangles are described with three points and therefore easy to store. It was not necessary to study tessellation for this implementation, it was enough to simply choose an implementation that met requirements; it works on integers and tracks boundary edges.

With the polygon broken down into triangles we can store a triangle and the `DocId` associated with the polygon. We also store boundary edges that the triangle shares with the tessellated polygon. We can test if a polygon is within a bounding box by testing if a boundary edge crosses the bounding box. If it does, the polygon is marked as excluded, if it doesn't it is included. Take the difference for the set of documents within the bounding box. For intersection all we need to know is whether the bounding box overlaps the triangle in any way.

In this way, this pull request implements an orthogonal range query to search for 2-dimensional shapes within axis-aligned partitions, i.e. bounding-box search.

---

To grasp the kd-concept quickly, it is enough to see [a simple tree and the space it partitions](https://courses.grainger.illinois.edu/cs225/sp2019/notes/kd-tree/). Otherwise, you could read a [Medium article](https://medium.com/@nickgerleman/the-bkd-tree-da19cf9493fb). You will come away with an understanding of how to partition two-dimensional space. You will know how to search for a _point_ in a two-dimensional kd-tree. It's a matter of descending the tree and choosing a path based on the depth of the tree. There are plenty of little examples of this 2-dimensional kd-tree of points in every programming language.

You will be left wondering how to search for a _triangle_ in two-dimensional space. Nowhere will you find a description of how to do this. A search for answers is going to trip over the fact that the word "dimension" is overloaded. Our minimal kd-tree stores 1-dimensional spatial data in a tree that it describes as 2-dimensional. When we use the term "dimension"  to describe a kd-tree we are describing the number of partitioning dimensions of the tree, not the spatial dimension of the indexed forms.

What I found, what finally clicked, was a single "teach a man to fish" answer to a [StackOverflow question](https://stackoverflow.com/a/17658530). A 4-dimensional point can represent a bounding box where the partitioning dimentions are the min and max and the x and y. I wrote a program to create a kd-tree with 4-dimensions and was able to search for a box with a box "using the range (a0,+inf) x (-inf, a1) x (b0, +inf) x (-inf, b1)".

Meanwhile, I had been stepping through the Lucene search where I'd see the bounding boxes in the nodes during search and knew that Lucene was depending on the properties of a kd-tree to build the tree, but not to search the tree. It did not search the tree using a 4-dimensional point, it used bounding boxes instead. It would build the bounding boxes as it partitioned the tree.

---

Another muddle on the way to understanding is the concept of a "block" kd-tree. Seeing `BKDReader.java` and `BKDWriter.java` in Lucene implied that Lucene has implemented a BKD-tree as defined in [Bkd-Tree: A Dynamic Scalable kd-Tree](https://users.cs.duke.edu/~pankaj/publications/papers/bkd-sstd.pdf) and so I assumed that whatever was in the paper was in Lucene. The paper describes an optimization based on creating LSM-like sets of progressively larger of kd-trees. Kd-trees are difficult to update, which is what this paper addresses, but Lucene and Tantivy already address this problem with segments. There's only one kd-tree for a spatial field per segment and that tree is rewritten when segments merge.

I never looked at the paper again once I began stepping through the Lucene code. There I saw the bounding box in kd-tree node used for search and affirmed my hyper-point as bounding box understanding by seeing that Lucene was indeed indexing the 2-demential (spatial) triangle bounding box using a 4-dimensional (partitions) tree. I saw too, that a single tree was built by bulk-partitioning. I never referenced the original paper myself, but here I am standing on the shoulders of giants, reading the theory in practice, so I'm not the one to ask how the paper influenced the design of Lucene, I can only talk about how Lucene influenced the design of this pull request.

However, since we are not creating a LSM-like structure nor using the serialization strategy presented in the paper, I'm going to rename `bkd.rs` to `kd.rs` to put an end to red herring for future readers. The term "block" in the paper references an I/O optimization to the serialized tree structure that is not employed by this pull request nor Lucene.

---

That's enough red herrings. The tree is a 4-dimensional kd-tree built with a recursive partitioning of an array of triangles. The 4-dimensions represent the four positions of the two points in a bounding box. We partition our array of triangles on the kd-tree dimension with the widest spread, the greatest distance between the min and max value. We calculate a bounding box to encapsulate the triangles in the array and then call our partition function on the left and right partitions. We create an inner node that has a bounding box and a left and right child reference. When an array is less than a certain size, that's a leaf.

If you look at the `write_leaf_pages` function, you'll see the partitioning. As of this commit, that is all the k-dimensional code, the rest is bounding box based. In fact, kd-tree is probably a misnomer. We're building a binary bounding box tree (a name I just now made up) using k-dimensional partitioning.

---

P.S. &mdash; As you can see, the implementation in this pull request does not implement a generic kd-tree like Lucene. We do not keep the partition dimension in the node, we throw it away and use just the bounding box. Lucene uses its kd-tree for range queries, but I believe Tantivy has "fast fields" for that use case. It has some code to do kNN with kd-trees, but the primary index it uses for kNN is a Hierarchical Navigable Small World index. I'd speculate that the kd-tree kNN was an offering that was superseded by HNSW.

src/indexer/merger.rs
```diff
@@ -520,6 +525,71 @@ impl IndexMerger {
...
+                use memmap2::MmapOptions;
+                let mmap = unsafe { MmapOptions::new().map_mut(temp_file.as_file())? };
+                // Cast to &[Triangle] slice
+                let triangle_count = mmap.len() / std::mem::size_of::<Triangle>();
+                let triangles = unsafe {
+                    std::slice::from_raw_parts_mut(mmap.as_ptr() as *mut Triangle, triangle_count)
+                };
+                // Get spatial writer and rebuild block kd-tree.
+                spatial_serializer.serialize_field(field, triangles)?;
+            }
```
fulmicoton: FYI  I imported your PR in https://github.com/quickwit-oss/tantivy/pull/2755 and started reviewing.

The test do not pass in your PR mainly due to :


```suggestion
            }
            spatial_serializer.close()?;
```

fulmicoton:
@flatheadmill Hello. I had time to have a deeper look at your PR.

For large PR like this, I usually modify the code as I look around.
You can check that work there:
https://github.com/quickwit-oss/tantivy/pull/2755/files

I will put my comments inline, you can also just pick the commits from #2755 when it makes sense to you.

src/index/index_meta.rs
```diff
@@ -142,6 +142,7 @@ impl SegmentMeta {
             SegmentComponent::FastFields => ".fast".to_string(),
             SegmentComponent::FieldNorms => ".fieldnorm".to_string(),
             SegmentComponent::Delete => format!(".{}.del", self.delete_opstamp().unwrap_or(0)),
+            SegmentComponent::Spatial => ".spatial".to_string(),
```
fulmicoton: General thing...

I think we should avoid creating this file if it is not used.
It might seem benigne, but tantivy creates those files on every single commit. Ideally we would fix that at the directory level, but I never managed to do it.

TBH it should be true for other components in tantivy (especially field norms).

I think I managed to get that in my #2755 .

src/indexer/merger.rs
```diff
@@ -520,6 +525,71 @@ impl IndexMerger {
...
+                        }
+                    }
+                }
+            }
+        }
+        if let Some(mut spatial_serializer) = serializer.extract_spatial_serializer() {
+            for (field, mut temp_file) in temp_files {
+                // Flush and sync triangles.
+                temp_file.flush()?;
+                temp_file.as_file_mut().sync_all()?;
```
fulmicoton: fsyncing is not useful for this usage.

src/indexer/merger.rs
```diff
@@ -520,6 +525,71 @@ impl IndexMerger {
...
+        let mut temp_files: HashMap<Field, NamedTempFile> = HashMap::new();
+
+        for (field, field_entry) in self.schema.fields() {
+            if matches!(field_entry.field_type(), FieldType::Spatial(_)) {
+                temp_files.insert(field, NamedTempFile::new()?);
+            }
+        }
+        for (segment_ord, reader) in self.readers.iter().enumerate() {
+            for (field, temp_file) in &mut temp_files {
+                let spatial_readers = reader.spatial_fields();
```
fulmicoton: we need the temp_file to be buffered.

src/query/spatial_query.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+pub struct SpatialWeight {
+    field: Field,
+    bounds: [(i32, i32); 2],
+    query_type: SpatialQueryType,
+}
+
+impl SpatialWeight {
+    fn new(field: Field, bounds: [(i32, i32); 2], query_type: SpatialQueryType) -> Self {
```
fulmicoton: I think (?) the integer bbox you end up using assume that the two points given here are the min corner and the max corner.
(because of the intersect code)

It might be worth enforcing that somewhere to avoid bugs.

Ideally I would prefer to have a IBox struct type that forces that invariant upon construction.

src/spatial/triangle.rs
```diff
@@ -0,0 +1,421 @@
...
+                    cy = by;
+                    ca = bc;
+                    bx = temp_x;
+                    by = temp_y;
+                    bc = temp_boundary;
+                }
+            }
+        }
+        // change orientation if CW
+        if orient2d(
```
fulmicoton: robust is imported just for orient2d.

I think we should remove the dep and use an adhoc implementation.

orient_2d is also needlessly complex. (We just want the sign of the determinant)

src/spatial/triangle.rs
```diff
@@ -0,0 +1,421 @@
...
+            Coord { y: ay, x: ax },
+            Coord { y: by, x: bx },
+            Coord { y: cy, x: cx },
+        ) < 0.0
+        {
+            let temp_x = bx;
```
fulmicoton:
```suggestion
            // We fix the orientation by swapping B and C.
            let temp_x = bx;
```

src/spatial/triangle.rs
```diff
@@ -0,0 +1,421 @@
...
+            let temp_y = by;
+            let temp_boundary = ab;
+            // ax and ay do not change, ab becomes bc
+            ab = bc;
```
fulmicoton: I believe this is a bug:
```suggestion
            ab = ac;
```

src/spatial/triangle.rs
```diff
@@ -0,0 +1,421 @@
...
+
+impl Triangle {
+    /// Encodes a triangle with the bounding box in the first four words for efficient spatial
+    /// pruning.
+    ///
+    /// Takes three vertices as `[y0, x0, y1, x1, y2, x2]` and edge boundary flags `[ab, bc, ca]`
+    /// indicating which edges are polygon boundaries. Returns a triangle struct with the bounding
+    /// box in the first four words as `[min_y, min_x, max_y, max_x]`. When decoded, the vertex
+    /// order may differ from the original input to `new()` due to normalized rotation.
+    pub fn new(doc_id: u32, triangle: [i32; 6], boundaries: [bool; 3]) -> Self {
```
fulmicoton: In several place in the code, we rely on implicit ordering to have `lon, lat` or `y, x`.

The code would benefit from introducing

`GeoPoint { x: f64, y: f64}`
and `IPoint { x: u32, u32 },
and `IBox { min_corner: IPoint, max_corner: IPoint }` or `IBox {min_x, min_y, max_x, max_y}`
This is zero cost in rust.

fulmicoton (changes requested):

flatheadmill:
Was I not supposed to cherry-pick your changes into the PR @fulmicoton ? The @fulmicoton-dd user is unexpected.

fulmicoton-dd:
@flatheadmill Yes if you can cherry-pick that works! Sorry both fulmicoton and fulmicoton-dd are myself. I had to create a separate work account for silly reasons.

fulmicoton-dd:
@philippemnoel I see that that you spotted that PR. One thing that might be missing here is support for GIS projections.

This PR treats geo search using long lat coordinates.
You can expect paranormal phenomenons near the poles, etc.
What you have in Postgres uses GIS projections. Maybe we should integrate that in tantivy (behind a feature flag to avoid bringing too many dependencies) or by adding a way to provide the mapping `(lat, lon) <-> (x, y)` as a trait.

philippemnoel:
> @philippemnoel I see that that you spotted that PR. One thing that might be missing here is support for GIS projections.
> 
> This PR treats geo search using long lat coordinates. You can expect paranormal phenomenons near the poles, etc. What you have in Postgres uses GIS projections. Maybe we should integrate that in tantivy (behind a feature flag to avoid bringing too many dependencies) or by adding a way to provide the mapping `(lat, lon) <-> (x, y)` as a trait.

Interesting. Yeah, Geo search has been requested a few times. In Postgres, PostGIS is very popular and we already package it as part of our Docker image, but finding a way to integrate it with Tantivy (or replace its functioality with Tantivy's Geo search) would be desirable to our users: #3719. A feature flag sounds like a great idea.

flatheadmill:
I've created a [demo program](https://github.com/flatheadmill/geo) with which I've indexed the GeoFabric OSM data for the United States.

PR #2736, opened 2025-11-18
fix: fixed nested aggregations deserialization

mdashti:
## What

Fixed deserialization of nested aggregations when the `aggs` field is embedded inside an aggregation variant's configuration object.

## Why

Previously, aggregation requests with nested `aggs` like `{"terms": {"field": "category", "aggs": {...}}}` failed to deserialize correctly. The nested aggregations were being ignored because the deserializer expected `aggs` at the top level, not inside the aggregation variant's config.

## How

Modified the `TryFrom<AggregationForDeserialization>` implementation to:
1. Scan the aggregation config object for nested `aggs` fields
2. Extract and parse them as sub-aggregations before deserializing the main variant
3. Assign the extracted nested aggregations to the `sub_aggregation` field

The implementation uses a functional approach to find and extract the first nested `aggs` field from the aggregation configuration.

## Tests

Added three test cases covering:
- Basic nested aggregations (terms → terms)
- Triple-nested aggregations (terms → terms → terms)
- Nested aggregations with metric sub-aggregations (terms → avg/max)

### DIFF
--- a/src/aggregation/agg_req.rs
+++ b/src/aggregation/agg_req.rs
@@ -81,7 +81,7 @@ impl TryFrom<AggregationForDeserialization> for Aggregation {
             mut aggs_remaining_json,
             mut sub_aggregation,
         } = value;
-        
+
         // Extract nested "aggs" from inside the aggregation variant
         // E.g., {"terms": {"field": "x", "aggs": {...}}} -> extract "aggs" from inside "terms"
         if let Some(obj) = aggs_remaining_json.as_object_mut() {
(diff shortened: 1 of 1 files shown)


trinity-1686a:
is that a standard syntax somewhere? I couldn't get ES to accept such an aggregation either, it complains with `x_content_parse exception: [terms] unknown field [aggs]`

mdashti:
@trinity-1686a Thanks for the comment. I actually realized I was expecting a wrong syntax to work. Closing this PR.

PR #2744, opened 2025-11-24
split Term in Term and IndexingTerm

PSeitz-dd:
split Term in Term and IndexingTerm
add JsonTermSerializer in json_postings_writer.rs

Follow up would be to replace the serialized form of Term with regular types, e.g. `Field`

minimal version of https://github.com/quickwit-oss/tantivy/pull/2366

### DIFF
--- a/src/postings/json_postings_writer.rs
+++ b/src/postings/json_postings_writer.rs
@@ -72,8 +72,7 @@ impl<Rec: Recorder> PostingsWriter for JsonPostingsWriter<Rec> {
         for (_field, path_id, term, addr) in ordered_term_addrs {
             if prev_term_id != path_id.path_id() {
                 term_buffer.clear();
-                term_buffer.append_path(ordered_id_to_path[path_id.path_id() as usize].as_bytes());
-                term_buffer.append_bytes(&[JSON_END_OF_PATH]);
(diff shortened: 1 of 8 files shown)


src/postings/json_postings_writer.rs
```diff
@@ -64,40 +65,39 @@ impl<Rec: Recorder> PostingsWriter for JsonPostingsWriter<Rec> {
...
         let mut buffer_lender = BufferLender::default();
-        term_buffer.clear_with_field_and_type(Type::Json, Field::from_field_id(0));
         let mut prev_term_id = u32::MAX;
         let mut term_path_len = 0; // this will be set in the first iteration
         for (_field, path_id, term, addr) in ordered_term_addrs {
             if prev_term_id != path_id.path_id() {
-                term_buffer.truncate_value_bytes(0);
+                term_buffer.clear();
                 term_buffer.append_path(ordered_id_to_path[path_id.path_id() as usize].as_bytes());
                 term_buffer.append_bytes(&[JSON_END_OF_PATH]);
```
fulmicoton-dd: maybe we should join this two calls and put that in term_buffer?

fulmicoton-dd (approved):

PR #2746, opened 2025-11-25
Add polish stemmer proposal

ptr07:
## What
This commit adds support for Polish language stemming.

## Why
The previously used rust-stemmers crate is abandoned and unmaintained, which blocked the addition of new languages. This change addresses a user request for Polish stemming to improve BM25 recall in their use case. The tantivy-stemmers crate is a modern, maintained alternative that also opens the door for supporting many other languages in the future.

## How
- Added the tantivy-stemmers crate as a dependency to the workspace, alongside the existing rust-stemmers dependency (for backward compatibility)
- Introduced an internal enum that can hold an algorithm from either rust-stemmers or tantivy-stemmers
- Added Polish to the main Language enum, mapped to the new tantivy-stemmers implementation
- Updated the token stream to handle both types of stemmers internally
- Added the POLISH variant to the stopwords list

## Tests
- Existing tests pass
- Added test_pl_tokenizer to verify that the Polish stemmer works correctly


fulmicoton:
Closing

I am not fond of the "let's depend on two different crates" approach, especially considering it is still possible to use tantivy with polish in any program by defining your own tokenizer.

I'm ok with switching to your crate though if you want to modify your PR to migrate all stemmers to yours.

PR #2750, opened 2025-11-27
Term and IndexingTerm cleanup

PSeitz:
Replace serialized field with proper `Field` in Term.
mark `Term::serialized_term` and `Term::wrap` as deprecated (to make the transition smooth: used in Quickwit)
add `Serialize` `Deserialize` to `Term`

Remove unused type tag in IndexingTerm (saves 1 byte per term in the termmap)
remove generic in `Term` and use derives instead of manual impls

fix: use IndexingTerm in postings_writer.rs

### DIFF
--- a/src/core/json_utils.rs
+++ b/src/core/json_utils.rs
@@ -406,7 +406,7 @@ mod tests {
         let mut term = Term::from_field_json_path(field, "color", false);
         term.append_type_and_str("red");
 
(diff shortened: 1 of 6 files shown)


fulmicoton (approved):
@PSeitz approved. Can you make sure to open a PR with the quickwit update rapidly after merging this?

PR #2760, opened 2025-12-04
fix: boolean query incorrectly dropping documents when AllScorer is present

mdashti:
## What

Fixed a bug where Boolean queries would incorrectly return 0 results (or fewer than expected) when one of the clauses produced an `AllScorer`—typically from range queries matching all documents.

## Why

When `BooleanWeight` processes scorers, it removes `AllScorer` instances as an optimization (since they match everything). However, the code wasn't properly accounting for these removed scorers in certain MUST/SHOULD combinations, causing:

1. **SHOULD + SHOULD with AllScorer**: The "match all" semantics were lost, leaving only the other SHOULD clause
2. **MUST (AllScorer) + SHOULD**: Empty `must_scorers` vector → `intersect_scorers([])` returned `EmptyScorer`
3. **Range queries in Boolean clauses**: When all docs match a range (e.g., `age > 50` when everyone is over 50), the query would silently fail

This commonly surfaced in queries like:
```
(age > 50 OR name = 'Alice') AND status = 'active'
```
when all documents happened to satisfy the range condition.

## How

Introduced two helper functions to centralize the AllScorer handling logic:

**`effective_must_scorer()`** — Returns the effective MUST scorer accounting for removed AllScorers:
- Empty list + had AllScorers → returns `AllScorer` (all docs match)
- Empty list + no AllScorers → returns `None` (no MUST constraint)
- Non-empty list → returns intersection of scorers

**`effective_should_scorer_for_union()`** — Restores "match all" semantics for SHOULD:
- If AllScorers were removed, unions the remaining scorer with `AllScorer`

The `complex_scorer()` match arms now use these helpers consistently:
- **Ignored**: Use effective MUST scorer or empty
- **Optional**: No MUST → promote SHOULD (with union fix); Has MUST → SHOULD only affects scoring
- **Required**: No MUST → SHOULD alone; Has MUST → intersect MUST with SHOULD

## Tests

Added 4 regression tests.

### DIFF
--- a/src/query/boolean_query/boolean_weight.rs
+++ b/src/query/boolean_query/boolean_weight.rs
@@ -288,7 +288,15 @@ impl<TScoreCombiner: ScoreCombiner> BooleanWeight<TScoreCombiner> {
                         should_scorer
                     }
                 } else {
-                    let must_scorer = intersect_scorers(must_scorers, num_docs);
+                    // When must_scorers is empty but there were AllScorers removed,
+                    // we need to use AllScorer instead of EmptyScorer
+                    let must_scorer: Box<dyn Scorer> = if must_scorers.is_empty() {
+                        // must_special_scorer_counts.num_all_scorers > 0 here
+                        // (otherwise we'd be in the first branch above)
+                        Box::new(AllScorer::new(reader.max_doc()))
+                    } else {
+                        intersect_scorers(must_scorers, num_docs)
+                    };
                     if self.scoring_enabled {
                         SpecializedScorer::Other(Box::new(RequiredOptionalScorer::<
                             _,
@@ -305,7 +313,21 @@ impl<TScoreCombiner: ScoreCombiner> BooleanWeight<TScoreCombiner> {
             }
             (ShouldScorersCombinationMethod::Required(should_scorer), mut must_scorers) => {
                 if must_scorers.is_empty() {
-                    should_scorer
+                    // When must_scorers is empty but there were AllScorers removed,
+                    // we need to union the should_scorer with AllScorer
+                    if must_special_scorer_counts.num_all_scorers > 0 {
+                        let all_scorers: Vec<Box<dyn Scorer>> = vec![
+                            into_box_scorer(should_scorer, &score_combiner_fn, num_docs),
+                            Box::new(AllScorer::new(reader.max_doc())),
+                        ];
+                        SpecializedScorer::Other(Box::new(BufferedUnionScorer::build(
+                            all_scorers,
+                            &score_combiner_fn,
+                            num_docs,
+                        )))
+                    } else {
+                        should_scorer
+                    }
                 } else {
                     must_scorers.push(into_box_scorer(should_scorer, &score_combiner_fn, num_docs));
                     SpecializedScorer::Other(intersect_scorers(must_scorers, num_docs))
--- a/src/query/boolean_query/boolean_weight.rs
+++ b/src/query/boolean_query/boolean_weight.rs
@@ -268,13 +268,21 @@ impl<TScoreCombiner: ScoreCombiner> BooleanWeight<TScoreCombiner> {
                 SpecializedScorer::Other(boxed_scorer)
             }
             (ShouldScorersCombinationMethod::Optional(should_scorer), must_scorers) => {
+                // Optional SHOULD: contributes to scoring but not required for matching.
+                //
+                // Case 1: No MUST clauses at all (empty list + no removed AllScorers)
+                //   => Promote SHOULD to required (at least one must match).
+                //
+                // Case 2: MUST had AllScorer(s) that were removed
+                //   => All docs match via MUST; SHOULD only affects scoring.
+                //
+                // When SHOULD contains AllScorer(s) that were removed, we must
+                // preserve the "all docs match" semantics by unioning with AllScorer.
                 if must_scorers.is_empty() && must_special_scorer_counts.num_all_scorers == 0 {
-                    // Optional options are promoted to required if no must scorers exists.
-                    // But if there were AllScorers removed from SHOULD, we need to union
-                    // the should_scorer with those matching all docs.
+                    // No MUST clauses: promote SHOULD to required.
                     if should_special_scorer_counts.num_all_scorers > 0 {
-                        // There was at least one AllScorer in SHOULD, so all docs match.
-                        // We still need to union with the remaining scorers for correct scoring.
+                        // Some SHOULD clauses were AllScorers (e.g., range queries matching
+                        // all docs). Restore the "match all" semantics via union.
                         let all_scorers: Vec<Box<dyn Scorer>> = vec![
                             into_box_scorer(should_scorer, &score_combiner_fn, num_docs),
                             Box::new(AllScorer::new(reader.max_doc())),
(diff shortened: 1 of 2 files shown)


src/query/boolean_query/mod.rs
```diff
@@ -374,4 +374,240 @@ mod tests {
         }
         Ok(())
     }
+
+    // =========================================================================
+    // AllScorer Preservation Regression Tests
```
stuhood, line 422: So, it looks like they didn't make it onto `main`, but I had added some property tests to a branch which were intended to cover this code: https://github.com/paradedb/tantivy/blob/2c904e12769eea0d496f8263bfcaad21a9f06858/src/query/boolean_query/mod.rs#L316-L446

I think that @PSeitz decided not to land the branch that those property tests landed on, which is fine. But IMO: that would be a reliable way to add coverage for all of these cases. They don't currently test `AllQuery` explicitly, but I think that that would be a small tweak.

stuhood (approved):
Thank you!

src/query/boolean_query/boolean_weight.rs
```diff
@@ -97,6 +97,55 @@ fn into_box_scorer<TScoreCombiner: ScoreCombiner>(
...
+/// For union semantics (OR): if any SHOULD clause was an AllScorer, the result
+/// should include all documents. We restore this by unioning with AllScorer.
+fn effective_should_scorer_for_union<TScoreCombiner: ScoreCombiner>(
+    should_scorer: SpecializedScorer,
+    removed_all_scorer_count: usize,
+    max_doc: DocId,
+    num_docs: u32,
+    score_combiner_fn: impl Fn() -> TScoreCombiner,
+) -> SpecializedScorer {
+    if removed_all_scorer_count > 0 {
```
trinity-1686a, line 138: this could probably return an AllScorer alone and drop the should_scorer when scoring is disabled

trinity-1686a:
interestingly, it seems it's not possible to trigger this bug with an AllQuery (only 1 of the added tests fails on previous commits, the range one) because it actually generates a `BoostScorer<AllScorer>`, not an AllScorer. It's possible to trigger it with a Range over fastfield, an exist query, and even a TermQuery.

This means we may be missing some optimisations in some case, without knowing it.

I also notice there seems to be places where scoring might be wrong due to AllScorer handling, notably FastFieldRangeWeight doesn't always honor boost (unbounded range on both sides), and BooleanWeight remove most AllScorers without compensating scores. That last one can cause result reordering on some score combiner inside a single segment, and on any combiner when considering multiple segments, where only some segment sees AllScorers, and other see a Range or Exist scorer

src/query/boolean_query/mod.rs
PSeitz: Actually the opposite, I asked for prop-tests to help land https://github.com/quickwit-oss/tantivy/pull/2538 and they have been merged here:
https://github.com/quickwit-oss/tantivy/pull/2690/files
Extending those tests to cover all/no hits would be nice I think

src/query/boolean_query/boolean_weight.rs
mdashti: Good idea. Applied it. Thanks.

src/query/boolean_query/mod.rs
mdashti: Cherry-picked @stuhood's commit with prop-tests and extended it to cover All and Range.

mdashti (commented):
@trinity-1686a @stuhood @PSeitz Thanks for the comments.

> interestingly, it seems it's not possible to trigger this bug with an AllQuery (only 1 of the added tests fails on previous commits, the range one) because it actually generates a BoostScorer<AllScorer>, not an AllScorer. It's possible to trigger it with a Range over fastfield, an exist query, and even a TermQuery.

You're right. Extended the test to better cover Range.

mdashti:
@fulmicoton @PSeitz any more comments?

src/query/boolean_query/mod.rs
```diff
@@ -374,4 +376,413 @@ mod tests {
...
+        /// Matches documents where the field has value "true"
+        Leaf {
+            field_idx: usize,
+        },
+        Union(Vec<BooleanQueryAST>),
+        Intersection(Vec<BooleanQueryAST>),
+    }
+
+    impl BooleanQueryAST {
+        fn matches(&self, doc_id: DocId, num_docs: usize) -> bool {
```
trinity-1686a: `num_docs` feels unused, what is it supposed to convey?

src/query/boolean_query/mod.rs
```diff
@@ -374,4 +376,413 @@ mod tests {
...
+        All,
+        /// Matches all documents via RangeQuery (returns bare AllScorer)
+        /// This is the actual trigger for the AllScorer preservation bug
+        RangeAll,
+        /// Matches documents where the field has value "true"
+        Leaf {
+            field_idx: usize,
+        },
+        Union(Vec<BooleanQueryAST>),
+        Intersection(Vec<BooleanQueryAST>),
```
trinity-1686a, line 703: ideally i think we'd want a `Vec<(Occur, BooleanQueryAST)>` to test the edge cases that can happen when mixing must and should. But some proptest is already good to have, and we can improve that later

src/query/boolean_query/mod.rs
```diff
@@ -374,4 +376,413 @@ mod tests {
...
+        // Leaf strategies: term queries, AllQuery, and RangeQuery matching all docs
+        let leaf = prop_oneof![
+            (0..num_fields).prop_map(|field_idx| BooleanQueryAST::Leaf { field_idx }),
+            Just(BooleanQueryAST::All),
+            Just(BooleanQueryAST::RangeAll),
+        ];
+        leaf.prop_recursive(
+            8,   // 8 levels of recursion
+            256, // 256 nodes max
+            10,  // 10 items per collection
```
trinity-1686a, line 806: i think these values are too high, and will lead to a very large search space with many redundant elements.
`3, 128, 8` feels like probably enough

trinity-1686a (approved):
i tried to poke hole in your changes, and couldn't find anything unexpected, except for [an issue](https://github.com/quickwit-oss/tantivy/issues/2773) that was already present before. Can you just clarify what `num_docs` is supposed to do in `BooleanQueryAST::matches` (and remove it if it's really not used) before i merge?

src/query/boolean_query/mod.rs
```diff
@@ -374,4 +376,413 @@ mod tests {
...
+                        .iter()
+                        .map(|child| (Occur::Must, child.to_query(fields, range_field)))
+                        .collect();
+                    Box::new(BooleanQuery::new(sub_queries))
+                }
+            }
+        }
+    }
+
+    fn create_index_with_boolean_permutations(num_fields: usize) -> (Index, Vec<Field>, Field) {
```
stuhood: It looks like you removed the `num_docs` parameter while cherry-picking this, but that turned out to be important for reproducing problems with certain optimizations around batching/blocking: https://github.com/quickwit-oss/tantivy/pull/2697

fulmicoton:
@stuhood Can you rebase/fix the conflict?

src/query/boolean_query/mod.rs
stuhood: Possibly... but the tests run at about 8 microseconds per case (so sub- 250 ms for the default 256 cases), so it doesn't seem like we need to prune here unless these get a lot slower?

stuhood:
> @stuhood Can you rebase/fix the conflict?

Done. Thanks!

src/query/boolean_query/mod.rs
trinity-1686a: the goal wouldn't be to improve perf, but to make the search space smaller, while hopefully keeping it big enough that it still contains instances of any interesting edge-cases. This makes it more likely to find something interesting.

As a made up example, if you test a function that evaluates to `f(x) = 1/(1+x)` randomly 256 times over the range `[-10^8; 10^8]`, you're unlikely to discover it can fail. Doing the same over `[-10^3; 10^3]`, you have a reasonable chance of discovering the edge case

stuhood: > As a made up example, if you test a function that evaluates to `f(x) = 1/(1+x)` randomly 256 times over the range `[-10^8; 10^8]`, you're unlikely to discover it can fail. Doing the same over `[-10^3; 10^3]`, you have a reasonable chance of discovering the edge case

I'm not sure that this case is comparable. Rather than choosing numbers in a distribution, we're choosing a tree structure. For example, if it takes a parent of type `X`, and a child of type `Y` to reproduce a particular bug, then the depth of that parent and child in the tree aren't particularly relevant, because the framework will shrink the repro as much as possible once it finds it.

We're choosing 256 items from a choice of N enums... so rather than choosing a single value in a huge space like `[-10^8; 10^8]`, this is more like choosing 256 numbers in the space of `[-10^3; 10^3]` (and then doing that 256 times, by default), some permutations of which might fail.

fulmicoton:
Thank you!

PR #2761, opened 2025-12-10
Handle JSON fields and columnar in space_usage

PSeitz-dd:
return field names in space_usage instead of `Field`
fixes JSON fields space usage
more detailed info for columns

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -216,9 +216,7 @@ use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize};
 
 pub use self::docset::{DocSet, COLLECT_BLOCK_BUFFER_LEN, TERMINATED};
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -216,9 +216,7 @@ use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize};
 
 pub use self::docset::{DocSet, COLLECT_BLOCK_BUFFER_LEN, TERMINATED};
-#[doc(hidden)]
```
trinity-1686a, line 219: is that on purpose?

trinity-1686a (approved):

src/lib.rs
PSeitz: yes `json_path_sep_to_dot` should be accessible

PR #2764, opened 2025-12-10
Remove Codecov

philippemnoel:
We removed Codecov from ParadeDB's fork of Tantivy because it was frequently failing, and we didn't want to maintain our own Codecov. I've noticed PRs often get merged with a failing Codecov check, and Pascal suggested raising this PR upstream.

Ultimately, it is up to the Tantivy project whether to use Codecov, but I am raising this PR per our discussion with Pascal.


PSeitz:
I instead just fixed it here, as it's still useful sometimes: https://github.com/quickwit-oss/tantivy/pull/2769

philippemnoel:
> I instead just fixed it here, as it's still useful sometimes: #2769

Thanks!

PR #2768, opened 2025-12-10
fix: added field validation helper for aggregations

### LINKED ISSUE #2767: aggregations silently return empty results for invalid fields
**Describe the bug**

- **What did you do?**
  
  Ran an aggregation query using a field name that doesn't exist in the schema:
  ```rust
  let agg_req: Aggregations = serde_json::from_str(r#"{
      "avg_price": {
          "avg": { "field": "invalid_field" }
      }
  }"#)?;
  
  let collector = AggregationCollector::from_aggs(agg_req, Default::default());
  let result = searcher.search(&AllQuery, &collector)?;
  ```

- **What happened?**
  
  The aggregation succeeded and returned empty/null results:
  ```json
  {"avg_price": {"value": null}}
  ```
  
  No error was raised, making it impossible to distinguish between:
  - A typo in the field name
  - A valid field with no data
  - A query that matched no documents

- **What was expected?**
  
  An error should be returned indicating the field doesn't exist:
  ```rust
  Err(TantivyError::FieldNotFound("invalid_field"))
  ```
  
  This matches the behavior of other query types (like `ExistsQuery`) and aligns with user expectations from SQL databases and Elasticsearch.

This affects the current main branch and all recent versions (tested on 0.26.0).

**To Reproduce**

```rust
use tantivy::aggregation::agg_req::Aggregations;
use tantivy::aggregation::AggregationCollector;
use tantivy::query::AllQuery;
use tantivy::schema::{Schema, FAST};
use tantivy::Index;

    // Create a simple schema
    schema_builder.add_f64_field("price", FAST);
    
    let index = Index::create_in_ram(schema);
    let mut index_writer = index.writer(50_000_000)?;
    index_writer.commit()?;
    
    let reader = index.reader()?;
    
    // Try to aggregate on a non-existent field
    let agg_req: Aggregations = serde_json::from_str(r#"{
        "avg_invalid": {
            "avg": { "field": "nonexistent_field" }
        }
    }"#)?;
    
    let collector = AggregationCollector::from_aggs(agg_req, Default::default());
    let result = searcher.search(&AllQuery, &collector)?;
    
    // Currently returns empty results instead of an error
    println!("{}", serde_json::to_string_pretty(&result)?);
    // Output: {"avg_invalid": {"value": null}}
    
    Ok(())
}
```

This issue affects all aggregation types: `avg`, `sum`, `min`, `max`, `terms`, `histogram`, `date_histogram`, `range`, etc.

mdashti:
# Ticket(s) Closed

- Closes #2767

## What

Adds a public `validate_aggregation_fields()` helper function that validates all fields in an aggregation request exist in the schema and are configured as fast fields. This allows users to validate aggregations upfront before execution.

## Why

Users were getting confusing empty results when aggregation field names had typos or didn't exist. However, some users (like Quickwit) need lenient behavior for schema evolution across different data segments.

This PR provides a validation helper that users can call explicitly when they want validation, while keeping the default aggregation behavior lenient for backward compatibility.

## How

Added `validate_aggregation_fields()` function in `agg_req.rs` that:
- Extracts all field names from the aggregation request using existing `get_fast_field_names()`
- Validates each field exists in the schema (directly or as part of a JSON field)
- Checks that each field is configured as a fast field
- Returns `FieldNotFound` or `SchemaError` if validation fails

Users can call this before executing aggregations:

```rust
// Validate upfront (optional)
validate_aggregation_fields(&agg_req, segment_reader)?;

// Execute aggregation
let collector = AggregationCollector::from_aggs(agg_req, Default::default());
let results = searcher.search(&query, &collector)?;
```

The validation is completely optional - existing code continues to work unchanged with lenient behavior.

## Tests

- Added `test_aggregation_field_validation_helper` testing both invalid and valid fields
- Includes working doctest example
- All 166 existing aggregation tests pass unchanged
- Fully backward compatible - no breaking changes


PSeitz:
The main difference is that we don't always have a fixed schema with JSON fields, which means a field could exist on one segment, but not another one.

Even if a requested field is not part of a JSON field, it would break use cases where you federate a query over different indices.

Adding some validation on top should be easy with the `get_fast_field_names` method.

fulmicoton:
I thought I commented here, but can't find my comment :(.
Not sure what happened.

In quickwit, we make it possible to change your schema.
In that case, we check that the request for the current version of the schema, but we run the query over older data, with a best effort approach.

If the fields are not fast fields, the aggregation result will be empty.

So this is not a bug, but a feature.
Can we make your validation optional (and not the default)?

mdashti:
@PSeitz @fulmicoton Thank you for the feedback! You're absolutely right - the strict enforcement would break important use cases like schema evolution.

I've updated the implementation to make field validation **optional and opt-in**:

#### Default Behavior (Unchanged)
- `strict_field_validation = false` by default
- Returns empty results for non-existent fields (current behavior)
- Supports your use case of querying across splits with different schemas

#### Strict Mode (Opt-in)
- `strict_field_validation = true` when explicitly enabled
- Returns `FieldNotFound` error for non-existent fields
- Useful for catching typos and configuration errors in development

#### Benefits

1. **Backward Compatible**: No breaking changes - default behavior is unchanged
2. **Flexible**: Users can choose strict or lenient based on their needs
3. **Explicit**: The flag name makes it clear what's happening
4. **No Performance Impact**: Validation only runs when explicitly enabled

Let me know if this works for your use case, or if you have other suggestions.

src/aggregation/accessor_helpers.rs
```diff
@@ -55,22 +55,48 @@ pub(crate) fn get_numeric_or_date_column_types() -> &'static [ColumnType] {
...
+                        return Err(crate::TantivyError::SchemaError(format!(
+                            "Field '{}' is not a fast field. Aggregations require fast fields.",
+                            field_name
+                        )));
+                    }
+                }
+
+                // Field doesn't exist at all or has no values in this segment
+                // Check if it exists in schema to provide a better error message
+                if schema.find_field(field_name).is_none() {
```
fulmicoton-dd: this code is correct but odd.

You already called find_field in line 75.

Could we make it a else statement?

src/aggregation/accessor_helpers.rs
```diff
@@ -55,22 +55,48 @@ pub(crate) fn get_numeric_or_date_column_types() -> &'static [ColumnType] {
...
-        .u64_lenient_for_type(allowed_column_types, field_name)?
-        .unwrap_or_else(|| {
-            (
+    let ff_field_with_type = ff_fields.u64_lenient_for_type(allowed_column_types, field_name)?;
+
+    match ff_field_with_type {
+        Some(field) => Ok(field),
+        None => {
+            if strict_validation {
+                // Check if the field exists in the schema but is not a fast field
```
fulmicoton-dd:
```suggestion
                // Here we check two things:
                // - the field is either directly in the schema or could be part of a json field present in the schema.
                // - the field (or the json field holding it) is a fast field.
```

src/aggregation/accessor_helpers.rs
```diff
@@ -89,16 +115,39 @@ pub(crate) fn get_dynamic_columns(
...
+    strict_validation: bool,
 ) -> crate::Result<Vec<(columnar::Column<u64>, ColumnType)>> {
     let ff_fields = reader.fast_fields();
     let mut ff_field_with_type =
         ff_fields.u64_lenient_for_type_all(allowed_column_types, field_name)?;
+
     if ff_field_with_type.is_empty() {
+        if strict_validation {
+            // Check if the field exists in the schema but is not a fast field
+            let schema = reader.schema();
```
fulmicoton-dd: that's the same code as above isn't it?  Can you make a function somewhere.

src/aggregation/mod.rs
```diff
@@ -180,18 +180,57 @@ pub type BucketId = u32;
...
+    /// If false (default), aggregations will return empty results for non-existent fields.
+    ///
+    /// Set to false for use cases where schema can change over time (e.g., Quickwit splits).
+    /// Set to true to catch typos and configuration errors early.
+    pub strict_field_validation: bool,
+}
+
+impl Default for AggContextParams {
+    fn default() -> Self {
+        Self {
```
fulmicoton-dd: Why not rely on the derive Default?

src/aggregation/mod.rs
```diff
@@ -180,18 +180,57 @@ pub type BucketId = u32;
...
+    pub fn with_strict_field_validation(
+        limits: AggregationLimitsGuard,
+        tokenizers: TokenizerManager,
+    ) -> Self {
+        Self {
+            limits,
+            tokenizers,
+            strict_field_validation: true,
+        }
+    }
```
fulmicoton-dd:
```suggestion
```

I don't think we need both a constructor and an accessor for this. This is too niche to require investing in syntactic sugar.

fulmicoton-dd:
I added a few comments @mdashti. 

The PR contains what looks like AI slop.
Using AI agent is ok, but make sure you do not push the burden of CR on us.

src/aggregation/accessor_helpers.rs
```diff
@@ -55,22 +55,48 @@ pub(crate) fn get_numeric_or_date_column_types() -> &'static [ColumnType] {
...
 ) -> crate::Result<(columnar::Column<u64>, ColumnType)> {
     let ff_fields = reader.fast_fields();
-    let ff_field_with_type = ff_fields
-        .u64_lenient_for_type(allowed_column_types, field_name)?
-        .unwrap_or_else(|| {
-            (
+    let ff_field_with_type = ff_fields.u64_lenient_for_type(allowed_column_types, field_name)?;
+
+    match ff_field_with_type {
+        Some(field) => Ok(field),
```
fulmicoton: field as a variable name for a couple is misleading.

src/aggregation/accessor_helpers.rs
```diff
@@ -55,22 +55,48 @@ pub(crate) fn get_numeric_or_date_column_types() -> &'static [ColumnType] {
...
+    strict_validation: bool,
 ) -> crate::Result<(columnar::Column<u64>, ColumnType)> {
     let ff_fields = reader.fast_fields();
-    let ff_field_with_type = ff_fields
-        .u64_lenient_for_type(allowed_column_types, field_name)?
-        .unwrap_or_else(|| {
-            (
+    let ff_field_with_type = ff_fields.u64_lenient_for_type(allowed_column_types, field_name)?;
+
+    match ff_field_with_type {
```
fulmicoton: if let Some(field) = ff_field_with_type {
    return Ok(field);
} 

// ...

PSeitz:
I would move it outside of the aggregation. You can fetch the fields from the aggregation request and do a validation in a helper function

src/aggregation/agg_req.rs
```diff
@@ -115,6 +115,54 @@ pub fn get_fast_field_names(aggs: &Aggregations) -> HashSet<String> {
     fast_field_names
 }
 
+/// Validates that all fields referenced in the aggregation request exist in the schema
```
mdashti: a

src/aggregation/accessor_helpers.rs
mdashti: Done in [90ceec6](https://github.com/quickwit-oss/tantivy/pull/2768/commits/90ceec62aa9832de5f2daaf042d95aab97f120c4), but then changed the approach based on your last comment.

src/aggregation/accessor_helpers.rs
mdashti: Applied in [90ceec6](https://github.com/quickwit-oss/tantivy/pull/2768/commits/90ceec62aa9832de5f2daaf042d95aab97f120c4), but then changed the approach based on your last comment.

src/aggregation/accessor_helpers.rs
mdashti: Done in [90ceec6](https://github.com/quickwit-oss/tantivy/pull/2768/commits/90ceec62aa9832de5f2daaf042d95aab97f120c4), but then changed the approach based on your last comment.

src/aggregation/mod.rs
mdashti: Removed in [90ceec6](https://github.com/quickwit-oss/tantivy/pull/2768/commits/90ceec62aa9832de5f2daaf042d95aab97f120c4), but then changed the approach based on your last comment.

src/aggregation/mod.rs
mdashti: Removed in [90ceec6](https://github.com/quickwit-oss/tantivy/pull/2768/commits/90ceec62aa9832de5f2daaf042d95aab97f120c4), but then changed the approach based on your last comment.

src/aggregation/accessor_helpers.rs
mdashti: Fixed in [90ceec6](https://github.com/quickwit-oss/tantivy/pull/2768/commits/90ceec62aa9832de5f2daaf042d95aab97f120c4), but then changed the approach based on your last comment.

src/aggregation/accessor_helpers.rs
mdashti: Done in [90ceec6](https://github.com/quickwit-oss/tantivy/pull/2768/commits/90ceec62aa9832de5f2daaf042d95aab97f120c4), but then changed the approach based on your last comment.

mdashti (commented):
@fulmicoton Thanks for the comments. Your idea of providing a helper function is a good one. Changed the approach. Please take another look.

src/aggregation/agg_req.rs
```diff
@@ -115,6 +115,66 @@ pub fn get_fast_field_names(aggs: &Aggregations) -> HashSet<String> {
...
+/// let searcher = reader.searcher();
+///
+/// // Validate fields before executing
+/// for segment_reader in searcher.segment_readers() {
+///     validate_aggregation_fields(&agg_req, segment_reader)?;
+/// }
+/// # Ok(())
+/// # }
+/// ```
+pub fn validate_aggregation_fields(
```
PSeitz:
```suggestion
pub fn validate_aggregation_fields_exist(
```

mdashti: Better choice. Renamed in 59fbc7e. While I was in there I also added a short doc note about why this lives outside the aggregation execution path (schema evolution / federated queries) so future readers don't have to dig through PR threads to find that context.

mdashti (commented):
@PSeitz Thanks for the comment. Please take another look.

PSeitz (approved):

PR #2770, opened 2025-12-11
Add an erased `SortKeyComputer` to sort on types which are not known until runtime

stuhood:
This change adds an erased `SortKeyComputer` which outputs `OwnedValue`. It allows for sorting by features/columns whose types are not known until runtime.

In ParadeDB, we support ordering by up to 3 columns: we specialize / code-gen on the first column, since it is the most performance sensitive. But to avoid generating all permutations of types for 3 columns, we erase the types of the latter two columns, and operate on `Option<u64>` segments with dynamic dispatch, and with `OwnedValue` `SortKey` outputs.

### DIFF
--- a/src/collector/sort_key/order.rs
+++ b/src/collector/sort_key/order.rs
@@ -548,7 +548,7 @@ mod tests {
         let i = OwnedValue::I64(10);
         let f = OwnedValue::F64(10.0);
 
-        let nc = NaturalComparator::default();
+        let nc = NaturalComparator;
         assert_eq!(nc.compare(&u, &i), Ordering::Equal);
         assert_eq!(nc.compare(&u, &f), Ordering::Equal);
         assert_eq!(nc.compare(&i, &f), Ordering::Equal);
--- a/src/collector/sort_key/order.rs
+++ b/src/collector/sort_key/order.rs
@@ -34,7 +34,34 @@ fn compare_owned_value<const NULLS_FIRST: bool>(lhs: &OwnedValue, rhs: &OwnedVal
         (OwnedValue::Facet(a), OwnedValue::Facet(b)) => a.cmp(b),
         (OwnedValue::Bytes(a), OwnedValue::Bytes(b)) => a.cmp(b),
         (OwnedValue::IpAddr(a), OwnedValue::IpAddr(b)) => a.cmp(b),
-        x => panic!("Unsupported comparison: {x:?}"),
+        (OwnedValue::U64(a), OwnedValue::I64(b)) => {
+            if *b < 0 {
+                Ordering::Greater
+            } else {
+                a.cmp(&(*b as u64))
+            }
+        }
+        (OwnedValue::I64(a), OwnedValue::U64(b)) => {
+            if *a < 0 {
+                Ordering::Less
+            } else {
+                (*a as u64).cmp(b)
+            }
+        }
+        (OwnedValue::U64(a), OwnedValue::F64(b)) => (*a as f64).to_u64().cmp(&b.to_u64()),
+        (OwnedValue::F64(a), OwnedValue::U64(b)) => a.to_u64().cmp(&(*b as f64).to_u64()),
+        (OwnedValue::I64(a), OwnedValue::F64(b)) => (*a as f64).to_u64().cmp(&b.to_u64()),
+        (OwnedValue::F64(a), OwnedValue::I64(b)) => a.to_u64().cmp(&(*b as f64).to_u64()),
+        (a, b) => {
+            let ord = a.discriminant_value().cmp(&b.discriminant_value());
+            // If the discriminant is equal, it's because a new type was added, but hasn't been
+            // included in this `match` statement.
+            assert!(
+                ord != Ordering::Equal,
+                "Unimplemented comparison for type of {a:?}, {b:?}"
+            );
+            ord
+        }
     }
 }
 
@@ -490,6 +517,7 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::schema::OwnedValue;
 
     #[test]
     fn test_natural_none_is_higher() {
@@ -512,5 +540,28 @@ mod tests {
 
         // compare(None, None) should be Equal.
         assert_eq!(comp.compare(&null, &null), Ordering::Equal);
+    }
+
+    #[test]
+    fn test_mixed_ownedvalue_compare() {
+        let u = OwnedValue::U64(10);
+        let i = OwnedValue::I64(10);
+        let f = OwnedValue::F64(10.0);
+
+        let nc = NaturalComparator::default();
+        assert_eq!(nc.compare(&u, &i), Ordering::Equal);
+        assert_eq!(nc.compare(&u, &f), Ordering::Equal);
+        assert_eq!(nc.compare(&i, &f), Ordering::Equal);
+
+        let u2 = OwnedValue::U64(11);
+        assert_eq!(nc.compare(&u2, &f), Ordering::Greater);
+
+        let s = OwnedValue::Str("a".to_string());
+        // Str < U64
+        assert_eq!(nc.compare(&s, &u), Ordering::Less);
+        // Str < I64
+        assert_eq!(nc.compare(&s, &i), Ordering::Less);
+        // Str < F64
+        assert_eq!(nc.compare(&s, &f), Ordering::Less);
     }
 }
--- a/src/collector/sort_key/order.rs
+++ b/src/collector/sort_key/order.rs
@@ -404,11 +404,12 @@ impl<TSegmentSortKeyComputer, TSegmentSortKey, TComparator> SegmentSortKeyComput
     for SegmentSortKeyComputerWithComparator<TSegmentSortKeyComputer, TComparator>
 where
     TSegmentSortKeyComputer: SegmentSortKeyComputer<SegmentSortKey = TSegmentSortKey>,
-    TSegmentSortKey: PartialOrd + Clone + 'static + Sync + Send,
+    TSegmentSortKey: Clone + 'static + Sync + Send,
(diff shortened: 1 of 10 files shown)


src/collector/sort_key/sort_by_owned.rs
```diff
@@ -0,0 +1,352 @@
...
+        Self::Field(column_name.to_string())
+    }
+
+    /// Creates a new sort key computer which will sort by score, with type erasure.
+    pub fn for_score() -> Self {
+        Self::Score
+    }
+}
+
+/// TODO: Rename to Boxed...? Or Owned?
```
stuhood: Thoughts welcome.

src/collector/sort_key/order.rs
```diff
@@ -1,11 +1,43 @@
...
+        (OwnedValue::PreTokStr(a), OwnedValue::PreTokStr(b)) => a.cmp(b),
+        (OwnedValue::U64(a), OwnedValue::U64(b)) => a.cmp(b),
+        (OwnedValue::I64(a), OwnedValue::I64(b)) => a.cmp(b),
+        (OwnedValue::F64(a), OwnedValue::F64(b)) => a.to_u64().cmp(&b.to_u64()),
+        (OwnedValue::Bool(a), OwnedValue::Bool(b)) => a.cmp(b),
+        (OwnedValue::Date(a), OwnedValue::Date(b)) => a.cmp(b),
+        (OwnedValue::Facet(a), OwnedValue::Facet(b)) => a.cmp(b),
+        (OwnedValue::Bytes(a), OwnedValue::Bytes(b)) => a.cmp(b),
+        (OwnedValue::IpAddr(a), OwnedValue::IpAddr(b)) => a.cmp(b),
+        x => panic!("Unsupported comparison: {x:?}"),
```
fulmicoton: I think  it is required to handle comparing u64, i64, f64.

When dealing with Json fields, tantivy automatically decides whether it should be a i64, a u64 or a f64 based on the values it encounterred. 

It means that two segments could end up with different types.

src/collector/sort_key/sort_by_owned.rs
```diff
@@ -0,0 +1,352 @@
...
+use crate::schema::{OwnedValue, Type};
+use crate::{DateTime, DocId, Score};
+
+/// Sort by the boxed / OwnedValue representation of either a fast field, or of the score.
+///
+/// Using the OwnedValue representation allows for type erasure, and can be useful when sort orders
+/// are not known until runtime. But it comes with a performance cost: wherever possible, prefer to
+/// use a SortKeyComputer implementation with a known-type at compile time.
+#[derive(Debug, Clone)]
+pub enum SortByOwnedValue {
```
fulmicoton: You used to have TypeErased in the name. I think it was nice and explicit.

src/collector/sort_key/order.rs
fulmicoton: Actually for other types we need a sort too (because of json).

For non-numerical stuff it is easier though. You can return a type based comparison.
Just make the 3 numerical enums are in the same "place" in that type ordering,  in order to maintain transitivity.

src/collector/sort_key/sort_by_owned.rs
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    fn segment_sort_key_computer(
+        &self,
+        segment_reader: &crate::SegmentReader,
+    ) -> crate::Result<Self::Child> {
+        let schema = segment_reader.schema();
+        let inner: Box<dyn AnySegmentSortKeyComputer> = match self {
+            Self::Field(column_name) => {
+                let field = schema.get_field(column_name)?;
```
fulmicoton: this does not support json (and this is sad)

src/collector/sort_key/sort_by_erased_type.rs
```diff
@@ -0,0 +1,352 @@
...
+                            "Field `{}` is of type {:?}, which is not supported for sorting by \
+                             owned value yet.",
+                            column_name,
+                            field_type.value_type()
+                        )))
+                    }
+                }
+            }
+            Self::Score => Box::new(ScoreSegmentSortKeyComputer {
+                segment_computer: SortBySimilarityScore,
```
fulmicoton, line 177: score is really lame as it is the only cause for the box that impacts performance.
I suggest we remove it from that scope, as well as ByOwnedValueColumnSegmentSortKeyComputer.

I understand you need to offer sort by score to your users but could it be done one layer above?
You are leaving a lot of performance on the table here. I did not dig much, possibly even losing block wand.

fulmicoton:
@stuhood this is very useful indeed. See comments inline.

src/collector/sort_key/sort_by_erased_type.rs
stuhood: > score is really lame as it is the only cause for the box that impacts performance. I suggest we remove it from that scope, as well as ByOwnedValueColumnSegmentSortKeyComputer.

Ah, I suppose that you are right... at the segment level, even the String implementation could operate entirely on a static u64 Column.

Having said that: `Column` itself is already doing dynamic dispatch to `ColumnValues`... https://github.com/quickwit-oss/tantivy/pull/2728 would reduce that cost, which would be great. And, in fact, https://github.com/quickwit-oss/tantivy/pull/2728 would _also_ reduce the cost of doing dynamic dispatch here, because we would build batches of `SortKey`s at once.

> I understand you need to offer sort by score to your users but could it be done one layer above? You are leaving a lot of performance on the table here. I did not dig much, possibly even losing block wand.

We definitely are, yea: but only when `Score` is _not_ the first sort column (which is fairly rare). And in cases where the `Score` is not the first sort column, block WAND is (usually) less important because the majority of rows are eliminated by the first column(s).

----

So, to my mind, looping back to tackle https://github.com/quickwit-oss/tantivy/pull/2728 would fix dynamic dispatch here, _and_ in `Column`. Would you be ok leaving this dynamic dispatch here, and I can commit to taking another look at that one over the holidays?

src/collector/sort_key/sort_by_owned.rs
stuhood: Totally. Will rename.

src/collector/sort_key/sort_by_owned.rs
stuhood: Would emulating https://github.com/quickwit-oss/tantivy/blob/e9020d17d4d9dfbc9d6a536bcb85ebae93fc4376/src/collector/sort_key/sort_by_static_fast_value.rs#L68-L72 get that?

src/collector/sort_key/order.rs
stuhood: Mm, all makes sense: thanks! Will fix.

stuhood (commented):
Thank you!

src/collector/sort_key/sort_by_owned.rs
fulmicoton: I think it would work yes.

You can keep the type `sort_column_type` to build you erased value at the end.

stuhood:
@fulmicoton : This is ready for another look. Thanks!

src/collector/sort_key/sort_by_erased_type.rs
fulmicoton: ok

src/schema/document/owned_value.rs
```diff
@@ -58,6 +58,29 @@ impl AsRef<OwnedValue> for OwnedValue {
...
+impl OwnedValue {
+    /// Returns a u8 discriminant value for the `OwnedValue` variant.
+    ///
+    /// This can be used to sort `OwnedValue` instances by their type.
+    pub fn discriminant_value(&self) -> u8 {
+        match self {
+            OwnedValue::Null => 0,
+            OwnedValue::Str(_) => 1,
+            OwnedValue::PreTokStr(_) => 2,
+            OwnedValue::U64(_) => 3,
```
fulmicoton, line 72:
```suggestion
            // It is key to make sure U64, I64, F64 are grouped together in there, otherwise we might be breaking transivity.
            OwnedValue::U64(_) => 3,
```

stuhood: @fulmicoton : So, given that we've turned this into a complete `PartialOrd` implementation, how would you feel about actually moving it to `impl PartialOrd for OwnedValue`?

I didn't initially, because it was (intentionally) incomplete.

src/collector/sort_key/order.rs
```diff
@@ -346,3 +424,32 @@ where
...
+    use super::*;
+    use crate::schema::OwnedValue;
+
+    #[test]
+    fn test_mixed_ownedvalue_compare() {
+        let u = OwnedValue::U64(10);
+        let i = OwnedValue::I64(10);
+        let f = OwnedValue::F64(10.0);
+
+        let nc = NaturalComparator::default();
```
fulmicoton: somehow the suggestion UI has been removed from github?

anyway can we change this to 

`let nc = NaturalComparator;`

(just to please clippy. I personally don't think it matters)

src/schema/document/owned_value.rs
fulmicoton: Sounds good to me. I'll approve this PR but you can add an extra commit and merge if you want.

(Can you confirm you have merge permissions?)

(Also this is an Ord implementation, isn't it?)

fulmicoton (approved):

src/schema/document/owned_value.rs
stuhood: > (Can you confirm you have merge permissions?)

I do not.

> (Also this is an Ord implementation, isn't it?)

No: currently just a static function used by the `Comparator` types.

Now that I think about it though, it has a const generic parameter to drive the nulls-first/nulls-last behavior... if we actually implemented `PartialOrd`, we'd have to choose one of those to be the default. Can wait until someone actually needs it.

PR #2777, opened 2025-12-16
Use a strict comparison in TopNComputer

### LINKED ISSUE #2775: Regression for TopN with many duplicates; suggested behavior change
Benchmarks in ParadeDB (see edits in https://github.com/paradedb/paradedb/pull/3752#pullrequestreview-3565551054) show that tiebreaking many equal values got 60% slower in https://github.com/quickwit-oss/tantivy/pull/2726.

I triaged this down to the behavior of the `TopNComputer::push` having changed:
* from: https://github.com/quickwit-oss/tantivy/blob/7d513a44c530ee6792a588c300802cf5cd3ded82/src/collector/top_score_collector.rs#L993-L1000
* to: https://github.com/quickwit-oss/tantivy/blob/e9020d17d4d9dfbc9d6a536bcb85ebae93fc4376/src/collector/top_score_collector.rs#L606

In short: we moved from eliminating items `less-than-or-equal-to` the threshold, to eliminating items `less-than` the threshold, which leads to many more items making it into the buffer, and many more quickselects. Swapping the comparison in the new code to `less-than-or-equal-to` recovers 45% of the regression.

----

The old code was not including the `DocId` in the threshold comparison, which was a determinism hazard: see the explanation in https://github.com/quickwit-oss/tantivy/pull/2672#pullrequestreview-3098342775 ... essentially, it meant that _only_ quickselect was considering the `DocId`/`DocAddress`: that meant that if all of the matches had a single feature value, then some might be eliminated non-deterministically based on the order that segments were collected.

The new code is still not including the `DocId` in the threshold comparison, but because it puts `equal-to` items in the buffer, it _is_ deterministic, because the quickselect will consider the `DocId`.

----

I'm opening a ticket here because I think that this relates to another consideration: `Score` based ordering probably wants/needs `DocId`/`DocAddress` tiebreaking (at least by default)... but for any other sort, it should probably be optional for performance reasons.

Currently, the inclusion of the `DocId` in the sort order is haphazard: if instead it was something controlled by the `SortKeyComputer`, it would enable this optimization, and also probably clean up the code considerably.

stuhood:
As discussed in #2775, this change switches (back) to a strict comparison on the threshold in `TopNComputer`.

To do this safely, we document the invariant that documents must be pushed to the `TopNComputer` in ascending `DocId`/`DocAddress` order, and that the output of `TopNComputer` will sort in ascending `DocId`/`DocAddress` order for duplicate `SortKey`s.

Fixes #2775.

### DIFF
--- a/src/collector/sort_key/order.rs
+++ b/src/collector/sort_key/order.rs
@@ -21,9 +21,9 @@ pub trait Comparator<T>: Send + Sync + std::fmt::Debug + Default {
         // TopNComputer sorts in descending order of the SortKey by default: we apply that ordering
         // here to ease comparison in testing.
         self.compare(&rhs.sort_key, &lhs.sort_key).then_with(|| {
-            // In case of a tie on the feature, we always sort by ascending `DocAddress` in order
-            // to ensure a stable sorting of the documents. See the TopNComputer docs for more
-            // information.
+            // In case of a tie on the sort key, we always sort by ascending `DocAddress` in order
+            // to ensure a stable sorting of the documents, regardless of the sort key's order.
+            // See the TopNComputer docs for more information.
             lhs.doc.cmp(&rhs.doc)
         })
     }
@@ -47,8 +47,8 @@ impl<T: PartialOrd> Comparator<T> for NaturalComparator {
 /// first.
 ///
 /// The ReverseComparator does not necessarily imply that the sort order is reversed compared
-/// to the NaturalComparator. In presence of a tie, both version will retain the documents based on
-/// descending `DocId`/`DocAddress`.
+/// to the NaturalComparator. In presence of a tie on the sort key, documents will always be
+/// sorted by ascending `DocId`/`DocAddress` in TopN results, regardless of the comparator.
 #[derive(Debug, Copy, Clone, Default, Serialize, Deserialize)]
 pub struct ReverseComparator;
 
--- a/src/collector/sort_key/order.rs
+++ b/src/collector/sort_key/order.rs
@@ -12,7 +12,7 @@ pub trait Comparator<T>: Send + Sync + std::fmt::Debug + Default {
     fn compare(&self, lhs: &T, rhs: &T) -> Ordering;
     /// Return the order between two ComparableDoc values.
     #[inline(always)]
-    fn compare_doc<D: PartialOrd>(
+    fn compare_doc<D: Ord>(
         &self,
         lhs: &ComparableDoc<T, D>,
         rhs: &ComparableDoc<T, D>,
@@ -21,7 +21,7 @@ pub trait Comparator<T>: Send + Sync + std::fmt::Debug + Default {
             // In case of a tie on the feature, we always sort by descending `DocAddress` in order
             // to ensure a stable sorting of the documents. See the TopNComputer docs for more
             // information.
-            rhs.doc.partial_cmp(&lhs.doc).unwrap_or(Ordering::Equal)
+            rhs.doc.cmp(&lhs.doc)
         })
     }
 }
--- a/src/collector/sort_key/order.rs
+++ b/src/collector/sort_key/order.rs
@@ -44,7 +44,8 @@ impl<T: PartialOrd> Comparator<T> for NaturalComparator {
 /// first.
 ///
 /// The ReverseComparator does not necessarily imply that the sort order is reversed compared
-/// to the NaturalComparator. In presence of a tie, both version will retain the higher doc ids.
+/// to the NaturalComparator. In presence of a tie, both version will retain the documents based on
+/// descending `DocId`/`DocAddress`.
 #[derive(Debug, Copy, Clone, Default, Serialize, Deserialize)]
 pub struct ReverseComparator;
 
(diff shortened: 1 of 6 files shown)


src/collector/top_score_collector.rs
```diff
@@ -628,11 +633,9 @@ where
...
     fn truncate_top_n(&mut self) -> TSortKey {
         // Use select_nth_unstable to find the top nth score
-        let (_, median_el, _) = self.buffer.select_nth_unstable_by(self.top_n, |lhs, rhs| {
-            self.comparator
-                .compare(&rhs.sort_key, &lhs.sort_key)
-                .then_with(|| lhs.doc.cmp(&rhs.doc))
-        });
+        let (_, median_el, _) = self
+            .buffer
+            .select_nth_unstable_by(self.top_n, |lhs, rhs| self.comparator.compare_doc(rhs, lhs));
```
stuhood: Note: This comparison (and the one below) was already reversed, and I preserved that. It's ... fairly confusing though. This might be a good opportunity to fix it.

src/collector/sort_key/order.rs
```diff
@@ -2,14 +2,28 @@ use std::cmp::Ordering;
...
+    fn compare_doc<D: PartialOrd>(
+        &self,
+        lhs: &ComparableDoc<T, D>,
+        rhs: &ComparableDoc<T, D>,
+    ) -> Ordering {
+        self.compare(&lhs.sort_key, &rhs.sort_key).then_with(|| {
+            // In case of a tie on the feature, we always sort by descending `DocAddress` in order
+            // to ensure a stable sorting of the documents. See the TopNComputer docs for more
+            // information.
+            rhs.doc.partial_cmp(&lhs.doc).unwrap_or(Ordering::Equal)
```
fulmicoton:
```suggestion
            rhs.doc.cmp(&lhs.doc)
```

src/collector/top_score_collector.rs
fulmicoton: I think this is ok like this...

The compare_doc is expressing the order relation we want to enforce.

However we are trying to get top K documents, so we call it in reverse.

src/collector/top_collector.rs
```diff
@@ -20,45 +13,12 @@ pub struct ComparableDoc<T, D, const REVERSE_ORDER: bool = false> {
...
     pub doc: D,
 }
-impl<T: std::fmt::Debug, D: std::fmt::Debug, const R: bool> std::fmt::Debug
-    for ComparableDoc<T, D, R>
-{
+
+impl<T: std::fmt::Debug, D: std::fmt::Debug> std::fmt::Debug for ComparableDoc<T, D> {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        f.debug_struct(format!("ComparableDoc<_, _ {R}").as_str())
+        f.debug_struct(format!("ComparableDoc").as_str())
```
fulmicoton: can you fix this?

fulmicoton (changes requested):

fulmicoton (changes requested):
I'm ok with the implementation. Can you fix tests to enforce the right behavior?

src/collector/top_score_collector.rs
stuhood: I ended up baking the reverse-ordering into `compare_doc`, because it makes all of the tests make more sense: the `Comparator` that you use to sort aligns with the `Comparator` passed to the `TopNComputer`.

I could imagine that maybe `Comparator::compare_doc` should maybe be something like `Comparator::compare_top_n`... thoughts?

stuhood:
> I'm ok with the implementation. Can you fix tests to enforce the right behavior?

Fixed.

Take a look at `test_topn_computer_duplicates`: the semantics that you proposed on the ticket work, but in practice, I found it easier to explain that "the `DocId`/`DocAddress` is always sorted in ascending order".

It's _technically_ `(SortKey, Reverse(DocId))`, but because the `TopNComputer` sorts in reverse order, what you get out is: `Reverse((SortKey, Reverse(DocId)))` ... i.e. `(Reverse(SortKey), DocId)`.

src/collector/top_score_collector.rs
fulmicoton: I don't like it at all. `compare_doc` should define an order relationship.

Not a reverse order relationship because it happens that it is used in a TopK collector and most of the call site actually want a reverse order.

fulmicoton:
> it easier to explain that "the DocId/DocAddress is always sorted in ascending order".

Agreed. 
Using the expression `ascending order` is a much better expression here (no ambiguity).

src/collector/top_score_collector.rs
fulmicoton: ah i think I misunderstood what you meant.

src/collector/sort_key/order.rs
```diff
@@ -2,14 +2,31 @@ use std::cmp::Ordering;
...
+    /// implemented by TopNComputer.
+    #[inline(always)]
+    fn compare_doc<D: Ord>(
+        &self,
+        lhs: &ComparableDoc<T, D>,
+        rhs: &ComparableDoc<T, D>,
+    ) -> Ordering {
+        // TopNComputer sorts in descending order of the SortKey by default: we apply that ordering
+        // here to ease comparison in testing.
+        self.compare(&rhs.sort_key, &lhs.sort_key).then_with(|| {
```
fulmicoton: noooooooo

fulmicoton (approved):

PR #2779, opened 2025-12-19
document 1bit hole and some queries supporting running with just fastfield

trinity-1686a:


fulmicoton (approved):

src/postings/skip.rs
```diff
@@ -6,17 +6,21 @@ use crate::{DocId, Score, TERMINATED};
...
+//         |1|2|3|  4  |
 // - 1: unused
 // - 2: is delta-1 encoded. 0 if not, 1, if yes
-// - 3: a 6 bit number in 0..=32, the actual bitwidth
+// - 3: unused
+// - 4: a 5 bit number in 0..32, the actual bitwidth. Bitpacking could in theory say this is 32
+//   (requiring a 6th bit), but the biggest doc_id we can want to encode is TERMINATED-1, which can
+//   be represented on 31b without delta encoding.
 fn encode_bitwidth(bitwidth: u8, delta_1: bool) -> u8 {
+    assert!(bitwidth < 32);
```
fulmicoton, line 17: I agree this is correct, but I don't understand. What is your objective here?

fulmicoton (changes requested):
Accepted but I don't quite see the point.

fulmicoton (approved):

src/postings/skip.rs
trinity-1686a: is the question about the assertion, or whether the one bit-hole can be useful?

fulmicoton: Ah I think I got it.

A bitdwidth of up to 32 bits is causing us to encode the bidwidth over 5 bits when really we never go above 31 so 4 bits is sufficient. My bad.

PR #2780, opened 2025-12-19
Add support for natural-order-with-none-highest in `TopDocs::order_by`

stuhood:
Add `ComparatorEnum::NaturalNoneHigher`, which matches Postgres's `DESC NULLS FIRST` behavior in `TopDocs::order_by`.

Expands comments on `Comparator` implementations to ensure that behavior for `None` is explicit.


stuhood:
FWIW: The four variants of `ComparatorEnum` have been proptested for equivalence to Postgres (for a few types) in https://github.com/paradedb/paradedb/pull/3805:
https://github.com/paradedb/paradedb/blob/970f6864507be1e6908673e773c34a5655e3eaee/pg_search/src/api/mod.rs#L291-L304

fulmicoton (approved):

PR #2782, opened 2025-12-26
Lazy loading of the first block of RangeDocSet

ChangRui-Ryan:
### Optimization for RangeDocSet's Initialization Overhead

**Problem Statement**
The `new` method of `RangeDocSet` always attempts to load at least one `doc id` that meets the range criteria. However, when computing the intersection between another `DocSet` and a `RangeDocSet`, if the other `DocSet` is empty or its first `doc id` has a relatively large value, the significant overhead incurred by the `fetch_block` method called in `RangeDocSet`'s `new` method is completely wasted.

As observed from the newly added `bool_queries` benchmark: in the worst-case scenario, `fetch_block` processing 10 million entries takes about 15 milliseconds — an overhead that is entirely avoidable.

**Optimization Solutions**
1. Removed the invocation of `fetch_block` from the `new` method of `RangeDocSet`, and instead triggers `fetch_block` **on-demand** in other methods of `RangeDocSet`.
2. Modified the `go_to_first_doc` method: it now only performs a `seek` operation on the `Scorer` of `RangeDocSet`, instead of calling the `doc` method immediately upon initialization.

before optimization (compare with after):
image

after optimization (compare with before):
image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -176,3 +176,7 @@ harness = false
 [[bench]]
 name = "and_or_queries"
 harness = false
+
+[[bench]]
+name = "bool_queries"
+harness = false
--- a/benches/bool_queries.rs
+++ b/benches/bool_queries.rs
@@ -1,0 +1,320 @@
+use binggan::{black_box, BenchGroup, BenchRunner};
+use rand::prelude::*;
+use rand::rngs::StdRng;
+use rand::SeedableRng;
+use tantivy::collector::{Collector, Count, DocSetCollector, TopDocs};
+use tantivy::query::{Query, QueryParser};
+use tantivy::schema::{Schema, FAST, INDEXED, TEXT};
+use tantivy::{doc, Index, Order, ReloadPolicy, Searcher};
+
+#[derive(Clone)]
+struct BenchIndex {
+    #[allow(dead_code)]
+    index: Index,
+    searcher: Searcher,
+    query_parser: QueryParser,
+}
+
+fn build_shared_indices(num_docs: usize, p_title_a: f32, distribution: &str) -> BenchIndex {
+    // Unified schema
+    let mut schema_builder = Schema::builder();
+    let f_title = schema_builder.add_text_field("title", TEXT);
+    let f_num_rand = schema_builder.add_u64_field("num_rand", INDEXED);
+    let f_num_asc = schema_builder.add_u64_field("num_asc", INDEXED);
+    let f_num_rand_fast = schema_builder.add_u64_field("num_rand_fast", INDEXED | FAST);
+    let f_num_asc_fast = schema_builder.add_u64_field("num_asc_fast", INDEXED | FAST);
+    let schema = schema_builder.build();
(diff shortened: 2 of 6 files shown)


fulmicoton-dd:
@ChangRui-Ryan Thanks for spotting that. Can we try and find a solution that does not special case this strongly on fast field range however?

fulmicoton:
@PSeitz Thoughts on this. It sounds like a side effect of my choice to make DocSet an "initialized iterator". `.doc()`  returns a valid DocId event before having called `.advance()` or `.seek()`...

We could add an optional seek target in the `.scorer` of weights, and some cost methods?

In this specific case, I suspect your `.sloppy_seek` (I forgot how it was called) optimization would really help, so I'd like to hear your thoughts.

PSeitz:
> @PSeitz Thoughts on this. It sounds like a side effect of my choice to make DocSet an "initialized iterator". `.doc()` returns a valid DocId event before having called `.advance()` or `.seek()`...
> 
> We could add an optional seek target in the `.scorer` of weights, and some cost methods?
> 
> In this specific case, I suspect your `.sloppy_seek` (I forgot how it was called) optimization would really help, so I'd like to hear your thoughts.

The `seek_into_the_dangerzone` [PR](https://github.com/quickwit-oss/tantivy/pull/2538) is similar, but it covers only the intersection part, not the initialization.

> We could add an optional seek target in the `.scorer` of weights, and some cost methods?

I think that should work. The `cost` method is currently on the scorer, but we could have another method on the `Weight` trait to check the initialization cost.

We could also remove the `.doc()` method and the initialization requirement on the docsets. In that case it would be the job of the caller to store the current doc (which is trivial except for intersections and unions).

ChangRui-Ryan:
> @ChangRui-Ryan Thanks for spotting that. Can we try and find a solution that does not special case this strongly on fast field range however?

This issue is quite prominent in my usage scenario, and this PR provides an excellent solution to it. I understand that the current changes are not very elegant at the code level, but they can at least serve as a reference for future refactoring and refinement.

fulmicoton:
@ChangRui-Ryan https://github.com/quickwit-oss/tantivy/pull/2785/changes Does this address your problem? Can you open a PR with your benchmark?

Also. We just merged a PR from @PSeitz that might have a strong impact on your use case. Unfortunately it is not yet unabled for fast field docset. I added it in the following branch https://github.com/quickwit-oss/tantivy/pull/2786

ChangRui-Ryan:
> @ChangRui-Ryan https://github.com/quickwit-oss/tantivy/pull/2785/changes Does this address your problem? Can you open a PR with your benchmark?
> 
> Also. We just merged a PR from @PSeitz that might have a strong impact on your use case. Unfortunately it is not yet unabled for fast field docset. I added it in the following branch #2786

It seems that this PR might have resolved my issue. Anyway, I'll submit a separate PR to add my benchmark first: https://github.com/quickwit-oss/tantivy/pull/2787. Thanks!

fulmicoton:
Closing for #2785

PR #2783, opened 2025-12-29
optimize RangeDocSet for non-overlapping query ranges

ChangRui-Ryan:
When there is no overlap between the query's range and the `[min_value, max_value]` of the column in `RangeDocSet`, we can optimize to **avoid any traversal** of the column in `RangeDocSet`.

Currently, the code only implements a similar optimization in `transform_range_before_linear_transformation`, which is **only valid for the Bitpacked encoding scenario**. In fact, this check should be moved to `RangeDocSet` to take effect for **all scenarios**. Consequently, the corresponding code in `transform_range_before_linear_transformation` can be removed.  
before optimization (compare with after):
img_v3_02te_acba53f2-17cf-41bd-a38c-f7caf5844ccg
after optimization (compare with before):
img_v3_02te_a9d05a49-dc39-4e35-8a23-c3ec9ba480dg

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -176,3 +176,7 @@ harness = false
 [[bench]]
 name = "and_or_queries"
 harness = false
+
+[[bench]]
+name = "range_queries"
+harness = false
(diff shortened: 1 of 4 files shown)


PSeitz (approved):
Nice catch, thanks!

PR #2784, opened 2025-12-30
fix flaky test

PSeitz:


fulmicoton (approved):

src/lib.rs
```diff
@@ -17,19 +17,15 @@
...
-//! #     // sake of this example
-//! #     if let Ok(dir) = TempDir::new() {
-//! #         run_example(dir.path()).unwrap();
-//! #         dir.close().unwrap();
-//! #     }
+//! #     let dir = RamDirectory::create();
+//! #     run_example(dir.path()).unwrap();
 //! # }
 //! #
 //! # fn run_example(index_path: &Path) -> tantivy::Result<()> {
```
fulmicoton, line 39: you need to fix this. This receives a Directory now.

Also note this is not great...

The point of this doctest is to generate a rustdoc that people can read and reuse.
If you use a RAM directory it defeats the purpose.

fulmicoton (changes requested):
See comment

src/lib.rs
PSeitz: The issue is that the tempdir is not empty sometimes, so I added an extra cleanup upfront

fulmicoton (approved):

PR #2785, opened 2025-12-30
Added seek_doc to weight

fulmicoton-dd:


src/query/boolean_query/boolean_weight.rs
```diff
@@ -207,10 +207,27 @@ impl<TScoreCombiner: ScoreCombiner> BooleanWeight<TScoreCombiner> {
...
-        for (occur, subweight) in &self.weights {
-            let sub_scorer: Box<dyn Scorer> = subweight.scorer(reader, boost)?;
+        let (must_weights, other_weights): (Vec<(Occur, _)>, Vec<(Occur, _)>) = self
+            .weights
+            .iter()
+            .map(|(occur, weight)| (*occur, weight))
+            .partition(|(occur, _weight)| *occur == Occur::Must);
+        // We start by must weights in order to get the best "seek_first_doc" so that we
+        // can skip the first few documents of the other scorers.
+        for (_, must_sub_weight) in must_weights {
```
fulmicoton-dd, line 234: Need to add a sort before this loop.

src/query/weight.rs
```diff
@@ -126,8 +139,14 @@ pub trait Weight: Send + Sync + 'static {
...
         callback: &mut dyn FnMut(DocId, Score) -> Score,
     ) -> crate::Result<()> {
-        let mut scorer = self.scorer(reader, 1.0)?;
+        let mut scorer = self.scorer(reader, 1.0, 0)?;
         for_each_pruning_scorer(scorer.as_mut(), threshold, callback);
         Ok(())
     }
+
+    /// Returns a priority number used to sort weights when running an
+    /// intersection.
```
PSeitz, line 148: We should add a comment here what higher or lower numbers mean

src/postings/block_segment_postings.rs
```diff
@@ -87,6 +87,31 @@ fn split_into_skips_and_postings(
     Ok((Some(skip_data), postings_data))
 }
 
+/// A block segment postings for which the first block has not been loaded yet.
+///
+/// You can either call `load_at_start` to load it its first block,
+/// or skip a few blocks by calling `seek_and_load`.
+pub(crate) struct BlockSegmentPostingsNotLoaded(BlockSegmentPostings);
```
PSeitz: Can't we require the docid seek on the BlockSegmentPostings instead and drop this intermediate struct?

src/query/weight.rs
fulmicoton: yes. Added a comment.

src/postings/block_segment_postings.rs
fulmicoton: It feels more safe to enforce that with a type.

PSeitz: I think having the seek_doc as parameter on the constructor is more readable and as safe

fulmicoton: ok I moved it to the constructor.

src/postings/segment_postings.rs
```diff
@@ -127,28 +128,29 @@ impl SegmentPostings {
...
 
-    /// Reads a Segment postings from an &[u8]
-    ///
-    /// * `len` - number of document in the posting lists.
-    /// * `data` - data array. The complete data is not necessarily used.
-    /// * `freq_handler` - the freq handler is in charge of decoding frequencies and/or positions
+    /// Creates a Segment Postings from a
+    /// - `BlockSegmentPostings`,
+    /// - a position reader
+    /// - a target document to seek to
```
PSeitz, line 145:
```suggestion
    /// - a target document to seek to within the passed block
```
not sure if that's correct (the comment and variable name don't match)

src/index/inverted_index_reader.rs
```diff
@@ -192,20 +193,24 @@ impl InvertedIndexReader {
         term: &Term,
         option: IndexRecordOption,
     ) -> io::Result<Option<BlockSegmentPostings>> {
-        self.get_term_info(term)?
-            .map(move |term_info| self.read_block_postings_from_terminfo(&term_info, option))
-            .transpose()
+        let Some(term_info) = self.get_term_info(term)? else {
+            return Ok(None);
+        };
+        let block_postings_not_loaded =
```
PSeitz, line 199:
```suggestion
        let block_postings =
```

PSeitz (approved):
left some comments

PR #2795, opened 2026-01-05
add benchmark for string search and get

ChangRui-Ryan:
This benchmark is designed to quantify the performance of retrieving Str-type values based on doc IDs, including two approaches: retrieving from TantivyDocument and from the column of FAST fields. The results indicate that retrieving Str-type field values incurs significant overhead. In particular, for FAST fields, when retrieving via StrColumn, the `ord_to_term` method requires repeatedly decoding incremental encoding from scratch for each value, leading to extremely high overhead—with latency reaching dozens of times that of retrieval from TantivyDocument.  

img_v3_02tl_18c9d0b1-1122-48fe-a0f4-f01657fdd39g

### DIFF
--- a/benches/str_search_and_get.rs
+++ b/benches/str_search_and_get.rs
@@ -1,0 +1,421 @@
+// This benchmark compares different approaches for retrieving string values:
+//
+// 1. Fast Field Approach: retrieves string values via term_ords() and ord_to_str()
+//
+// 2. Doc Store Approach: retrieves string values via searcher.doc() and field extraction
+//
+// The benchmark includes various data distributions:
+// - Dense Sequential: Sequential document IDs with dense data
+// - Dense Random: Random document IDs with dense data
+// - Sparse Sequential: Sequential document IDs with sparse data
+// - Sparse Random: Random document IDs with sparse data
+use std::ops::Bound;
+
+use binggan::{black_box, BenchGroup, BenchRunner};
+use rand::prelude::*;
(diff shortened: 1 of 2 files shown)


benches/str_search_and_get.rs
```diff
@@ -0,0 +1,421 @@
...
+    searcher: Searcher,
+}
+
+fn build_shared_indices(num_docs: usize, distribution: &str) -> BenchIndex {
+    // Schema with string fast field and stored field for doc access
+    let mut schema_builder = Schema::builder();
+    let f_str_fast = schema_builder.add_text_field("str_fast", STRING | STORED | FAST);
+    let f_str_stored = schema_builder.add_text_field("str_stored", STRING | STORED);
+    let schema = schema_builder.build();
+    let index = Index::create_in_ram(schema.clone());
```
PSeitz, line 37: I think the test would be fairer if the docstore is zstd compressed (maybe with 1MB blocks)

ChangRui-Ryan: Are you saying that ztsd is slower? But since LZ4 is the default, I think it’s better to run the tests with the default settings. Additionally, what I actually want to say is that it’s not just the string value retrieval from fast field column that’s slow—retrieving docs from doc store is also not fast enough.

fulmicoton: Whatever compression scheme makes sense to you is fine in this becnchamrk.

PR #2798, opened 2026-01-09
feat: Add support for Exists leaf in query parser

Darkheir:
A while ago the Exists syntax was added to the query grammar: #2170.
It was never added to the query parser and this PR aims to implement this part.

### DIFF
--- a/src/query/query_parser/logical_ast.rs
+++ b/src/query/query_parser/logical_ast.rs
@@ -162,7 +162,7 @@ impl fmt::Debug for LogicalLiteral {
                 ref pattern,
                 ref field,
             } => write!(formatter, "Regex({field:?}, {pattern:?})"),
-            LogicalLiteral::Exists { ref field } => write!(formatter, "exists:{field}"),
+            LogicalLiteral::Exists { ref field } => write!(formatter, "Exists({field})"),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/query/query_parser/logical_ast.rs
```diff
@@ -159,6 +162,7 @@ impl fmt::Debug for LogicalLiteral {
                 ref pattern,
                 ref field,
             } => write!(formatter, "Regex({field:?}, {pattern:?})"),
+            LogicalLiteral::Exists { ref field } => write!(formatter, "exists:{field}"),
```
xingtanzjr: That may be ambuigous if we put operator `exists` before `:`. For example, if user gives this query

```
'exists:title'
```
it may be treated as something like finding `title` term in feild `exists`.

xingtanzjr:
> _ No description provided. _

Maybe we can add some description here

src/query/query_parser/logical_ast.rs
Darkheir: Fair enough, I updated the PR to follow the same syntax as for regexes.

Darkheir:
> Maybe we can add some description here

Added a small description, but there is not much to say :-)

xingtanzjr (approved):

PSeitz (approved):
Thanks!

Issue #2796, opened 2026-01-07
Update `lru` to >=0.16.3


rkrishn7:
`cargo deny` reports an unsound advisory with the current version of `lru`:

```
    ├ ID: RUSTSEC-2026-0002
    ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2026-0002
    ├ Affected versions of this crate contain a soundness issue in the `IterMut`
      iterator implementation. The `IterMut::next` and `IterMut::next_back`
      methods temporarily create an exclusive reference to the key when
      dereferencing the internal node pointer.
      
      This invalidates the shared pointer held by the internal `HashMap`,
      violating Stacked Borrows rules.
```

lucieleblanc:
Will there be a new release that includes #2799? I'm happy to see it was merged, and it'd be great to have an official release containing the fix so downstream dependencies can update to include it.

fulmicoton:
I think @PSeitz is planning an imminent release.

PR #2799, opened 2026-01-10
Bump lru to 0.16.3 per dependabot

jollygreenlaser:
https://github.com/quickwit-oss/tantivy/issues/2796


stuhood:
Please run `cargo check` and commit the resulting `Cargo.lock` edit as well.

trinity-1686a:
there is no committed Cargo.lock in tantivy? They aren't really relevant for libraries

trinity-1686a (approved):

PR #2803, opened 2026-01-14
fix(aggs): top_hits panics with unknown fields

Totodore:
_Related to this comment:_
https://github.com/quickwit-oss/quickwit/pull/6088#issuecomment-3748772985

* The current implementation top hits panics when incorrect fields are specified in `docvalue_fields`. This leads to non-ergonomic errors in quickwit for example when we query non-existent `docvalue_fields`.
* fixes a bug where non-glob fast-fields where incorrectly matched because the dot notation of columns was incorrectly escaped.
* Make allowed fields of `TopHitsAggregationReq` pub so it is possible to instantiate and modify a `TopHitsAggregationReq`.

### DIFF
--- a/src/aggregation/metric/top_hits.rs
+++ b/src/aggregation/metric/top_hits.rs
@@ -19,7 +19,7 @@ use crate::aggregation::{AggregationError, BucketId};
 use crate::collector::sort_key::ReverseComparator;
 use crate::collector::TopNComputer;
 use crate::schema::OwnedValue;
-use crate::{DocAddress, DocId, SegmentOrdinal};
+use crate::{DocAddress, DocId, SegmentOrdinal, TantivyError};
 
 /// Contains all information required by the TopHitsSegmentCollector to perform the
 /// top_hits aggregation on a segment.
@@ -149,7 +149,9 @@ impl Serialize for KeyOrder {
 
 impl<'de> Deserialize<'de> for KeyOrder {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where D: Deserializer<'de> {
+    where
+        D: Deserializer<'de>,
+    {
         let mut key_order = <HashMap<String, Order>>::deserialize(deserializer)?.into_iter();
         let (field, order) = key_order.next().ok_or(serde::de::Error::custom(
             "Expected exactly one key-value pair in sort parameter of top_hits, found none",
@@ -218,12 +220,16 @@ impl TopHitsAggregationReq {
             .doc_value_fields
             .iter()
             .map(|field| {
-                if !field.contains('*')
-                    && reader
+                if !field.contains('*') {
+                    return reader
                         .iter_columns()?
-                        .any(|(name, _)| name.as_str() == field)
-                {
-                    return Ok(vec![field.to_owned()]);
+                        .find(|(name, _)| name == field)
+                        .map(|_| vec![field.to_owned()])
+                        .ok_or_else(|| {
+                            TantivyError::SchemaError(format!(
+                                "Field '{field}' in docvalue_fields does not exist"
+                            ))
+                        });
                 }
 
                 let pattern = globbed_string_to_regex(field)?;
@@ -235,10 +241,13 @@ impl TopHitsAggregationReq {
                     })
                     .filter(|name| pattern.is_match(name))
                     .collect::<Vec<_>>();
-                assert!(
-                    !fields.is_empty(),
-                    "No fields matched the glob '{field}' in docvalue_fields"
-                );
+
+                if fields.is_empty() {
+                    return Err(TantivyError::SchemaError(format!(
+                        "No fields matched the glob '{field}' in docvalue_fields"
+                    )));
+                }
+
                 Ok(fields)
             })
             .collect::<crate::Result<Vec<_>>>()?
@@ -663,6 +672,7 @@ mod tests {
     use crate::collector::ComparableDoc;
     use crate::query::AllQuery;
     use crate::schema::OwnedValue;
+    use crate::TantivyError;
 
     fn invert_order(cmp_feature: DocValueAndOrder) -> DocValueAndOrder {
         let DocValueAndOrder { value, order } = cmp_feature;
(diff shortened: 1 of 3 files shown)


src/aggregation/metric/top_hits.rs
```diff
@@ -218,12 +225,16 @@ impl TopHitsAggregationReq {
...
             .map(|field| {
-                if !field.contains('*')
-                    && reader
+                if !field.contains('*') {
+                    return reader
                         .iter_columns()?
-                        .any(|(name, _)| name.as_str() == field)
-                {
-                    return Ok(vec![field.to_owned()]);
+                        .find(|(name, _)| &name.replace(JSON_PATH_SEGMENT_SEP_STR, ".") == field)
```
fulmicoton: that's not the way to do things. Can you look into the code and see how it is done elsewhere?

Totodore: I used `FastFieldReaders::resolve_field` which required to make it pub(crate), as I don't know the type of the column when checking if the field exists so I cant use `column_opt`. Is it Ok? or should I use something else?

fulmicoton: There are a bunch of aggregations relying on columns.

fulmicoton: There are a bunch of aggregations relying on columns.

Totodore: Could you elaborate?

src/aggregation/metric/top_hits.rs
fulmicoton, line 225: this is pretty terrible.

Can you answer the question what is "self.doc_value_fields"?

Can you investigate whether we can fix this subaggregation in a deeper way?

src/aggregation/metric/top_hits.rs
```diff
@@ -218,27 +226,33 @@ impl TopHitsAggregationReq {
             .doc_value_fields
             .iter()
             .map(|field| {
-                if !field.contains('*')
-                    && reader
-                        .iter_columns()?
-                        .any(|(name, _)| name.as_str() == field)
-                {
+                if !field.contains('*') {
+                    reader.resolve_field(field)?.ok_or_else(|| {
```
fulmicoton, line 230: I suspect dynamic_column_handles is better (as it is public). I am not sure though

src/aggregation/metric/top_hits.rs
```diff
@@ -218,27 +226,33 @@ impl TopHitsAggregationReq {
...
+                        ))
+                    })?;
+
                     return Ok(vec![field.to_owned()]);
                 }
 
                 let pattern = globbed_string_to_regex(field)?;
                 let fields = reader
+                    .columnar()
                     .iter_columns()?
```
fulmicoton, line 242: I think you mean to use it in Quickwit? I suspect it might not work if you have a large number of columns.

src/aggregation/metric/top_hits.rs
```diff
@@ -218,27 +226,33 @@ impl TopHitsAggregationReq {
...
                     })
                     .filter(|name| pattern.is_match(name))
                     .collect::<Vec<_>>();
-                assert!(
-                    !fields.is_empty(),
-                    "No fields matched the glob '{field}' in docvalue_fields"
-                );
+
+                if fields.is_empty() {
+                    return Err(TantivyError::SchemaError(format!(
```
fulmicoton, line 251: thx

src/aggregation/metric/top_hits.rs
Totodore: You mean the full reallocation? In that case I agree and I can refactor that.

The only purpose of this seems to map globbed patterns to real fields, non glob-patterns should not trigger clones though.

src/aggregation/metric/top_hits.rs
```diff
@@ -218,27 +226,33 @@ impl TopHitsAggregationReq {
...
                     return Ok(vec![field.to_owned()]);
                 }
 
                 let pattern = globbed_string_to_regex(field)?;
                 let fields = reader
+                    .columnar()
                     .iter_columns()?
                     .map(|(name, _)| {
                         // normalize path from internal fast field repr
                         name.replace(JSON_PATH_SEGMENT_SEP_STR, ".")
```
fulmicoton, line 245: This seems wrong too

src/aggregation/metric/top_hits.rs
Totodore: `dynamic_column_handles` does a bit more work and internally calls `resolve_field`:
https://docs.rs/tantivy/latest/src/tantivy/fastfield/readers.rs.html#238-251

fulmicoton:
@Totodore the whole sub aggregation is terrible. We should have never merged this PR originally. Can you approach it with clean eyes and clean it up?

src/aggregation/metric/top_hits.rs
fulmicoton: ok, but can it is not public :-/

Changing to pub(crate) is not so bad but unnecessary here.

src/aggregation/metric/top_hits.rs
fulmicoton: There are a many aggregations relying on columns, including aggregation that do not know beforehands the type of the targeted columns. Term aggregation for instance. Before increasing the visibility of a method, could you see if you can do what is done there?

The thing I was missing is that this aggregation type implementation is terrible  and misleading.

src/aggregation/metric/top_hits.rs
fulmicoton: No the reallocation is not the problem. 

We need to be quantitative in this kind of discussion: this allocation will happen once per segment. This is negligible. Same thing for  FxHashMap vs HashMap in quickwit and your concern about calling a  method that is public but does "more".

On the other hand, the following code (again this is not your doing) is a massive footgun.

If you try to explain to someone what `doc_value_fields` is it would look something like this.

"well before we run validation, it is a human readable string that describes a user input, but after validation it is a string that tantivy uses to address the regular field or a json field that it uses internally"

It is even exposed to the external world because aggregations make it possible to get the list of columns used in an aggregation as it used for IO in quickwit.

Totodore:
Thanks for your feedbacks @fulmicoton I'll try to provide a full refacto based on what I can learn from other aggregation implementations (terms for example,) rather than a small patch.

fulmicoton:
@Totodore That would be awesome! Thank you!

For the most obscure part:
Quickwit need to pre-fetch all columns from S3 before running search. For this reason, aggregation need to explicitly declare the list of columns they rely on. We need to make sure that the function returns the column in the format expected.

Totodore:
@fulmicoton  I dig a bit and I don't really see how we could provide the dynamic field with pattern matching functionality available in `docvalue_fields` while pre-fetching columns in quickwit.

I think there are three solutions:
* Remove the pattern matching on column names feature entirely.
* Keep it but make it unusable from quickwit as we cannot pre-fetch the fast fields if there are wildcards.
* Update [`get_fast_field_names`](https://docs.rs/tantivy/latest/tantivy/aggregation/agg_req/fn.get_fast_field_names.html) to take a handle to the columns list to resolve the regexes dynamically when quickwit is getting all the fastfields on the aggregation.

PR #2809, opened 2026-01-22
fix: Fix hang in rare intersection case

stuhood:
In a rare case demonstrated by the repro on https://github.com/paradedb/paradedb/pull/3989, intersection could enter an infinite loop trying to `seek_into_the_danger_zone(TERMINATED)`.

This change adds two tests reproducing that behavior, and then has `seek_into_the_danger_zone` call `seek` when called with `TERMINATED` to fix both tests.

### DIFF
--- a/src/query/intersection.rs
+++ b/src/query/intersection.rs
@@ -304,6 +304,42 @@ mod tests {
         assert_eq!(intersection.doc(), TERMINATED);
     }
 
+    #[test]
+    #[ignore]
+    fn test_intersection_termination() {
+        use crate::query::score_combiner::DoNothingCombiner;
+        use crate::query::{BufferedUnionScorer, ConstScorer, VecDocSet};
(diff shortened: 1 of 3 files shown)


stuhood:
So, I really don't love this repro, but I failed to incorporate it into the existing proptests. Ideas welcome!

fulmicoton:
thank you so much!

fulmicoton:
Summing up my understanding:
The contract says `seek_into_the_danger_zone` returns true if do is part of the docset.

The blanket implementation goes like this.
```rust
let current_doc = self.doc();
if current_doc < target {
     self.seek(target);
}
self.doc() == target
```

So it will return true if target is `TERMINATED`, where really TERMINATED does not belong to the docset.

fulmicoton:
Merged as part of #2810

PR #2818, opened 2026-01-28
fix(query-grammar): Fix regexes between parentheses

Darkheir:
Handle case where regexes are defined between parentheses.

### DIFF
--- a/query-grammar/src/user_input_ast.rs
+++ b/query-grammar/src/user_input_ast.rs
@@ -66,6 +66,7 @@ impl UserInputLeaf {
             }
             UserInputLeaf::Range { field, .. } if field.is_none() => *field = Some(default_field),
             UserInputLeaf::Set { field, .. } if field.is_none() => *field = Some(default_field),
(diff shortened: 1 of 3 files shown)


query-grammar/src/user_input_ast.rs
```diff
@@ -66,6 +66,7 @@ impl UserInputLeaf {
             }
             UserInputLeaf::Range { field, .. } if field.is_none() => *field = Some(default_field),
             UserInputLeaf::Set { field, .. } if field.is_none() => *field = Some(default_field),
+            UserInputLeaf::Regex { field, .. } if field.is_none() => *field = Some(default_field),
```
rdettai-sk, line 69: I think it is worth testing this and the parentheses thing in query_parser.rs (test_regex())

Darkheir: Done :-)

fulmicoton:
@trinity-1686a can you have a look & merge if it looks ok?

trinity-1686a (approved):

PR #2825, opened 2026-01-30
faster exclude queries

### LINKED ISSUE #2822: Performance improvement on MUST NOT
tantivy is terrible on queries like 

`(+something -subquerya -subqueryb)`

That query will be executed as 
`(+something -(subquerya OR subqueryb))`

It would be better to interpret as something closer to 
`((+something -subquerya) -subqueryb))`
in spirit.

Although it is probably preferrable to not implement that way but just to have the Exclude scorer allow for several exclude DocSet.

Also we might want to remove specialization and add buffering on it.

PSeitz:
Faster exclude queries with multiple terms.

Changes `Exclude` to be able to exclude multiple DocSets, instead of putting the DocSets into a union.
Use `seek_danger` in `Exclude`.

closes #2822

Negated
Screenshot 2026-01-30 at 15 11 34

Negated (one term)
Screenshot 2026-01-30 at 15 11 24

Negated (two terms)
Screenshot 2026-01-30 at 15 11 29

### DIFF
--- a/src/query/boolean_query/boolean_weight.rs
+++ b/src/query/boolean_query/boolean_weight.rs
@@ -291,18 +291,6 @@ impl<TScoreCombiner: ScoreCombiner> BooleanWeight<TScoreCombiner> {
             }
         };
 
-        let exclude_scorer_opt: Option<Box<dyn Scorer>> = if exclude_scorers.is_empty() {
-            None
-        } else {
-            let exclude_specialized_scorer: SpecializedScorer =
-                scorer_union(exclude_scorers, DoNothingCombiner::default, num_docs);
-            Some(into_box_scorer(
-                exclude_specialized_scorer,
-                DoNothingCombiner::default,
-                num_docs,
-            ))
-        };
-
         let include_scorer = match (should_scorers, must_scorers) {
             (ShouldScorersCombinationMethod::Ignored, must_scorers) => {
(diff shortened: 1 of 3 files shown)


src/query/boolean_query/boolean_weight.rs
```diff
@@ -380,15 +368,32 @@ impl<TScoreCombiner: ScoreCombiner> BooleanWeight<TScoreCombiner> {
...
+
+        let include_scorer_boxed = into_box_scorer(include_scorer, &score_combiner_fn, num_docs);
+        if exclude_scorers.len() == 1 {
+            let exclude_scorer = exclude_scorers.pop().unwrap();
+            if exclude_scorer.is::<TermScorer>() {
+                // Optimize for single term exclusion.
+                return Ok(SpecializedScorer::Other(Box::new(Exclude::new(
+                    include_scorer_boxed,
+                    *(exclude_scorer
+                        .downcast::<TermScorer>()
```
fulmicoton-dd: nitpick:

downcast return an error that wraps the original pointer.
You could remove the unwrap and the call to `.is::<>()` and use the error instead.

fulmicoton-dd (approved):

src/query/boolean_query/boolean_weight.rs
PSeitz: replaced unwrap with `match`

PR #2827, opened 2026-02-03
feat(aggregation): expose extra fields for aggregations

congx4:
## Summary

This change modifies the Average aggregation to return `sum` and `count` alongside the computed average value, enabling downstream systems to properly merge results across multiple query steps.

## Changes

- Add `AverageMetricResult` struct with `value`, `sum`, and `count` fields
- Add `sum()` and `count()` getter methods to `IntermediateAverage`
- Update `MetricResult::Average` to use `AverageMetricResult`
- Update finalization to populate sum/count from intermediate result
- Update tests to expect new JSON format

## JSON Output Change

**Before:**
```json
{ "value": 2.5 }
```

**After:**
```json
{ "value": 2.5, "sum": 15.0, "count": 6 }
```

### DIFF
--- a/src/aggregation/agg_result.rs
+++ b/src/aggregation/agg_result.rs
@@ -10,7 +10,8 @@ use serde::{Deserialize, Serialize};
 
 use super::bucket::GetDocCount;
 use super::metric::{
-    ExtendedStats, PercentilesMetricResult, SingleMetricResult, Stats, TopHitsMetricResult,
+    AverageMetricResult, ExtendedStats, PercentilesMetricResult, SingleMetricResult, Stats,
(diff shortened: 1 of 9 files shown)


PSeitz:
`IntermediateAverage` is already used in quickwit as an intermediate result to merge results

congx4:
> `IntermediateAverage` is already used in quickwit as an intermediate result to merge results

While reusing `IntermediateAverage` avoids boilerplate, my concern is backward compatibility. Exposing the intermediate struct as the final result produces a breaking change for downstream consumers like Quickwit. Introducing `AverageMetricResult` preserves the expected `value` field.

PR #2829, opened 2026-02-06
add intermediate result accessors

congx4:
add intermediate result accessors in order to group multi steps queries in upstream.


src/aggregation/metric/average.rs
```diff
@@ -55,6 +55,11 @@ impl IntermediateAverage {
     pub(crate) fn from_stats(stats: IntermediateStats) -> Self {
         Self { stats }
     }
+
+    pub fn stats(&self) -> &IntermediateStats {
```
PSeitz, line 60: missing documentation

PSeitz (approved):

trinity-1686a (approved):

PR #2839, opened 2026-02-12
DataFusion Tantviy Table Providers

alexanderbianchi:
THIS IS A DRAFT

I'm still running benchmarks to try and find the best way to structure plans and what optimizations we can make

So far Df is slower, but not much slower. Tantivy is purposely single threaded for a given segment I've experimented with both keeping it for DF and changing it. 

I'm also working on a distributed datafusion Quickwit layer that will use these primitives

the goal is twofold
1. execute existing elastic search queries through datafusion (mostly for benchmarking and correctness testing)
2. expose full SQL capabilities on top of Quickwit/tantivy

Before pushing for review I will need to do a deep review myself, this is the result of a day long moms-basement style hackathon with claude.

we need https://github.com/apache/datafusion/pull/20246 for the three way joins to work. Also considering making this just it's own repo with quickwit + tantivy stuff together but i'm not sure how the dep tree winds up looking with that.


gene-bordegaray:
The PR mentioned will be here now: https://github.com/apache/datafusion/pull/20331 😄

tantivy-datafusion/src/fast_field_reader.rs
```diff
@@ -0,0 +1,370 @@
...
+                        DataFusionError::Internal(format!("str fast field '{name}' not found"))
+                    })?;
+
+                // Build dictionary values once (one string per unique term)
+                let num_terms = str_col.num_terms();
+                let mut dict_builder = StringBuilder::with_capacity(num_terms, num_terms * 16);
+                let mut buf = String::new();
+                for ord in 0..num_terms {
+                    buf.clear();
+                    str_col
```
fulmicoton, line 151: this thing can probably be considerably accelerated by streaming the dictionary directly.

tantivy-datafusion/datafusion_limitations.md
```diff
@@ -0,0 +1,53 @@
...
+    FastFieldDataSource
+
+-- Aggregations: needs ALL matching docs
+AggregateExec: gby=[category], aggr=[...]
+  HashJoinExec
+    InvertedIndexDataSource(topk=None)      ← scan 2
+    FastFieldDataSource
+```
+
+DataFusion currently has no mechanism for one subplan to feed multiple downstream consumers. Each `DataFrame` produces an independent physical plan tree, so forking a base DataFrame results in duplicated scans.
```
fulmicoton, line 23: Ah that's a bummer. I was told otherwise.

tantivy-datafusion/datafusion_limitations.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Impact
+
+- The inverted index posting lists are traversed twice (once with Block-WAND pruning for hits, once fully for aggregations)
+- Fast field columns are scanned twice (though dynamic filters prune both sides)
```
fulmicoton, line 28: Right now I think the way aggregation or maybe collection works in tantivy.
Instead of every consume fetching data in situ, I'd rather have them declare what they need upfront.

Then the flow would change into (pseudo code):
```
for batch_of_docs in query.stream_batch_of_docs() {
    fastfield.fill_values_for_all_cols(value_buffers);
    apply_aggregation(value_buffers)
}
```

It would avoid fetching values from the same column twice (usually timestamp is the problem).
More importantely, it would decouple aggregation from fetching data, and make it easier to iterate/optimize on both separately.

Introducing datafusion, or using vortex for the columnar would then be easier.

tantivy-datafusion/datafusion_limitations.md
```diff
@@ -0,0 +1,53 @@
...
+2. **Materialize the base** — collect the unfiltered join result into an in-memory table, then run both TopK sort and aggregations from materialized data:
+   ```rust
+   let base_batches = base_df.collect().await?;
+   let mem_table = MemTable::try_new(schema, vec![base_batches])?;
+   ctx.register_table("base", Arc::new(mem_table))?;
+   // Now both hits and aggs read from memory
+   ```
+   Trades memory for avoiding double scan. Good when the matching doc set is bounded.
+
+3. **Separate the concerns** — run TopK hits through the optimized inverted index path (with Block-WAND), and aggregations through fast-field-only scans (no join needed if aggs don't require `_score`). This is often the right split since aggregations rarely need BM25 scores.
```
fulmicoton, line 44: for logs, we do not use blockwand. Things are sorted by timestamps (and we have never worked on accelerating this queries but that's a shame).

tantivy-datafusion/df_limitations.md
```diff
@@ -0,0 +1,143 @@
...
+   columnar data as Arrow arrays without copying (via shared memory or FFI
+   buffers) would eliminate the materialization cost.
+
+## Category 2: Terms & Histogram (GROUP BY aggregations)
+
+**Pushdown: enabled** — 1.3–2.3x faster than DataFusion.
+
+### Why DF is slow for terms/histogram
+
+DataFusion's `AggregateExec` uses a generic hash-based GROUP BY:
```
fulmicoton, line 49: Even  if the key is a small integer? Maybe that's something we should fix in datafusion?

tantivy-datafusion/df_limitations.md
```diff
@@ -0,0 +1,143 @@
...
+1. Hash each row's group key (string hashing for terms)
+2. Probe/insert into a hash table per group
+3. Two-phase: Partial hash tables → repartition → Final hash merge
+
+For terms with 7 distinct values over 1M rows, this means 1M string hash
```
fulmicoton, line 54: string hashes? Does the group by really happen on strings (as opposed to tantivy term ord , or arrow term id)? (I don't think this is possible as this would be way slower than just 2x).

tantivy-datafusion/df_limitations.md
```diff
@@ -0,0 +1,143 @@
...
+operations into a 7-entry table. Tantivy's native collector uses
+**ordinal-indexed counting** — each term has a dense integer ordinal, and the
+collector increments `counts[ordinal]` directly. No hashing.
+
+### Why pushdown still has a ~2x gap vs native
+
+- **Doc ID collection overhead**: `collect_block` iterates `0..max_doc`, whereas
+  native tantivy drives iteration more efficiently via the `Collector` trait
+- **Arrow conversion**: After tantivy produces results, converting to Arrow
+  `RecordBatch` (string arrays for keys, i64 for counts) adds allocation
```
fulmicoton, line 64: strings will be reallly tough.

tantivy-datafusion/df_limitations.md
```diff
@@ -0,0 +1,143 @@
...
+
+Currently, string fast fields produce `DictionaryArray<Int32, Utf8>` where
+dictionary keys are **per-segment ordinals**. Different segments have different
+ordinal→string mappings, so DataFusion must resolve to strings for cross-segment
+grouping.
+
+If we introduced a **global dictionary** across segments:
+1. Build a merged dictionary at reader open time (union of per-segment dicts)
+2. Remap per-segment ordinals to global ordinals during batch generation
+3. All batches share the same dictionary → DF can GROUP BY integer keys directly
```
fulmicoton, line 78: Is there a way to push down aggregation in a way that groups by tantivy term ord, and the results themselves are mapped to a  global dictionary? 

(the benefit i'm shooting for here is that for selective queries, we might come across way less unique terms than the entire column)

tantivy-datafusion/df_limitations.md
```diff
@@ -0,0 +1,143 @@
...
+
+This would eliminate string hashing entirely for terms aggregations, making DF's
```
fulmicoton, line 80: today string hashing happens above segments?

tantivy-datafusion/df_limitations.md
```diff
@@ -0,0 +1,143 @@
...
+
+Single scan with hash GROUP BY on string bucket keys. This is correct and
+produces a normal `AggregateExec(GROUP BY bucket)` that the pushdown optimizer
+can replace with `TantivyAggregateExec`. However, hashing 1M string keys adds
+overhead that offsets the single-scan benefit.
+
+### Why pushdown doesn't help much for range
+
+The CASE path does 1 scan but hashes string keys for every row. Tantivy's native
+collector uses `RangeDocSet` with SIMD columnar scans — no hashing, no string
```
fulmicoton, line 132: I'm interested in how we perform against raw Datafusion with a modern columnar format in a silly

```
SELECT sum(some_int_col)
WHEN price >= 3 AND price < 5
```
especially where the price column is sparse or multivalued.

We do not have any index to accelerate range queries.
We run like a columnar would do it.

This is super important for us, because that's how we deal with timestamp range queries.
We **think** we are fast, but really we have never benched it against actual columnars.

tantivy-datafusion/df_limitations.md
```diff
@@ -0,0 +1,143 @@
...
+allocation. The gap is the same doc ID collection + Arrow conversion overhead
+as for terms, plus the string hashing cost of the CASE GROUP BY.
+
+### Potential improvements
+
+1. **Integer bucket keys in CASE**: Instead of `THEN 'cheap'`, use `THEN 0`,
+   `THEN 1`, etc., and resolve to string labels after aggregation. Integer
+   hashing is ~10x faster than string hashing.
```
fulmicoton, line 140: string hashing is really happening here two?

tantivy-datafusion/quickwit-tantivy-guide.md
```diff
@@ -0,0 +1,415 @@
...
+1M docs, 1 segment, target_partitions=10:
+  partition 0: docs 0..100K
+  partition 1: docs 100K..200K
+  ...
+  partition 9: docs 900K..1M
+```
+
+**Correctness invariants**:
+- Dictionary stability: All chunks from the same segment produce the same dictionary values (same `StrColumn.ord_to_str` results). Only the keys differ per chunk.
+- Alive bitset: Each chunk filters by `alive_bitset` within its range.
```
fulmicoton, line 246: we don't have any deletes in quickwit. You can omit them

tantivy-datafusion/df_limitations.md
alexanderbianchi: The problem is you can't guarantee that strings[key] = string is unique across recordbatches because they could have different strings arrays. I need to see if we can fix that though. I might be wrong.

tantivy-datafusion/df_limitations.md
alexanderbianchi: I would take some of this with a grain of salt, i benched many permutations some included full aggregation pushdown.

tantivy-datafusion/df_limitations.md
alexanderbianchi: yes, this should be possible

tantivy-datafusion/quickwit-tantivy-guide.md
alexanderbianchi: sounds good

tantivy-datafusion/df_limitations.md
alexanderbianchi: The default unless optimized is largely string hashing, so it'll take some elbow grease (which I am putting in) but yeah fighting string hashes is common right now.

alexanderbianchi:
Closing - now pursuing a single table provider instead of the decomposed one found in this PR. You can see the work in progress [here](https://github.com/alexanderbianchi/tantivy-datafusion), will push to the official repo when ready.

PR #2842, opened 2026-02-18
Use sketches-ddsketch fork with Java-compatible binary encoding

congx4:
## Description

Reference the `quickwit-oss/rust-sketches-ddsketch` fork (via `git` + `rev` in Cargo.toml) which adds native Java-compatible binary serialization for DDSketch. This enables Rust applications to produce DDSketch bytes that Java consumers can directly deserialize and merge via `DDSketchWithExactSummaryStatistics.decode()` from the `sketches-java` library.

### Why?

The upstream `sketches-ddsketch` Rust crate only supports serde-based serialization (JSON), while Java's `sketches-java` library uses a custom binary wire format. For distributed aggregation pipelines where Rust search nodes produce intermediate DDSketch results consumed by Java query orchestrators, binary compatibility is required.

### What changed

- Removed the vendored `sketches-ddsketch/` directory from the tantivy workspace
- Changed the `sketches-ddsketch` dependency from `path = "./sketches-ddsketch"` to `git + rev` pointing to [quickwit-oss/rust-sketches-ddsketch@555caf1](https://github.com/quickwit-oss/rust-sketches-ddsketch/tree/java-binary-encoding)
- The fork contains:
  - `encoding.rs`: Java-compatible binary encode/decode using signed/unsigned varint and VarDouble encoding
  - Cross-language golden-byte tests verifying binary compatibility with Java's DDSketch
  - Config/Store/DDSketch modifications to support `encode()`/`decode()` methods

### Testing

- All existing tantivy aggregation tests pass (percentiles, cardinality, etc.)
- All sketches-ddsketch tests pass including cross-language golden-byte tests


sketches-ddsketch/src/encoding.rs
```diff
@@ -0,0 +1,813 @@
...
+//! serialization so that sketches produced in Rust can be deserialized
+//! and merged by Java consumers.
+
+use std::fmt;
+
+use crate::config::Config;
+use crate::ddsketch::DDSketch;
+use crate::store::Store;
+
+// ---------------------------------------------------------------------------
```
congx4: This file is generated by Opus and it generates tests to ensure it is fully correct. See more details from the pr description.

PSeitz:
I don't think this should be in the tantivy repo. 

The changes should be in your `sketches-ddsketch` fork, where we try to upstream it.

fulmicoton-dd:
@congx4 I am not sure @PSeitz comment was clear. Can you just point to your repo with `{git =, rev=}`

congx4:
> @congx4 I am not sure @PSeitz comment was clear. Can you just point to your repo with `{git =, rev=}`

Thanks, but looks like we don't have permissions to folk it under Datadog, I will just folk it under quickwit-oss.

src/aggregation/metric/percentiles.rs
```diff
@@ -610,11 +616,11 @@ mod tests {
 
         assert_eq!(
             res["range_with_stats"]["buckets"][0]["percentiles"]["values"]["1.0"],
-            5.0028295751107414
+            5.002829575110705
```
fulmicoton-dd: nitpick: assert_eq! on float, especially after serialization in json quite brittle. You might want to check just that the number is nearly equal. I suspect there are macros to do that in the project.

Cargo.toml
```diff
@@ -64,7 +64,7 @@ query-grammar = { version = "0.25.0", path = "./query-grammar", package = "tanti
 tantivy-bitpacker = { version = "0.9", path = "./bitpacker" }
 common = { version = "0.10", path = "./common/", package = "tantivy-common" }
 tokenizer-api = { version = "0.6", path = "./tokenizer-api", package = "tantivy-tokenizer-api" }
-sketches-ddsketch = { version = "0.3.0", features = ["use_serde"] }
+sketches-ddsketch = { git = "https://github.com/quickwit-oss/rust-sketches-ddsketch.git", rev = "555caf1", features = ["use_serde"] }
```
fulmicoton-dd, line 67: do we still need serde?

fulmicoton-dd (approved):
Approved, but please have a look at the comments and see if they make sense

Cargo.toml
congx4: Yes, we still need it. `PercentilesCollector` derives `Serialize, Deserialize` for its inner `sketches_ddsketch::DDSketch` field, and IntermediateMetricResult derives `#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]` which is required for postcard intermediate result serialization during distributed search.

fulmicoton-dd (approved):

PR #2843, opened 2026-02-20
use FxHashSet in DocSetCollector

PSeitz:


fulmicoton-dd:
@PSeitz isn't it just for tests?

PSeitz:
> @PSeitz isn't it just for tests?

No, this is exported https://docs.rs/tantivy/latest/tantivy/collector/struct.DocSetCollector.html

fulmicoton:
"This collector is mostly useful for tests."

@PSeitz I suggest we do not export it. It is rather dumb to use a hashset here no? A Vec would be more efficient (at the very least at the SegmentCollector level).

I don't see the point of this PR.

PR #2854, opened 2026-03-18
fix: deduplicate doc counts in term aggregation for multi-valued fields

### LINKED ISSUE #2721: `doc_count` in aggregation result is not the actual number of documents
I noticed that in term aggregations, `doc_count` isn't the actual number of documents that match the term but the number of times the term appears. For example, in [this test](https://github.com/quickwit-oss/tantivy/blob/main/src/aggregation/bucket/term_agg.rs#L1813-L1900), some documents have the term `text_field => "Hello Hello"` twice and are therefore counted twice. 

In Lucene, I think you can get the right `doc_count` by deduplicating thanks to SortedSetDocValues which stores multi valued terms for a document in sorted order. My understanding is that ES uses this to display actual document counts.

Could something similar be achieved in Tantivy? do we want to? If not, I think we should document [doc_count](https://docs.rs/tantivy/latest/tantivy/aggregation/agg_result/struct.BucketEntry.html#structfield.doc_count) to clarify this.

nuri-yoo:
## Summary

- Term aggregation was counting **term occurrences** instead of **documents** for multi-valued fields. A document with the same value appearing multiple times would inflate `doc_count`.
- Added `fetch_block_with_missing_unique_per_doc` to `ColumnBlockAccessor` that deduplicates `(doc_id, value)` pairs after fetching, and used it in term aggregation collection.
- Fixed existing test assertions that were asserting the buggy counts.

## Details

The root cause is in `ColumnBlockAccessor::fetch_block`, which calls `row_ids_for_docs`. For multi-valued columns, this can return the same `(doc_id, value)` pair multiple times. `term_entry()` then increments `bucket.count` for each occurrence, inflating `doc_count`.

The fix adds a deduplication step that removes consecutive duplicate `(doc_id, value)` pairs in O(n), which is safe because `row_ids_for_docs` returns entries sorted by doc_id.

This also fixes the sub-aggregation path — `CachedSubAggs::push(bucket_id, doc)` was receiving the same doc multiple times for the same bucket.

## Test plan

- [x] Existing `terms_aggregation_missing_multi_value` test now asserts correct doc counts (4 instead of 5, 2 instead of 3)
- [x] All 169 aggregation tests pass
- [x] `cargo +nightly fmt --all` clean

Fixes #2721

### DIFF
--- a/columnar/src/block_accessor.rs
+++ b/columnar/src/block_accessor.rs
@@ -69,21 +69,44 @@ impl<T: PartialOrd + Copy + std::fmt::Debug + Send + Sync + 'static + Default>
         docs: &[u32],
         accessor: &Column<T>,
         missing: Option<T>,
-    ) {
+    ) where
+        T: Ord,
+    {
         self.fetch_block_with_missing(docs, accessor, missing);
         if accessor.index.get_cardinality().is_multivalue() {
             self.dedup_docid_val_pairs();
         }
     }
 
-    /// Removes consecutive duplicate (doc_id, value) pairs from the caches.
+    /// Removes duplicate (doc_id, value) pairs from the caches.
     ///
-    /// After `fetch_block`, entries for the same doc are adjacent, so duplicates
-    /// (same doc, same value) are consecutive and can be removed in O(n).
-    fn dedup_docid_val_pairs(&mut self) {
+    /// After `fetch_block`, entries are sorted by doc_id, but values within
+    /// the same doc may not be sorted (e.g. `(0,1), (0,2), (0,1)`).
+    /// We group consecutive entries by doc_id, sort values within each group
+    /// if it has more than 2 elements, then deduplicate adjacent pairs.
+    fn dedup_docid_val_pairs(&mut self)
+    where T: Ord {
         if self.docid_cache.len() <= 1 {
             return;
         }
+
+        // Sort values within each doc_id group so duplicates become adjacent.
+        let mut start = 0;
+        while start < self.docid_cache.len() {
+            let doc = self.docid_cache[start];
+            let mut end = start + 1;
+            while end < self.docid_cache.len() && self.docid_cache[end] == doc {
+                end += 1;
+            }
+            if end - start > 2 {
+                self.val_cache[start..end].sort();
+            } else if end - start == 2 && self.val_cache[start] > self.val_cache[start + 1] {
+                self.val_cache.swap(start, start + 1);
+            }
+            start = end;
+        }
+
+        // Now duplicates are adjacent — deduplicate in place.
         let mut write = 0;
         for read in 1..self.docid_cache.len() {
             if self.docid_cache[read] != self.docid_cache[write]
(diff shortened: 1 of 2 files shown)


columnar/src/block_accessor.rs
```diff
@@ -58,6 +58,49 @@ impl<T: PartialOrd + Copy + std::fmt::Debug + Send + Sync + 'static + Default>
...
+    /// where multi-valued fields can produce duplicate entries that inflate counts.
+    #[inline]
+    pub fn fetch_block_with_missing_unique_per_doc(
+        &mut self,
+        docs: &[u32],
+        accessor: &Column<T>,
+        missing: Option<T>,
+    ) {
+        self.fetch_block_with_missing(docs, accessor, missing);
+        if !accessor.index.get_cardinality().is_full() {
```
PSeitz: it's only necessary to deduplicate for multivalue cardinality

nuri-yoo: Fixed, Thanks for the review

columnar/src/block_accessor.rs
```diff
@@ -58,6 +58,49 @@ impl<T: PartialOrd + Copy + std::fmt::Debug + Send + Sync + 'static + Default>
...
+    /// After `fetch_block`, entries for the same doc are adjacent, so duplicates
+    /// (same doc, same value) are consecutive and can be removed in O(n).
+    fn dedup_docid_val_pairs(&mut self) {
+        if self.docid_cache.len() <= 1 {
+            return;
+        }
+        let mut write = 0;
+        for read in 1..self.docid_cache.len() {
+            if self.docid_cache[read] != self.docid_cache[write]
+                || self.val_cache[read] != self.val_cache[write]
```
PSeitz, line 119: I think we should check only for duplicate docids, not for duplicate values?

Can you extend the tests to capture this?

PSeitz: Ah nevermind, we need to check both, so that termid only filters duplicate values on the same docid, but still handles multi-values

PSeitz: It does not cover pairs if the values are not consecutive, e.g. :

(0, 1), (0, 2), (0, 1)

It's a bit more expensive. I think we could 
* make groups of consecutive docids (just start and end pos)
* if the group contains more than 2 elements sort by value
* then do the same algorithm as now

nuri-yoo: Fixed. Values within each doc_id group are now sorted before deduplicating, and added unit tests.

columnar/src/block_accessor.rs
```diff
@@ -58,6 +58,72 @@ impl<T: PartialOrd + Copy + std::fmt::Debug + Send + Sync + 'static + Default>
...
+        // Sort values within each doc_id group so duplicates become adjacent.
+        let mut start = 0;
+        while start < self.docid_cache.len() {
+            let doc = self.docid_cache[start];
+            let mut end = start + 1;
+            while end < self.docid_cache.len() && self.docid_cache[end] == doc {
+                end += 1;
+            }
+            if end - start > 2 {
+                self.val_cache[start..end].sort();
```
PSeitz, line 110: We only need to "rewrite" the data if this block actually has multi values.
The block below does this in every case, but we could just do it if we actually have duplicates.

E.g. by setting a dedup flag.

The benchmarks we have in the repo are pathological for this use case:

`cargo bench --bench agg_bench multivalue_term`

```
multivalue
terms_7                                        Memory: 35.8 KB              Avg: 6.8315ms (+24.30%)     Median: 6.8226ms (+25.36%)     [6.7555ms .. 6.9741ms]
terms_all_unique                               Memory: 14.4 MB (+12.35%)    Avg: 11.5699ms (+10.31%)    Median: 11.4794ms (+9.81%)     [11.3364ms .. 12.3389ms]
terms_150_000                                  Memory: 3.0 MB (+0.55%)      Avg: 11.1931ms (+15.22%)    Median: 11.2063ms (+15.86%)    [10.8713ms .. 11.4624ms]
terms_many_top_1000                            Memory: 5.2 MB (+0.02%)      Avg: 14.2467ms (+11.21%)    Median: 14.2215ms (+11.22%)    [13.7755ms .. 14.8058ms]
terms_many_order_by_term                       Memory: 3.0 MB (+0.55%)      Avg: 9.8394ms (+16.38%)     Median: 9.8637ms (+17.31%)     [9.5194ms .. 10.1383ms]
terms_many_with_top_hits                       Memory: 49.9 MB (+1.90%)     Avg: 110.6834ms (+3.53%)    Median: 109.9098ms (+5.59%)    [98.3478ms .. 137.3680ms]
terms_all_unique_with_avg_sub_agg              Memory: 56.4 MB (+2.88%)     Avg: 26.5355ms (+2.60%)     Median: 26.6464ms (+4.64%)     [24.7364ms .. 29.9060ms]
terms_many_with_avg_sub_agg                    Memory: 13.5 MB (+0.01%)     Avg: 25.5032ms (+1.03%)     Median: 25.4548ms (+2.08%)     [22.9880ms .. 29.1043ms]
terms_status_with_avg_sub_agg                  Memory: 92.6 KB (-0.47%)     Avg: 13.1254ms (+16.84%)    Median: 13.1155ms (+16.91%)    [12.5845ms .. 13.7508ms]
terms_status_with_histogram                    Memory: 108.8 KB (+0.42%)    Avg: 12.8820ms (+18.11%)    Median: 12.7837ms (+17.62%)    [12.5241ms .. 13.4927ms]
terms_zipf_1000                                Memory: 68.2 KB (+0.00%)     Avg: 6.8137ms (+26.97%)     Median: 6.7798ms (+27.19%)     [6.6736ms .. 7.1462ms]
terms_zipf_1000_with_histogram                 Memory: 1.3 MB (-3.39%)      Avg: 32.5976ms (+8.00%)     Median: 32.3621ms (+7.79%)     [31.9863ms .. 34.1550ms]
terms_zipf_1000_with_avg_sub_agg               Memory: 465.3 KB (+0.69%)    Avg: 17.5710ms (+12.78%)    Median: 17.4528ms (+13.23%)    [17.0784ms .. 19.2251ms]
terms_many_json_mixed_type_with_avg_sub_agg    Memory: 20.3 MB (+14.09%)    Avg: 31.1680ms (+0.97%)     Median: 30.8948ms (+0.65%)     [30.1223ms .. 34.5088ms]
terms_status_with_cardinality_agg              Memory: 4.7 MB (+0.01%)      Avg: 39.3074ms (-3.40%)     Median: 39.0745ms (-7.32%)     [38.8384ms .. 41.4017ms]
```

columnar/src/block_accessor.rs
```diff
@@ -58,6 +58,72 @@ impl<T: PartialOrd + Copy + std::fmt::Debug + Send + Sync + 'static + Default>
...
+            } else if end - start == 2 && self.val_cache[start] > self.val_cache[start + 1] {
+                self.val_cache.swap(start, start + 1);
```
PSeitz: this is nonsensical, the algorithm below doesn't need sorted values only grouped ones (which is always the case for two values)

nuri-yoo:
Fixed both issues:

1. Added early return when no consecutive doc_ids are equal, skipping sort and dedup entirely for blocks without multivalue entries.
2. Removed the unnecessary 2-element swap.

Benchmark results after the fix (`cargo bench --bench agg_bench multivalue_term`):

| Benchmark | Before (regression) | After fix |
|-----------|-------------------|-----------|
| terms_7 | +24.30% | 5.94ms |
| terms_all_unique | +10.31% | 10.45ms |
| terms_150_000 | +15.22% | 10.29ms |
| terms_many_top_1000 | +11.21% | 13.27ms |
| terms_many_order_by_term | +16.38% | 8.90ms |
| terms_zipf_1000 | +26.97% | 5.87ms |
| terms_status_with_histogram | +18.11% | 11.77ms |
| terms_status_with_avg_sub_agg | +16.84% | 12.07ms |

No measurable regression compared to main.

PSeitz (approved):
Thanks!

PR #2865, opened 2026-03-27
Optimizing top K using Adrien Grand's ideas

fulmicoton:
https://jpountz.github.io/2025/08/28/compiled-vs-vectorized-search-engine-edition.html


PSeitz:
I read it as grand ideas :)

fulmicoton:
It works both ways :D

jpountz:
Nice, computing scores eagerly and using them to skip by score before doing membership checks on follow-up clauses helped a lot when it was merged in Lucene (https://github.com/apache/lucene/pull/12382). Have you had a chance to run benchmarks already?

fulmicoton:
Yes it already improves perf greatly, but I need to add vectorization in.

src/query/boolean_query/block_wand_intersection.rs
```diff
@@ -0,0 +1,440 @@
...
+        // This ensures the block_max values are valid for all docs in [doc, window_end].
+        // Different scorers have independently aligned blocks, so we must use the
+        // smallest window where all block_max values hold.
+        let mut window_end: DocId = leader.last_doc_in_block();
+
+        let mut secondary_block_max_sum: Score = 0.0;
+        for secondary in secondaries.iter_mut() {
+            secondary.seek_block(doc);
+            secondary_block_max_sum += secondary.block_max_score();
+            window_end = window_end.min(secondary.last_doc_in_block());
```
jpountz, line 75: For the record I remember seeing some missed performance opportunities with taking the min `last_doc_in_block` across clauses, as it could result in small-ish windows, so I ended up computing the window based on the `leader` for conjunctive queries, and the min `last_doc_in_block` across essential clauses for disjunctive queries (ignoring non-essential clauses).

(Also a reason why Lucene stores max scores on 2 levels, though if I were to do it again, I'd probably look into making the 2nd level docId-aligned (e.g. every 65k docs) rather than block-aligned.)

src/query/boolean_query/block_wand_intersection.rs
```diff
@@ -0,0 +1,440 @@
...
+
+        for idx in start_idx..end_idx {
+            let candidate_doc = block_docs[idx];
+            let fieldnorm_id = fieldnorm_reader.fieldnorm_id(candidate_doc);
+            let term_freq = block_freqs[idx];
+            let leader_score = bm25_weight.score(fieldnorm_id, term_freq);
+            candidate_doc_ids[num_candidates] = candidate_doc;
+            candidate_scores[num_candidates] = leader_score;
+            num_candidates += (leader_score > score_threshold) as usize;
+        }
```
jpountz, line 125: Oh nice, Lucene first computes scores in buffers and then filters in place, but you're saving one copy by doing scoring+filtering in a single pass.

jpountz (approved):
Now with vectorization, very exciting!

jpountz:
Out of curiosity, how does Tantivy mask deleted documents? I was expected to see something about them in the new scorer but maybe Tantivy doesn't need to because it handles them differently?

fulmicoton:
@jpountz It is a bitset. 

We kind of cheat. We never do that at the scorer level. If there are deleted documents, we omit them at collection level.
It used to be pushed down further, but it was annoying code wise, and hurting if there was few deletes.

fulmicoton:
@stuhood can you review? This increases intersection perf on top K bm25.

fulmicoton:
On top 10, this increases intersection significantly. 
622 μs -> 355 μs

src/query/boolean_query/block_wand_intersection.rs
```diff
@@ -0,0 +1,461 @@
...
+/// For each window, the sum of block_max_scores is compared to the current threshold;
+/// if the block can't beat it, the entire block is skipped.
+///
+/// Within non-skipped blocks, individual documents are pruned by checking whether
+/// leader_score + sum(secondary block_max_scores) can exceed the threshold before
+/// performing the expensive intersection membership check (seeking into secondary scorers).
+///
+/// # Preconditions
+/// - `scorers` has at least 2 elements
+/// - `scorers` has less than 16 elements
```
stuhood: This seems like a very reasonable constraint, but is it worth calling out the reason for it? Guaranteeing that we register/stack allocate the scores? It seems like you could potentially make it a generic parameter and generate a few copies of this method for interesting sizes (e.g. 4, 8, 16).

src/query/boolean_query/block_wand_intersection.rs
```diff
@@ -0,0 +1,461 @@
...
+            }
+            window_end = window_end.min(secondary.last_doc_in_block());
+            let bms = secondary.block_max_score();
+            secondary_block_max_scores[idx] = bms;
+            secondary_block_max_sum += bms;
+        }
+
+        // Precompute suffix sums: suffix[i] = sum of block_max for secondaries[i+1..].
+        // Used in Phase 2 to prune candidates that can't beat threshold even with
+        // remaining secondaries contributing their block_max.
```
stuhood: If the `leader_block_max + secondary_block_max_sum <= threshold` condition doesn't pass, then these precomputed values are unused: can move them after that case.

Additionally, you can skip computing these if `num_candidates` is 0 I think.

src/query/boolean_query/block_wand_intersection.rs
```diff
@@ -0,0 +1,461 @@
...
+                }
+            }
+
+            // All secondaries matched.
+            if total_score > threshold {
+                threshold = callback(candidate_doc, total_score);
+
+                if leader_max_score + secondaries_global_max_sum <= threshold {
+                    return;
+                }
```
stuhood, line 173: Is this worth checking per-document rather than per-block?

stuhood (approved):
This looks great, thanks!

src/query/boolean_query/block_wand_intersection.rs
fulmicoton: absolutely.

src/query/boolean_query/block_wand_intersection.rs
fulmicoton: it was to keep a bunch of arrays on the stack but it does not really matter actually. I moved these to the heap.

src/query/boolean_query/block_wand_intersection.rs
fulmicoton: I don't understand the question.

stuhood: Just that this check is currently inside of the inner loop over documents, as opposed to the outer loop over (leader) blocks. Not a big deal.

PSeitz:
Not sure if it's coming from this, but we seem to have a regression on `intersection_union` for TOP_100 between the last release and current main `4480cf0`

intersection_union_top_100

fulmicoton:
Probably more the topK collector that is based on binary heap now?

PR #2868, opened 2026-03-30
Abstract tantivy's data storage behind traits for pluggable backends

PSeitz:
Extract trait interfaces from tantivy's core reader types so that alternative storage backends (e.g. Quickwit) can provide their own implementations while tantivy's query engine works through dynamic dispatch.

Reader trait extraction:

- SegmentReader is now a trait; the concrete implementation is renamed to TantivySegmentReader.
- DynInvertedIndexReader trait for object-safe dynamic dispatch, plus a typed InvertedIndexReader trait with associated Postings/DocSet types for static dispatch.  The concrete reader becomes TantivyInvertedIndexReader.
- StoreReader is now a trait; the concrete implementation is renamed to TantivyStoreReader.  get() returns TantivyDocument directly instead of requiring a generic DocumentDeserialize bound.

Typed downcast for performance-critical paths:

- try_downcast_and_call() + TypedInvertedIndexReaderCb allow query weights (TermWeight, PhraseWeight) to attempt a downcast to the concrete TantivyInvertedIndexReader, obtaining typed postings for zero-cost scoring, and falling back to the dynamic path otherwise.
- TermScorer<TPostings> is now generic over its postings type.
- PostingsWithBlockMax trait enables block-max WAND acceleration through the trait boundary.
- block_wand() and block_wand_single_scorer() are generic over PostingsWithBlockMax, and for_each_pruning is dispatched through the SegmentReader trait so custom backends can provide their own block-max implementations.

Searcher decoupled from Index:

- New SearcherContext holds schema, executor, and tokenizers.
- Searcher can be constructed from Vec<Arc<dyn SegmentReader>> via Searcher::from_segment_readers(), without needing an Index.
- Searcher::index() is deprecated in favor of Searcher::context().

Postings and DocSet changes:

- Postings trait gains doc_freq() -> DocFreq (Exact/Approximate) and has_freq().
- RawPostingsData struct carries raw postings bytes across the trait boundary for custom reader implementations.
- BlockSegmentPostings::open() takes OwnedBytes instead of FileSlice.
- DocSet gains fill_bitset() method.

Scorer improvements:

- Scorer trait absorbs for_each, for_each_pruning, and explain (previously free functions or on Weight).
- box_scorer() helper avoids double-boxing Box<dyn Scorer>.
- BoxedTermScorer wraps a type-erased term scorer.
- BufferedUnionScorer initialization fixed to avoid an extra advance() on construction.

Other changes:

- Document::to_json() now returns serde_json::Value; the old string serialization is renamed to to_serialized_json().
- DocumentDeserialize removed from the store reader public API.


src/core/searcher.rs
```diff
@@ -71,9 +168,66 @@ pub struct Searcher {
...
+    pub fn from_segment_readers<Ctx: Into<SearcherContext>>(
+        context: Ctx,
+        segment_readers: Vec<Arc<dyn SegmentReader>>,
+    ) -> crate::Result<Searcher> {
+        Self::from_segment_readers_with_generation_id(context, segment_readers, 0)
+    }
+
+    /// Same as [`Searcher::from_segment_readers`] but allows setting
+    /// a custom generation id.
+    pub fn from_segment_readers_with_generation_id<Ctx: Into<SearcherContext>>(
```
fulmicoton-dd, line 185: Is there a real use case for exposing a public function with the generation id?

src/core/searcher.rs
```diff
@@ -85,7 +239,7 @@ impl Searcher {
     ///
     /// The searcher uses the segment ordinal to route the
     /// request to the right `Segment`.
-    pub fn doc<D: DocumentDeserialize>(&self, doc_address: DocAddress) -> crate::Result<D> {
+    pub fn doc(&self, doc_address: DocAddress) -> crate::Result<TantivyDocument> {
```
fulmicoton-dd, line 227: This is reverting a public feature. Can we check this is not hurting people? Do we use it in Quickwit?
Can you make sure this shows up in the changelog?

src/index/segment_reader.rs
```diff
@@ -6,17 +6,101 @@ use common::{ByteCount, HasLen};
...
+    fn max_doc(&self) -> DocId;
+
+    /// Returns the number of alive documents. Deleted documents are not counted.
+    fn num_docs(&self) -> DocId;
+
+    /// Returns the schema of the index this segment belongs to.
+    fn schema(&self) -> &Schema;
+
+    /// Performs a for_each_pruning operation on the given scorer.
+    fn for_each_pruning(
```
fulmicoton-dd: What is the justification to have for_each_pruning in there?
The goal is to allow custom project to have their own specialized impl?

src/index/index_meta.rs
```diff
@@ -287,7 +287,6 @@ pub struct IndexMeta {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub payload: Option<String>,
 }
```
fulmicoton-dd, line 289: new line was not bad useful

src/core/searcher.rs
PSeitz: currently not that I'm aware of. Should be probably moved to `SearcherContext` anyway

src/core/searcher.rs
PSeitz: we use `TantivyDocument` in quickwit currently. It's unlikely relevant for performance, I'll ask in the chat

src/index/segment_reader.rs
PSeitz: This is actually the reason I converted it to a draft to remove it from there first. We don't need it with the downcast approach.

src/indexer/merger.rs
```diff
@@ -355,7 +368,8 @@ impl IndexMerger {
                          indexed. Have you modified the schema?",
         );
 
-        let mut segment_postings_containing_the_term: Vec<(usize, SegmentPostings)> = vec![];
+        let mut segment_postings_containing_the_term: Vec<(usize, Box<dyn Postings>)> =
```
fulmicoton-dd, line 371: Is there really no regression here!?

src/index/inverted_index_reader.rs
```diff
@@ -10,37 +15,268 @@ use itertools::Itertools;
...
 ///
 /// It is safe to delete the segment associated with
-/// an `InvertedIndexReader`. As long as it is open,
+/// an `InvertedIndexReader` implementation. As long as it is open,
 /// the [`FileSlice`] it is relying on should
 /// stay available.
 ///
-/// `InvertedIndexReader` are created by calling
+/// `TantivyInvertedIndexReader` instances are created by calling
 /// [`SegmentReader::inverted_index()`](crate::SegmentReader::inverted_index).
```
fulmicoton-dd, line 242: why was it moved here?

benches/fill_bitset.rs
```diff
@@ -0,0 +1,106 @@
...
+        let inverted_index = &inverted_index;
+        let term_info = &term_info;
+        // This uses BlockSegmentPostings directly, bypassing SegmentPostings entirely.
+        group.register("BlockSegmentPostings direct", move |_| {
+            let raw = inverted_index
+                .read_raw_postings_data(term_info, IndexRecordOption::Basic)
+                .unwrap();
+            let mut block_postings = BlockSegmentPostings::open(
+                term_info.doc_freq,
+                raw.postings_data,
```
fulmicoton-dd, line 68: That's code smell here:  inverted index at this point is a DynInvertedIndex, yet you use the fact that you expect it to be the standard one to interpret the raw postings data.

src/index/inverted_index_reader.rs
```diff
@@ -10,37 +15,268 @@ use itertools::Itertools;
...
+    fn terms(&self) -> &TermDictionary;
+
+    /// Return the fields and types encoded in the dictionary in lexicographic order.
+    /// Only valid on JSON fields.
+    ///
+    /// Notice: This requires a full scan and therefore **very expensive**.
+    fn list_encoded_json_fields(&self) -> io::Result<Vec<InvertedIndexFieldSpace>>;
+
+    /// Returns the raw postings bytes and metadata for a term.
+    fn read_raw_postings_data(
```
fulmicoton-dd: Why do we need read_raw_postings_data? It seems very strange.

Also, I expected InvertedIndexReader: DynInvertedIndexReader

src/index/inverted_index_reader.rs
```diff
@@ -10,37 +15,268 @@ use itertools::Itertools;
...
+/// Attempts to downcast a `DynInvertedIndexReader` to tantivy's concrete
+/// `TantivyInvertedIndexReader` before falling back to the dynamic path.
+///
+/// The body is compiled twice: once with the concrete reader (yielding typed
+/// postings such as `SegmentPostings`) and once with the dynamic reader
+/// (yielding `Box<dyn Postings>`).  The body must therefore be generic
+/// enough to work with both postings types.
+///
+/// # Example
+///
```
fulmicoton-dd, line 193: Does this thing help much? It keeps things dynamic but attempts to replace the dynamic part by a if-statement in some case.

PSeitz: The example is a bit silly. The main use case is to build query types with concrete Postlinglists types e.g. `TermScorer<TantivySegmentPostings>`

src/postings/block_segment_postings.rs
```diff
@@ -88,19 +79,18 @@ fn split_into_skips_and_postings(
 }
 
 impl BlockSegmentPostings {
-    /// Opens a `BlockSegmentPostings`.
+    /// Opens a `StandardPostingsReader`.
```
fulmicoton-dd: ?

src/postings/loaded_postings.rs
```diff
@@ -25,16 +25,16 @@ impl LoadedPostings {
     /// Creates a new `LoadedPostings` from a `SegmentPostings`.
     ///
     /// It will also preload positions, if positions are available in the SegmentPostings.
-    pub fn load(segment_postings: &mut SegmentPostings) -> LoadedPostings {
-        let num_docs = segment_postings.doc_freq() as usize;
+    pub fn load(postings: &mut Box<dyn Postings>) -> LoadedPostings {
```
fulmicoton-dd, line 28: &mut dyn Postings

I don't think there is a case for the Box, is there?

src/postings/loaded_postings.rs
```diff
@@ -25,16 +25,16 @@ impl LoadedPostings {
...
+        let num_docs: usize = u32::from(postings.doc_freq()) as usize;
         let mut doc_ids = Vec::with_capacity(num_docs);
         let mut positions = Vec::with_capacity(num_docs);
         let mut position_offsets = Vec::with_capacity(num_docs);
-        while segment_postings.doc() != TERMINATED {
+        while postings.doc() != TERMINATED {
```
fulmicoton-dd, line 33: that will be slow. If it is used in RegexPhraseQuery, we might want to push this to Postings.

Although... The current impl is pretty terrible: we should probably not load positions but instead save something that makes it easy to load them later. Like an offset or something.

src/postings/mod.rs
```diff
@@ -16,22 +23,53 @@ mod recorder;
...
+    serialize_postings, IndexingPosition, PostingsWriter, PostingsWriterEnum,
+};
 pub use self::segment_postings::SegmentPostings;
 pub use self::serializer::{FieldSerializer, InvertedIndexSerializer};
-pub(crate) use self::skip::{BlockInfo, SkipReader};
 pub use self::term_info::TermInfo;
 
+/// Raw postings bytes and metadata read from storage.
+#[derive(Debug, Clone)]
+pub struct RawPostingsData {
```
fulmicoton-dd, line 46: Do we need this to be pub?

src/query/union/buffered_union.rs
```diff
@@ -285,7 +315,7 @@ where
         let mut is_hit = false;
         let mut min_new_target = TERMINATED;
 
-        for docset in self.docsets.iter_mut() {
+        for docset in self.scorers.iter_mut() {
```
fulmicoton-dd, line 318: rename scorer maybe?

src/query/union/buffered_union.rs
```diff
@@ -239,16 +271,14 @@ where
             doc
         } else {
             // clear the buffered info.
-            for obsolete_tinyset in self.bitsets.iter_mut() {
-                *obsolete_tinyset = TinySet::empty();
-            }
+            self.bitsets.fill(TinySet::empty());
```
fulmicoton-dd, line 274: I wonder if we should make the compiler's life easier by have a const EMPTY: TinySet

src/query/union/buffered_union.rs
```diff
@@ -81,51 +83,81 @@ fn refill<TScorer: Scorer, TScoreCombiner: ScoreCombiner>(
...
+                i += 1;
+            }
+        }
+        let first_score: Score = score_combiner_cloned.score();
+        BufferedUnionScorer {
+            scorers: non_empty_docsets,
             bitsets: Box::new([TinySet::empty(); HORIZON_NUM_TINYBITSETS]),
-            scores: Box::new([score_combiner_fn(); HORIZON as usize]),
+            scores: Box::new([score_combiner; HORIZON as usize]),
             bucket_idx: HORIZON_NUM_TINYBITSETS,
```
fulmicoton-dd, line 136: I think you borrowed that code from my PR. Just so this is in your radar, the scary part is that to avoid regression I made it so that the Scorer does not buffer docs on initialization.

(for instance on a TopK union, it would be a waste to load the first block )

This is probably the highest risk of regression, so if you can have an extra look at the logic it might be worth for you to do it.

fulmicoton (changes requested):
See comments.

src/index/inverted_index_reader.rs
PSeitz: I'm not sure what you are referring to? `TantivyInvertedIndexReader` ?

src/index/inverted_index_reader.rs
PSeitz: agree on both, removed `read_raw_postings_data` and added `InvertedIndexReader: DynInvertedIndexReader`

src/indexer/merger.rs
PSeitz: I didn't apply the downcast trick yet here, currently there's a regression.

src/postings/block_segment_postings.rs
PSeitz: left over from a refactor

PR #2878, opened 2026-04-02
fix: return None from SUM when no values were collected

mdashti:
## What
Return `None` from `IntermediateSum::finalize()` when `count == 0`.

## Why
`IntermediateSum::finalize()` returned `Some(0.0)` even when all documents had missing/NULL values, unlike MIN, MAX, and AVG which all return `None` for `count == 0`. The `0.0` came from `IntermediateStats`' default sum initialization. Consumers (like ParadeDB) that map `None` to SQL `NULL` were incorrectly getting `0` for SUM on all-NULL groups.

## How
Check `count == 0` before returning the sum value, returning `None` instead of `Some(0.0)`.

## Tests
- Verify SUM returns `None` when all values are missing/NULL
- Verify SUM still returns correct values for non-empty groups
- Verify MIN, MAX, AVG behavior unchanged

### DIFF
--- a/src/aggregation/intermediate_agg_result.rs
+++ b/src/aggregation/intermediate_agg_result.rs
@@ -377,7 +377,22 @@ impl IntermediateMetricResult {
                 MetricResult::ExtendedStats(intermediate_stats.finalize())
             }
             IntermediateMetricResult::Sum(intermediate_sum) => {
-                MetricResult::Sum(intermediate_sum.finalize().into())
+                // By default match Elasticsearch: empty / all-missing sum
+                // buckets serialize as `"value": 0`, not `"value": null`.
+                // The non-ES `none_if_no_match` flag on `SumAggregation`
+                // opts into SQL-style `null` for downstream consumers.
+                let none_if_no_match = req
+                    .agg
+                    .as_sum()
+                    .and_then(|sum| sum.none_if_no_match)
+                    .unwrap_or(false);
+                let value = intermediate_sum.finalize();
+                let coerced = if none_if_no_match {
+                    value
+                } else {
+                    Some(value.unwrap_or(0.0))
+                };
+                MetricResult::Sum(coerced.into())
             }
             IntermediateMetricResult::Percentiles(percentiles) => MetricResult::Percentiles(
                 percentiles
--- a/src/aggregation/intermediate_agg_result.rs
+++ b/src/aggregation/intermediate_agg_result.rs
@@ -387,12 +387,12 @@ impl IntermediateMetricResult {
                     .and_then(|sum| sum.none_if_no_match)
                     .unwrap_or(false);
                 let value = intermediate_sum.finalize();
-                let coerced = if none_if_no_match {
-                    value
+                if none_if_no_match {
+                    MetricResult::Sum(value.into())
                 } else {
-                    Some(value.unwrap_or(0.0))
-                };
-                MetricResult::Sum(coerced.into())
+                    let value = Some(value.unwrap_or(0.0));
+                    MetricResult::Sum(value.into())
+                }
             }
             IntermediateMetricResult::Percentiles(percentiles) => MetricResult::Percentiles(
                 percentiles
--- a/src/aggregation/metric/sum.rs
+++ b/src/aggregation/metric/sum.rs
@@ -59,8 +59,54 @@ impl IntermediateSum {
     pub fn merge_fruits(&mut self, other: IntermediateSum) {
         self.stats.merge_fruits(other.stats);
     }
-    /// Computes the final minimum value.
+    /// Computes the final sum value.
+    ///
+    /// Returns `None` when no values were collected (all documents had
+    /// missing/NULL values for the field), matching the behavior of
+    /// `IntermediateMin`, `IntermediateMax`, and `IntermediateAvg`.
     pub fn finalize(&self) -> Option<f64> {
-        Some(self.stats.finalize().sum)
+        let stats = self.stats.finalize();
+        if stats.count == 0 {
+            None
+        } else {
+            Some(stats.sum)
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_sum_finalize_returns_none_when_no_values() {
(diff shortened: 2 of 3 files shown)


src/aggregation/metric/sum.rs
```diff
@@ -59,8 +59,54 @@ impl IntermediateSum {
...
     }
-    /// Computes the final minimum value.
+    /// Computes the final sum value.
+    ///
+    /// Returns `None` when no values were collected (all documents had
+    /// missing/NULL values for the field), matching the behavior of
+    /// `IntermediateMin`, `IntermediateMax`, and `IntermediateAvg`.
     pub fn finalize(&self) -> Option<f64> {
-        Some(self.stats.finalize().sum)
+        let stats = self.stats.finalize();
```
PSeitz, line 68: Shouldn't we change `sum` on the `Stats` struct to Option instead?
Update: actually that would make de/serialization a bit weird with serde

fulmicoton: I don't understand the spec. Sum of 0 elements is defined. Why do we return an Option?
What does elastic return?

PSeitz: on the struct itself we have some freedom how to express null, only the serialization needs to follow elastic search spec

fulmicoton: I meant the sum of no number is defined. It is 0. Why are we having this discussion? Is this about reproducing a glitch in Elasticsearch?

PSeitz: sum of NULL is NULL in SQL

fulmicoton: Confirmation of the behavior
https://onecompiler.com/postgresql/44jr6em28

Aggregation are not meant to align themselves with SQL. We want elasticsearch behavior by default.

PSeitz: Yes, but we match elastic search behavior only on the serialized input and output, not on the intermediate formats. 

Currently the change is a breaking change, because of 
```rust
    /// Sum metric result.
    Sum(SingleMetricResult),

/// Single-metric aggregations use this common result structure.
///
/// Main reason to wrap it in value is to match elasticsearch output structure.
#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
pub struct SingleMetricResult {
    /// The value of the single value metric.
    pub value: Option<f64>,
}
```

Currently sum is weird (on main), SingleMetricResult is `Option`, but it's never `None`.
I think we should fix that one way or the other.

@mdashti Can you check how elasticsearch behaves?

mdashti: On flipping `Stats.sum` itself to `Option`, agree the serde dance isn't worth it. Kept `IntermediateStats` untouched and branched only inside `IntermediateSum::finalize` so the broader stats path stays as-is.

mdashti: Checked ES on this. Sum on an empty / all-missing bucket comes back as `"value": 0`, while min/max/avg return `"value": null`. So strict ES parity does mean `Some(0.0)` for sum, and you're right that this PR diverges on the wire.

The reason we leaned toward `None`: ParadeDB's SQL `SUM` over zero rows is `NULL`, and `Some(0.0)` is indistinguishable from a bucket that genuinely sums to zero, so the SQL boundary loses that signal. Hence the asymmetry feels worse than the ES divergence from where we sit.

Re the breaking-change point, fair. `SingleMetricResult.value` is already `Option<f64>` for sum but never `None` on main, so consumers that already match on `Option` won't blow up. Anything coercing `null → 0` to mirror the ES wire format will see a diff though.

Pulled in 44861669 a small doc note on `finalize` calling out the ES divergence and the SQL motivation, so future readers see the trade-off without rediscovering this thread.

Happy to take this whichever way you want, keep `None`, revert to `Some(0.0)` and just document the `Option`-but-never-`None` quirk, or gate the `None` behind something opt-in.

mdashti (commented):
@fulmicoton @PSeitz Thanks for the comments.

trinity-1686a:
we could add a `none_if_no_match: Option<bool>` to let users opt-in into a postgres-like behavior if they want. We don't want to differ from ES, but we could add our own extensions. wdyt?

src/aggregation/metric/sum.rs
mdashti: > we could add a none_if_no_match: Option<bool> to let users opt-in into a postgres-like behavior if they want. We don't want to differ from ES, but we could add our own extensions. wdyt?

@trinity-1686a Love the opt-in framing. Pushed 4daa35d4 in that shape.

- `SumAggregation` gets a new `none_if_no_match: Option<bool>` (skipped on serialize when unset, so JSON stays clean for the ES default path).
- `IntermediateSum::finalize` still returns `Option<f64>` internally so the Rust side stays parallel to min/max/avg.
- The ES-vs-SQL choice now lives at the boundary in `IntermediateMetricResult::into_final_metric_result`: `None` is coerced to `Some(0.0)` by default, and the flag opts out of that coercion. Added `AggregationVariants::as_sum()` to read it.
- Two end-to-end tests on an empty index, one for the ES default (`"value": 0`) and one for `none_if_no_match: true` (`"value": null`). Full `aggregation::` suite green locally.

So out of the box the wire format is back to matching ES, and pg-like behaviour just sets the flag on its side.

mdashti (commented):
@trinity-1686a @fulmicoton Please take another look.

src/aggregation/intermediate_agg_result.rs
```diff
@@ -377,7 +377,22 @@ impl IntermediateMetricResult {
...
+                    .as_sum()
+                    .and_then(|sum| sum.none_if_no_match)
+                    .unwrap_or(false);
+                let value = intermediate_sum.finalize();
+                let coerced = if none_if_no_match {
+                    value
+                } else {
+                    Some(value.unwrap_or(0.0))
+                };
+                MetricResult::Sum(coerced.into())
```
PSeitz-dd:
```suggestion
                if none_if_no_match {
                    MetricResult::Sum(value.into())
                } else {
                    let value = Some(value.unwrap_or(0.0));
                    MetricResult::Sum(value.into())
                }
```

PSeitz (approved):

PR #2889, opened 2026-04-07
Paul.masurel/2879 nested cardinality perf

fulmicoton:
This PR improves performance on one specific edge case.

When a string cardinality aggregation is nested it end up being applied to different buckets.
Dictionary encoding relies on a different dictionaries for each segment. 

As a result, during segment collection, we only collect term ordinals in a HashSet, and decode them in the
term dictionary at the end of collection.

Before this PR, this decoding phase was done once for each bucket, causing the same work to be done over and over. This PR introduce a coupon cache. The HLL sketch relies on a hash of the string values.

We populate the cache before bucket collection, and get our values from it.

This PR also rename "caching" "buffering" in aggregation (it was never caching), and does several cleanups.

```
full
cardinality_agg                           Memory: 4.7 MB (+33.23%)      Avg: 29.2733ms (-0.67%)     Median: 29.2294ms (-0.46%)     [28.7135ms .. 30.2274ms]
terms_status_with_cardinality_agg         Memory: 90.6 KB (-9.74%)      Avg: 6.3083ms (-1.76%)      Median: 6.2933ms (-1.18%)      [6.1373ms .. 6.5452ms]
terms_100_buckets_with_cardinality_agg    Memory: 19.9 MB (+103.92%)    Avg: 95.7822ms (-92.59%)    Median: 95.7456ms (-92.58%)    [93.7563ms .. 99.0150ms]
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ regex = { version = "1.5.5", default-features = false, features = [
     "std",
     "unicode",
 ] }
+murmurhash32 = "0.3"
 aho-corasick = "1.0"
 tantivy-fst = "0.5"
 memmap2 = { version = "0.9.0", optional = true }
@@ -65,7 +66,7 @@ tantivy-bitpacker = { version = "0.10", path = "./bitpacker" }
 common = { version = "0.11", path = "./common/", package = "tantivy-common" }
 tokenizer-api = { version = "0.7", path = "./tokenizer-api", package = "tantivy-tokenizer-api" }
 sketches-ddsketch = { version = "0.4", features = ["use_serde"] }
-datasketches = "0.2.0"
+datasketches = { git = "https://github.com/fulmicoton-dd/datasketches-rust", rev = "eb4ad64" }
 futures-util = { version = "0.3.28", optional = true }
 futures-channel = { version = "0.3.28", optional = true }
 fnv = "1.0.7"
--- a/src/aggregation/bucket/term_agg.rs
+++ b/src/aggregation/bucket/term_agg.rs
@@ -1022,19 +1022,20 @@ where
 
             let (term_ids, buckets): (Vec<u64>, Vec<Bucket>) = entries.into_iter().unzip();
 
-            let mut intermediate_entries: Vec<IntermediateTermBucketEntry> =
-                Vec::with_capacity(buckets.len());
-            for bucket in buckets {
-                intermediate_entries.push(into_intermediate_bucket_entry(
-                    bucket,
-                    reborrow_opt_collector(&mut sub_agg_collector),
-                    agg_data,
-                )?);
-            }
+            let intermediate_entries: Vec<IntermediateTermBucketEntry> = buckets
+                .into_iter()
+                .map(|bucket| {
+                    into_intermediate_bucket_entry(
+                        bucket,
+                        reborrow_opt_collector(&mut sub_agg_collector),
+                        agg_data,
+                    )
+                })
+                .collect::<crate::Result<_>>()?;
 
             let mut intermediate_entry_it = intermediate_entries.into_iter();
 
-            term_dict.sorted_ords_to_term_cb(term_ids.into_iter(), |term| {
+            term_dict.sorted_ords_to_term_cb(&term_ids[..], |term| {
                 let intermediate_entry = intermediate_entry_it.next().unwrap();
                 dict.insert(
                     IntermediateKey::Str(
--- a/src/aggregation/bucket/term_agg.rs
+++ b/src/aggregation/bucket/term_agg.rs
@@ -1,5 +1,4 @@
 use std::fmt::Debug;
-use std::io;
 use std::net::Ipv6Addr;
 
 use columnar::column_values::CompactSpaceU64Accessor;
@@ -18,7 +17,8 @@ use crate::aggregation::agg_data::{
 use crate::aggregation::agg_limits::MemoryConsumption;
 use crate::aggregation::agg_req::Aggregations;
 use crate::aggregation::buffered_sub_aggs::{
-    BufferedSubAggs, HighCardSubAggBuffer, LowCardBufferedSubAggs, LowCardSubAggBuffer, SubAggBuffer,
+    BufferedSubAggs, HighCardSubAggBuffer, LowCardBufferedSubAggs, LowCardSubAggBuffer,
+    SubAggBuffer,
 };
 use crate::aggregation::intermediate_agg_result::{
     IntermediateAggregationResult, IntermediateAggregationResults, IntermediateBucketResult,
@@ -790,8 +790,14 @@ impl<TermMap: TermAggregationMap, B: SubAggBuffer> SegmentAggregationCollector
         let term_req = &self.terms_req_data;
         let name = term_req.name.clone();
 
-        let bucket =
-            Self::into_intermediate_bucket_result(term_req, &mut self.sub_agg, bucket, agg_data)?;
+        let bucket = Self::into_intermediate_bucket_result(
+            term_req,
+            self.sub_agg
+                .as_mut()
+                .map(BufferedSubAggs::get_sub_agg_collector),
+            bucket,
+            agg_data,
+        )?;
         results.push(name, IntermediateAggregationResult::Bucket(bucket))?;
         Ok(())
     }
@@ -907,6 +913,34 @@ fn extract_missing_value<T>(
     Some((key, bucket))
 }
 
+fn reborrow_opt_collector<'a>(
+    opt: &'a mut Option<&mut dyn SegmentAggregationCollector>,
+) -> Option<&'a mut dyn SegmentAggregationCollector> {
+    match opt {
+        Some(inner) => Some(*inner),
+        None => None,
+    }
+}
+
+fn into_intermediate_bucket_entry(
+    bucket: Bucket,
+    sub_agg_collector: Option<&mut dyn SegmentAggregationCollector>,
+    agg_data: &AggregationsSegmentCtx,
+) -> crate::Result<IntermediateTermBucketEntry> {
+    let mut sub_aggregation_res = IntermediateAggregationResults::default();
+    if let Some(sub_agg_collector) = sub_agg_collector {
+        sub_agg_collector.add_intermediate_aggregation_result(
+            agg_data,
+            &mut sub_aggregation_res,
+            bucket.bucket_id,
+        )?;
+    }
+    Ok(IntermediateTermBucketEntry {
+        doc_count: bucket.count,
+        sub_aggregation: sub_aggregation_res,
+    })
+}
+
 impl<TermMap, B> SegmentTermCollector<TermMap, B>
 where
     TermMap: TermAggregationMap,
@@ -922,7 +956,7 @@ where
     #[inline]
     pub(crate) fn into_intermediate_bucket_result(
         term_req: &TermsAggReqData,
-        sub_agg: &mut Option<BufferedSubAggs<B>>,
+        mut sub_agg_collector: Option<&mut dyn SegmentAggregationCollector>,
         term_buckets: TermMap,
         agg_data: &AggregationsSegmentCtx,
     ) -> crate::Result<IntermediateBucketResult> {
@@ -965,31 +999,6 @@ where
         let mut dict: FxHashMap<IntermediateKey, IntermediateTermBucketEntry> = Default::default();
         dict.reserve(entries.len());
 
-        let into_intermediate_bucket_entry =
-            |bucket: Bucket,
-             sub_agg: &mut Option<BufferedSubAggs<B>>|
-             -> crate::Result<IntermediateTermBucketEntry> {
-                if let Some(sub_agg) = sub_agg {
-                    let mut sub_aggregation_res = IntermediateAggregationResults::default();
-                    sub_agg
-                        .get_sub_agg_collector()
-                        .add_intermediate_aggregation_result(
-                            agg_data,
-                            &mut sub_aggregation_res,
-                            bucket.bucket_id,
-                        )?;
-                    Ok(IntermediateTermBucketEntry {
-                        doc_count: bucket.count,
-                        sub_aggregation: sub_aggregation_res,
-                    })
-                } else {
-                    Ok(IntermediateTermBucketEntry {
-                        doc_count: bucket.count,
-                        sub_aggregation: Default::default(),
-                    })
-                }
-            };
-
         if term_req.column_type == ColumnType::Str {
             let fallback_dict = Dictionary::empty();
             let term_dict = term_req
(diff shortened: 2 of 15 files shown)


src/aggregation/metric/cardinality.rs
```diff
@@ -196,12 +246,24 @@ impl SegmentCardinalityCollectorBucket {
...
     }
 }
 
+fn append_to_sketch(
+    term_ords: &FxHashSet<u64>,
+    coupon_cache: &impl CouponCache,
+    sketch: &mut CardinalityCollector,
+) {
+    for &term_ord in term_ords {
+        let coupon: Coupon = coupon_cache.get_coupon(term_ord).unwrap_or(0);
```
hushy: **Note: All comments in that review were made doing a guided review with codex to get an understanding of the code base. Please correct my understanding if that is nonsense.** 
Potential Edge case : 
append_to_sketch currently inserts coupon_cache.get_coupon(term_ord).unwrap_or(0).
For the missing sentinel, the cache lookup fails (populate_coupon_cache removes missing) and we insert coupon 0, then later also insert the configured missing value:

```
if has_missing {
    let missing_key = req_data.req.missing.as_ref().expect(...);
    match missing_key {
        Key::Str(missing) => { self.cardinality.insert(missing.as_str()); }

```
This can add an unintended extra distinct element.

fulmicoton-dd: you are right!

Cargo.toml
```diff
@@ -65,7 +66,7 @@ tantivy-bitpacker = { version = "0.10", path = "./bitpacker" }
 common = { version = "0.11", path = "./common/", package = "tantivy-common" }
 tokenizer-api = { version = "0.7", path = "./tokenizer-api", package = "tantivy-tokenizer-api" }
 sketches-ddsketch = { version = "0.4", features = ["use_serde"] }
-datasketches = "0.2.0"
+datasketches = { git = "https://github.com/fulmicoton-dd/datasketches-rust", rev = "eb4ad64" }
```
Mallets, line 69: Nit: the `datasketches` dependency is pinned to a personal fork (`fulmicoton-dd/datasketches-rust` rev `eb4ad64`). This adds `update_with_coupon` and renames `get_result` → `to_sketch`. Is there a plan to upstream these changes before merging? A rev-pinned git dep on a personal fork will block `cargo publish` and is fragile if the fork is ever removed.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ regex = { version = "1.5.5", default-features = false, features = [
     "std",
     "unicode",
 ] }
+murmurhash32 = "0.3"
```
Mallets, line 27: Curiosity question: why `murmurhash32` over something like `ahash` (AES-NI accelerated)? Determinism across runs isn't needed here since the coupons are computed and consumed within a single segment collection pass (never serialized or stored). Is it because datasketches expects _murmurhash-formatted_ coupons in `update_with_coupon`, or just convention from the reference HLL implementations?

src/aggregation/metric/cardinality.rs
fulmicoton-dd: Actually this is more than one bug: We do not use the same hash for the missing key value if the key is missing or if the missing key value is present.

src/aggregation/bucket/term_agg.rs
```diff
@@ -907,10 +913,38 @@ fn extract_missing_value<T>(
     Some((key, bucket))
 }
 
-impl<TermMap, C> SegmentTermCollector<TermMap, C>
+fn reborrow_opt_collector<'a>(
```
Mallets, line 916: I love how the borrow checker is still not smart enough to detect this lifetime bound...

src/aggregation/metric/cardinality.rs
```diff
@@ -120,7 +124,54 @@ impl CardinalityAggregationReq {
...
+}
+
+impl CouponCacheEnum {
+    fn new(term_ords: Vec<u64>, coupons: Vec<Coupon>) -> CouponCacheEnum {
+        let num_terms = term_ords.len();
+        assert_eq!(num_terms, coupons.len());
+        if term_ords.is_empty() {
+            return CouponCacheEnum::Dense(Vec::new());
+        }
+        let highest_term_ord = term_ords.last().copied().unwrap_or(0u64);
```
Mallets, line 156: `.unwrap_or(0u64)` is unnecessary. The `if term_ords.is_empty() { ... }` block could be rewritten more clearly like:
```rust
let Some(highest_term_ord) = term_ords.last().copied() else {
    return CouponCacheEnum::Dense(Vec::new());
};
```
If last is `None` then _term_ords_ is empty.

src/aggregation/metric/cardinality.rs
```diff
@@ -120,7 +124,54 @@ impl CardinalityAggregationReq {
...
+        if highest_term_ord < 1_000_000u64 || highest_term_ord < num_terms as u64 * 10u64 {
```
Mallets: Just to be sure, this check is to use _Dense_ if either the max ordinal is under _1M_, or the ordinals are at least _10% dense_ (_max_ord < 10 * count_)? A comment would help here explaining the logic.

Moreover, the density heuristic uses `OR`:
```rust
if highest_term_ord < 1_000_000u64 || highest_term_ord < num_terms as u64 * 10u64 {
```
this means the _1M_ cap alone is sufficient to pick _Dense_, regardless of how sparse the data is. 
E.g., 2 terms with ordinals `[0, 999_999]` allocates a _4MB_ vec for 2 entries. Should this be `AND`?

src/aggregation/metric/cardinality.rs
```diff
@@ -120,7 +124,54 @@ impl CardinalityAggregationReq {
...
+                term_ords.into_iter().zip(coupons.into_iter()).collect();
+            CouponCacheEnum::Sparse(coupon_cache_sparse)
+        }
+    }
+}
+
+/// Coupon cache defines a generic implementation of the datastructure that
+/// could hold a term ordinal -> Coupon (a hash used in the cardinality sketch)
+/// mapping.
+trait CouponCache {
```
Mallets: Nit: `CouponCache` is only used in `append_to_sketch`. The trait, two impls on `Vec`/`FxHashMap`, and the manual enum unpacking at the call site could be simplified to a single method on `CouponCacheEnum:`

```rust
impl CouponCacheEnum {
   fn get_coupon(&self, term_ord: u64) -> Option<Coupon> {
        match self {
            Self::Dense(v) => v.get(term_ord as usize).copied(),
            Self::Sparse(m) => m.get(&term_ord).copied(),
        }
    }
}
```
Then `append_to_sketch` takes `&CouponCacheEnum` directly without needing to match the enum at the call site. The trait gives _monomorphization_ for the inner loop, but this isn't the hot path (the
 dictionary walk was, which is now cached), and branch prediction on the match should be perfect since the variant is constant across the loop.

src/aggregation/metric/cardinality.rs
```diff
@@ -196,12 +246,24 @@ impl SegmentCardinalityCollectorBucket {
                     }
                 }
             }
+        } else {
+            assert_ne!(req_data.column_type, ColumnType::Str);
```
Mallets: Is an `assert` the good choice here or it should return an `Error`? Or maybe a `debug_assert_ne`?

src/aggregation/metric/cardinality.rs
```diff
@@ -215,7 +277,43 @@ impl SegmentCardinalityCollector {
             accessor_idx,
             accessor,
             missing_value_for_accessor,
+            coupon_cache: None,
+        }
+    }
+
+    fn populate_coupon_cache(&mut self, dictionary: &Dictionary) -> io::Result<()> {
```
Mallets: The API signature accepts a `&Dictionary` parameter on every call, but the `is_some()` guard silently ignores it after the first invocation. These two things are in contrast: the signature suggests _populate with this dictionary_, but the behavior is _populate with the first dictionary, silently discard all subsequent ones_.

The current call chain guarantees the dictionary is always the same (same `AggregationsSegmentCtx` throughout segment finalization), so this is correct today. But if the call chain ever changes, e.g. the collector is reused across segments, or `add_intermediate_aggregation_result` is called with a different `agg_data`, the `Ok(())` return silently hides a stale cache. The caller has no way to know its dictionary was ignored.

src/aggregation/metric/cardinality.rs
```diff
@@ -215,7 +277,43 @@ impl SegmentCardinalityCollector {
...
+            || self.column_type != ColumnType::Str
+            || self.buckets.is_empty()
+        {
+            return Ok(());
         }
+        let mut term_ords: Vec<u64> = if self.buckets.len() == 1 {
+            self.buckets[0].entries.iter().copied().collect()
+        } else {
+            let term_ords_capacity: usize =
+                self.buckets.iter().map(|bucket| bucket.entries.len()).sum();
```
Mallets: Nit: capacity over-estimate: the sum of all bucket sizes is an upper bound. E.g., with _100_ buckets of _10k_ entries each and _80%_ overlap, we allocate for _1M_ entries but only use _~200k_. The HashSet will be _5x_ over-sized. Not a correctness issue, but for very high bucket counts this wastes memory. 

I'm not sure whether this is a real problem but I simply want to bring it up.

src/aggregation/metric/cardinality.rs
```diff
@@ -215,7 +277,43 @@ impl SegmentCardinalityCollector {
...
+                    term_ords_union.insert(term_ord);
+                }
+            }
+            term_ords_union.into_iter().collect()
+        };
+        term_ords.sort_unstable();
+
+        term_ords.pop_if(|highest_term_ord| *highest_term_ord == MISSING_TERM_SENTINEL);
+
+        let mut coupons: Vec<Coupon> = Vec::with_capacity(term_ords.capacity());
```
Mallets:
```suggestion
        let mut coupons: Vec<Coupon> = Vec::with_capacity(term_ords.len());
```

src/aggregation/metric/cardinality.rs
```diff
@@ -215,7 +277,43 @@ impl SegmentCardinalityCollector {
...
+        dictionary.sorted_ords_to_term_cb(&term_ords, |term_bytes| {
+            let coupon: Coupon = murmurhash32::murmurhash2(term_bytes);
+            coupons.push(coupon);
+        })?;
```
Mallets: `sorted_ords_to_term_cb` returns `io::Result<bool>` where the bool indicates whether all ordinals were found in the dictionary. Here the `?` propagates the `io::Result` but the bool is silently dropped. If an ordinal is missing from the dictionary, the callback fires fewer times than `term_ords.len()`, so `coupons.len() < term_ords.len()`, and `CouponCacheEnum::new` panics on `assert_eq!(num_terms,
  coupons.len())`.

This should hold in practice (ordinals come from the same segment's column), but if it ever doesn't, the panic gives no context. I think it's worth checking the bool and return an error instead of panicking.

sstable/src/dictionary.rs
```diff
@@ -538,33 +540,36 @@ impl<TSSTable: SSTable> Dictionary<TSSTable> {
...
+            let next_ord = loop {
+                let Some(next_ord) = ords.next() else {
+                    return Ok(true);
+                };
+                if next_ord == ord {
+                    // This is the same ordinal, let's just call the callback directly.
+                    cb(&bytes);
+                } else {
+                    // we checked it was sorted beforehands
+                    assert!(next_ord > ord);
```
Mallets: Nit: `assert!(ords.is_sorted());` already verifies this. If you want to keep it, it could be downgraded to `debug_assert`.

Mallets (approved):
PR looks good overall. The coupon cache optimization seems solid (besides the already [identified bug](https://github.com/quickwit-oss/tantivy/pull/2889#discussion_r3056012516)) and the _92%_ latency win on nested cardinality is very encouranging.

One suggestion: `sorted_ords_to_term_cb` changed how duplicate ordinals are handled (the new inner loop calls the callback directly instead of falling through to the outer loop). The behavior is  equivalent, but there's no specific test covering duplicate ordinals. A small test case like `ords = [100_000, 100_000, 100_001]` asserting it would nail this down.

src/aggregation/metric/cardinality.rs
fulmicoton: Marking as resolved as this part of the code does not exist anymore.

src/aggregation/metric/cardinality.rs
fulmicoton-dd: no this was meant to be an OR. I will add a comment.

Cargo.toml
fulmicoton-dd: my XP is that murmurhash32 is actually faster on small strings.

src/aggregation/bucket/term_agg.rs
fulmicoton-dd: yeah, really annoying.

src/aggregation/metric/cardinality.rs
```diff
@@ -120,85 +121,194 @@ impl CardinalityAggregationReq {
...
-                .as_ref()
-                .map(|el| el.dictionary())
-                .unwrap_or_else(|| &fallback_dict);
-            let mut has_missing = false;
-
-            // TODO: replace FxHashSet with something that allows iterating in order
-            // (e.g. sparse bitvec)
-            let mut term_ids = Vec::new();
-            for term_ord in self.entries.into_iter() {
-                if term_ord == u64::MAX {
```
fulmicoton-dd, line 164: (this was incorrect before, the sentinel value is not u64::MAX)

src/aggregation/metric/cardinality.rs
fulmicoton-dd: I think an assert is correct here, but let me check!

fulmicoton: you are right, the contract justifying it is really remote. I removed the assert entirely.

src/aggregation/metric/cardinality.rs
fulmicoton: You are right adn this is a big problem. I meant max, thanks

Cargo.toml
fulmicoton: > s there a plan to upstream these changes before merging

yes. I would also like to upstream u64 coupon. u32 is inappropriate for large cardinality.

src/aggregation/metric/cardinality.rs
fulmicoton: this was removed.

That kind of code is not  about branch prediction. 
Removing the condition from the loop allow for vectorization.

src/aggregation/metric/cardinality.rs
fulmicoton: I moved the "is_none()" logic to the caller site.

src/aggregation/metric/cardinality.rs
fulmicoton: agreed. I added an assert here.

src/aggregation/metric/cardinality.rs
```diff
@@ -120,85 +121,203 @@ impl CardinalityAggregationReq {
...
+
+    term_ords.pop_if(|highest_term_ord| *highest_term_ord >= dictionary.num_terms() as u64);
+
+    let mut coupons: Vec<Coupon> = Vec::with_capacity(term_ords.len());
+    let all_term_ords_found: bool =
+        dictionary.sorted_ords_to_term_cb(&term_ords, |term_bytes| {
+            let coupon: Coupon = murmurhash32::murmurhash2(term_bytes);
+            coupons.push(coupon);
+        })?;
+    assert!(all_term_ords_found);
```
PSeitz, line 272: do we really want to panic here instead of returning an error?

PR #2898, opened 2026-04-20
Optimize single term BooleanQuery to use BMW

jamessewell:
When there is a single term, delegate directly to the child weight's `for_each_pruning`. This ensures that a `BooleanQuery` wrapping a single `TermQuery` still benefits from block-max WAND pruning (via `TermWeight::for_each_pruning`), rather than falling through to the generic f`or_each_pruning_scorer` loop.

This isn't hit in Tantivy proper (single term queries are always collapsed to `TermQuery`, but is in pg_search using the `|||` operator, and could be by other future implementations.

If this isn't desired in Tantivy we are happy to just fix in `pg_search` by collapsing in the same way, but it feels like a bit of a footgun atm.

### DIFF
--- a/src/query/boolean_query/boolean_weight.rs
+++ b/src/query/boolean_query/boolean_weight.rs
@@ -519,6 +519,17 @@ impl<TScoreCombiner: ScoreCombiner + Sync> Weight for BooleanWeight<TScoreCombin
         reader: &SegmentReader,
         callback: &mut dyn FnMut(DocId, Score) -> Score,
     ) -> crate::Result<()> {
+        // When there is a single include clause, delegate directly to the child
+        // weight's for_each_pruning. This ensures that e.g. a BooleanQuery
+        // wrapping a single TermQuery still benefits from block-max WAND
+        // pruning (via TermWeight::for_each_pruning), rather than falling
+        // through to the generic for_each_pruning_scorer loop.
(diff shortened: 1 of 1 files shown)


fulmicoton:
We have different place to plug this in, this is not the right one.

Summing up the situation. We have a block max wand impl for bool union, and for single terms.

Currently the single term impl only kicks in if for a term query.
We do not realize that a booleanquery sometimes has a single term.
(A) This PR spots that and calls the implementation. 

One level below we convert weight into scorers. In presence of EmptyScorer, it is possible to have more cases where the number of scorer is now one and the optimization does not kick in anymore. We could spot this at that level (B)

Finally, as we consume scorers, the number of active scorers decreases to 1.
We could also optimize the wand algorithm to detect that and be as fast as that. 
We could spot it at that level (C).

Implementing things at the level (B) seems easy and covers everything in (A).

Implementing things at level (C) requires a little bit more care, but would be the best solution.

fulmicoton:
(C) itself has different variant. We can either exit the loop when a single scorer is remaining, or attempt to improve find_pivot so that it works off the skip list first.

fulmicoton:
See https://github.com/quickwit-oss/tantivy/pull/2912

fulmicoton:
Solved as B in https://github.com/quickwit-oss/tantivy/pull/2912. Attempted to do it as C but it resulted in a small regression.

PR #2900, opened 2026-04-20
Optimize nested term aggregations with a shared term ordinal cache

### LINKED ISSUE #2892: Apply optimization introduced in #2879 for other aggregations
In #2879, we introduced a performance improvement for cardinality over string in a nested aggregations.

In reality, this optimization is relevant for ANY nested aggregations that require dictionary encoding.
If we have time, this coudl also prompt for a refactoring of aggregations.

Anything calling sorted_ords_to_term_cb is a probably a good candidate.

fulmicoton:
When a string term aggregation is nested inside another bucket
aggregation, each parent bucket previously resolved term ordinals
to strings independently via `sorted_ords_to_term_cb`. This was
redundant since sibling buckets share the same term dictionary.

Introduce `TermOrdToStrCache` which resolves all observed term
ordinals once across all parent buckets, then stores the results
in a `StringArena` (a single contiguous String buffer) with
compact `StringRef` handles. The cache uses either a dense
`Vec<Option<StringRef>>` or a sparse `FxHashMap<u64, StringRef>`
depending on the ordinal range.

Also add `collect_term_ords` to the `TermAggregationMap` trait so
each map variant can export its live ordinals into a shared set.

Closes #2892

### DIFF
--- a/benches/agg_bench.rs
+++ b/benches/agg_bench.rs
@@ -206,6 +206,7 @@ fn terms_7(index: &Index) {
     });
     execute_agg(index, agg_req);
 }
+
 fn terms_all_unique(index: &Index) {
     let agg_req = json!({
         "my_texts": { "terms": { "field": "text_all_unique_terms" } },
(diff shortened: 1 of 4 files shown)


fulmicoton:
This actually causes a regression, probably because the term ords do not get filtered out via `cut_off_buckets` anymore, including for the first level of aggregation.

PR #2901, opened 2026-04-20
Test demonstrating that nested aggregation are insanely slow

fulmicoton:
Test highlighting our performance problem on nested aggregation


trinity-1686a:
fixed by https://github.com/quickwit-oss/tantivy/pull/2902

PR #2902, opened 2026-04-20
perf fix: compute memory consumption only for current bucket

PSeitz:
`SegmentTermCollector::collect` and `SegmentCompositeCollector::collect` called `get_memory_consumption` for all parent_buckets. Only the current parent_bucket_id is mutated, so the it should compute only the memory change for the current bucket. 

```rust
full
terms_zipf_1000_with_terms_status_sub_agg    Avg: 13.6910ms (-94.17%)    Median: 13.7017ms (-94.14%)    [13.3138ms .. 14.2339ms]    
dense
terms_zipf_1000_with_terms_status_sub_agg    Avg: 18.2570ms (-92.40%)    Median: 18.2083ms (-92.39%)    [18.0665ms .. 18.6569ms]    
sparse
terms_zipf_1000_with_terms_status_sub_agg    Avg: 16.4531ms (-37.61%)    Median: 16.3728ms (-37.29%)    [16.0743ms .. 17.1176ms]    
multivalue
terms_zipf_1000_with_terms_status_sub_agg    Avg: 22.1713ms (-91.15%)    Median: 22.1696ms (-91.13%)    [21.6588ms .. 22.8247ms]    
```

### DIFF
--- a/src/aggregation/bucket/term_agg.rs
+++ b/src/aggregation/bucket/term_agg.rs
@@ -809,7 +809,7 @@ impl<TermMap: TermAggregationMap, B: SubAggBuffer> SegmentAggregationCollector
         docs: &[crate::DocId],
         agg_data: &mut AggregationsSegmentCtx,
     ) -> crate::Result<()> {
-        let mem_pre = self.get_memory_consumption();
+        let mem_pre = self.get_memory_consumption(parent_bucket_id);
 
         let req_data = &mut self.terms_req_data;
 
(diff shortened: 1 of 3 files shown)


fulmicoton-dd (approved):

src/aggregation/bucket/term_agg.rs
```diff
@@ -946,11 +946,8 @@ where
     TermMap: TermAggregationMap,
     B: SubAggBuffer,
 {
-    fn get_memory_consumption(&self) -> usize {
-        self.parent_buckets
-            .iter()
-            .map(|b| b.get_memory_consumption())
-            .sum()
+    fn get_memory_consumption(&self, parent_bucket_id: BucketId) -> usize {
```
fulmicoton-dd, line 950: I think you could remove the function and inline its impl maybe.

fulmicoton-dd:
Can you fix Composite too? It has the same problem

src/aggregation/bucket/term_agg.rs
PSeitz-dd: tested, but I think it's slightly more readable like this

PR #2905, opened 2026-04-21
Fix O(2^n) query parser regression for deeply-nested queries

### LINKED ISSUE #2498: Possible perf regression in Query Parser with deeply nested queries
# Overview

When upgrading tantivy from 0.21 to 0.22 (used within tantivy-py), we observed a performance regression in **query parsing**.  Rolling back resolved the issue.  The issue occurred with queries that were deeply nested.

# Reproducer

It's been difficult to get an exact reproducer, but I think I have a small example that demonstrates the issue:

```
Tag 0.22.0

 depth  leading+      parse time query                         
    20     false         0.64593 ((((((((((((((((((((title:test))))))))))))))))))))
    21     false         1.29425 (((((((((((((((((((((title:test)))))))))))))))))))))
    22     false         2.64207 ((((((((((((((((((((((title:test))))))))))))))))))))))
    23     false         5.31082 (((((((((((((((((((((((title:test)))))))))))))))))))))))
    24     false        10.62795 ((((((((((((((((((((((((title:test))))))))))))))))))))))))
    25     false        21.20079 (((((((((((((((((((((((((title:test)))))))))))))))))))))))))
 depth  leading+      parse time query                         
    20      true         0.68443 +((((((((((((((((((((title:test))))))))))))))))))))
    21      true         1.35804 +(((((((((((((((((((((title:test)))))))))))))))))))))
    22      true         2.70420 +((((((((((((((((((((((title:test))))))))))))))))))))))
    23      true         5.37000 +(((((((((((((((((((((((title:test)))))))))))))))))))))))
    24      true        10.82749 +((((((((((((((((((((((((title:test))))))))))))))))))))))))
    25      true        21.23502 +(((((((((((((((((((((((((title:test)))))))))))))))))))))))))

Tag 0.21.1

 depth  leading+       parse (s) query                         
    20     false         0.59977 ((((((((((((((((((((title:test))))))))))))))))))))
    21     false         1.20016 (((((((((((((((((((((title:test)))))))))))))))))))))
    22     false         2.39481 ((((((((((((((((((((((title:test))))))))))))))))))))))
    23     false         4.79820 (((((((((((((((((((((((title:test)))))))))))))))))))))))
    24     false         9.53619 ((((((((((((((((((((((((title:test))))))))))))))))))))))))
    25     false        19.15066 (((((((((((((((((((((((((title:test)))))))))))))))))))))))))
 depth  leading+       parse (s) query                         
    20      true         0.30958 +((((((((((((((((((((title:test))))))))))))))))))))
    21      true         0.62179 +(((((((((((((((((((((title:test)))))))))))))))))))))
    22      true         1.24262 +((((((((((((((((((((((title:test))))))))))))))))))))))
    23      true         2.49942 +(((((((((((((((((((((((title:test)))))))))))))))))))))))
    24      true         4.94994 +((((((((((((((((((((((((title:test))))))))))))))))))))))))
    25      true         9.89061 +(((((((((((((((((((((((((title:test)))))))))))))))))))))))))

main branch (55b0b52457c87df45250243c3310b4ffad85b088)

 depth  leading+       parse (s) query                         
    20     false         0.52478 ((((((((((((((((((((title:test))))))))))))))))))))
    21     false         1.03610 (((((((((((((((((((((title:test)))))))))))))))))))))
    22     false         2.09806 ((((((((((((((((((((((title:test))))))))))))))))))))))
    23     false         4.16839 (((((((((((((((((((((((title:test)))))))))))))))))))))))
    24     false         8.33352 ((((((((((((((((((((((((title:test))))))))))))))))))))))))
    25     false        16.73920 (((((((((((((((((((((((((title:test)))))))))))))))))))))))))
 depth  leading+       parse (s) query                         
    20      true         0.52754 +((((((((((((((((((((title:test))))))))))))))))))))
    21      true         1.05164 +(((((((((((((((((((((title:test)))))))))))))))))))))
    22      true         2.10110 +((((((((((((((((((((((title:test))))))))))))))))))))))
    23      true         4.19262 +(((((((((((((((((((((((title:test)))))))))))))))))))))))
    24      true         8.38012 +((((((((((((((((((((((((title:test))))))))))))))))))))))))
    25      true        16.75604 +(((((((((((((((((((((((((title:test)))))))))))))))))))))))))
```

It turns out, on 0.21, we had been benefitting from the faster path that happens when a leading `+` is in front of the query text.  In 0.22, and in main branch, that benefit is gone.  With 0.22, we observed some queries that took several minutes to parse. It is somewhat difficult to get those exact queries, but if necessary I can try.

It is also clear from this simple benchmark that the query parsing time is exponential in the depth of nesting, even for the very simple query above. That is its own issue, but my main question is whether the performance profile of 0.21 can be restored in 0.22.x, or main branch. As is, with 0.22 we will be unable to upgrade without additional work on our query generation code.

Here are the files to reproduce the project:

```toml
# Cargo.toml
[package]
name = "deepquery"
version = "0.1.0"
edition = "2021"

[dependencies]
zstd-sys = "=2.0.9"
# tantivy = { git = "https://github.com/quickwit-oss/tantivy.git", tag = "0.22.0" }
tantivy = { git = "https://github.com/quickwit-oss/tantivy.git", tag = "0.21.1" }
# tantivy = { git = "https://github.com/quickwit-oss/tantivy.git", tag = "0.21" }
# tantivy = { git = "https://github.com/quickwit-oss/tantivy.git", branch = "main" }
```

```rust
// lib.rs
#[cfg(test)]
mod tests {
    use std::time::Instant;
    use tantivy::schema::*;

    #[test]
    fn test_parse() -> Result<(), Box<dyn std::error::Error>> {
        let mut schema_builder = SchemaBuilder::new();
        schema_builder.add_u64_field("order", STORED | FAST);
        schema_builder.add_text_field("title", TEXT | STORED);
        let title = schema.get_field("title").unwrap();

        let mut index = tantivy::Index::create_in_ram(schema.clone());
        let query_parser = QueryParser::for_index(&index, vec![title]);

        for leading_plus in [false, true].into_iter() {
            println!(
                "{:>6}{:>10}{:>16} {:<30}",
                "depth", "leading+", "parse (s)", "query"
            );
            for depth in 20..=25 {
                let query_text = make_query_text(depth, leading_plus);
                let t0 = Instant::now();
                {
                    let query = query_parser.parse_query(&query_text)?;
                }
                let t1 = Instant::now();
                println!(
                    "{:>6?}{:>10?}{:>16.5} {:<30}",
                    depth,
                    leading_plus,
                    (t1 - t0).as_secs_f64(),
                    query_text
                );
            }
        }

        Ok(())
    }

    fn make_query_text(count: usize, leading_plus: bool) -> String {
        let leading = vec!['('; count].iter().collect::<String>();
        let trailing = vec![')'; count].iter().collect::<String>();
        let front = if leading_plus { "+" } else { "" };
        let q_text = vec![front, &leading, "title:test", &trailing].join("");
        q_text
    }
}
```

On rust 1.81, linux x86_64, I run the test case like this:

```
$ cargo test --release -- --nocapture
```

wallies:
## Summary

Fixes #2498.

The top-level `ast()` parser used `alt((boolean_expr, single_leaf))` at every group level. When the group contained a single leaf with no trailing operand, `boolean_expr` would parse `occur_leaf` (recursing into the inner group), fail at `multispace1`, backtrack, and then `single_leaf` would re-parse `occur_leaf` from scratch. Every nesting level doubled the work, giving O(2^n) time for queries like `(((((title:test)))))`.

Parse `occur_leaf` once and peek ahead for a trailing operand instead of backtracking. This keeps parsing O(n) and also avoids the duplicate parse for simple single-leaf queries. `Err::Error` from the lookahead falls back to a single leaf (matching the old `alt` semantics); `Err::Failure` and `Err::Incomplete` propagate.

## Benchmark

A new `benches/query_parser_nested.rs` (binggan) exercises the depths from the issue. Run with:

```
cargo bench --bench query_parser_nested
```

Results on my machine with the fix:

```
parse_nested_depth_20_plain   Avg: 1715ns   Median: 1707ns
parse_nested_depth_20_plus    Avg: 1700ns   Median: 1688ns
parse_nested_depth_21_plain   Avg: 1757ns   Median: 1756ns
parse_nested_depth_21_plus    Avg: 1818ns   Median: 1757ns
```

For reference, the issue reporter measured **0.87 s at depth 20 and 1.72 s at depth 21** under the regression - roughly a 500,000x improvement at depth 20 and 1,000,000x at depth 21. Scaling with the fix is linear (depth 2000 parses in ~200 µs).

Non-pathological queries are unaffected or slightly faster (single-leaf cases benefit from not being double-parsed):

| query | before | after |
|-------|-------:|------:|
| `hello` | 650 ns | 308 ns |
| `title:rust` | 667 ns | 325 ns |
| `a AND b AND c` | 1380 ns | 1364 ns |
| `a OR b OR c` | 1424 ns | 1374 ns |
| `+a -b c` | 1397 ns | 1403 ns |
| `title:rust AND (body:web OR body:api) -tag:draft` | 3426 ns | 3460 ns |

## Test Plan

- [x] All 53 existing `tantivy-query-grammar` tests pass.
- [x] All 56 `tantivy` `query_parser` tests pass.
- [x] `cargo clippy -p tantivy-query-grammar --all-targets -- -D warnings` clean.
- [x] `cargo clippy -p tantivy --bench query_parser_nested -- -D warnings` clean.
- [x] New regression test `test_parse_deeply_nested_query` at depth 60 - this would not complete under the old parser (~10^18 operations) and passes instantly with the fix.
- [x] New `query_parser_nested` binggan benchmark covering depths 20/21 from the issue, with and without the leading `+`.

### DIFF
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -1045,18 +1045,37 @@ fn operand_leaf(inp: &str) -> IResult<&str, (Option<BinaryOperand>, Option<Occur
 }
 
 fn ast(inp: &str) -> IResult<&str, UserInputAst> {
-    let boolean_expr = map_res(
-        separated_pair(occur_leaf, multispace1, many1(operand_leaf)),
-        |(left, right)| aggregate_binary_expressions(left, right),
-    );
-    let single_leaf = map(occur_leaf, |(occur, ast)| {
-        if occur == Some(Occur::MustNot) {
-            ast.unary(Occur::MustNot)
-        } else {
-            ast
-        }
-    });
-    delimited(multispace0, alt((boolean_expr, single_leaf)), multispace0)(inp)
+    // Parse `occur_leaf` once, then conditionally extend into a boolean
+    // expression. The previous implementation used `alt((boolean_expr,
+    // single_leaf))` which, when the input was a single leaf with no
+    // following operand, would parse `occur_leaf` once for `boolean_expr`,
+    // fail at `multispace1`, backtrack, then re-parse `occur_leaf` for
+    // `single_leaf`. With recursively-nested groups like `(+(+(+a)))`, that
+    // doubling at every level produced O(2^n) parse time. Parsing once and
+    // peeking ahead for the operand keeps it O(n).
+    delimited(
+        multispace0,
+        |inp| {
+            let (rest, first) = occur_leaf(inp)?;
+            match preceded(multispace1, many1(operand_leaf))(rest) {
+                Ok((rest, more)) => {
+                    let combined = aggregate_binary_expressions(first, more)
+                        .map_err(|_| nom::Err::Error(Error::new(inp, ErrorKind::MapRes)))?;
+                    Ok((rest, combined))
+                }
+                Err(_) => {
+                    let (occur, ast) = first;
+                    let single = if occur == Some(Occur::MustNot) {
+                        ast.unary(Occur::MustNot)
+                    } else {
+                        ast
+                    };
+                    Ok((rest, single))
+                }
+            }
+        },
+        multispace0,
+    )(inp)
 }
 
 fn ast_infallible(inp: &str) -> JResult<&str, UserInputAst> {
@@ -1890,5 +1909,24 @@ mod test {
             "field : 'happy tax payer' AND other_field  : 1",
             r#"(+"field":'happy tax payer' +"other_field":1)"#,
         );
+    }
+
+    // Regression test for https://github.com/quickwit-oss/tantivy/issues/2498:
+    // deeply nested parenthesized queries used to take O(2^n) time because the
+    // top-level `ast()` parser tried `boolean_expr` first and re-parsed the
+    // inner `occur_leaf` when it backtracked to `single_leaf`. Depth 60 would
+    // take ~10^18 operations under the regression; with the fix it parses
+    // instantly. We use `test_parse_query_to_ast_helper` so this test would
+    // never finish if the regression returned.
+    #[test]
+    fn test_parse_deeply_nested_query() {
+        let depth = 60;
+        let leading: String = "(".repeat(depth);
+        let trailing: String = ")".repeat(depth);
+        let query = format!("{leading}title:test{trailing}");
+        test_parse_query_to_ast_helper(&query, r#""title":test"#);
+
+        let query_with_plus = format!("+{leading}title:test{trailing}");
+        test_parse_query_to_ast_helper(&query_with_plus, r#""title":test"#);
     }
 }
--- a/query-grammar/src/query_grammar.rs
+++ b/query-grammar/src/query_grammar.rs
@@ -1929,4 +1929,35 @@ mod test {
         let query_with_plus = format!("+{leading}title:test{trailing}");
         test_parse_query_to_ast_helper(&query_with_plus, r#""title":test"#);
     }
+
+    // Manual benchmark matching the depths from issue #2498 (originally
+    // 0.87 s at depth 20 and 1.72 s at depth 21 under the regression).
+    // Run with:
+    //   cargo test -p tantivy-query-grammar --release bench_deeply_nested \
+    //       -- --ignored --nocapture
+    #[test]
+    #[ignore]
+    fn bench_deeply_nested_query() {
+        use std::time::Instant;
+        for depth in [20, 21] {
+            let leading: String = "(".repeat(depth);
+            let trailing: String = ")".repeat(depth);
+            for (label, query) in [
+                ("plain", format!("{leading}title:test{trailing}")),
+                ("plus ", format!("+{leading}title:test{trailing}")),
+            ] {
+                // Warm up.
+                for _ in 0..100 {
+                    parse_to_ast(&query).unwrap();
+                }
+                let iters = 10_000;
+                let t0 = Instant::now();
+                for _ in 0..iters {
+                    parse_to_ast(&query).unwrap();
+                }
+                let ns_per_parse = t0.elapsed().as_nanos() as f64 / iters as f64;
+                println!("depth {depth:>2} {label} {ns_per_parse:>10.0} ns/parse");
+            }
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


PSeitz:
Can you add a benchmark for deeply nested queries? Maybe for 20 and 21 depth

query-grammar/src/query_grammar.rs
```diff
@@ -1045,18 +1045,37 @@ fn operand_leaf(inp: &str) -> IResult<&str, (Option<BinaryOperand>, Option<Occur
...
+        multispace0,
+        |inp| {
+            let (rest, first) = occur_leaf(inp)?;
+            match preceded(multispace1, many1(operand_leaf))(rest) {
+                Ok((rest, more)) => {
+                    let combined = aggregate_binary_expressions(first, more)
+                        .map_err(|_| nom::Err::Error(Error::new(inp, ErrorKind::MapRes)))?;
+                    Ok((rest, combined))
+                }
+                Err(_) => {
```
trinity-1686a: nom has 3 kinds or Err: Error/Failure/Incomplete
alt only capture Error to try other arms, and propagate Failure/Incomplete. I don't know if this changes anything in our case (we don't emit the later, but do we emit the former?). Can you match on `Err(nom::Err::Error(_))` just in case?

query-grammar/src/query_grammar.rs
```diff
@@ -1891,4 +1910,54 @@ mod test {
...
+    }
+
+    // Manual benchmark matching the depths from issue #2498 (originally
+    // 0.87 s at depth 20 and 1.72 s at depth 21 under the regression).
+    // Run with:
+    //   cargo test -p tantivy-query-grammar --release bench_deeply_nested \
+    //       -- --ignored --nocapture
+    #[test]
+    #[ignore]
+    fn bench_deeply_nested_query() {
```
PSeitz: That's not a benchmark. can you add a binggan benchmark in the benches folder?
e.g. like `benches/fill_bitset.rs`

wallies: @PSeitz added, thanks for pointing me in the right direction

wallies:
@PSeitz is there anything else you think this needs? We have it in our fork and it works

trinity-1686a (approved):

PR #2910, opened 2026-04-24
Composite aggregations: send after key on last page

Abdul-Andha:
### Motivation
- For composite aggregations, tantivy does not send an `after_key` if there is no more data on the next page. So its possible to have buckets in the response and no after key. This was causing an error when querying with trino.
- I ran an elasticsearch container locally ([image](https://www.docker.elastic.co/r/elasticsearch/elasticsearch:9.3.3)) and tested composite aggregations with pages. When there is no more data on the next page, the response still gives an after key. Using that after key yields a response with no buckets and no after key. Seeing no buckets indicates to trino (and possibly other clients) that pagination is over

### Changes
- This PR changes tantivy behavior to match elasticsearch (send after key on last page with data)

### Tests
- Updated tests accordingly
- Tested locally

### DIFF
--- a/src/aggregation/bucket/composite/mod.rs
+++ b/src/aggregation/bucket/composite/mod.rs
@@ -722,6 +722,7 @@ mod tests {
         });
         let agg_req: Aggregations = serde_json::from_value(agg_req_json).unwrap();
         let res = exec_request(agg_req.clone(), &index).unwrap();
+        assert!(res["my_composite"].get("after_key").is_none());
         assert_eq!(
             res["my_composite"]["buckets"],
             json!([]),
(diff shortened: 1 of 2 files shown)


src/aggregation/bucket/composite/mod.rs
```diff
@@ -711,8 +707,27 @@ mod tests {
                 {"key": {"myterm": "terme"}, "doc_count": 1}
             ])
         );
-        assert!(res["my_composite"].get("after_key").is_none());
```
trinity-1686a, line 714: can you keep that assert (below the empty page)

trinity-1686a (approved):

congx4 (approved):

congx4:
Sweet!

PR #2913, opened 2026-04-27
fix: Missing sentinel is now considered in term agg requests with exclude + missing

barbarj:
The missing sentinel value is computed as (this name is descriptive, not a reference ->) max_term_ordinal + 1. When there are zero unique terms, this causes the sentinel value to be 1 (because max_term_ordinal is 0 by default), which means the allowed terms bitset never contains it (because it will only have capacity 1 in this case).

This fix adds logic to handle that by also passing the number of non-nil values in the column, and in the case that is 0, setting the sentinel value as 0 directly. 

(Ideally we'd have distinct representations for the max term ordinal of zero unique terms and one unique term, but that would be much broader refactor)

### DIFF
--- a/src/aggregation/accessor_helpers.rs
+++ b/src/aggregation/accessor_helpers.rs
@@ -7,6 +7,18 @@ use columnar::{Column, ColumnType};
 use crate::aggregation::{f64_to_fastfield_u64, Key};
 use crate::index::SegmentReader;
 
+/// For string columns, the missing value is always represented as column_max_value + 1, but in
+/// the case where we have no values for this column, that causes the missing value to be
+/// erroneously set to 1 (which would imply we had one additional value represented as 0).
+/// This checks for that case that case and returns 0 instead.
+fn missing_val_for_str_cols(column_max_value: u64, column_num_values: u32) -> u64 {
+    if column_num_values == 0 {
+        0
+    } else {
+        column_max_value + 1
+    }
+}
+
 /// Get the missing value as internal u64 representation
 ///
 /// For terms we use u64::MAX as sentinel value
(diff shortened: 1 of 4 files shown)


src/aggregation/accessor_helpers.rs
```diff
@@ -16,16 +16,16 @@ use crate::index::SegmentReader;
...
     let missing_val = match missing {
-        Key::Str(_) if column_type == ColumnType::Str => Some(column_max_value + 1),
-        // Allow fallback to number on text fields
-        Key::F64(_) if column_type == ColumnType::Str => Some(column_max_value + 1),
-        Key::U64(_) if column_type == ColumnType::Str => Some(column_max_value + 1),
-        Key::I64(_) if column_type == ColumnType::Str => Some(column_max_value + 1),
+        Key::Str(_) | Key::F64(_) | Key::U64(_) | Key::I64(_) if column_type == ColumnType::Str => {
+            // For strings, we use the max term ordinal + 1, which happens to always be the same as
+            // the number of values in the column
+            Some(column_val_count as u64)
```
fulmicoton: !? no? num_vals is the number of non nil values in the column isn't it?

barbarj: Right. And the max term ordinal for string columns is always 1 less than the num_vals (since term ordinals are zero-indexed), meaning num_vals always equals the desired ordinal for the nil sentinel. 

Or have I missed something entirely about how this works for strings that the unit tests also didn't catch?

src/aggregation/agg_data.rs
```diff
@@ -967,7 +967,7 @@ fn build_terms_or_cardinality_nodes(
         let missing_value_for_accessor = if use_special_missing_agg {
             None
         } else if let Some(m) = missing.as_ref() {
-            get_missing_val_as_u64_lenient(column_type, accessor.max_value(), m, field_name)?
+            get_missing_val_as_u64_lenient(column_type, accessor.values.num_vals(), m, field_name)?
```
fulmicoton: see comment above.

src/aggregation/accessor_helpers.rs
barbarj: Hey @fulmicoton, I revisited this and now understand what you were getting at. I've changed it to use `max_value` as before, with `num_vals` added so we can handle the zero non-nil terms case. (I've also updated the tests to include duplicate values so that the problem you pointed out with the original code is covered)

PSeitz-dd:
Wouldn't it be simpler to update `missing_sentinel_adjustment` for this case? Otherwise the logic around it is in different places

```rust
    let mut allowed: Option<BitSet> = None;
    let missing_sentinel_adjustment = if reserve_missing_sentinel { 1 } else { 0 };
    let allowed_capacity = str_col.dictionary().num_terms() as u32 + missing_sentinel_adjustment;
```

The correct fix IMO would be to make return `max_value` and `min_value` an `Option`. I'd guess there are other places where we may misuse it.

barbarj:
> Wouldn't it be simpler to update `missing_sentinel_adjustment` for this case? Otherwise the logic around it is in different places
> 
> ```rust
>     let mut allowed: Option<BitSet> = None;
>     let missing_sentinel_adjustment = if reserve_missing_sentinel { 1 } else { 0 };
>     let allowed_capacity = str_col.dictionary().num_terms() as u32 + missing_sentinel_adjustment;
> ```
I don't think so. The `missing_sentinel_adjustment` is adjusting the size of the `allowed` terms bitset. The bug is in the calculation of the missing sentinel value itself, which is ultimately used to index into the bitset. We could account for this in the computed `missing_sentinel_adjustment` by making it `2` in this case, but then we'd be 1) adjusting for a bug somewhere other than it's source, and 2) would have to have an unused bit in the 0 slot.

> The correct fix IMO would be to make return max_value and min_value an Option. I'd guess there are other places where we may misuse it.

Agreed, though it may be a while before I can do that. A quick attempt at it shows that the blast radius is a lot wider than I can quickly tackle.

PR #2931, opened 2026-05-14
Fix opening positions file error

JinheLin:
Within `open_with_custom_alive_set`, `segment.open_read(Positions)` must differentiate the file does not exist error from other I/O errors.

### DIFF
--- a/src/index/segment_reader.rs
+++ b/src/index/segment_reader.rs
@@ -6,6 +6,7 @@ use common::{ByteCount, HasLen};
 use fnv::FnvHashMap;
 use itertools::Itertools;
 
+use crate::directory::error::OpenReadError;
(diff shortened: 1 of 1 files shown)


PSeitz:
Do you have a test case for this, that fails without this change?

JinheLin:
@PSeitz 
I use Tantivy to implement full-text indexing internally in database. In scenarios with a large number of tables and indexes, **mmap may fail when opening files**, as the number of mmap mappings exceeds system limits.

Simply falling back to `CompositeFile::empty()` here defers the error until query time. Moreover, the returned error is **"Data corrupted"**, which is highly misleading.

PSeitz (approved):

PSeitz:
makes sense, thanks for the PR!

PR #2937, opened 2026-05-19
Implement exponential search for rank -> docid batch lookup

waamm:
This PR addresses an almost 2 year old TODO by @PSeitz: instead of scanning doc-by-doc to find which document owns a given value rank, we now bracket the answer exponentially and finish with a monotone binary search. Quick informal timings suggest this is roughly **~5×** faster on some workloads.

Tests include cases against the old linear-scan implementation.

Edit: On a 10M-document multivalued column where every doc has 2 values, a single `select_batch_in_place` call converting 10k evenly spaced row ranks (on an M4 Max):

```
10M nonempty / 10k ranks exponential
select_batch_single_batch        Avg: 2.2253ms (-3.87%)    Median: 2.2268ms (-3.83%)    [2.1889ms .. 2.2818ms]    
10M nonempty / 10k ranks linear
select_batch_single_batch        Avg: 11.9561ms (-3.13%)    Median: 11.8640ms (-3.45%)    [11.7694ms .. 14.0043ms] 
```

### DIFF
--- a/columnar/src/column_index/multivalued_index.rs
+++ b/columnar/src/column_index/multivalued_index.rs
@@ -193,7 +193,12 @@ impl MultiValueIndexV1 {
         for i in 0..ranks.len() {
             let pos = ranks[i];
             if cumulative_end_for_doc_idx(col, cur_doc) <= pos {
-                cur_doc = exponential_search_first_end_gt(col, cur_doc, pos, self.num_docs().saturating_sub(1));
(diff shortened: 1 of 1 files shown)


fulmicoton:
Are there benchmark confirming this is improving something? Can you paste the results in the PR description?

waamm:
Hi! I added a description of the benchmark I had in mind (I can also add the code).

PR #2940, opened 2026-05-27
Add sve + neon filter vec implementation as spotted by Adam

fulmicoton:
```
bitpacking_read        Avg: 0.1190ms (+0.86%)    Median: 0.1190ms (+0.86%)    [0.1190ms .. 0.1193ms]
blockedbitp_read        Avg: 0.0505ms (-0.29%)    Median: 0.0505ms (-0.21%)    [0.0505ms .. 0.0507ms]
blockedbitp_create        Avg: 0.0845ms (-0.47%)    Median: 0.0845ms (-0.46%)    [0.0843ms .. 0.0846ms]
filter_vec_dense        Avg: 0.0387ms (-60.77%)    Median: 0.0387ms (-60.76%)    [0.0386ms .. 0.0388ms]
filter_vec_sparse        Avg: 0.0399ms (-59.99%)    Median: 0.0399ms (-59.98%)    [0.0398ms .. 0.0400ms]
filter_vec_full        Avg: 0.0373ms (-61.94%)    Median: 0.0373ms (-61.92%)    [0.0372ms .. 0.0373ms]
```


bitpacker/benches/bench.rs
```diff
@@ -62,4 +61,124 @@ mod tests {
...
+    const MAX_VAL: u32 = 1_000;
+
+    fn make_values(n: usize, max_val: u32) -> Vec<u32> {
+        (0..n as u32)
+            .map(|i| (i as u64 * max_val as u64 / n as u64) as u32)
+            .collect()
+    }
+
+    #[bench]
+    fn bench_filter_vec_dense(b: &mut Bencher) {
```
PSeitz: Can you use binggan to show to difference before and after?

fulmicoton: done

fulmicoton:
@PSeitz done, can you review?

adamtobey (commented):
This looks good to me. I have a small addition to the prop tests that I'd like to push to your branch if that's OK. That is the main way I'll feel confident in the change (code makes sense, but not easy to find bugs or anything just by eye). Related to that, do we actually have any real SVE test coverage? We have the test code there, but do we have any CI or dev machines that actually have SVE? I'm wondering if we should punt on SVE for now. SVE capable machines should also support NEON, and that's easily testable today

fulmicoton:
@adamtobey I don't think we have that but that would be nice. I will run your proptest on a datadog workspace

fulmicoton:
I iterated a bit more on proptesting to:
-  test all available instruction sets
- explore values in a more opinionated way increasing the chances to spot bugs.

bitpacker/src/filter_vec/mod.rs
```diff
@@ -1,15 +1,28 @@
...
+
 mod scalar;
 
 #[derive(Clone, Copy, Eq, PartialEq, Debug)]
 #[repr(u8)]
 enum FilterImplPerInstructionSet {
     #[cfg(target_arch = "x86_64")]
     AVX2 = 0u8,
+    #[cfg(all(target_arch = "aarch64", not(target_vendor = "apple")))]
+    Sve = 3u8,
```
PSeitz: nitpick: SVE

PSeitz:
Would be good to know how the SVE assembly performs vs the NEON one for 128bit (graviton 4)

SVE support with more than 128bit is pretty rare. Graviton 3 has 256Bit, but Graviton 4 only has 128bit. 
As of today the SVE code may not be worth it.

fulmicoton:
There was a bench for that pascal, and for all instruction sets independently. I removed it because you asked me to get binggan before/after output, which required to bench the public function with the implementation dispatch
NEON is halfway between scalar and SVE

bitpacker/src/filter_vec/mod.rs
fulmicoton: Renaming yes

bitpacker/src/filter_vec/mod.rs
```diff
@@ -141,25 +223,59 @@ mod tests {
...
-                FilterImplPerInstructionSet::Scalar.filter_vec_in_place(start..=end, offset, &mut vals_clone);
-                assert_eq!(&vals, &vals_clone);
-            }
+            mut vals in vals_strategy()) {
+                for implementation in available_impls() {
+                    if implementation == FilterImplPerInstructionSet::Scalar {
+                        continue;
+                    }
+                    let mut vals_clone = vals.clone();
+                    implementation.filter_vec_in_place(start..=end, offset, &mut vals);
```
trinity-1686a, line 275: this clobbers `vals` so that the next implementation is tested on an already filtered slice

bitpacker/src/filter_vec/neon.rs
```diff
@@ -0,0 +1,113 @@
...
+        // SAFETY: mask is always in [0, 15] by construction (max sum of [1,2,4,8]).
+        // BYTE_SHUFFLE_TABLE has 16 entries, so this is always in bounds.
+        let shuffle = BYTE_SHUFFLE_TABLE.get_unchecked(mask as usize);
+        let shuffle_vec = vld1q_u8(shuffle.as_ptr());
+        vreinterpretq_u32_u8(vqtbl1q_u8(vreinterpretq_u8_u32(data), shuffle_vec))
+    }
+}
+
+#[inline(never)]
+pub fn filter_vec_in_place(range: RangeInclusive<u32>, offset: u32, output: &mut Vec<u32>) {
```
trinity-1686a, line 21: semantically, this function should be unsafe, it doesn't hold the safety invariant of `filter_vec_neon_aux` by itself (though its only caller, `FilterImplPerInstructionSet::filter_vec_in_place` does check that invariant before calling already)

bitpacker/src/filter_vec/neon.rs
```diff
@@ -0,0 +1,113 @@
...
+    [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 0, 1, 2, 3], // 0b0111: lanes 0,1,2
+    [12, 13, 14, 15, 0, 1, 2, 3, 0, 1, 2, 3, 0, 1, 2, 3], // 0b1000: lane 3
+    [0, 1, 2, 3, 12, 13, 14, 15, 0, 1, 2, 3, 0, 1, 2, 3], // 0b1001: lanes 0,3
+    [4, 5, 6, 7, 12, 13, 14, 15, 0, 1, 2, 3, 0, 1, 2, 3], // 0b1010: lanes 1,3
+    [0, 1, 2, 3, 4, 5, 6, 7, 12, 13, 14, 15, 0, 1, 2, 3], // 0b1011: lanes 0,1,3
+    [8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2, 3, 0, 1, 2, 3], // 0b1100: lanes 2,3
+    [0, 1, 2, 3, 8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2, 3], // 0b1101: lanes 0,2,3
+    [4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2, 3], // 0b1110: lanes 1,2,3
+    [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15], // 0b1111: all lanes
+];
```
trinity-1686a, line 113: my understanding is that most of the `0, 1, 2, 3` could be `16, 16, 16, 16`, which would more clearly convey that we don't care about the value. it took me some time to understand what the omnipresent 0,1,2,3 was for, and why 0000 and 0001 had the same table

bitpacker/src/filter_vec/sve.rs
```diff
@@ -0,0 +1,258 @@
...
+
+//         let mut input = input;
+//         let mut output_tail = output;
+
+//         // Unrolled ×2: both cntp calls have independent inputs and execute in parallel.
+//         // The two output_tail updates are sequential but together cost 4+1+1=6 cy per
+//         // pair vs 5+5=10 cy for two scalar iterations, breaking the cntp latency chain.
+//         let num_pairs = num_words / 2;
+//         for _ in 0..num_pairs {
+//             let word_a = svld1_u32(all_true, input);
```
trinity-1686a, line 218: my reading on all these operations is that we could have a last iteration that has its svbool param be not all true (possibly even being all false for one of the two vectors if we want to keep using the unrolled trick), and remove non-simd handling of ends that do not divide cleanly

bitpacker/src/filter_vec/sve.rs
```diff
@@ -0,0 +1,258 @@
...
+//             // cntp_a and cntp_b have independent inputs: OOO engine issues them in parallel.
+//             let added_len_a = svcntp_b32(all_true, in_range_a) as usize;
+//             let added_len_b = svcntp_b32(all_true, in_range_b) as usize;
+
+//             // Write the full vector — only the first added_len slots are valid.
+//             // Subsequent iterations overwrite the trailing zeros before truncate.
+//             svst1_u32(all_true, output_tail, compacted_a);
+//             output_tail = output_tail.add(added_len_a);
+//             svst1_u32(all_true, output_tail, compacted_b);
+//             output_tail = output_tail.add(added_len_b);
```
trinity-1686a, line 238: i wonder if this optimise better. I would somewhat expect the compiler to reorder, or the cpu to out of order execute, but i'm not sure how much intrinsics prevent reordering by compiler
```suggestion
//             let output_tail_2 = output_tail.add(added_len_a);
//             svst1_u32(all_true, output_tail, compacted_a);
//             svst1_u32(all_true, output_tail_2, compacted_b);
//             output_tail = output_tail_2.add(added_len_b);
```

trinity-1686a: nevermind, this was reordered by the compiler into
```
"str z5, [{out}]", //first store
"st1w {{z6.s}}, p0, [{out}, {cnt_a}, lsl #2]", // 2nd store, with added offset
"add {out}, {out}, {cnt_a}, lsl #2", // 1st add
"add {out}, {out}, {cnt_b}, lsl #2", // 2nd add

File: bitpacker/src/filter_vec/mod.rs [hunk 8]
```diff
@@ -141,25 +223,59 @@ mod tests {
...
+                    FilterImplPerInstructionSet::Scalar.filter_vec_in_place(start..=end, offset, &mut vals_clone);
+                    assert_eq!(&vals, &vals_clone);
+                }
```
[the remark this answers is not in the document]
fulmicoton: https://github.com/quickwit-oss/tantivy/pull/2952

bitpacker/src/filter_vec/neon.rs
fulmicoton: comment valid for sve, addressed in https://github.com/quickwit-oss/tantivy/pull/2952

bitpacker/src/filter_vec/neon.rs
fulmicoton: agreed. addressed in https://github.com/quickwit-oss/tantivy/pull/2952

bitpacker/src/filter_vec/sve.rs
fulmicoton: i d rather play it safe and keep it scalar.

PR #2952, opened 2026-06-06
CR comments from https://github.com/quickwit-oss/tantivy/pull/2940

fulmicoton:


trinity-1686a (approved):

fulmicoton:
thank you @trinity-1686a !

PR #2957, opened 2026-06-11
2x faster terms × histogram aggregation via a specialized collector

PSeitz-dd:
## What problem(s) was I solving?

`terms` × `histogram`/`date_histogram` (e.g. "status terms, each broken into a date histogram over time") is one of the most common aggregations in log/observability workloads, but it ran entirely through the general nested-aggregation machinery:

- The terms collector buffers doc-ids per bucket, then the histogram sub-collector **re-fetches the histogram column scattered by those doc-ids** (random access) with per-flush dynamic dispatch.
- The histogram collector stored every bucket in an `FxHashMap` keyed by bucket position — **hashing on every value** even when buckets densely cover a contiguous range.
- Every histogram bucket carried a `BucketId` **even with no sub-aggregation** to consume it.

## How I implemented it

1. **Specialized terms × histogram collector** — new `src/aggregation/bucket/term_agg/term_histogram.rs`. For top-level low-cardinality `terms` over a full column × a single `histogram`/`date_histogram` leaf (nothing nested) over a full column, with a small grid (capped at 2¹⁴ cells). It keeps a flat dense 2-D `u32` grid `counts[term * num_time_buckets + bucket]`; one pass zips both columns positionally and bumps counters directly — no doc-id buffering, no scattered re-fetch, no per-flush dynamic dispatch, no per-bucket key/id storage during collection. At result time it expands back via `SegmentHistogramCollector::from_dense_rows` into the shared intermediate builders, so cross-segment merge / ordering / cut-off / dict handling are identical to the general path.

   - **Derive per-term counts from the histogram grid.** With no hard bounds, every doc lands in the dense range, so a term's total `doc_count` equals the sum of its histogram row. Rather than keep a separate per-doc term counter (`term_counts[t] += 1` on every document), we drop that work entirely during collection and recover the totals as grid row-sums at the end. This removes a loop-carried store-to-load dependency on the few hot term addresses — **~17%** on the benches. (With `hard_bounds`, out-of-range docs still count toward the term but not the histogram, so the per-doc counter is retained.)

2. **Dense histogram bucket storage** (`histogram.rs`) — `HistogramBuckets` is now `Sparse(FxHashMap) | Dense { base_pos, Vec }`. Starts sparse; `maybe_densify` (at block boundaries) switches to a `Vec` indexed by `bucket_pos - base_pos` only once that uses ≤ the map's current memory, so peak memory never rises and a large-but-sparse range never crosses over. Removes hashing from the dense hot loop. `compute_dense_range` derives the range from column min/max clamped to bounds.

3. **Zero-sized bucket id when there are no sub-aggregations** (`histogram.rs`) — `SegmentHistogramBucketEntry<B>` is generic over a `BucketIdSlot`: `BucketId` with sub-aggs, `()` without. `()` makes id assignment a no-op and drops 8 bytes/bucket (24→16). `build_segment_histogram_collector` picks the slot by whether the node has children.

4. **Split histogram hot loop by sub-agg presence** — the no-sub-agg path uses the cheaper `iter_vals` (no doc-ids); the sub-agg path keeps `iter_docid_vals`.

5. **Contiguous-block fast path in the columnar accessor** (`block_accessor.rs`, `column_values/mod.rs`) — `fetch_block` detects a contiguous ascending doc run and fetches via `get_range` (bulk slice decode, no per-value dynamic dispatch) instead of `get_vals` gather; `get_range`'s default impl now decodes 4-wide.

6. **Misc** — `#[inline]` on `SubAggBuffer::push`; `term_agg.rs` → `term_agg/mod.rs` to host the submodule; Histogram/DateHistogram now route through `build_segment_histogram_collector` in `agg_data.rs`.

# Scope

Currently the specialized collector only covers terms-> histogram aggregations, but we could also cover histogram -> terms with it for some cases.

# Benchmark
`terms_status_with_histogram` is 2x faster than before and on par with a pure term aggregation like `terms_7`.

```
full
average_u64                                           Memory: 21.8 KB (-0.01%)      Avg: 3.0488ms (+0.24%)      Median: 3.0495ms (+1.25%)      [2.9689ms .. 3.1085ms]       
average_f64                                           Memory: 21.5 KB (-3.36%)      Avg: 3.1754ms (+0.01%)      Median: 3.1552ms (-0.11%)      [3.1036ms .. 3.3260ms]       
average_f64_u64                                       Memory: 22.1 KB (-3.02%)      Avg: 5.9575ms (+1.54%)      Median: 5.9325ms (+1.49%)      [5.8536ms .. 6.2756ms]       
stats_f64                                             Memory: 22.8 KB (+4.45%)      Avg: 3.1777ms (+0.08%)      Median: 3.1654ms (+0.59%)      [3.1200ms .. 3.2973ms]       
extendedstats_f64                                     Memory: 23.1 KB (-0.49%)      Avg: 3.3235ms (-0.85%)      Median: 3.3084ms (-0.57%)      [3.2781ms .. 3.4639ms]       
percentiles_f64                                       Memory: 39.4 KB (+0.31%)      Avg: 7.0153ms (-1.05%)      Median: 6.9986ms (-0.61%)      [6.9162ms .. 7.2479ms]       
terms_7                                               Memory: 37.2 KB               Avg: 2.3745ms (-1.50%)      Median: 2.3700ms (-1.24%)      [2.3510ms .. 2.4221ms]       
terms_all_unique                                      Memory: 12.7 MB               Avg: 6.8086ms (-1.52%)      Median: 6.8867ms (+0.40%)      [6.1163ms .. 7.6669ms]       
terms_150_000                                         Memory: 3.0 MB                Avg: 6.4853ms (-1.94%)      Median: 6.4640ms (-1.32%)      [6.3474ms .. 6.7107ms]       
terms_many_top_1000                                   Memory: 5.2 MB                Avg: 9.3724ms (-0.49%)      Median: 9.3168ms (-0.58%)      [9.1974ms .. 9.9923ms]       
terms_many_order_by_term                              Memory: 3.0 MB                Avg: 5.3237ms (-1.55%)      Median: 5.2654ms (-1.68%)      [5.1582ms .. 5.7237ms]       
terms_many_with_top_hits                              Memory: 48.9 MB               Avg: 91.3558ms (-3.88%)     Median: 91.1049ms (-2.45%)     [77.6622ms .. 120.1151ms]    
terms_all_unique_with_avg_sub_agg                     Memory: 54.7 MB (-0.00%)      Avg: 18.4627ms (+0.17%)     Median: 18.8469ms (+0.40%)     [15.8393ms .. 21.2119ms]     
terms_many_with_avg_sub_agg                           Memory: 13.5 MB (+0.00%)      Avg: 17.4354ms (+1.13%)     Median: 17.6457ms (+1.63%)     [14.7116ms .. 20.5066ms]     
terms_status_with_avg_sub_agg                         Memory: 90.7 KB               Avg: 5.1456ms (-1.74%)      Median: 5.1092ms (-1.80%)      [5.0377ms .. 5.4171ms]       
terms_status_with_terms_zipf_1000_sub_agg             Memory: 203.8 KB              Avg: 4.1112ms (-2.98%)      Median: 4.1000ms (-2.05%)      [4.0760ms .. 4.2348ms]       
terms_zipf_1000_with_terms_status_sub_agg             Memory: 682.3 KB              Avg: 12.6181ms (-12.90%)    Median: 12.5871ms (-12.79%)    [12.4670ms .. 12.9818ms]     
terms_status_with_histogram                           Memory: 129.7 KB (+16.57%)    Avg: 2.4701ms (-49.01%)     Median: 2.4535ms (-49.18%)     [2.4164ms .. 2.6420ms]       
terms_status_with_date_histogram                      Memory: 125.7 KB (+20.39%)    Avg: 2.4710ms (-47.13%)     Median: 2.4571ms (-47.09%)     [2.4264ms .. 2.5937ms]       
terms_status_with_date_histogram_and_sibling_terms    Memory: 126.4 KB (+5.79%)     Avg: 3.9756ms (-36.27%)     Median: 3.9637ms (-35.94%)     [3.9328ms .. 4.0482ms]       
terms_zipf_1000                                       Memory: 69.7 KB               Avg: 2.2323ms (-1.86%)      Median: 2.2230ms (-1.98%)      [2.1847ms .. 2.4241ms]       
terms_zipf_1000_with_histogram                        Memory: 1.2 MB (-10.06%)      Avg: 20.4613ms (-7.77%)     Median: 20.1618ms (-8.04%)     [19.9110ms .. 28.3846ms]     
terms_zipf_1000_with_avg_sub_agg                      Memory: 466.6 KB              Avg: 9.2337ms (+5.18%)      Median: 9.1955ms (+5.83%)      [9.1335ms .. 9.6850ms]       
terms_many_json_mixed_type_with_avg_sub_agg           Memory: 17.9 MB               Avg: 28.0062ms (+3.81%)     Median: 27.5667ms (+3.28%)     [26.1262ms .. 38.6095ms]     
composite_term_many_page_1000                         Memory: 619.9 KB              Avg: 25.3132ms (+0.51%)     Median: 25.0434ms (+0.30%)     [24.8027ms .. 28.2588ms]     
composite_term_many_page_1000_with_avg_sub_agg        Memory: 4.2 MB                Avg: 25.8580ms (-3.62%)     Median: 25.8224ms (-2.74%)     [25.3591ms .. 27.0475ms]     
composite_term_few                                    Memory: 37.2 KB               Avg: 12.2783ms (-13.84%)    Median: 12.1950ms (-13.95%)    [11.8739ms .. 12.9316ms]     
composite_histogram                                   Memory: 309.0 KB              Avg: 17.5778ms (-15.27%)    Median: 17.4555ms (-16.21%)    [17.0569ms .. 18.6302ms]     
composite_histogram_calendar                          Memory: 21.1 KB (-0.29%)      Avg: 26.6866ms (+3.88%)     Median: 26.4930ms (+4.32%)     [23.5668ms .. 29.8123ms]     
cardinality_agg                                       Memory: 5.9 MB                Avg: 13.0570ms (-5.35%)     Median: 13.0484ms (-3.40%)     [12.7178ms .. 13.6398ms]     
cardinality_agg_high_card                             Memory: 24.0 MB               Avg: 66.9179ms (-0.62%)     Median: 66.8118ms (-0.65%)     [66.3455ms .. 68.6998ms]     
cardinality_agg_low_card                              Memory: 21.0 KB               Avg: 1.2359ms (-3.91%)      Median: 1.2090ms (-5.35%)      [1.1834ms .. 1.9768ms]       
terms_status_with_cardinality_agg                     Memory: 92.2 KB               Avg: 3.2289ms (-3.48%)      Median: 3.2162ms (-2.13%)      [3.1695ms .. 3.4494ms]       
terms_100_buckets_with_cardinality_agg                Memory: 9.7 MB                Avg: 49.7009ms (-3.86%)     Median: 49.6114ms (-2.90%)     [48.9692ms .. 51.0129ms]     
terms_many_with_single_term_order_by_card             Memory: 48.9 MB               Avg: 78.2022ms (-18.09%)    Median: 79.1280ms (-15.08%)    [69.7685ms .. 92.2715ms]     
terms_many_with_single_term_2_order_by_card           Memory: 40.6 MB (+0.00%)      Avg: 53.8404ms (-13.33%)    Median: 53.8191ms (-12.87%)    [45.8436ms .. 78.0856ms]     
range_agg                                             Memory: 25.5 KB (+0.31%)      Avg: 3.3144ms (-3.21%)      Median: 3.3037ms (-1.84%)      [3.2787ms .. 3.4476ms]       
range_agg_with_avg_sub_agg                            Memory: 93.7 KB               Avg: 6.9874ms (-1.87%)      Median: 6.9612ms (-1.36%)      [6.9177ms .. 7.2324ms]       
range_agg_with_term_agg_status                        Memory: 110.1 KB              Avg: 6.3740ms (-2.62%)      Median: 6.3541ms (-1.53%)      [6.2933ms .. 6.6485ms]       
range_agg_with_term_agg_many                          Memory: 6.9 MB                Avg: 13.1165ms (-1.57%)     Median: 12.9684ms (-1.91%)     [12.7785ms .. 14.9161ms]     
histogram                                             Memory: 22.5 KB (+1.16%)      Avg: 2.8934ms (-4.08%)      Median: 2.8674ms (-3.83%)      [2.8401ms .. 3.0533ms]       
histogram_hard_bounds                                 Memory: 21.2 KB (+1.17%)      Avg: 1.6378ms (-2.19%)      Median: 1.6340ms (-2.53%)      [1.5854ms .. 1.7588ms]       
histogram_with_avg_sub_agg                            Memory: 122.1 KB (+0.05%)     Avg: 8.4063ms (-10.13%)     Median: 8.3718ms (-9.82%)      [8.3107ms .. 8.7760ms]       
histogram_with_term_agg_status                        Memory: 496.7 KB (-0.37%)     Avg: 9.5842ms (+0.93%)      Median: 9.5203ms (+1.42%)      [9.4448ms .. 10.1218ms]      
avg_and_range_with_avg_sub_agg                        Memory: 82.7 KB               Avg: 9.8139ms (-3.28%)      Median: 9.7739ms (-3.49%)      [9.7269ms .. 10.1989ms]      
filter_agg_all_query_count_agg                        Memory: 128.8 KB (-0.44%)     Avg: 5.6002ms (+15.99%)     Median: 5.6153ms (+16.74%)     [5.3823ms .. 5.8560ms]       
filter_agg_term_query_count_agg                       Memory: 129.0 KB (-0.32%)     Avg: 8.0399ms (+10.79%)     Median: 8.0387ms (+11.03%)     [7.7664ms .. 8.6361ms]       
filter_agg_all_query_with_sub_aggs                    Memory: 151.1 KB (-0.17%)     Avg: 10.5435ms (+8.24%)     Median: 10.5488ms (+8.57%)     [10.2309ms .. 10.9686ms]     
filter_agg_term_query_with_sub_aggs                   Memory: 147.4 KB (-0.17%)     Avg: 12.9841ms (+6.54%)     Median: 12.9893ms (+7.05%)     [12.5942ms .. 13.3322ms]     
```

# Overhead
Preallocating a vec has some overhead. Here's the cutoff in number of hits:

| grid    | crossover |
| -------- | ------- |
| 256  | ~16 hits    |
| 1024 | ~30 hits     |
| 4096    | ~100 hits    |
| 16384    | ~1000 hits    |

# Future work

Accessing values is now taking 46% of the time in `terms_status_with_date_histogram`. The way we access them is very wasteful. All values are returned as u64, even if they are internally bitpacked to a few bits.

Currently we require both to be full, but only the histogram needs to be full. terms could be optional.

### DIFF
--- a/src/aggregation/bucket/histogram/histogram.rs
+++ b/src/aggregation/bucket/histogram/histogram.rs
@@ -338,24 +338,41 @@ impl SegmentAggregationCollector for SegmentHistogramCollector {
         agg_data
             .column_block_accessor
             .fetch_block(docs, &req.accessor);
-        for (doc, val) in agg_data
-            .column_block_accessor
-            .iter_docid_vals(docs, &req.accessor)
-        {
-            let val = f64_from_fastfield_u64(val, req.field_type);
-            let bucket_pos = get_bucket_pos(val);
-            if bounds.contains(val) {
-                let bucket = buckets.entry(bucket_pos).or_insert_with(|| {
-                    let key = get_bucket_key_from_pos(bucket_pos as f64, interval, offset);
-                    SegmentHistogramBucketEntry {
-                        key,
-                        doc_count: 0,
-                        bucket_id: self.bucket_id_provider.next_bucket_id(),
-                    }
-                });
-                bucket.doc_count += 1;
-                if let Some(sub_agg) = &mut self.sub_agg {
+        // special path for nested buckets
+        if let Some(sub_agg) = &mut self.sub_agg {
+            for (doc, val) in agg_data
+                .column_block_accessor
+                .iter_docid_vals(docs, &req.accessor)
+            {
+                let val = f64_from_fastfield_u64(val, req.field_type);
+                let bucket_pos = get_bucket_pos(val);
+                if bounds.contains(val) {
+                    let bucket = buckets.entry(bucket_pos).or_insert_with(|| {
+                        let key = get_bucket_key_from_pos(bucket_pos as f64, interval, offset);
+                        SegmentHistogramBucketEntry {
+                            key,
+                            doc_count: 0,
+                            bucket_id: self.bucket_id_provider.next_bucket_id(),
+                        }
+                    });
+                    bucket.doc_count += 1;
                     sub_agg.push(bucket.bucket_id, doc);
+                }
+            }
+        } else {
+            for val in agg_data.column_block_accessor.iter_vals() {
+                let val = f64_from_fastfield_u64(val, req.field_type);
+                let bucket_pos = get_bucket_pos(val);
+                if bounds.contains(val) {
+                    let bucket = buckets.entry(bucket_pos).or_insert_with(|| {
+                        let key = get_bucket_key_from_pos(bucket_pos as f64, interval, offset);
+                        SegmentHistogramBucketEntry {
+                            key,
+                            doc_count: 0,
+                            bucket_id: self.bucket_id_provider.next_bucket_id(),
+                        }
+                    });
+                    bucket.doc_count += 1;
                 }
             }
         }
--- a/src/aggregation/bucket/histogram/histogram.rs
+++ b/src/aggregation/bucket/histogram/histogram.rs
@@ -244,7 +244,7 @@ impl Display for HistogramBounds {
 }
 
 impl HistogramBounds {
-    fn contains(&self, val: f64) -> bool {
+    pub(crate) fn contains(&self, val: f64) -> bool {
         val >= self.min && val <= self.max
     }
 }
@@ -317,11 +317,11 @@ impl<B: BucketIdSlot> SegmentHistogramBucketEntry<B> {
 /// the histogram bounds). Buckets in `[base_pos, base_pos + len)` can be stored in a flat `Vec`
 /// indexed by `bucket_pos - base_pos`, avoiding the hash map on the hot path.
 #[derive(Clone, Copy, Debug)]
-struct DenseRange {
+pub(crate) struct DenseRange {
     /// `bucket_pos` mapped to index 0 of the dense `Vec`.
-    base_pos: i64,
+    pub(crate) base_pos: i64,
     /// Number of bucket positions in the range.
-    len: usize,
+    pub(crate) len: usize,
 }
 
 /// Storage for the histogram buckets of a single parent bucket.
@@ -444,8 +444,6 @@ pub struct SegmentHistogramCollector<B> {
     /// Theoretical bucket range derived from the column min/max, if dense `Vec` storage is
     /// viable. `None` keeps every parent bucket in the sparse hash map.
     dense_range: Option<DenseRange>,
-
-    small_column_block_accessor: columnar::ColumnBlockAccessor<u32>,
 }
 
 impl<B: BucketIdSlot> SegmentAggregationCollector for SegmentHistogramCollector<B> {
@@ -601,15 +599,7 @@ impl<B: BucketIdSlot> SegmentHistogramCollector<B> {
             None
         };
         let mut req_data = agg_data.per_request.histogram_req_data[node.idx_in_req_data].clone();
-        req_data.req.validate()?;
-        if req_data.field_type == ColumnType::DateTime && !req_data.is_date_histogram {
-            req_data.req.normalize_date_time();
-        }
-        req_data.bounds = req_data.req.hard_bounds.unwrap_or(HistogramBounds {
-            min: f64::MIN,
-            max: f64::MAX,
-        });
-        req_data.offset = req_data.req.offset.unwrap_or(0.0);
+        normalize_histogram_req(&mut req_data)?;
         agg_data
             .context
             .limits
@@ -629,9 +619,89 @@ impl<B: BucketIdSlot> SegmentHistogramCollector<B> {
             req_data,
             bucket_id_provider: BucketIdProvider::default(),
             dense_range,
-            small_column_block_accessor: columnar::ColumnBlockAccessor::default(),
         })
     }
+}
+
+impl SegmentHistogramCollector<()> {
+    /// Builds a histogram collector whose parent `t` is a dense histogram filled from
+    /// `counts[t * num_time_buckets .. (t + 1) * num_time_buckets]` (row-major). Used by the fused
+    /// terms×histogram collector to turn its flat 2D counters into the regular intermediate result,
+    /// so cross-segment merging is shared with the general path.
+    pub(crate) fn from_dense_rows(
+        req_data: HistogramAggReqData,
+        base_pos: i64,
+        num_time_buckets: usize,
+        counts: &[u32],
+    ) -> Self {
+        let interval = req_data.req.interval;
+        let offset = req_data.offset;
+        let num_parents = if num_time_buckets == 0 {
+            0
+        } else {
+            counts.len() / num_time_buckets
+        };
+        let parent_buckets = (0..num_parents)
+            .map(|t| {
+                let row = &counts[t * num_time_buckets..(t + 1) * num_time_buckets];
+                let buckets = row
+                    .iter()
+                    .enumerate()
+                    .map(|(b, &doc_count)| SegmentHistogramBucketEntry {
+                        key: get_bucket_key_from_pos(
+                            (base_pos + b as i64) as f64,
+                            interval,
+                            offset,
+                        ),
+                        doc_count: doc_count as u64,
+                        bucket_id: (),
+                    })
+                    .collect();
+                HistogramBuckets::Dense { base_pos, buckets }
+            })
+            .collect();
+        Self {
+            parent_buckets,
+            sub_agg: None,
+            req_data,
+            bucket_id_provider: BucketIdProvider::default(),
+            dense_range: None,
+        }
+    }
+}
+
+/// Validates and normalizes a histogram request in place: applies date ns-normalization (for a
+/// `histogram` on a date column) and resolves `bounds`/`offset` from the request.
+fn normalize_histogram_req(req_data: &mut HistogramAggReqData) -> crate::Result<()> {
+    req_data.req.validate()?;
+    if req_data.field_type == ColumnType::DateTime && !req_data.is_date_histogram {
+        req_data.req.normalize_date_time();
+    }
+    req_data.bounds = req_data.req.hard_bounds.unwrap_or(HistogramBounds {
+        min: f64::MIN,
+        max: f64::MAX,
+    });
+    req_data.offset = req_data.req.offset.unwrap_or(0.0);
+    Ok(())
+}
+
+/// Clones and normalizes (resolving interval/offset/bounds) the histogram request at `node`, and
+/// returns it together with its dense bucket range — or `None` if the column has no usable range.
+/// Used by the fused terms×histogram collector, which then owns the normalized request.
+pub(crate) fn prepare_histogram_dense_range(
+    agg_data: &AggregationsSegmentCtx,
+    node: &AggRefNode,
+) -> crate::Result<Option<(HistogramAggReqData, DenseRange)>> {
+    let mut req_data = agg_data.per_request.histogram_req_data[node.idx_in_req_data].clone();
+    normalize_histogram_req(&mut req_data)?;
+    let dense_range = compute_dense_range(
+        &req_data.accessor,
+        req_data.field_type,
+        req_data.req.interval,
+        req_data.offset,
+        req_data.bounds,
+    );
+    Ok(dense_range.map(|range| (req_data, range)))
 }
 
 /// Builds a boxed histogram (or date histogram) segment collector, picking the bucket-id storage
@@ -653,7 +723,7 @@ pub(crate) fn build_segment_histogram_collector(
 }
 
 #[inline]
-fn get_bucket_pos_f64(val: f64, interval: f64, offset: f64) -> f64 {
+pub(crate) fn get_bucket_pos_f64(val: f64, interval: f64, offset: f64) -> f64 {
     ((val - offset) / interval).floor()
 }
 
--- a/src/aggregation/bucket/histogram/histogram.rs
+++ b/src/aggregation/bucket/histogram/histogram.rs
@@ -682,6 +682,22 @@ fn normalize_histogram_req(req_data: &mut HistogramAggReqData) -> crate::Result<
         max: f64::MAX,
     });
     req_data.offset = req_data.req.offset.unwrap_or(0.0);
+    // Drop `hard_bounds` that can't exclude any value (the column's range already sits inside
+    // them): the per-doc `bounds.contains` check is then a no-op, so collapsing to the unbounded
+    // sentinel lets the histogram hot loop skip it and the fused term×histogram path derive
+    // per-term counts from the grid. Only this collect-time filter is touched — empty-bucket
+    // emission reads `req.hard_bounds` directly (see `get_req_min_max`), and `hard_bounds` only
+    // ever clips that range, so a wider-than-data bound leaves the result unchanged.
+    if req_data.req.hard_bounds.is_some() {
+        let col_min = f64_from_fastfield_u64(req_data.accessor.min_value(), req_data.field_type);
+        let col_max = f64_from_fastfield_u64(req_data.accessor.max_value(), req_data.field_type);
+        if col_min >= req_data.bounds.min && col_max <= req_data.bounds.max {
+            req_data.bounds = HistogramBounds {
+                min: f64::MIN,
+                max: f64::MAX,
+            };
+        }
+    }
     Ok(())
 }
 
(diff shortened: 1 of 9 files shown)


src/aggregation/bucket/term_agg/term_histogram.rs
```diff
@@ -0,0 +1,477 @@
...
+        return Ok(None);
+    }
+
+    // Clone + normalize the histogram request and get its dense bucket range; only take the fused
+    // path when the flat `num_terms × num_time_buckets` grid is small enough.
+    let Some((hist_req_data, range)) = prepare_histogram_dense_range(agg_data, &node.children[0])?
+    else {
+        return Ok(None);
+    };
+    let num_terms = (max_term_id + 1) as usize;
```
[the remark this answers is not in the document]
PSeitz-dd: we'll never have that many terms

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -338,30 +485,42 @@ impl SegmentAggregationCollector for SegmentHistogramCollector {
...
                 }
             }
         }
 
-        let mem_delta = self.get_memory_consumption(parent_bucket_id) - mem_pre;
-        if mem_delta > 0 {
+        // `checked_sub` is `None` when densifying shrank the accounted memory; only account growth.
+        if let Some(mem_delta) = self
+            .get_memory_consumption(parent_bucket_id)
+            .checked_sub(mem_pre)
```
trinity-1686a, line 522: does that mean we may think we're consuming more memory than we actually do when we densify?

src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -441,35 +599,178 @@ impl SegmentHistogramCollector {
...
+    /// so cross-segment merging is shared with the general path.
+    pub(crate) fn from_dense_rows(
+        req_data: HistogramAggReqData,
+        base_pos: i64,
+        num_time_buckets: usize,
+        counts: &[u32],
+    ) -> Self {
+        let interval = req_data.req.interval;
+        let offset = req_data.offset;
+        let num_parents = if num_time_buckets == 0 {
```
trinity-1686a: clippy would prefer  `counts.len().checked_div(num_time_buckets).unwrap_or(0)`

src/aggregation/bucket/term_agg/term_histogram.rs
```diff
@@ -0,0 +1,543 @@
...
+pub(crate) struct SegmentTermHistogramCollector {
+    /// The counts, indexed by term id.
+    ///
+    /// Only used when we can't derive the term counts from `counts`.
+    term_counts: Vec<u32>,
+    /// Flattened `[num_terms * num_time_buckets]` histogram counters (`u32`, see
+    /// `term_counts`).
+    ///
+    /// Each term id get its own contiguous slice of `num_time_buckets` histogram counter.
+    /// When we count all docs (#nofilter), we can derive the per-term total as the sum over that
```
trinity-1686a, line 52: when is there filtering inside aggregations? When setting hard bounds i guess?

src/aggregation/bucket/term_agg/term_histogram.rs
```diff
@@ -0,0 +1,543 @@
...
+                );
+                counts[term_id * num_time_buckets + bucket] += 1;
+            }
+        } else {
+            for (term_id, hist_raw) in self.term_block.iter_vals().zip(self.hist_block.iter_vals())
+            {
+                let term_id = term_id as usize;
+                term_counts[term_id] += 1;
+                let val = f64_from_fastfield_u64(hist_raw, field_type);
+                if bounds.contains(val) {
```
trinity-1686a: would it make sense to have 
```
if bounds.contains(val) {
   //... count the bucket
} else {
  term_counts[term_id] += 1;
}
```
and do `sum(counts[term_id *num_time_buckets..(term_id + 1) * num_time_buckets]) + term_counts[term_id]` when getting the per term result?

i think this would remove some mostly duplicated code, and i don't think this would impact performance

src/aggregation/bucket/term_agg/term_histogram.rs
```diff
@@ -0,0 +1,543 @@
...
+    is_top_level: bool,
+) -> crate::Result<Option<Box<dyn SegmentAggregationCollector>>> {
+    // Both columns must be full (one value per doc) so their values align positionally with `docs`
+    // and we can zip them. Requiring full columns also makes the terms agg's `missing` config a
+    // no-op (`fetch_block_with_missing` early-returns on full columns), so we needn't check for it.
+    //
+    // We don't cap the term cardinality here: the flat grid is bounded by the total cell count
+    // (`num_terms * num_time_buckets <= MAX_FUSED_GRID_BUCKETS`) checked below, which subsumes it.
+    let fuseable = is_top_level
+        && terms_req_data.allowed_term_ids.is_none()
```
trinity-1686a, line 230: i feel like this constraint could be removed easily (maybe it's not worth doing, but can be worth a comment)

src/aggregation/bucket/term_agg/term_histogram.rs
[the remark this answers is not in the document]
trinity-1686a: codex is right, for a u64 column (not a string column), we can end up storing high values. It's not really a term_id then, just a value, but it could be u64::MAX, cause an underflow to zero, allocate a 0 bucket `counts`, and panics as soon as collection starts

src/aggregation/bucket/histogram/histogram.rs
PSeitz-dd: Yes, the difference is minor though

src/aggregation/bucket/term_agg/term_histogram.rs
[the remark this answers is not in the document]
PSeitz-dd: yes, the variable name is wrong, renamed and fixed

src/aggregation/bucket/histogram/histogram.rs
fulmicoton-dd: +1

src/aggregation/bucket/term_agg/term_histogram.rs
```diff
@@ -0,0 +1,543 @@
...
+
+        // Both columns are full (checked at construction), so values align with `docs` positionally
+        // and are read together in one pass.
+        if self.derive_term_counts_from_histogram {
+            // No hard bounds: every doc lands in a bucket, so we skip the per-doc `term_counts`
+            // increment (derived from the grid at flush) and the always-true bounds check.
+            for (term_id, hist_val) in self.term_block.iter_vals().zip(self.hist_block.iter_vals())
+            {
+                let term_id = term_id as usize;
+                let val = f64_from_fastfield_u64(hist_val, field_type);
```
fulmicoton-dd: I am not fond of this part of the code.
It feels like this f64_from_fastfield_u64 should be done as a batch when we fetch the column values.

I am not entirely how to organize the code cleanly, but the existence of f64_from_fastfield_u64 itself is quite a code smell too.

src/aggregation/bucket/histogram/histogram.rs
PSeitz: fixed

src/aggregation/bucket/term_agg/term_histogram.rs
fulmicoton-dd: (although tbh if it does not impact the benchmark perf I guess I'm ok with it)

src/aggregation/bucket/term_agg/term_histogram.rs
PSeitz-dd: Yes updated, but with an additional check `if all_docs_in_bounds || bounds.contains(val) {`. This has the advantage that histograms with bounds may also benefit from the single vec count optimization.

src/aggregation/bucket/term_agg/term_histogram.rs
PSeitz-dd: I added a comment, it's top level only, because we don't don't know how many buckets there will be.

trinity-1686a: my comment was on the "allowed_term_ids" line

src/aggregation/bucket/term_agg/term_histogram.rs
PSeitz-dd: Yes, `f64_from_fastfield_u64` costs 13%. Accessing the values in the column costs 46%. So addressing those 2 will have the biggest impact. 

Currently accessing values is quite wasteful, even if the values would fit in u32 with millisecond precision we access them as u64 nanoprecision (or i64?) and then convert f64.
Similar for terms, they would fit in u8.

I'm thinking of having some accessor for an intermediate smaller representation and then a converter to convert to the final value at the end of collection, e.g. u32 millisecond precision accessor for DateTime or the bitpacked u8 values for terms. (not in this PR though)

src/aggregation/bucket/term_agg/term_histogram.rs
PSeitz-dd: ah yes, we can support that easily

PSeitz-dd: add a TODO

fulmicoton-dd (approved):

Issue #2352, opened 2024-04-09
Planned removal of index sorting in 0.23.0


PSeitz:
Index sorting provides currently little to no benefit, and there's some confusion about it from tantivy users on how to use it. (e.g. that it is a way to get sorted search results, which is not really the case)
It also adds considerable complexity to the indexing and merging parts of the code.

On the potential pros: 
- Range queries may be accelerated by using binary search instead of a full scan, but we don't do that currently and it is limited to one field you would need to pick in advance.
- Compression may be improved, but we don't have much data about that though.
- order_by queries on the same field may be accelerated. On a single segment that would be trivial and remove the need to fetch fast field values. On multiple segments this would be more difficult.

There will also be a deprecation warning added to `0.22.0` with a link to this issue.

If you have a use case which requires index sorting, please explain your use case here.

Related https://github.com/quickwit-oss/tantivy/issues/2335

sujayakar:
I was planning on using index sorting for our [full text search](https://docs.convex.dev/text-search) at Convex to ensure that results are sorted in a deterministic order that's independent of how documents are split across segments.

We're an OLTP database that lets users define text search indexes on their existing tables, much like Postgres with its GIN-based indexes. Indexes are fully transactional, with older data stored across multiple Tantivy segments and recent data stored in a multiversioned in-memory index. Furthermore, we want read transactions in our system to be deterministic so we can cache them automatically and precisely detect when they're invalidated.

So, we'd like our text search queries (which are just TopK queries by BM25 of an OR of a set of search terms) to be deterministic, even if there's been an index rebuild since the query was executed. Elsewhere in our system, our rows have a natural sort order based on their system-assigned ID and insertion time. Then, during query execution, ideally each segment would return its top K results based on `(BM25, natural order)`, independent of how `DocId`s are assigned within each segment, and then we'd merge the results to get a deterministic global result.

We've already done some of the work here (e.g. adjusting BM25 statistics to be deterministic as of the transaction's timestamp: #1855, [Convex `Bm25StatisticsProvider` implementation](https://github.com/get-convex/convex-backend/blob/26f9b0db4f22abd9de3d55c9a6ee36062fdde92f/crates/search/src/scoring.rs#L44)), but I would love to make this 100% accurate, and embedding the natural row order in the `DocId` order seems like a good path to do this efficiently at query time without having to load a fast field on every hit.

### Future features

[...]

PR #105, opened 2017-03-20
Avoid using make for building simdcomp

lnicola:
Fixes #69 

I kept the aggressive optimizations, `-msse4.1` and `-march=native` from the `simdcomp` `Makefile`s. We might want to revisit this later.

Issue #104, opened 2017-03-20
Windows support


lnicola:
- [x] [Flushing a directory fails on Windows](https://github.com/tantivy-search/tantivy/pull/103)
 - [x] [`simdcompression` feature doesn't build on Windows](https://github.com/tantivy-search/tantivy/pull/105)
 - [x] [Test suite expects deleting a memory-mapped file to succeed](https://github.com/tantivy-search/tantivy/pull/106)
 - [x] [AppVeyor CI](https://github.com/tantivy-search/tantivy/pull/108)
 - [x] ~~Investigate `lz4` build failure with `-gnu`~~ invalid, had to do a clean build to recompile the native libs
 - [x] Edit README.md, add the appveyor badge

fulmicoton:
Thanks for the very clear ticket.

fulmicoton:
@vchugreev Windows is not officially supported on windows just yet, but it should be working. We will change the README message when we have proper continuous integration.

KodrAus:
@fulmicoton I'll set up AppVeyor over the next few days, and then this should be rounded off!

fulmicoton:
Closing. 
I'll edit README with the release.

Good job guys!

PR #2959, opened 2026-06-13
feat: Restore index sorting

rebasedming:
We ([ParadeDB](https://github.com/paradedb/paradedb)) have restored and been using the removed [index sorting](https://github.com/quickwit-oss/tantivy/issues/2352) feature in our Tantivy fork.

Our use case is sorting the index by Postgres' internal `ctid` identifier. Results returned from Tantivy must be checked against Postgres' visibility map, and checking them in ctid order is much more cache friendly, resulting in up to 80% speedups for certain queries.

This PR is split into 5 commits, corresponding to the index sorting reversal plus bug fixes we uncovered during our usage of index sorting.

| Commit | Maps to | What it does |
|---|---|---|
| `2aea0ad9f` | foundation ([#104](https://github.com/paradedb/tantivy/pull/104)) | Restore `SegmentComponent::TempStore` (revert of upstream #2815). Subsumes fork PR [#104](https://github.com/paradedb/tantivy/pull/104)'s CI fix. |
| `9205bcb0c` | [#92](https://github.com/paradedb/tantivy/pull/92) | Restore sort-by-field (single-segment + merge paths). |
| `39c790f0f` | [#101](https://github.com/paradedb/tantivy/pull/101) | Enable `sort_by` for `Str`/`Bytes` fast fields. |
| `9c4341a87` | [#105](https://github.com/paradedb/tantivy/pull/105) | Native typed numeric sort-key comparison (precision/NULL fix). |
| `2d9ba2418` | [#106](https://github.com/paradedb/tantivy/pull/106) | Preserve NULL ordering in numeric segment merges. |

We have discussed with the Tantivy maintainers and they indicated they would be open to this PR. Another motivation for landing this PR is we are planning on contributing a significant refactor that makes Tantivy's segment components extensible, and landing that without index sorting leads to too many conflicts.


stuhood (approved):
We have been using this patch downstream in ParadeDB, and have found it to be necessary for some of our use cases.

----

The primary/essential benefit that this enables for us is receiving Query results in sorted order. But we have also explored some additional Query optimizations which are enabled by this feature, and have found some significant wins which we would be interested in upstreaming separately (including optimizations for set and range queries on sorted columns), if there is interest.

src/postings/recorder.rs
```diff
@@ -181,18 +194,35 @@ impl Recorder for TermFrequencyRecorder {
...
         self.stack.read_to_end(arena, buffer);
         let mut u32_it = VInt32Reader::new(&buffer[..]);
-        let mut prev_doc = 0;
-        while let Some(delta_doc_id) = u32_it.next() {
-            let doc_id = prev_doc + delta_doc_id;
-            prev_doc = doc_id;
-            let term_freq = u32_it.next().unwrap_or(self.current_tf);
-            serializer.write_doc(doc_id, term_freq, &[][..]);
+        if let Some(doc_id_map) = doc_id_map {
+            let mut doc_id_and_tf = vec![];
```
stuhood: This will allocate a new vector per term.

src/postings/recorder.rs
```diff
@@ -238,12 +268,14 @@ impl Recorder for TfAndPositionRecorder {
...
         &self,
         arena: &MemoryArena,
+        doc_id_map: Option<&DocIdMapping>,
         serializer: &mut FieldSerializer<'_>,
         buffer_lender: &mut BufferLender,
     ) {
         let (buffer_u8, buffer_positions) = buffer_lender.lend_all();
         self.stack.read_to_end(arena, buffer_u8);
         let mut u32_it = VInt32Reader::new(&buffer_u8[..]);
+        let mut doc_id_and_positions = vec![];
```
stuhood: Ditto, except this will be per-document.

stuhood (approved):
Looks good.

I ran the existing indexing and merging benchmarks, and performance is not impacted when index sorting is not enabled.

I did see two cases that will significantly increase allocation traffic, but only when sorting is enabled: will push a fix for those.

PR #2962, opened 2026-06-15
fix flaky test

PSeitz:
floating point issue in blockwand intersection test


trinity-1686a (approved):
can you add a comment explaining why this order of additions is "more correct" (is the same as in the actual code)

PR #2964, opened 2026-06-17
Fix MoreLikeThis panic on indexes with deleted documents

stumpylog:
## What

`MoreLikeThisQuery` panics with `doc_count >= doc_freq` (in `idf()`, `src/query/bm25.rs`) when the searched index contains soft-deleted documents.

```
thread '<unnamed>' panicked at src/query/bm25.rs:53:5:
95 >= 100
```

## Why

In `MoreLikeThis::create_score_term` the idf denominator is computed from `SegmentReader::num_docs()`, which counts **alive documents only**:

```rust
let num_docs = searcher
    .segment_readers()
    .iter()
    .map(|segment_reader| segment_reader.num_docs() as u64)
    .sum::<u64>();
...
let doc_freq = searcher.doc_freq(term)?;
let idf = idf(doc_freq, num_docs);
```

`searcher.doc_freq(term)` reads `doc_freq` straight from the term dictionary, and that value keeps counting soft-deleted documents until a merge expunges them. So after documents are deleted (e.g. a bulk import followed by deletions, with no merge in between), a common term's `doc_freq` can exceed the alive-only `num_docs`, and `idf()` trips its `assert!(doc_count >= doc_freq)`.

Ordinary search does not hit this because the standard BM25 statistics provider uses `Searcher::total_num_docs()`, which sums `SegmentReader::max_doc()` (deleted documents included) — consistent with how `doc_freq` is counted.

## Fix

Use `max_doc()` instead of `num_docs()` for the MLT idf denominator, so the document count and `doc_freq` are counted consistently (both include deleted documents). This mirrors `total_num_docs()` used by the regular BM25 scorer.

## Test

Adds `test_more_like_this_query_with_deleted_documents`: indexes six documents sharing body terms (with `NoMergePolicy` so postings survive), deletes two, and runs a MoreLikeThis query against a survivor. It panics on `main` and passes with this change.

## Context

Discovered downstream in paperless-ngx, where "More Like This" returns HTTP 500 after bulk import/delete cycles: https://github.com/paperless-ngx/paperless-ngx/issues/13024

Developed with assistance from Claude Code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Tantivy 0.26 (Unreleased)
 ================================
 
 ## Bugfixes
+- Fix `MoreLikeThis` panic on indexes with deleted documents [#2964](https://github.com/quickwit-oss/tantivy/pull/2964)(@stumpylog)
 - Align float query coercion during search with the columnar coercion rules [#2692](https://github.com/quickwit-oss/tantivy/pull/2692)(@fulmicoton)
 - Fix lenient elastic range queries with trailing closing parentheses [#2816](https://github.com/quickwit-oss/tantivy/pull/2816)(@evance-br)
 - Fix intersection `seek()` advancing below current doc id [#2812](https://github.com/quickwit-oss/tantivy/pull/2812)(@fulmicoton)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Tantivy 0.26 (Unreleased)
 ================================
 
 ## Bugfixes
+- Fix `MoreLikeThis` panic on indexes with deleted documents [#2964](https://github.com/quickwit-oss/tantivy/pull/2964)(@stumpylog)
```
stumpylog, line 11: I was a little confused about the released vs not state.  There's a 0.26.1 tag (https://github.com/quickwit-oss/tantivy/releases/tag/0.26.1) but the header here says 0.26.0 is (Unreleased).  So just let me know if this needs moving/updating.

stumpylog:
We've shipped our release using this library (with a workaround for the underlying issue on our end), and would love to see this fix merged upstream. @PSeitz whenever you get a chance, no rush.

PR #2966, opened 2026-06-18
fix(query-parser): honor phrase prefix and slop on JSON fields

DavIvek:
Phrase prefix (`"..."*`) and slop (`"..."~N`) were silently dropped on JSON fields. `generate_literals_for_json_object` hard-coded `slop: 0` and `prefix: false`, so a query like `data.name:"foo bar"*` degraded to an exact phrase on the JSON field — even though the underlying `PhrasePrefixQuery` already supports JSON-path terms.

This threads `slop`/`prefix` through the JSON branch, exactly as the `Str` branch (`generate_literals_for_str`) already does, and adds the "phrase prefix requires at least two terms" guard for the single-token case.

### Tests

Added as JSON analogues of the existing text-field tests (`test_phrase_prefix`, `test_phrase_prefix_too_short`, `test_phrase_slop`):

- `test_phrase_prefix_on_json_field`
- `test_phrase_prefix_too_short_on_json_field`
- `test_phrase_slop_on_json_field`

This aligns JSON fields with the phrase `~`/`*` behavior already documented for `QueryParser`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Tantivy 0.26 (Unreleased)
 - Fix integer overflow in segment sorting and merge policy truncation [#2846](https://github.com/quickwit-oss/tantivy/pull/2846)(@anaslimem)
 - Fix merging of intermediate aggregation results [#2719](https://github.com/quickwit-oss/tantivy/pull/2719)(@PSeitz)
 - Fix deduplicate doc counts in term aggregation for multi-valued fields [#2854](https://github.com/quickwit-oss/tantivy/pull/2854)(@nuri-yoo)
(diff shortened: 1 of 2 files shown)


Josipmrden:
Nice!

DavIvek:
Hi @PSeitz ! Sorry to ping — just wanted to gently bump this PR in case it slipped through. No rush at all, but whenever someone has a chance to take a look I'd really appreciate it. Happy to rebase or make any changes if that helps.

PSeitz:
Can you add a test that makes sure it works with actual data?

DavIvek:
> Can you add a test that makes sure it works with actual data?

Added!

PSeitz (approved):
Thanks for the PR!

PR #2978, opened 2026-06-26
feat: add custom doc id mapping finalization

Mallets:
## What

Adds manual doc id mapping finalization for segment writers. Callers can provide a checked `DocIdMapping` at finalize time, and Tantivy serializes postings, fast fields, fieldnorms, docstore, and opstamps in the mapped doc id order.

## Why

Index sorting already remaps doc ids internally from `sort_by_field`. This PR exposes the same finalization machinery for callers that compute their own doc id order, while keeping the temporary docstore path opt-in.

## Changes

- Expose `DocIdMapping` as public type.
- Add `DocIdMapping::new_permutation(...)` as the public constructor for caller-provided mappings.
- Add `SegmentWriter::finalize_with_doc_id_mapping(&DocIdMapping)`.
- Add `SingleSegmentIndexWriter::finalize_with_doc_id_mapping(&DocIdMapping)`.
- Add hidden, non-persisted `IndexSettings::manual_doc_id_mapping` (`#[serde(skip)]`) to opt into the temporary docstore path needed by manual mapping.
- Extend `SegmentSerializer::for_segment(...)` so it writes docs to `TempStore` when either `sort_by_field` is configured or `manual_doc_id_mapping` is enabled.
- Make `common::BitSet::insert()` return whether the set changed, and use that for duplicate detection in mapping validation.

## Design notes

`manual_doc_id_mapping` is intentionally a hidden, non-persisted `IndexSettings` flag. It is only used at segment construction time to select the temporary docstore path required by manual remapping, and is skipped from `meta.json`.

An alternative would be to avoid putting this runtime-only option in `IndexSettings` and instead add separate `SegmentWriter::for_segment` / `SegmentSerializer::for_segment` variants that explicitly request the temp docstore path. That would make the dependency more local, but it also requires threading a new construction mode through the segment writer/serializer setup and splitting the existing single constructor path. Given that index sorting already uses the same remapping machinery, keeping one constructor path and gating the temp docstore behavior through a hidden runtime flag is the smaller change.

To keep this from becoming a durable index invariant:

- `manual_doc_id_mapping` is `#[serde(skip)]`.
- It cannot be combined with `sort_by_field`.
- It is cleared from the returned `Index` after `finalize_with_doc_id_mapping`.

## Notes

- `manual_doc_id_mapping` is a runtime construction flag only; it is not written to `meta.json`.

## Test plan

- [x] `cargo test`

### DIFF
--- a/src/core/tests.rs
+++ b/src/core/tests.rs
@@ -1,10 +1,10 @@
 use crate::collector::Count;
 use crate::directory::{RamDirectory, WatchCallback};
-use crate::index::SegmentId;
+use crate::index::{SegmentComponent, SegmentId};
 use crate::indexer::{DocIdMapping, LogMergePolicy, NoMergePolicy};
 use crate::postings::Postings;
 use crate::query::TermQuery;
-use crate::schema::{Field, IndexRecordOption, Schema, Value, INDEXED, STORED, STRING, TEXT};
+use crate::schema::{Field, IndexRecordOption, Schema, Value, FAST, INDEXED, STORED, STRING, TEXT};
 use crate::tokenizer::TokenizerManager;
 use crate::{
     Directory, DocAddress, DocSet, Index, IndexBuilder, IndexReader, IndexSettings, IndexWriter,
@@ -342,6 +342,49 @@ fn test_single_segment_index_writer_with_doc_id_mapping() -> crate::Result<()> {
         doc_2.get_first(text_field).and_then(|val| val.as_str()),
         Some("alpha beta")
     );
+
+    assert!(!index.settings().manual_doc_id_mapping);
+    let segment_metas = index.searchable_segment_metas()?;
+    let segment_meta = &segment_metas[0];
+    assert!(!segment_meta
+        .list_files()
+        .contains(&segment_meta.relative_path(SegmentComponent::TempStore)));
+
+    let mut index_writer = index.writer_for_tests()?;
+    index_writer.add_document(doc!(text_field=>"delta"))?;
+    index_writer.commit()?;
+
+    Ok(())
+}
+
+#[test]
+fn test_single_segment_index_writer_with_sort_by_field_untracks_tempstore() -> crate::Result<()> {
+    let mut schema_builder = Schema::builder();
+    let sort_field = schema_builder.add_u64_field("sort", FAST | STORED);
+    let schema = schema_builder.build();
+    let sort_field_name = schema.get_field_entry(sort_field).name().to_string();
+    let directory = RamDirectory::default();
+    let settings = IndexSettings {
+        sort_by_field: Some(crate::IndexSortByField {
+            field: sort_field_name,
+            order: crate::Order::Asc,
+        }),
+        ..Default::default()
+    };
+    let mut single_segment_index_writer = Index::builder()
+        .schema(schema)
+        .settings(settings)
+        .single_segment_index_writer(directory, 15_000_000)?;
+
+    single_segment_index_writer.add_document(doc!(sort_field=>2u64))?;
+    single_segment_index_writer.add_document(doc!(sort_field=>1u64))?;
+    let index = single_segment_index_writer.finalize()?;
+
+    let segment_metas = index.searchable_segment_metas()?;
+    let segment_meta = &segment_metas[0];
+    assert!(!segment_meta
+        .list_files()
+        .contains(&segment_meta.relative_path(SegmentComponent::TempStore)));
     Ok(())
 }
 
@@ -365,6 +408,30 @@ fn test_single_segment_index_writer_finalize_rejects_manual_doc_id_mapping() -> 
     let error = single_segment_index_writer.finalize().unwrap_err();
     assert!(matches!(error, TantivyError::InvalidArgument(_)));
     Ok(())
+}
+
+#[test]
+fn test_index_builder_rejects_manual_doc_id_mapping_with_sort_by_field() {
+    let mut schema_builder = Schema::builder();
+    schema_builder.add_text_field("text", TEXT | STORED);
+    let sort_field = schema_builder.add_u64_field("sort", STORED | FAST);
+    let schema = schema_builder.build();
+    let sort_field_name = schema.get_field_entry(sort_field).name().to_string();
+    let settings = IndexSettings {
+        manual_doc_id_mapping: true,
+        sort_by_field: Some(crate::IndexSortByField {
+            field: sort_field_name,
+            order: crate::Order::Asc,
+        }),
+        ..Default::default()
+    };
+
+    let error = Index::builder()
+        .schema(schema)
+        .settings(settings)
+        .create_in_ram()
+        .unwrap_err();
+    assert!(matches!(error, TantivyError::InvalidArgument(_)));
 }
 
 #[test]
--- a/src/core/tests.rs
+++ b/src/core/tests.rs
@@ -346,9 +346,11 @@ fn test_single_segment_index_writer_with_doc_id_mapping() -> crate::Result<()> {
     assert!(!index.settings().manual_doc_id_mapping);
     let segment_metas = index.searchable_segment_metas()?;
     let segment_meta = &segment_metas[0];
+    let temp_store_path = segment_meta.relative_path(SegmentComponent::TempStore);
     assert!(!segment_meta
         .list_files()
-        .contains(&segment_meta.relative_path(SegmentComponent::TempStore)));
+        .contains(&temp_store_path));
+    assert!(!index.directory().exists(&temp_store_path)?);
 
     let mut index_writer = index.writer_for_tests()?;
     index_writer.add_document(doc!(text_field=>"delta"))?;
@@ -382,9 +384,11 @@ fn test_single_segment_index_writer_with_sort_by_field_untracks_tempstore() -> c
 
     let segment_metas = index.searchable_segment_metas()?;
     let segment_meta = &segment_metas[0];
+    let temp_store_path = segment_meta.relative_path(SegmentComponent::TempStore);
     assert!(!segment_meta
         .list_files()
-        .contains(&segment_meta.relative_path(SegmentComponent::TempStore)));
+        .contains(&temp_store_path));
+    assert!(!index.directory().exists(&temp_store_path)?);
     Ok(())
 }
 
--- a/src/core/tests.rs
+++ b/src/core/tests.rs
@@ -347,9 +347,7 @@ fn test_single_segment_index_writer_with_doc_id_mapping() -> crate::Result<()> {
     let segment_metas = index.searchable_segment_metas()?;
     let segment_meta = &segment_metas[0];
     let temp_store_path = segment_meta.relative_path(SegmentComponent::TempStore);
-    assert!(!segment_meta
-        .list_files()
-        .contains(&temp_store_path));
+    assert!(!segment_meta.list_files().contains(&temp_store_path));
     assert!(!index.directory().exists(&temp_store_path)?);
 
     let mut index_writer = index.writer_for_tests()?;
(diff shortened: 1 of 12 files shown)


Mallets:
@codex review

src/indexer/single_segment_index_writer.rs
```diff
@@ -38,9 +38,29 @@ impl<D: Document> SingleSegmentIndexWriter<D> {
...
+
+    pub fn finalize_with_doc_id_mapping(self, mapping: &DocIdMapping) -> crate::Result<Index> {
+        let Self {
+            segment,
+            segment_writer,
+            ..
+        } = self;
+        let max_doc = segment_writer.max_doc();
+        segment_writer.finalize_with_doc_id_mapping(mapping)?;
+        Self::finalize_inner(segment, max_doc)
```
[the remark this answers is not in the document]
Mallets: Fixed in b7234c153. `finalize_with_doc_id_mapping` now clears the runtime-only `manual_doc_id_mapping` setting before saving metadata and before returning the in-memory `Index`, and the regression test verifies the returned index can be reused for a normal write.

src/indexer/single_segment_index_writer.rs
```diff
@@ -38,9 +38,29 @@ impl<D: Document> SingleSegmentIndexWriter<D> {
...
+    }
+
+    pub fn finalize_with_doc_id_mapping(self, mapping: &DocIdMapping) -> crate::Result<Index> {
+        let Self {
+            segment,
+            segment_writer,
+            ..
+        } = self;
+        let max_doc = segment_writer.max_doc();
+        segment_writer.finalize_with_doc_id_mapping(mapping)?;
```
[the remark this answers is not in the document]
Mallets: Fixed in b7234c153. The combination is now rejected both at `IndexBuilder::validate` and again in `SegmentWriter::finalize_with_doc_id_mapping` to cover post-builder `settings_mut()` changes. Added regression coverage for both paths.

src/indexer/segment_serializer.rs
```diff
@@ -25,17 +25,18 @@ impl SegmentSerializer {
         // If the segment is going to be sorted, we stream the docs first to a temporary file.
         // In the merge case this is not necessary because we can kmerge the already sorted
         // segments
-        let remapping_required = segment.index().settings().sort_by_field.is_some() && !is_in_merge;
         let settings = segment.index().settings().clone();
+        let remapping_required =
+            (settings.sort_by_field.is_some() || settings.manual_doc_id_mapping) && !is_in_merge;
```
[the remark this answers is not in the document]
Mallets: Fixed in b7234c153. `SingleSegmentIndexWriter` now untracks `SegmentComponent::TempStore` for any remapped single-segment finalize path, including manual mappings and `sort_by_field`, mirroring the normal `IndexWriter` metadata cleanup boundary. Added regression coverage for both cases.

Mallets:
@codex review

src/indexer/single_segment_index_writer.rs
```diff
@@ -38,19 +38,53 @@ impl<D: Document> SingleSegmentIndexWriter<D> {
...
+
+    fn finalize_inner(
+        segment: Segment,
+        max_doc: u32,
+        remapping_required: bool,
+        index_settings: IndexSettings,
+    ) -> crate::Result<Index> {
+        let segment: Segment = segment.with_max_doc(max_doc);
+        if remapping_required {
+            segment.meta().untrack_temp_docstore();
```
PSeitz: I think you just want to directly delete the temp docstore, but I'm not sure what effect it has on garbage collection.

Mallets: An alternative is to just `untrack` it  and call the garbage collection straight away, but even there I'm not sure on the impact of forcing the `garbage_collection` directly and whether _garbage collection_ does something more than simply deleting the file.

Mallets: https://github.com/quickwit-oss/tantivy/pull/2978/changes/6f7cc10a9b32adc7554d6f739458a22892fc120e calls `garbage_collect` on `SingleSegmentIndexWriter::finalize`.

src/core/tests.rs
```diff
@@ -300,6 +300,140 @@ fn test_single_segment_index_writer() -> crate::Result<()> {
     Ok(())
 }
 
+#[test]
```
PSeitz: These tests should be moved to the doc mapping code

src/index/index_meta.rs
```diff
@@ -547,6 +555,18 @@ mod tests {
                 serde_json::from_value(index_settings_json).unwrap();
             assert_eq!(index_settings_deser, index_settings);
         }
+        {
+            index_settings.manual_doc_id_mapping = true;
```
PSeitz, line 560:
```suggestion
            // manual_doc_id_mapping should not be persistet 
            index_settings.manual_doc_id_mapping = true;
```

src/core/tests.rs
Mallets: Done in https://github.com/quickwit-oss/tantivy/pull/2978/commits/56bd23bafecc4acd4a12614c7c2ffe06d43a8ae3

src/index/index_meta.rs
Mallets: Done in https://github.com/quickwit-oss/tantivy/pull/2978/commits/0a071c55f7c24e708c1efee7ce70e65fb7093293

src/indexer/doc_id_mapping.rs
```diff
@@ -71,7 +70,34 @@ pub struct DocIdMapping {
 }
 
 impl DocIdMapping {
-    pub fn from_new_id_to_old_id(new_doc_id_to_old: Vec<DocId>) -> Self {
+    /// Creates a `DocIdMapping` from a mapping of new doc ids to old doc ids, with permutation
+    /// validation. The mapping is validated by checking that every old doc id appears exactly
+    /// once in the mapping. I.e., doc ids must be consecutive from `0` to
+    /// `new_doc_id_to_old.len() - 1`, inclusive.
+    pub fn new_permutation(new_doc_id_to_old: Vec<DocId>) -> crate::Result<Self> {
```
PSeitz, line 81: This is the same as `from_new_id_to_old_id`, but with validation to be a permutation

A shared impl with a parameter with `is_permutation` or an enum would be better

src/indexer/segment_writer.rs
```diff
@@ -136,10 +136,21 @@ impl SegmentWriter {
...
+        // Ensure the segment writer was created in remap mode so the docstore can be reordered.
+        if self
+            .segment_serializer
+            .segment()
+            .index()
+            .settings()
+            .manual_doc_id_mapping
+        {
+            return Err(TantivyError::InvalidArgument(
+                "IndexSettings::manual_doc_id_mapping must be set to false".to_string(),
```
PSeitz:
```suggestion
                "IndexSettings::manual_doc_id_mapping must be set to false. With manual_doc_id_mapping, you need to call finalize_with_doc_id_mapping".to_string(),
```

src/indexer/doc_id_mapping.rs
Mallets: Done in https://github.com/quickwit-oss/tantivy/pull/2978/changes/2c967e5d207fc928eb50dc9b8abec6f88ac6b853

src/indexer/doc_id_mapping.rs
```diff
@@ -81,43 +111,50 @@ impl DocIdMapping {
             .unwrap_or(0);
         let mut old_doc_id_to_new = vec![0; old_max_doc as usize];
```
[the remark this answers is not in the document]
Mallets: The behavior of previous `from_new_id_to_old_id` is preserved in this case. If the user passes `u32::MAX` I would treat it as a business logic issue.

src/indexer/single_segment_index_writer.rs
```diff
@@ -38,19 +38,65 @@ impl<D: Document> SingleSegmentIndexWriter<D> {
...
+            segment,
+            segment_writer,
+            ..
+        } = self;
+        let max_doc = segment_writer.max_doc();
+        segment_writer.finalize_with_doc_id_mapping(mapping)?;
+        let mut index_settings = segment.index().settings().clone();
+        index_settings.manual_doc_id_mapping = false;
+        let mut index = Self::finalize_inner(segment, max_doc, true, index_settings)?;
+        index.settings_mut().manual_doc_id_mapping = false;
```
PSeitz: why not just set it before on the original, instead of cloning and setting it twice

src/indexer/single_segment_index_writer.rs
```diff
@@ -38,19 +38,65 @@ impl<D: Document> SingleSegmentIndexWriter<D> {
...
+        Ok(index)
+    }
+
+    fn finalize_inner(
+        segment: Segment,
+        max_doc: u32,
+        remapping_required: bool,
```
PSeitz:
```suggestion
        did_remapping: bool,
```

I think the remapping is already done here, required suggests otherwise

src/indexer/doc_id_mapping.rs
```diff
@@ -71,7 +74,34 @@ pub struct DocIdMapping {
...
+
+    /// Creates a `DocIdMapping` from a mapping of new doc ids to old doc ids.
+    pub(crate) fn from_new_id_to_old_id(new_doc_id_to_old: Vec<DocId>) -> Self {
+        Self::from_new_id_to_old_id_inner(new_doc_id_to_old, |_| Ok::<(), Infallible>(()))
+            .unwrap_infallible()
+    }
+
+    fn from_new_id_to_old_id_inner<E: Error>(
+        new_doc_id_to_old: Vec<DocId>,
+        mut check: impl FnMut(DocId) -> Result<(), E>,
```
PSeitz:
```suggestion
        mut validate: impl FnMut(DocId) -> Result<(), E>,
```

src/indexer/doc_id_mapping.rs
```diff
@@ -71,7 +74,34 @@ pub struct DocIdMapping {
...
+    /// Creates a `DocIdMapping` from a mapping of new doc ids to old doc ids, with permutation
+    /// validation. The mapping is validated by checking that every old doc id appears exactly
+    /// once in the mapping. I.e., doc ids must be consecutive from `0` to
+    /// `new_doc_id_to_old.len() - 1`, inclusive.
+    pub fn new_permutation(new_doc_id_to_old: Vec<DocId>) -> crate::Result<Self> {
+        let max_doc = new_doc_id_to_old.len() as DocId;
+        let mut seen_doc_ids: BitSet = BitSet::with_max_value(max_doc);
+
+        Self::from_new_id_to_old_id_inner(new_doc_id_to_old, |old_doc_id| {
+            if old_doc_id >= max_doc || !seen_doc_ids.insert(old_doc_id) {
```
PSeitz, line 86: I think a closure is a bit overkill here instead of moving the check to the inner function behind a bool flag

PSeitz (approved):

src/indexer/doc_id_mapping.rs
Mallets: It's not an overkill in the sense that `from_new_id_to_old_id_inner` returns `Self` while `new_permutation` returns a `Result`. The closure with the `Infallible` error ensures that the `unwrap` will never panic and it's elided by the compiler.

src/indexer/doc_id_mapping.rs
Mallets: Done in https://github.com/quickwit-oss/tantivy/pull/2978/commits/2554072e8f48817003b73eba97fa96957e930da8

src/indexer/single_segment_index_writer.rs
Mallets: Done in https://github.com/quickwit-oss/tantivy/pull/2978/commits/2554072e8f48817003b73eba97fa96957e930da8

src/indexer/single_segment_index_writer.rs
Mallets: Done in https://github.com/quickwit-oss/tantivy/pull/2978/commits/2554072e8f48817003b73eba97fa96957e930da8

PR #2981, opened 2026-06-28
Fix union performance regression

PSeitz:
search-benchmark-game shows TOP_100_COUNT regression on union queries 

The regression came from allowing single-term boolean unions to become TermUnion for Block-WAND. https://github.com/quickwit-oss/tantivy/pull/2915
When such a scorer is used as the optional side of RequiredOptionalScorer, boxing converted the lone term into BufferedUnionScorer.

Also it seems https://github.com/quickwit-oss/tantivy/pull/2865 had a side-effect, that caused `refill` to be inlined.
With a `inline(never)` the performance regression is fully fixed.


stuhood (approved):
Thank you!

PR #2990, opened 2026-07-06
add method to prune intermediate agg results

trinity-1686a:


src/aggregation/intermediate_agg_result.rs
```diff
@@ -821,6 +958,102 @@ impl IntermediateTermBucketResult {
...
+        use_segment_size: bool,
+    ) -> crate::Result<()> {
+        let req_internal = TermsAggregationInternal::from_req(req);
+        let size = if use_segment_size {
+            req_internal.segment_size as usize
+        } else {
+            req_internal.size as usize
+        };
+
+        if !use_segment_size {
```
PSeitz-dd: I don't understand why we do this only when we not using `segment_size`

trinity-1686a: that's something that was asked for by codex [here](https://github.com/quickwit-oss/tantivy/pull/2990#discussion_r3531231511), and i think the reasoning makes sense
(initially this wasn't done at all, but doing `prune_intermediate_results(segment_size=false).finalize()` would have changed returned results, including potentially returning less buckets than requested)

PSeitz (approved):

PR #2993, opened 2026-07-08
feat: Extensible segment components via plugin trait

rebasedming:
## Motivation

Today every segment component — postings, fast fields, field norms, store — is hardcoded into several places. Adding a new per-segment data structure means forking Tantivy and editing each of those sites.

This PR introduces a `SegmentPlugin` trait that lets a custom component participate in the full segment lifecycle — write, serialize, merge, garbage collection, space usage — through the same interface the built-ins use, without touching Tantivy internals. The four built-in components are themselves reimplemented as plugins.

We (ParadeDB) plan on using this trait for 1) additional segment metadata for partitioning 2) custom vector index.

## Plugin Trait

Two traits. The first is the `SegmentPlugin` factory:

```rust
pub trait SegmentPlugin: Send + Sync + 'static {
    /// File extensions this component owns, e.g. ["idx", "pos", "term"] for postings.
    fn extensions(&self) -> &[&str];

    /// Create a writer for the indexing path.
    fn create_writer(&self, ctx: &PluginWriterContext) -> crate::Result<Box<dyn PluginWriter>>;

    /// Merge this component across several source segments into the target segment.
    fn merge(&self, ctx: PluginMergeContext) -> crate::Result<()>;

    /// Report on-disk space usage, keyed by component name. Has a default impl.
    fn space_usage(&self, reader: &SegmentReader)
        -> crate::Result<BTreeMap<String, ComponentSpaceUsage>>;
}
```

A `SegmentPlugin` owns one or more file extensions and knows how to (a) build a writer for the indexing path and (b) merge itself across segments.

The second trait is the segment writer:

```rust
pub trait PluginWriter: Send + Any {
    /// Called once per document, in doc-id order, for every plugin writer.
    fn add_document(&mut self, doc_id: DocId, doc: &TantivyDocument, schema: &Schema)
        -> crate::Result<()> { Ok(()) }

    /// Serialize accumulated data to segment files (honoring an optional doc-id remap).
    fn serialize(&mut self, segment: &Segment, doc_id_map: Option<&DocIdMapping>) -> crate::Result<()>;

    fn close(self: Box<Self>) -> crate::Result<()>;
    fn mem_usage(&self) -> usize;

    fn as_any(&self) -> &dyn Any;       // downcast support, Rust 1.86
    fn as_any_mut(&mut self) -> &mut dyn Any;
}
```

The write path no longer has any by-name wiring: `SegmentWriter` hands every document to every plugin writer's `add_document`, and `finalize()` calls `serialize` then close on each. 

## Key Design Decisions

1. The index — not the segment — owns the plugin set. The set of custom plugins is recorded once, at index creation, in `IndexMeta`. `#[serde(default)]` makes this backward compatible.

2. Plugins are registered, like tokenizers — and re-registration is enforced fail-closed. Plugins are not serialized; they're re-attached on every `Index::open` via `register_plugin`, exactly like custom tokenizers. To prevent consumers from accidentally forgetting to register a plugin, we validate the registered plugin set against the persisted set when the index is first used for a write/merge/GC operation.

3. Registration order is the write/merge order. Built-ins come first (field norms → postings → fast fields → store), then custom plugins.

4. The read side needs no plugin hook. Custom component data is read back through the existing public surface `SegmentReader::open_read`.

5. Backwards compatibility — behavior for existing indexes is unchanged.


src/fastfield/plugin.rs
```diff
@@ -0,0 +1,193 @@
...
+        Ok(BTreeMap::from([(
+            FAST_FIELDS.to_string(),
+            ComponentSpaceUsage::PerField(usage),
+        )]))
+    }
+}
+
+pub struct FastFieldsPluginWriter {
+    /// The inner writer is wrapped in an `Option` because `FastFieldsWriter::serialize`
+    /// takes `self` by value. We `.take()` it during serialize.
```
fulmicoton: this is useless here. It is ok to change the prototype of serialize to avoid this indirection.

src/fastfield/plugin.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub writer: Option<FastFieldsWriter>,
+    fast_field_write: Option<WritePtr>,
```
fulmicoton: there are no reason for fast_field_write to be an Option as far as I can tell.

src/index/index.rs
```diff
@@ -275,24 +341,99 @@ impl IndexBuilder {
...
         Ok(index)
     }
 }
 
+/// An [`Index`] whose registered plugin set has been validated against the index's
+/// recorded plugin set: no two plugins claim the same extension, none claim a reserved
+/// one, and the registered custom plugins exactly match the extensions the index was
+/// created with. The plugin set is part of the index's structure, like the schema, and
+/// is fixed at creation.
+pub(crate) struct PluginCheckedIndex(Index);
```
fulmicoton: a bit too much overengineering maybe?

a `validate_plugins()`method would do the job I think

src/indexer/segment_updater.rs
```diff
@@ -139,6 +138,11 @@ fn merge(
 /// This function does NOT check or take the `IndexWriter` is running. It is not
 /// meant to work if you have an `IndexWriter` running for the origin indices, or
 /// the destination `Index`.
+///
+/// It is the caller's responsibility to ensure every custom
```
fulmicoton: Is this a feature you use? If not, shouldn't we run a rapid check? ( checksum is ok)

fulmicoton: It is not entirely clear what happens if all indices being merge have different list of plugins being registered.

Is it the first one that matters? the intersection? the union?

fulmicoton: Look at the index merger code, it seems like the first one is the one that matters.

src/indexer/segment_writer.rs
```diff
@@ -89,9 +89,15 @@ impl SegmentWriter {
...
-        let tokenizer_manager_fast_field = segment.index().fast_field_tokenizer().clone();
         let table_size = compute_initial_table_size(memory_budget_in_bytes)?;
-        let segment_serializer = SegmentSerializer::for_segment(segment, false)?;
+        let ctx = PluginWriterContext {
+            segment: &segment,
+        };
+        let plugin_writers = segment
+            .index()
+            .all_plugins()
+            .map(|p| p.create_writer(&ctx))
```
fulmicoton:
```suggestion
            .map(|plugin| plugin.create_writer(&ctx))
```

generally speaking, let's avoid 1 letter variable names.

cmp(a,b) -> cmp(lhs, rhs)

etc.

Of course indices are exception to that rule.

src/indexer/segment_writer.rs
```diff
@@ -243,8 +280,11 @@ impl SegmentWriter {
                         );
                     }
                     if field_entry.has_fieldnorms() {
-                        self.fieldnorms_writer
-                            .record(doc_id, field, indexing_position.num_tokens);
+                        self.plugin_writer_mut::<FieldNormsPluginWriter>().record(
```
fulmicoton: plugin_writer_mut does a linear search.

If I get things correctly, this is only used for the FieldNormsPluginWriter.
I think we need to either refactor things more heavily (e.g. group fieldnorms, termdictionary, postings and positions together as an inverted Index plugin), or at least cache this.

src/indexer/segment_writer.rs
```diff
@@ -378,11 +425,14 @@ impl SegmentWriter {
...
+        let document = document.into_tantivy_document();
+        let doc_id = self.max_doc;
         self.doc_opstamps.push(opstamp);
-        self.fast_field_writers.add_document(&document)?;
         self.index_document(&document)?;
-        let doc_writer = self.segment_serializer.get_store_writer();
-        doc_writer.store(&document, &self.schema)?;
+        let schema = self.schema.clone();
+        for writer in &mut self.plugin_writers {
+            writer.add_document(doc_id, &document, &schema)?;
```
fulmicoton: Ok so if I understand correctly, all inverted index related stuff are exceptions.
Because they interact with each other, you kept them outside of the loop.

fulmicoton (changes requested):
Main comment:

You went for a 1 file per plugin design, which forced you to isolate indexing aside.
I think we could have one InvertedIndex plugin which covers several files.

What do you think?

fulmicoton (changes requested):
The inverted index stuff, due to their interactions, end up being pulled apart.
Should we group them (termdict, postings, positions, fieldnorms) as a single plugin maybe?

src/plugin.rs
```diff
@@ -0,0 +1,445 @@
...
+/// during finalization.
+pub trait PluginWriter: Send + Any {
+    /// Records a single document during indexing.
+    ///
+    /// Called once per document added to the segment, in doc-id order, for every plugin
+    /// writer. The default is a no-op; override it to accumulate per-document state.
+    fn add_document(
+        &mut self,
+        _doc_id: DocId,
+        _doc: &TantivyDocument,
```
fulmicoton: You removed the point of having a Document trait.

I was not much involved with its introduction. It was mainly lead by @ChillFish8 . I can check if we rely on it on Quickwit side. Can you check with @ChillFish8 if this is still in use in his projects?
If not, we need to also assess removing the Document trait entirely.

fulmicoton (changes requested):
This PR (silently) removes the purpose of the Document trait. This might be ok (well not mentioning in the PR description is not ok) but we need to investigate.

src/plugin.rs
ChillFish8: I use a custom built system now, however, I believe some other users make use of this trait going of some previous discussions in the discord.

I think I would lean towards a solution which can still support custom document types, although I think current design of that custom doc system isn't perfect and might want some changes to make it more ergonomic.

src/indexer/segment_updater.rs
rebasedming: We don't merge across indices -- fixed to error if the indexes do not have consistent plugin sets

src/plugin.rs
rebasedming: Good point. We don't make use of custom documents so I forgot to flag this when putting up this PR.

Given the current pull-based `Document` approach, we'd need a pretty significant refactor to make this generic for all plugins. Perhaps a middle ground could be: bring back the generic `add_document` for the builtin plugins (inverted index, fast field, store), but make it so that `TantivyDocument`s are materialized for custom plugins (and we document that as a perf limitation).

Thoughts?

rebasedming: Okay, let me know what you think of this...

Builtin plugins index the documents generically (like what they do today). Custom plugins get an erased `ErasedDocument` which has an `ErasedValue` that mirrors `ReferenceValue`, and can be downcast to a custom `Document`.

ChillFish8: Have you checked how heavy the performance impact of that change is? The additional scan of the document fields plus the heavy boxing concerns me.

rebasedming: Yes, but two clarifications here: 

1. Only custom plugins receive the erased document — built-ins index it generically. 
2. If the plugin knows its concrete document type at compile time, say `TantivyDocument`, it's just a downcast with no walk/allocations. 

The walk is only needed when a plugin must handle document types it wasn't compiled against. In my local microbenchmarks the cost was ~6ns/field, but the alternative for supporting a custom document type there would be materializing it into a `TantivyDocument`, which would be a full copy.

src/indexer/single_segment_index_writer.rs
```diff
@@ -42,8 +42,10 @@ impl<D: Document> SingleSegmentIndexWriter<D> {
         self.segment_writer.finalize()?;
         let segment: Segment = self.segment.with_max_doc(max_doc);
         let index = segment.index();
+        let previous_meta = index.load_metas()?;
         let index_meta = IndexMeta {
             index_settings: index.settings().clone(),
+            persisted_custom_extensions: previous_meta.persisted_custom_extensions.clone(),
```
fulmicoton: I have similar comment as Codex here, but maybe human words could make more sense.

SingleSegmentIndexWriter is a weird API that makes it possible to build a segment from your running thread.
It is meant to help you build ONE INDEX WITH ONE SEGMENT.
This is used in quickwit.

SingleSegmentIndexWriter::new should not have been public to begin with. It gives you the impression it is meant to overwrite an existing index.

I suggest we make SingleSegmentIndexWriter::new  crate private, leaving user the ability to create a SingleSegmentIndexWriter via IndexBuilder::single_segment_index_writer.

Calling load_metas should be removed.

The list of plugins is the one defined in the IndexBuilder/Index used to create the SingleSegmentIndexWriter

src/postings/indexing_context.rs
```diff
@@ -1,6 +1,28 @@
...
 
 use crate::indexer::path_to_unordered_id::PathToUnorderedId;
+use crate::postings::compute_table_memory_size;
+
+/// Computes the initial size of the term hash table.
+///
+/// Returns the recommended initial table size as a power of 2.
+///
+/// Note this is a very dumb way to compute log2, but it is easier to proofread that way.
+pub(crate) fn compute_initial_table_size(per_thread_memory_budget: usize) -> crate::Result<usize> {
```
fulmicoton: is this the right place? I only see it called in inverted_index_plugin and in unit test in segment_writer.

the unit test, and compute_initial_table_size should just be moved to inverted_index_plugin.
It can also be made private.

src/store/plugin.rs
```diff
@@ -0,0 +1,237 @@
...
+                .close()
+                .map_err(|e| crate::TantivyError::InternalError(e.to_string()))?;
+        }
+        Ok(())
+    }
+
+    fn mem_usage(&self) -> usize {
+        self.store_writer
+            .as_ref()
+            .map_or(0, |writer| writer.mem_usage())
```
fulmicoton: nitpick,
advance combinator like Option::map_or do not help readability.

.map(_).unwrap_of or even 

```
if let Some(writer) = self.store_writer {
} else {
}
```

or 

```
let Some(writer) = self.store_writer {
} else {
    return 0;
};
writer.mem_usage();
```

does the job better imo

src/indexer/segment_writer.rs
```diff
@@ -378,11 +138,23 @@ impl SegmentWriter {
...
+
+        // Built-in writers index the document generically — zero-copy for any `Document` type.
+        self.inverted_index.index_document(doc_id, &document)?;
+        self.fast_fields.index_document(&document)?;
+        self.store_writer.store(&document, &self.schema)?;
+
+        // Custom plugins are trait objects, so they take the document type-erased. Leaves stay
+        // borrowed; a plugin that knows the concrete type recovers it for free via `as_any`.
+        if !self.custom_plugins.is_empty() {
+            let erased: &dyn ErasedDocument = &document;
```
fulmicoton, line 192: ok... So if I get that right, you kept the concrete type for builtin plugins, but use a dynamic dispatch for custom_plugins. That works for me.

fulmicoton:
@rebasedming Have you checked there are no regression for this?

rebasedming:
> @rebasedming Have you checked there are no regression for this?

Yep, here are the results of the `index-bench` and `merge_segments` benchmarks:


Benchmark | main | PR | Δ
-- | -- | -- | --
index-hdfs/only-indexed-with-commit | 188.8 ms | 190.1 ms | +0.7%
index-hdfs/only-fast-with-commit | 68.3 ms | 70.4 ms | +3.0%
index-hdfs/dynamic-with-commit | 270.0 ms | 265.7 ms | −1.6%
index-wiki (no-commit / with-commit) | 8.62 ms | 8.68 ms | +0.7% / −0.4%


Config | main | PR | Δ
-- | -- | -- | --
8 seg × 50k docs | 81.39 ms | 81.45 ms | +0.1%
16 seg × 50k | 162.9 ms | 162.1 ms | −0.5%
16 seg × 100k | 324.8 ms | 326.0 ms | +0.4%
8 seg × 50k, vocab 3.2M | 1.228 s | 1.259 s | +2.6%

Additionally, already shipped these changes into ParadeDB and it's held up against our tests/real world use.

fulmicoton (approved):

PR #2995, opened 2026-07-10
Convert BMW implementations into publicly-available PruningScorers

### LINKED ISSUE #2390: Implementing Block WAND optimization for more queries

I only have one particular query in my system: a `BooleanQuery` of `TermQuery`s for the should, must, and must not clauses where only the should clauses contribute to scoring via BM25. 

I noticed that a `BooleanQuery` of `TermQuery`s implements the Block WAND optimization within `BooleanWeight::for_each_pruning` when it only has should clauses, but it falls back to the generic `for_each_pruning_scorer` when there are any must or must not clauses.

I have a prototype for our usage of Tantivy within Convex that is able to utilize Block WAND even with should or must not clauses, and I'd be curious if you would be interested in us upstreaming it!

The high level idea is to move the `for_each_pruning` API from the weights to the scorers. First, we add an optional method to the `Scorer` trait:
```rust
pub trait Scorer: DocSet {
    fn adjust_threshold(&mut self, threshold: Score);
}
```
This method mutates the document set, filtering out documents that have score less than or equal to `threshold` past the current position. The caller must call this method with monotonically increasing thresholds over the lifetime of the scorer.

Then, the protocol between the collector and the scorer moves from being callback-based (like `for_each_pruning_scorer` today) to alternating between advancing the document set, updating the collector, and updating the thresholds.
```rust
impl Collector for TopDocs {
    fn collect_segment(...) -> crate::Result<...> {
        ...
        let mut top_n = TopNComputer::new(heap_len);
        loop {
            let doc = scorer.doc();
            if doc == TERMINATED {
                break;
            }
            top_n.push(scorer.score());

            let threshold = top_n.threshold.unwrap_or(Score::MIN);
            scorer.adjust_threshold(threshold);
        }
    }
}          
```
Next, we implement a specialized `WeakAnd` scorer that takes most of the logic from today's `block_wand.rs`. 
```rust
pub struct WeakAnd {
    scorers: Vec<TermScorerWithMaxScore>,
    threshold: Score,

    // Invariant: If `scorers` is nonempty, this is present with the current
    // match.    
    current_match: Option<WeakAndMatch>,
}

struct WeakAndMatch {
    len: usize,
    doc: DocId,
    score: Score,
}
```
(Full source: https://gist.github.com/sujayakar/c742f07dd99089f4c529a242c3b5ac9c)

The idea here is that we run the Block WAND algorithm for generating pivot candidates in both `DocSet::advance` and `DocSet::seek`. So, a layer above can call `WeakAnd::seek` on some `doc: DocId`, and it will then resume where it left off, finding the next pivot after `doc` that's better than the current `threshold`.

Then, this approach composes well with `query::intersection::Intersection`: If we're computing the intersection of `WeakAnd` with a single `TermScorer` (which we'll assume has a smaller `size_hint`), the intersection will...
1. Iterate to the next document on the `TermScorer` to get a `candidate`
2. Advance the `WeakAnd` to this `candidate`, returning `seek_doc`. This step potentially skips over many `DocId`s in the union if the intersection is more selective.
3. If `seek_doc > candidate`, advance the `TermScorer` and retry. In this condition, the `WeakAnd` having a high threshold potentially lets us skip over many `DocId`s in the intersection.
4. Otherwise, emit `candidate`.

A similar approach works for `Exclude`, which we use for the must not clauses.

I haven't tested this approach rigorously yet, and I also haven't evaluated how it performs compared to the current callback-based API. But if you are interested in this work, I can clean it up and start evaluating it. I'd also be curious how this `adjust_threshold` approach works with other score combiners too.

barbarj:
This PR closes #2390 by allowing external callers to choose if they want pruning or regular behavior from a scorer while keeping the applicability of the BMW optimization inside of the `Weight`. We (ParadeDB) recently used this change to apply BMW to certain joins where we use the scorer directly. (See [here](https://github.com/paradedb/paradedb/pull/5487)) 

This works by:
- Introduces a `PruningScorer` trait that extends a `Scorer` to support
setting a threshold.
- Adds a `pruning_scorer` method to the `Weight` trait, with a default
implementation returning a `BasicPruningScorer`.
- Moves the logic for all 3 (union, single, and intersection)
block-max-wand variants into `PruningScorer` implementations so that
they can be driven externally. I did this with as minimal changes as possible,. only changing what was necessary for external iteration. (mostly transferring loop state into a struct)
- Have `BooleanWeight::pruning_scorer` drive the BMW choice.
- Changes the visibility of the BMW scorers and their inputs to be
available to consumers of the library.

The internal code-paths to drive BMW are unchanged at the call to `for_each_pruning` and above.

Local runs of the relevant in-repo benchmarks show ~no performance differences.

### DIFF
--- a/src/indexer/delete_queue.rs
+++ b/src/indexer/delete_queue.rs
@@ -245,12 +245,21 @@ mod tests {
 
     use super::{DeleteOperation, DeleteQueue};
     use crate::index::SegmentReader;
-    use crate::query::{Explanation, Scorer, Weight};
+    use crate::query::{Explanation, PruningScorer, Scorer, Weight};
     use crate::{DocId, Score};
 
     struct DummyWeight;
     impl Weight for DummyWeight {
         fn scorer(&self, _reader: &SegmentReader, _boost: Score) -> crate::Result<Box<dyn Scorer>> {
+            Err(crate::TantivyError::InternalError("dummy impl".to_owned()))
+        }
+
+        fn pruning_scorer(
+            &self,
+            _reader: &SegmentReader,
+            _boost: Score,
+            _init_threshold: Score,
+        ) -> crate::Result<Box<dyn PruningScorer>> {
             Err(crate::TantivyError::InternalError("dummy impl".to_owned()))
         }
 
--- a/src/query/boolean_query/block_wand_intersection.rs
+++ b/src/query/boolean_query/block_wand_intersection.rs
@@ -1,7 +1,21 @@
+use crate::fieldnorm::FieldNormReader;
 use crate::postings::compression::COMPRESSION_BLOCK_SIZE;
+use crate::query::scorer::PruningScorer;
 use crate::query::term_query::TermScorer;
-use crate::query::Scorer;
+use crate::query::{Bm25Weight, Scorer};
 use crate::{DocId, DocSet, Score, TERMINATED};
+
+#[cfg(test)]
+pub(crate) fn block_wand_intersection(
+    scorers: Vec<TermScorer>,
+    threshold: Score,
+    callback: &mut dyn FnMut(DocId, Score) -> Score,
+) {
+    use crate::query::weight::for_each_pruning_scorer;
+
+    let mut scorer = BlockWandIntersectionScorer::new(scorers, threshold);
+    for_each_pruning_scorer(&mut scorer, callback);
+}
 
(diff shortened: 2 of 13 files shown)


barbarj:
~Converted to draft after realizing there's some tweaks to the BMW implementations that aren't in our fork. I'll get them into my code and then re-open this.~

Nevermind, I just misread something.

PR #3016, opened 2026-07-28
Optimize very low-cardinality terms aggregations

PSeitz-dd:
## Summary

For very low cardinality aggregations, e.g. field `status` with `INFO`, `WARN` etc., where most logs are INFO and single memory address gets hammered repeatedly. The CPU cannot reorder writes, so the pipeline is underutilized.
We create multiple LANES are merge them together at the end, so the CPU has multiple write locations.

## Benchmarks

Full aggregation benchmark comparison:

| Benchmark | Average change | Memory |
|---|---:|---:|
| `terms_7` | **-47.51%** | +0.02% |
| `terms_status_with_avg_sub_agg` | -2.96% | unchanged |
| `terms_status_with_terms_zipf_1000_sub_agg` | -1.71% | unchanged |
| `terms_zipf_1000_with_terms_status_sub_agg` | -0.97% | unchanged |
| `terms_status_with_histogram` | -2.67% | unchanged |
| `terms_status_with_date_histogram` | -0.75% | unchanged |
| `terms_status_with_date_histogram_hard_bounds` | +1.45% | unchanged |
| `terms_status_with_date_histogram_and_sibling_terms` | **-14.97%** | unchanged |

## Validation

- `cargo +nightly fmt --all -- --check`
- `cargo clippy -p tantivy --tests`
- `cargo test -p tantivy aggregation::` (248 passed)

### DIFF
--- a/benches/agg_bench.rs
+++ b/benches/agg_bench.rs
@@ -765,7 +765,9 @@ fn get_collector(agg_req: Aggregations) -> AggregationCollector {
 fn get_test_index_bench(cardinality: Cardinality) -> tantivy::Result<Index> {
     // Flag to reuse an on-disk index across runs. The generated data differs per cardinality, so
     // the path must include the cardinality — otherwise one cardinality reuses another's index.
-    let reuse_index = std::env::var("REUSE_AGG_BENCH_INDEX").is_ok();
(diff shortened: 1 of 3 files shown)


PSeitz:
Reopened as #3017 under @PSeitz.

PR #3018, opened 2026-07-29
Up to 2x faster fused term-histogram aggregation

PSeitz:
When executing the term -> datehistogram aggregation on splits, some splits data fall completely into one bucket. In these cases we can completely skipping fetching data from the histogram column (usually the timestamp field).

If we have only few buckets, we can precompute the boundaries in the `u64` space instead of converting them to f64 during collection.

## Summary
- split counters into lanes for small fused term × histogram grids to reduce write dependencies on hot cells
- skip histogram-column reads for single-bucket ranges and use precomputed encoded boundaries for ranges with up to eight buckets
- retain computed bucket resolution for larger ranges and hard bounds, while limiting lane replication to grids that fit the collector's counter and memory budgets
- add benchmark cases for one, four, and eight histogram buckets, plus coverage for the linear resolver

```
full
terms_status_with_histogram                           Memory: 139.9 KB (+0.32%)    Avg: 2.4363ms (-0.63%)     Median: 2.3999ms (-1.64%)     [2.3794ms .. 2.7160ms]
terms_status_with_date_histogram                      Memory: 137.5 KB (+0.61%)    Avg: 2.3556ms (-1.49%)     Median: 2.3362ms (-1.84%)     [2.3021ms .. 2.5198ms]
terms_status_with_date_histogram_single_bucket        Memory: 55.3 KB (-0.40%)     Avg: 1.2849ms (-56.67%)    Median: 1.2716ms (-57.01%)    [1.2619ms .. 1.3878ms]
terms_status_with_date_histogram_4_buckets            Memory: 57.4 KB (+1.52%)     Avg: 1.9660ms (-33.24%)    Median: 1.9563ms (-33.25%)    [1.9340ms .. 2.0592ms]
terms_status_with_date_histogram_8_buckets            Memory: 58.4 KB (+2.18%)     Avg: 2.0031ms (-31.20%)    Median: 1.9884ms (-31.13%)    [1.9664ms .. 2.1497ms]
terms_status_with_date_histogram_hard_bounds          Memory: 136.1 KB             Avg: 2.5072ms (-0.47%)     Median: 2.4902ms (-0.58%)     [2.4551ms .. 2.7265ms]
terms_status_with_date_histogram_and_sibling_terms    Memory: 138.3 KB (+0.75%)    Avg: 3.2796ms (-0.81%)     Median: 3.2806ms (-0.68%)     [3.2182ms .. 3.3387ms]
terms_zipf_1000_with_histogram                        Memory: 1.2 MB               Avg: 20.2501ms (-1.22%)    Median: 20.1891ms (-1.57%)    [20.0173ms .. 20.6549ms]
```

### DIFF
--- a/src/aggregation/bucket/term_agg/mod.rs
+++ b/src/aggregation/bucket/term_agg/mod.rs
@@ -337,19 +337,23 @@ impl TermsAggregationInternal {
 }
 
 /// Threshold for using dense `Vec` storage ([`VecTermBuckets`] or
-/// [`LowCardVecTermBuckets`]) for term buckets. Below this many term ordinals the `Vec`
+/// [`VecTermBucketsWithLanes`]) for term buckets. Below this many term ordinals the `Vec`
 /// (direct-indexed, no hashing/paging) beats the paged/hashed storages. It preallocates
 /// `num_terms` slots, so the ceiling also bounds its memory.
 ///
 /// TODO: Benchmark to validate the threshold
 pub const MAX_NUM_TERMS_FOR_VEC: u64 = 20_000;
 
-/// Threshold below which a terms agg without sub-aggregations uses [`LowCardVecTermBuckets`], whose
-/// independent count lanes improve throughput when many consecutive values address the same few
-/// counters. With sub-aggregations, this threshold selects [`LowCardSubAggBuffer`] (which buffers
-/// docs in a per-bucket `Vec`) while retaining scalar term counters. Both specialized layouts only
-/// pay off for a handful of buckets, so this is far lower than [`MAX_NUM_TERMS_FOR_VEC`].
-pub const MAX_NUM_TERMS_FOR_LOWCARD_SUBAGG: u64 = 100;
+/// Maximum number of logical counters for which replicating count lanes is worthwhile.
+const MAX_NUM_BUCKETS_FOR_COUNT_LANES: usize = 100;
+
+/// Threshold below which a terms agg without sub-aggregations uses
+/// [`VecTermBucketsWithLanes`], whose independent count lanes improve throughput when many
+/// consecutive values address the same few counters. With sub-aggregations, this threshold selects
+/// [`LowCardSubAggBuffer`] (which buffers docs in a per-bucket `Vec`) while retaining scalar term
+/// counters. Both specialized layouts only pay off for a handful of buckets, so this is far lower
+/// than [`MAX_NUM_TERMS_FOR_VEC`].
+pub const MAX_NUM_TERMS_FOR_LOWCARD_SUBAGG: u64 = MAX_NUM_BUCKETS_FOR_COUNT_LANES as u64;
 
 /// Threshold for using [`PagedTermMap`] storage; larger term-id spaces use
 /// [`HashMapTermBuckets`] instead.
@@ -845,16 +849,17 @@ const NUM_LOW_CARD_COUNT_LANES: usize = 8;
 /// A very-low-cardinality term bucket with split count storage and one shared sub-aggregation id.
 /// Only the counters are replicated: `bucket_id` remains unique per logical term.
 #[derive(Clone, Copy, Debug)]
-struct TermBucketWithLanes<B> {
-    count_lanes: [u32; NUM_LOW_CARD_COUNT_LANES],
+struct TermBucketWithLanes<B, const LANES: usize> {
+    count_lanes: [u32; LANES],
     bucket_id: B,
 }
 
-impl<B: BucketIdSlot> TermBucketWithLanes<B> {
+impl<B: BucketIdSlot, const LANES: usize> TermBucketWithLanes<B, LANES> {
     #[inline(always)]
     fn new(bucket_id_provider: &mut BucketIdProvider) -> Self {
+        assert!(LANES > 0, "a term bucket needs at least one count lane");
         Self {
-            count_lanes: [0; NUM_LOW_CARD_COUNT_LANES],
+            count_lanes: [0; LANES],
             bucket_id: B::assign(bucket_id_provider),
         }
     }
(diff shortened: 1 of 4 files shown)


src/aggregation/bucket/term_agg/mod.rs
```diff
@@ -886,9 +891,11 @@ impl<B: BucketIdSlot> TermAggregationMap for VecTermBucketsWithLanes<B> {
     }
 
     fn new(num_terms: u64, bucket_id_provider: &mut BucketIdProvider) -> Self {
-        let buckets = std::iter::repeat_with(|| TermBucketWithLanes::new(bucket_id_provider))
-            .take(num_terms as usize)
-            .collect();
+        assert!(LANES > 0, "a term map needs at least one count lane");
```
trinity-1686a: we can maybe even put this into a const block so it's a compilation error and not a runtime one

src/aggregation/bucket/term_agg/term_histogram.rs
```diff
@@ -19,19 +24,302 @@ use crate::aggregation::intermediate_agg_result::{
...
+    /// Bounded variant. Only the computed resolver is constructed for binding hard bounds; the
+    /// default remains useful for the statically unbounded resolvers.
+    fn collect_block_with_bounds<const LANES: usize>(
+        &mut self,
+        term_ids: impl Iterator<Item = u64>,
+        counts: &mut [[u32; LANES]],
+        _term_counts: &mut [[u32; LANES]],
+    ) {
+        self.collect_block(term_ids, counts);
+    }
```
trinity-1686a: doesn't this "returns" the wrong result if some docs are out of hard bounds? If so, it should probably not be a default implementation, so we can make the statement of "this resolver doesn't support hard bounds" explicit

src/aggregation/bucket/term_agg/term_histogram.rs
```diff
@@ -19,19 +24,302 @@ use crate::aggregation::intermediate_agg_result::{
...
+                hist_req_data,
+                base_pos,
+            );
+            boundaries[bucket - 1] = bucket_start;
+        }
+        Some(Self {
+            hist_block: ColumnBlockAccessor::default(),
+            next_count_lane: 0,
+            accessor: hist_req_data.accessor.clone(),
+            is_full: hist_req_data.accessor.get_cardinality().is_full(),
```
trinity-1686a: i think that's always true currently (that or we instantiate SingleBucketResolver in situations where we cannot)

src/aggregation/bucket/term_agg/term_histogram.rs
```diff
@@ -248,43 +521,150 @@ pub(super) fn maybe_build_collector(
...
+    terms_req_data: &TermsAggReqData,
+    hist_req_data: HistogramAggReqData,
+    num_terms: usize,
+    num_time_buckets: usize,
+    base_pos: i64,
+) -> crate::Result<Box<dyn SegmentAggregationCollector>> {
+    assert!(LANES > 0, "a fused grid needs at least one count lane");
+
     let all_docs_in_bounds =
         hist_req_data.bounds.min == f64::MIN && hist_req_data.bounds.max == f64::MAX;
```
trinity-1686a, line 592: this could be tightened by looking at the mix/max of the actual column

src/aggregation/bucket/term_agg/mod.rs
```diff
@@ -337,19 +337,23 @@ impl TermsAggregationInternal {
...
+/// Maximum number of logical counters for which replicating count lanes is worthwhile.
+const MAX_NUM_BUCKETS_FOR_COUNT_LANES: usize = 100;
+
+/// Threshold below which a terms agg without sub-aggregations uses
+/// [`VecTermBucketsWithLanes`], whose independent count lanes improve throughput when many
+/// consecutive values address the same few counters. With sub-aggregations, this threshold selects
+/// [`LowCardSubAggBuffer`] (which buffers docs in a per-bucket `Vec`) while retaining scalar term
+/// counters. Both specialized layouts only pay off for a handful of buckets, so this is far lower
+/// than [`MAX_NUM_TERMS_FOR_VEC`].
+pub const MAX_NUM_TERMS_FOR_LOWCARD_SUBAGG: u64 = MAX_NUM_BUCKETS_FOR_COUNT_LANES as u64;
```
trinity-1686a: are these really related, or do we happen to pick the same value for both?

src/aggregation/bucket/term_agg/term_histogram.rs
```diff
@@ -348,6 +728,51 @@ mod tests {
         Ok(())
     }
 
+    /// Four and eight histogram buckets take their corresponding fixed-size linear resolvers.
+    /// Negative values also exercise monotonic `f64` fast-field boundaries on both sides of zero.
+    #[test]
+    fn fused_term_histogram_linear_bucket_resolver() -> crate::Result<()> {
```
trinity-1686a, line 784: can we get a similar test for the one bucket case?

trinity-1686a (approved):

src/aggregation/bucket/term_agg/mod.rs
PSeitz: yes switched to const assert

src/aggregation/bucket/term_agg/term_histogram.rs
PSeitz: It's never constructed currently. I switched it to a panic instead of the incorrect forwarding.

src/aggregation/bucket/term_agg/term_histogram.rs
PSeitz-dd: yes, I'll change to a static value (which may help a bit with inlining)

src/aggregation/bucket/term_agg/term_histogram.rs
PSeitz-dd: added a single bucket test

src/aggregation/bucket/term_agg/term_histogram.rs
PSeitz: they are already normalized

src/aggregation/bucket/term_agg/mod.rs
PSeitz: no, I'll separate them

PR #3024, opened 2026-08-03
perf: Implement early termination for AutomatonWeight-based queries

### LINKED ISSUE #2974: `AutomatonWeight`-based queries have no early termination: `TopDocs::with_limit(N)` is `O(matches)`, not `O(N)`

`AutomatonWeight`-based queries (`RegexQuery`, `FuzzyQuery`, `TermSetQuery`, `RegexPhraseQuery`) have no early termination path. A `TopDocs::with_limit(N)` search over them does work proportional to the **total number of matches**, not to `N`.

Two facts in the current code (`main` @ `02e34508e`, v0.26.0) combine to cause this.

1. They are **constant scored**. `AutomatonWeight::scorer` returns `ConstScorer::new(BitSetDocSet, boost)` (`src/query/automaton_weight.rs`), and `ConstScorer::score()` returns the constant `self.score`. So there is no ranking that would require visiting every match. Any `N` hits are as good as the "top" `N`.
2. Yet `AutomatonWeight::scorer` **eagerly materializes every matching doc into a `BitSet` sized to `max_doc` before collection begins** (`src/query/automaton_weight.rs:88`). It streams all automaton-matching terms, OR's every posting into the bitset, then wraps it in `ConstScorer`. `AutomatonWeight` implements only `scorer` (no `for_each*` override), and the default `Weight::for_each`, `for_each_no_score`, and `for_each_pruning` all call `self.scorer()`, so every collection path pays the full O(matches) cost up front. There is also no stop signal in the `Collector` API (`collect(&mut self, doc, score)` returns `()`), so a collector cannot terminate early either.

By contrast, scored `TermQuery` already prunes within the API via block max WAND (`TermWeight::for_each_pruning` calls `block_wand_single_scorer`). So the *constant scored* automaton case, the one that is trivially early terminable, is the one left unoptimized.

**Reproduction**:

```rust
use std::time::Instant;
use tantivy::collector::{Count, TopDocs};
use tantivy::query::RegexQuery;
use tantivy::schema::{Schema, TEXT};
use tantivy::{doc, Index};

    let mut sb = Schema::builder();
    let f = sb.add_text_field("f", TEXT);
    let index = Index::create_in_ram(sb.build());

    let mut w = index.writer(256_000_000)?;
    for i in 0u32..1_000_000 {
        w.add_document(doc!(f => format!("k{i:07}")))?; // one unique token per doc
    }
    w.commit()?;
    let searcher = index.reader()?.searcher();

    // (1) constant score: every hit scores 1.0
    let q = RegexQuery::from_pattern("k000000.*", f)?;
    let scores: Vec<f32> = searcher
        .search(&q, &TopDocs::with_limit(10).order_by_score())?
        .iter().map(|(s, _)| *s).collect();
    println!("(1) top-10 regex scores: {scores:?}");

    // (2) fixed limit=10, growing match count, so time scales with MATCHES, not the limit
    for pat in ["k000000.*", "k00000.*", "k0000.*", "k000.*", "k00.*", "k.*"] {
        let q = RegexQuery::from_pattern(pat, f)?;
        let matches = searcher.search(&q, &Count)?;
        for _ in 0..3 { searcher.search(&q, &TopDocs::with_limit(10).order_by_score())?; }
        let iters = 20;
        let t = Instant::now();
        for _ in 0..iters { searcher.search(&q, &TopDocs::with_limit(10).order_by_score())?; }
        println!("{pat:<12} matches {matches:>8}  limit10 {:.3} ms",
                 t.elapsed().as_secs_f64() * 1000.0 / iters as f64);
    }
    Ok(())
}
```

Output (1M doc in-RAM index, `TopDocs::with_limit(10)`):

```
(1) top-10 regex scores: [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
pattern        matches  limit10 (ms)
k000000.*           10        0.014
k00000.*           100        0.028
k0000.*           1000        0.146
k000.*           10000        1.362
k00.*           100000       12.704
k.*            1000000      119.394
```

With the limit pinned at 10, the search time grows linearly with the match count from about 1k upward (ten times the matches gives roughly ten times the time). Returning 10 hits from a regex that matches 1M docs takes about 119 ms. With early termination it could cost about as little as when only 10 docs match (about 0.014 ms).

A way for `AutomatonWeight`-based queries to avoid O(all matches) work when only a small `limit` is requested. Likely directions.

1. A **lazy union scorer** for `AutomatonWeight` 
2. An **early termination** mechanism for constant score queries with a limit. 

alex09x:
Fixes #2974

This PR implements `for_each_pruning` for `AutomatonWeight` to allow early termination and constant score scaling. It evaluates matching terms lazily and maintains a local `BitSet` to avoid O(N) allocation when we only need the top K documents.

Benchmark verification for 1,000,000 matches limit-10 Regex query:
- Before: 431.629 ms
- After: 0.173 ms

### DIFF
--- a/src/query/automaton_weight.rs
+++ b/src/query/automaton_weight.rs
@@ -110,6 +110,43 @@ where
         Ok(Box::new(const_scorer))
     }
 
+    fn for_each_pruning(
+        &self,
+        mut threshold: Score,
+        reader: &SegmentReader,
+        callback: &mut dyn FnMut(DocId, Score) -> Score,
+    ) -> crate::Result<()> {
+        if threshold >= 1.0 {
+            return Ok(());
+        }
+        let max_doc = reader.max_doc();
+        let mut visited = BitSet::with_max_value(max_doc);
+        let inverted_index = reader.inverted_index(self.field)?;
+        let term_dict = inverted_index.terms();
+        let mut term_stream = self.automaton_stream(term_dict)?;
+        while term_stream.advance() {
+            let term_info = term_stream.value();
+            let mut block_segment_postings = inverted_index
+                .read_block_postings_from_terminfo(term_info, IndexRecordOption::Basic)?;
+            loop {
+                let docs = block_segment_postings.docs();
+                if docs.is_empty() {
+                    break;
+                }
+                for &doc in docs {
+                    if visited.insert(doc) {
+                        threshold = callback(doc, 1.0);
+                        if threshold >= 1.0 {
+                            return Ok(());
+                        }
+                    }
+                }
+                block_segment_postings.advance();
+            }
+        }
+        Ok(())
+    }
+
     fn explain(&self, reader: &SegmentReader, doc: DocId) -> crate::Result<Explanation> {
         let mut scorer = self.scorer(reader, 1.0)?;
         if scorer.seek(doc) == doc {
@@ -195,6 +232,7 @@ mod tests {
         assert_eq!(scorer.advance(), TERMINATED);
         Ok(())
     }
+
 
     #[test]
     fn test_automaton_weight_boost() -> crate::Result<()> {


alex09x:
@chatgpt-codex-connector Thanks for the catch. Since AutomatonWeight processes terms lexicographicaly, yielding docs immediately breaks the ascending doc ID requirement of `TopNHeap`. This means for constant score queries, we'd lose exact DocId tie-breaking and potentially drop lower DocIds if they appear in later terms.

I will revert this PR or close it, as early termination without doc-order merging is fundamentally incompatible with the current `TopDocs` tie-breaking contract.

fulmicoton:
@stuhood can you review? We don't use bm25 at all.

stuhood:
> @stuhood can you review? We don't use bm25 at all.

Will do. Sorry, it's been a busy week.

stuhood:
> I will revert this PR or close it, as early termination without doc-order merging is fundamentally incompatible with the current `TopDocs` tie-breaking contract.

Yea, unfortunately this is correct.

PR #3028, opened 2026-08-04
Replace unmaintained rust-stemmers with frostem

Xuanwo:
## Why

`rust-stemmers` is effectively unmaintained: it still ships old Snowball algorithms and does not track upstream ([CurrySoftware/rust-stemmers#27](https://github.com/CurrySoftware/rust-stemmers/issues/27)). Staying on that crate means Tantivy inherits known stemmer bugs and cannot pick up algorithm fixes without a manual dependency swap.

A concrete example is the Greek multibyte-suffix panic fixed in [lance-format/lance#8183](https://github.com/lance-format/lance/pull/8183): `rust-stemmers` 1.2.0 can keep stale UTF-8 byte offsets after shortening a Greek word and then panic while slicing. Current upstream Snowball (and therefore current generated stemmers) handle that input correctly.

## Change

Switch the optional `stemmer` feature from `rust-stemmers` to [`frostem`](https://github.com/Xuanwo/frostem).

`frostem` is a pre-built Snowball facade that:

- **Automatically tracks upstream** [`snowballstem/snowball`](https://github.com/snowballstem/snowball) `main`
- **Automatically regenerates and publishes** the latest algorithm sources (no Snowball compiler at consumer build time)
- Keeps a small, stable Rust API (`Algorithm` / `Stemmer`) close to what Tantivy already uses

Only the languages already exposed by `Language` are enabled as Cargo features, so the dependency stays scoped to the existing public surface.

Public Tantivy API (`Language`, `Stemmer` token filter, feature flag name) is unchanged. Stemming results for the languages Tantivy already supported match the existing tests; a regression test covers the Greek multibyte case above.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,26 @@ fs4 = { version = "0.13.1", optional = true }
 levenshtein_automata = "0.2.1"
 uuid = { version = "1.0.0", features = ["v4", "serde"] }
 crossbeam-channel = "0.5.4"
-rust-stemmers = { version = "1.2.0", optional = true }
+frostem = { version = "1.20260804.0", optional = true, default-features = false, features = [
+    "arabic",
+    "danish",
+    "dutch",
+    "english",
+    "finnish",
+    "french",
+    "german",
+    "greek",
+    "hungarian",
+    "italian",
+    "norwegian",
+    "portuguese",
+    "romanian",
+    "russian",
+    "spanish",
+    "swedish",
+    "tamil",
+    "turkish",
+] }
 downcast-rs = "2.0.1"
 bitpacking = { version = "0.9.3", default-features = false, features = [
     "bitpacker4x",
@@ -115,7 +134,7 @@ overflow-checks = true
 
 [features]
 default = ["mmap", "stopwords", "lz4-compression", "columnar-zstd-compression", "stemmer"]
-stemmer = ["rust-stemmers"]
+stemmer = ["frostem"]
 mmap = ["fs4", "tempfile", "memmap2"]
 stopwords = []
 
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -37,7 +37,26 @@ fs4 = { version = "0.13.1", optional = true }
 levenshtein_automata = "0.2.1"
 uuid = { version = "1.0.0", features = ["v4", "serde"] }
 crossbeam-channel = "0.5.4"
-rust-stemmers = { version = "1.2.0", optional = true }
+frostem = { version = "1.20260804.0", optional = true, default-features = false, features = [
```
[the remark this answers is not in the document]
Xuanwo: I'm open to change into `~1.20260804.0` to make sure the generated snowball logic is the same. Let me know what do you think, @PSeitz.

Btw, thank you for maintaining this lib!

fulmicoton, line 40: what the hell is this semver O_O

fulmicoton:
good to go, but because it is @Xuanwo ...
Note to other maintains @PSeitz @stuhood 

We need to be super careful about PR that go:
"hey your dep XXX is unmaintained. I forked it and maintain it there. Look I fixed one bug." 

If I was an attacker, I'd probably use an LLM to plant dormant supply chain attacks everywhere.
That would really look that kind of PR.

(@Xuanwo has a long/impressive track record contributing to OSS)

fulmicoton (approved):

Cargo.toml
Xuanwo: It's the latest commit from snowball.

My current design is to get frostem follow snowball's algo changes automanticly. So we don't get a stale dep like rust-stemmers anymore.

There is an explaintion:

image

PR #3030, opened 2026-08-04
fix(sstable): report the real term ordinal when an automaton prunes blocks

flow3d:
## Problem

`Streamer::term_ord()` derives the ordinal by counting `delta_reader.advance()` calls
from a seed, and that seed is only taken from a block's `first_ordinal` when the stream
has a key lower bound:

```rust
let first_term = match start_key {
    Bound::Included(key) | Bound::Excluded(key) => ...first_ordinal,
    Bound::Unbounded => 0,
};


fulmicoton:
@trinity-1686a can you have a look?

trinity-1686a (approved):
LGTM, thanks!

PR #3040, opened 2026-08-13
Faster term aggregation result merges

PSeitz:
## Summary                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                               
   Speed up terms aggregation result merging across by replacing repeated hash-map reconstruction with a flat bucket vector and a key-to-position index.                                                                       
                                                                                                                                                                                                                                                               
   This targets queries over indexes with many segments, where intermediate term buckets are folded repeatedly and the previous merge path rehashed the accumulated result on every merge. 

   ## Changes                                                                                                                                                                                                                                                  
   - Store intermediate term buckets in a flat `Vec` rather than an `FxHashMap`.                                                                                                                                                                               
   - Build a transient key-to-position map on the first merge and reuse it while folding subsequent segment results.                                                                                                                                                                                                                                                                                                                                                                                                                        
   - Add deterministic key-based tie-breaking when term buckets have equal document counts.                                                                                                                                                                    
   - Extend `agg_bench` with a 100-segment index 
   - Give aggregation keys a total order, using `f64::total_cmp` so ordering is consistent with bitwise equality for NaNs and signed zero.

   ## Benchmark             
```
100_segments
terms_7             Memory: 871.4 KB (-1.20%)    Avg: 13.9813ms (-11.11%)    Median: 13.8848ms (-10.01%)    [13.4502ms .. 16.5813ms]    
terms_zipfs_1000    Memory: 7.0 MB (-15.73%)     Avg: 18.7053ms (-11.35%)    Median: 18.5579ms (-10.23%)    [18.0121ms .. 20.3877ms]    
terms_150_000       Memory: 13.6 MB (-14.20%)    Avg: 40.1980ms (-24.13%)    Median: 40.1449ms (-24.45%)    [39.2021ms .. 41.7115ms]    
terms_all_unique    Memory: 31.1 MB (+30.08%)    Avg: 40.9190ms (-33.76%)    Median: 40.9192ms (-34.39%)    [40.3646ms .. 41.5972ms]    
```


## Alternatives considered
An alternative would be k-merge on sorted Vecs. Results are merged incrementally in a distributed scenario, which may make this alternative more expensive.

## Profile

<img width="1610" height="703" alt="Screenshot 2026-08-13 at 08 15 34" src="https://github.com/user-attachments/assets/6c13568c-b26b-4642-aa6a-aef710aaa56f" />

### DIFF
--- a/benches/agg_bench.rs
+++ b/benches/agg_bench.rs
@@ -8,6 +8,7 @@ use rand_distr::Distribution;
 use serde_json::json;
 use tantivy::aggregation::agg_req::Aggregations;
 use tantivy::aggregation::AggregationCollector;
+use tantivy::indexer::NoMergePolicy;
 use tantivy::query::{AllQuery, Query, TermQuery};
 use tantivy::schema::{IndexRecordOption, Schema, TextFieldIndexing, FAST, STRING};
 use tantivy::{doc, DateTime, Index, Term};
@@ -52,6 +53,46 @@ fn main() {
         runner.set_name(input_name);
         bench_agg(&mut runner, &index, execute_filtered);
     }
+
+    bench_many_segments();
+}
+
+fn bench_many_segments() {
+    let mut runner = BenchRunner::new();
+    runner.add_plugin(PeakMemAllocPlugin::new(GLOBAL));
+    let index =
+        get_test_index_bench_with_num_segments(Cardinality::Full, 100).unwrap();
+    let mut group = runner.new_group();
+    group.set_name("100_segments");
+    for (benchmark_name, agg_req) in [
+        (
+            "terms_7",
+            terms_on_field("text_few_terms_status"),
+        ),
+        (
+            "terms_zipfs_1000",
+            terms_on_field("text_1000_terms_zipf"),
+        ),
+        (
+            "terms_150_000",
+            terms_on_field("text_many_terms"),
+        ),
+        (
+            "terms_all_unique",
+            terms_on_field("text_all_unique_terms"),
+        ),
+    ] {
+        group.register_with_input(benchmark_name, &index, move |index| {
+            execute_agg(index, agg_req.clone())
+        });
+    }
+    group.run();
+}
+
+fn terms_on_field(field: &str) -> AggregationRequest {
+    json!({
+        "terms": { "terms": { "field": field, "size": 100 } }
+    })
 }
 
 fn bench_agg(runner: &mut BenchRunner, index: &Index, execute_filtered: AggregationExecutor) {
@@ -985,12 +1026,26 @@ fn get_collector(agg_req: Aggregations) -> AggregationCollector {
 }
 
 fn get_test_index_bench(cardinality: Cardinality) -> tantivy::Result<Index> {
-    // Flag to reuse an on-disk index across runs. The generated data differs per cardinality, so
-    // the path must include the cardinality — otherwise one cardinality reuses another's index.
+    get_test_index_bench_with_num_segments(cardinality, 1)
+}
+
+fn get_test_index_bench_with_num_segments(
+    cardinality: Cardinality,
+    num_segments: usize,
+) -> tantivy::Result<Index> {
+    assert!(num_segments > 0);
+    assert!(num_segments == 1 || cardinality == Cardinality::Full);
+    assert_eq!(1_000_000 % num_segments, 0);
+    // Flag to reuse an on-disk index across runs. The generated data differs per cardinality and
+    // segment count, so the path must include both.
     let reuse_index = std::env::var("REUSE_AGG_BENCH_INDEX")
         .map(|v| v == "true")
         .unwrap_or(true);
-    let index_dir = format!("agg_bench/{cardinality:?}");
+    let index_dir = if num_segments == 1 {
+        format!("agg_bench/{cardinality:?}")
+    } else {
+        format!("agg_bench/{cardinality:?}_{num_segments}_segments")
+    };
     let mut schema_builder = Schema::builder();
     let text_fieldtype = tantivy::schema::TextOptions::default()
         .set_indexing_options(
@@ -1020,7 +1075,7 @@ fn get_test_index_bench(cardinality: Cardinality) -> tantivy::Result<Index> {
 
     if reuse_index && std::path::Path::new(&index_dir).try_exists()? {
         let index = Index::open_in_dir(&index_dir)?;
-        if index.schema() == schema {
+        if index.schema() == schema && index.searchable_segment_ids()?.len() == num_segments {
             return Ok(index);
         }
         drop(index);
@@ -1068,7 +1123,12 @@ fn get_test_index_bench(cardinality: Cardinality) -> tantivy::Result<Index> {
     {
         let mut rng = StdRng::from_seed([1u8; 32]);
         let mut filter_rng = StdRng::from_seed([2u8; 32]);
-        let mut index_writer = index.writer_with_num_threads(1, 200_000_000)?;
+        let mut index_writer = index.writer_with_num_threads(1, 400_000_000)?;
+        if num_segments > 1 {
+            index_writer.set_merge_policy(Box::new(NoMergePolicy));
+        }
+        let docs_per_segment = 1_000_000 / num_segments;
+        let mut num_indexed_docs = 0usize;
         // 1% steady-state match rate, with random clusters averaging four documents.
         const MATCH_RATE: f64 = 0.01;
         const CLUSTER_END_PROBABILITY: f64 = 0.25;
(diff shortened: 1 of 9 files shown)


src/aggregation/bucket/term_agg/mod.rs
```diff
@@ -1460,22 +1473,22 @@ where
...
                     let val: NumericalValue = val.into();
 
                     match val.normalize() {
                         NumericalValue::U64(val) => {
-                            dict.insert(IntermediateKey::U64(val), intermediate_entry);
+                            out.push((IntermediateKey::U64(val), intermediate_entry));
                         }
                         NumericalValue::I64(val) => {
-                            dict.insert(IntermediateKey::I64(val), intermediate_entry);
+                            out.push((IntermediateKey::I64(val), intermediate_entry));
```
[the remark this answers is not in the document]
PSeitz: Fixed. We normalize f64 columns now, when converting to intermediate results

benches/agg_bench.rs
```diff
@@ -1083,6 +1143,10 @@ fn get_test_index_bench(cardinality: Cardinality) -> tantivy::Result<Index> {
             };
             document.add_text(filter_field, if filter_matches { "a" } else { "b" });
             index_writer.add_document(document)?;
+            num_indexed_docs += 1;
+            if num_segments > 1 && num_indexed_docs.is_multiple_of(docs_per_segment) {
```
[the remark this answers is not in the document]
PSeitz: bumped to 1.87

src/aggregation/bucket/term_agg/mod.rs
```diff
@@ -1337,6 +1340,14 @@ where
                 .map(|el| el.dictionary())
                 .unwrap_or_else(|| &fallback_dict);
 
+            // Collect into a map to dedup by key, then flush into `out`. Two cases need it: a real
+            // term may equal the `missing` placeholder, and the min_doc_count==0 fill must skip
+            // already-collected terms. A single-segment query returns this result directly (the
+            // cross-segment merge that would otherwise dedup never runs), so duplicate keys here
+            // would reach the final result unmerged.
+            let mut dict: FxHashMap<IntermediateKey, IntermediateTermBucketEntry> =
+                FxHashMap::with_capacity_and_hasher(entries.len(), Default::default());
```
trinity-1686a, line 1349: (possibly outside of scope for this PR)
i feel like we could get ride of this dict:
- extract missing value from `entries` and resolve it
- sort `entries` (this is already done today)
- use sorted_ords_to_term_cb to fill `out` directly
  - after this step, `out` is sorted
- iterate both `out` and the stream of min_doc_count=0 values (observation: they are both sorted). For any value found in min_doc_count=0 but not out, push it to the end of `out`. For extra speed, it should be possible to compare term_ids instead of strings
- binary search both sub-sequences in `out` for the missing value, if it's found, merge the buckets, if not, push to the end of `out`

src/aggregation/intermediate_agg_result.rs
```diff
@@ -877,17 +923,71 @@ pub struct IntermediateRangeBucketResult {
...
 }
 
-#[derive(Default, Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Default, Clone, Debug, Serialize, Deserialize)]
 /// Term aggregation including error counts
 pub struct IntermediateTermBucketResult {
-    pub(crate) entries: FxHashMap<IntermediateKey, IntermediateTermBucketEntry>,
+    /// Bucket entries as a flat `Vec`, in unspecified order. The final result is sorted by the
+    /// requested order (see `into_final_result`).
+    pub(crate) entries: Vec<(IntermediateKey, IntermediateTermBucketEntry)>,
```
trinity-1686a, line 931: there is no safe rollout strategy for a quickwit search pool with that change. Ideally we'd encode/decode that the same way we would a Map

src/aggregation/intermediate_agg_result.rs
```diff
@@ -877,17 +923,71 @@ pub struct IntermediateRangeBucketResult {
...
+    pub(crate) key_to_idx_in_entries: FxHashMap<IntermediateKey, u32>,
+}
+
+// `key_to_idx_in_entries` is a transient acceleration structure and is excluded from equality.
+impl PartialEq for IntermediateTermBucketResult {
+    fn eq(&self, other: &Self) -> bool {
+        self.entries == other.entries
+            && self.sum_other_doc_count == other.sum_other_doc_count
+            && self.doc_count_error_upper_bound == other.doc_count_error_upper_bound
+    }
```
trinity-1686a: i don't like this definition of equality, things we think of as equal may be recorded as not. I wonder if intermediate results should really implement PartialEq, it appears to be used in only 3 tests

src/aggregation/intermediate_agg_result.rs
```diff
@@ -936,18 +1036,29 @@ impl IntermediateTermBucketResult {
...
-                        right.key.partial_cmp(&left.key)
+                        right.key.cmp(&left.key)
                     }
-                    .expect("expected type string, which is always sortable")
                 });
             }
             OrderTarget::Count => {
+                // Tie-break equal counts by key ascending so the output is deterministic
+                // regardless of the merge's (insertion-order) output.
+                let key_tie = |left: &BucketEntry, right: &BucketEntry| left.key.cmp(&right.key);
```
trinity-1686a, line 1039: this should also be done at the segment level to get a deterministic set of buckets

src/aggregation/intermediate_agg_result.rs
```diff
@@ -1045,15 +1158,38 @@ impl IntermediateTermBucketResult {
...
                     } else {
-                        keyed.select_nth_unstable_by(size, |(k1, _), (k2, _)| {
-                            k1.partial_cmp(k2)
-                                .expect("expected type string, which is always sortable")
-                        });
+                        keyed.select_nth_unstable_by(size, |(k1, _), (k2, _)| k1.cmp(k2));
+                    }
+                    entries = keyed.into_iter().map(|(_, entry)| entry).collect();
+                }
+                OrderTarget::Count if mode == PruneMode::Final => {
```
trinity-1686a: why only when PruneMode::Final?

trinity-1686a (changes requested):
some of this should be replicated on multi-terms too (not necessarily here)

src/aggregation/intermediate_agg_result.rs
PSeitz: good catch, I'm not sure if this broken already if we e.g. add a new aggregation. We use postcard currently

trinity-1686a: i think it's fine as long as we don't reorder anything (we fail to parse the new aggregation, but there is no miss-interpretation, which is as good as you can get when parsing data from a future version)

PSeitz: I changed it to use `tuple_vec_map` and added a fixture test

src/aggregation/intermediate_agg_result.rs
PSeitz: I removed the `PartialEq` implementation

src/aggregation/intermediate_agg_result.rs
PSeitz: removed the `PruneMode::Final` condition

src/aggregation/bucket/term_agg/mod.rs
```diff
@@ -1449,36 +1462,59 @@ where
                 )?;
                 let val: u128 = compact_space_accessor.compact_to_u128(val as u32);
                 let val = Ipv6Addr::from_u128(val);
-                dict.insert(IntermediateKey::IpAddr(val), intermediate_entry);
+                out.push((IntermediateKey::IpAddr(val), intermediate_entry));
+            }
+        } else if term_req.column_type == ColumnType::F64 {
+            // Distinct f64 encodings can normalize to the same public key, notably -0.0 and +0.0.
```
trinity-1686a, line 1468: this should only be +/-0.0, and NaNs (though arguably, every single NaN is different, even to itself). I don't recall whether we allow storing NaNs. I think we could reasonably special-case -0.0 => +0.0 so that we don't need much more machinery than usual u64

src/aggregation/intermediate_agg_result.rs
```diff
@@ -1087,10 +1196,10 @@ impl IntermediateTermBucketResult {
                 self.doc_count_error_upper_bound += cutoff_doc_count;
             }
             entries.truncate(size);
-            self.entries = entries.into_iter().collect();
+            self.entries = entries;
         }
 
-        for entry in self.entries.values_mut() {
+        for (_, entry) in &mut self.entries {
```
trinity-1686a, line 1202: i don't understand this change

trinity-1686a (approved):

src/aggregation/intermediate_agg_result.rs
PSeitz: it's a vec now

src/aggregation/bucket/term_agg/mod.rs
PSeitz: I think we don't filter NaN. I'll merge it like this, we can do a simplification in a follow-up

PR #3043, opened 2026-08-14
SIMD batch unpacking for the u64 fast field codecs

marcbachmann:
## Motivation

Reading fast field values is the hot inner loop of aggregations, range queries and
sorting. `ColumnValues::get_range(start, &mut [T])` is the right API for it, and
`ColumnBlockAccessor` already routes contiguous doc blocks there — but no codec
implemented it. It was the trait default everywhere: one virtual `get_val` per row,
each recomputing its bit address for its own load, shift and mask. The monotonic
wrapper every u64-mapped column is loaded behind didn't forward it either, so a
codec that did implement it couldn't be reached through the public column type.

This makes the bulk read path actually bulk, for all three u64 codecs, and reuses
the same kernels for range filtering. **No on-disk format change** — same bytes,
different read path.

The design is introspection-free: rather than codecs advertising whether their
`get_range` is worth calling, every codec defines a batched one and the wrapper
forwards unconditionally. The `ColumnValues` trait is unchanged.

## What's in it

**`perf(bitpacker): SIMD batch decode with fused affine sinks`** — NEON, AVX2 and
SVE kernels behind `BitUnpacker::get_batch`/`get_batch_affine`. Eight consecutive
values occupy exactly `bit_width` bytes, so every group of 8 starts byte-aligned
with a (byte offset, bit shift) pattern constant per width — that is what lets a
kernel unpack 8 at once. `min + gcd * val` is applied in-register through `Store`
sinks instead of a second pass over the block. Range filtering rides the same
kernels: values are range-tested in register as they unpack, so only surviving row
ids reach memory. Replaces `filter_vec` and the crate's `bitpacking` dependency —
`tantivy-bitpacker` now has no dependencies.

**`perf(columnar): batch-decode get_range in the u64 codecs`** — every reader gets a
`decode_range` on those kernels; the linear codecs add their line back in a
vectorizable pass. The monotonic wrapper forwards in place for the identity mapping,
via a stack scratch for i64/f64/date. Block fetches gather single docs instead of
ranging them.

**`perf(columnar): read sub-8-row ranges through get_val`** — below
`MIN_BATCH_ROWS = 8` the batch setup outweighs the rows. With that floor inside each
codec, `get_range` is never slower than a `get_val` loop at any length, which is what
makes unconditional forwarding safe

**`perf(columnar): batched range filter for the blockwise linear codec`** — the
default `get_row_ids_for_value_range` walks one `get_val` per row and branches on
[...]

PR #3044, opened 2026-08-14
Add BlockFor fast field codec + SIMD batch decode (ALP included, unwired)

marcbachmann:
> **Update:** this PR now carries the whole series as five self-contained commits — the
> batch-decode work from #3043 is commits 1–2 (rewritten), the codecs follow. If you have
> reviewed #3043, start at `feat(columnar): BlockFor per-block frame-of-reference codec`.
> The ALP codec is no longer wired into the format (details below).

## Motivation

The u64 codec roster has a structural gap: every codec prices the whole column
at once. `Bitpacked` uses one bit width for all values, so a handful of
outliers inflates every row — a million 42_000s with 0.1% random 48-bit
outliers pack at 48 bits/value.

This PR first routes `ColumnValues::get_range` through SIMD kernels that
unpack 128 fixed-width values per call (NEON on aarch64, AVX2 on x86_64), then
adds a codec that prices each 128-value block independently on top of the same
kernels:

- **BlockFor (code 3)** — per-block frame of reference: each block stores its
  own minimum and bit width, so outliers only inflate the block they live in.
  Block metadata is bitpacked into fixed-width records read in place off the
  mapped bytes (O(1) load, zero resident metadata), and the range filter
  prunes whole blocks by their metadata before decoding.

An **ALP-style pseudo-decimal codec for f64 columns** (following Afroozeh et
al.) ships as code but is deliberately **not wired into `CodecType`**: it
needs more validation before production columns serialize with it. `alp.rs`
is compiled and fully tested through its estimator/loader; no code byte is
reserved, and wiring it up later is a small change.

## Commits

1. `perf(bitpacker): SIMD block decode kernels for bitpacked values` — narrow
   u32-lane kernel for widths 1..=25, wide u64-lane kernel for 26..=56, scalar
   fallback; drops the external `bitpacking` dependency.
2. `perf(columnar): batch get_range through whole-block decode` — a shared
   `BlockDecode` trait drives Bitpacked/Linear/BlockwiseLinear; the monotonic
   wrapper gains a three-way `get_range` (per-value below the codec-reported
   `min_batch_rows` crossover, in-place forward when `Input == Output`,
   block-buffered otherwise); `ColumnBlockAccessor` serves 1-doc fetches with
   a direct `get_val`.
3. `feat(columnar): BlockFor per-block frame-of-reference codec` — joins the
   default selection with a ~7.7% size penalty against its pricier random
   access.
4. `feat(columnar): experimental ALP pseudo-decimal codec for f64 columns` —
   not in `CodecType`.
5. `test(columnar): extend codec benchmarks for batch decode paths`

## Numbers

Measured on Apple M4 (NEON) and a 32-core x86_64 host (AVX2, native — not
Rosetta). In-process A/B where possible; `TANTIVY_BITPACKER_SCALAR=1` forces
the scalar decode path.

### Kernels (`kernel_sweep`, one 128-value block, vs scalar unpack)

| kernel | widths | M4 / NEON | x86 / AVX2 |
|---|---|---|---|
| narrow (u32 lanes) | 1..=25 | 3.16–3.28× | 3.48–3.56× |
| wide (u64 lanes) | 26..=56 | 2.79–2.96× | 3.99–4.29× |

Every packed width reaches a kernel (pinned by a test); widths 0/64 are
served by fill/plain loads.

### Batch `get_range` vs the per-value default it replaces (`bench_bitpacked_batch`, medians)

| shape | M4 | x86 |
|---|---|---|
| full scan, widths 8..64 | 4.0–4.5× | 5.8–7.1× |
| scattered 64-row takes | 2.0× | 2.2× |
| scattered 16-row takes | 1.9× | 2.2× |
| f64 column, 128-row chunks (buffered wrapper path) | 1.6–1.9× | 2.5× |

1-row takes deliberately fall back to the per-value loop; the production
1-doc path (`ColumnBlockAccessor`) short-circuits to `get_val`.

### BlockFor (2M-row shapes, `bench_column_values_get`, medians, M4)

The selection still picks BlockwiseLinear on these shapes on size; BlockFor's
win is the metadata-pruned range filter and whole-block decode:

| timestamps shape | BlockwiseLinear | BlockFor |
|---|---|---|
| `filter_1pct_span` | 1543 µs | **57 µs (27×)** |
| `filter_walk_only` (no hits) | 1539 µs | **32 µs (49×)** |
| `get_range_seq128` | 919 µs | 492 µs (1.9×) |
| `get_val_random` (100k gets) | 288 µs | 569 µs (0.5×) |

Random single-value access is where BlockFor pays (block-meta load per get) —
hence the 13/12 selection penalty: it must win on size by ≥ ~7.7%, so a
near-tie never trades a 55–80% random-get cost for a noise-level size win.
Write side: BlockFor serializes at ~ BlockwiseLinear speed (502 vs 515 µs on
the 130k-value bench input).

### End-to-end regression check (this branch vs its merge-base, same machine)

- `agg_bench` (1M-doc indices built by the *base* binary, so both sides read
  identical bytes): median delta −1.9% across the metrics/cardinality group,
  everything inside the cross-binary noise band.
- `range_query`, `str_search_and_get`, `merge_segments` and the columnar
  `bench_values_u64` / `bench_access` / `bench_merge`: medians flat (−0.3% to
  +2.6% per suite); the fast-field range-filter cells improve ~15%.
- Index size, `agg_bench` corpus rebuilt by each side: **77.2 MB vs 80.3 MB
  (−3.9%)** with BlockFor in the selection; `.fast` files 33.0 vs 33.5 MB.

## Format impact

Codec code 3 (BlockFor) is new: columns written after this PR may not be
readable by older readers. Everything previously written stays readable — the
existing codecs and their code bytes are untouched. ALP reserves no code.

## Verification

- `tantivy-bitpacker` 25 tests, `tantivy-columnar` 233 tests: pass on
  aarch64/NEON **and natively on x86_64/AVX2**, each also with
  `TANTIVY_BITPACKER_SCALAR=1` forcing the scalar path.
- Each of the five commits compiles warning-free and passes the crate suites
  individually.
- New tests: proptests for `get_batch`/`get_batch_u32s` across all widths and
  block phases; per-width kernel-vs-scalar equality plus a gate test proving
  no packed width silently falls back to scalar; every codec × every block
  phase `get_range` equality (raw + wrapped); wrapper-forwarding pin tests
  (`min_batch_rows`, `specializations_survive_arc`); BlockFor
  footer-layout/estimate-exactness tests; ALP round-trip incl. NaN/±0/inf,
  full-exception blocks, and opt-out on non-decimal / full-precision data.
- Threshold calibration is reproducible via the ignored `gate_sweep` /
  `kernel_sweep` tests (commands in their doc comments); the x86 sweep
  confirms the shipped `min_batch_rows` values hold there too.

### DIFF
--- a/bitpacker/Cargo.toml
+++ b/bitpacker/Cargo.toml
@@ -15,7 +15,6 @@ homepage = "https://github.com/quickwit-oss/tantivy"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-bitpacking = { version = "0.9.2", default-features = false, features = ["bitpacker1x"] }
 
 [dev-dependencies]
 binggan = "0.17.0"
--- a/bitpacker/src/bitpacker.rs
+++ b/bitpacker/src/bitpacker.rs
@@ -1,7 +1,7 @@
 use std::io;
 use std::ops::{Range, RangeInclusive};
 
-use bitpacking::{BitPacker as ExternalBitPackerTrait, BitPacker1x};
+use crate::block_decode::{BLOCK_LEN, Lane, decode_block_lanes};
 
 pub struct BitPacker {
     mini_buffer: u64,
@@ -122,6 +122,10 @@ impl BitUnpacker {
         val_shifted & self.mask
     }
 
+    pub fn get_batch(&self, start_idx: u32, data: &[u8], output: &mut [u64]) {
+        self.get_batch_lanes(start_idx, data, output);
+    }
+
     // Decodes the range of bitpacked `u32` values with idx
     // in [start_idx, start_idx + output.len()).
     //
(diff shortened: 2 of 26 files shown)


marcbachmann:
Found the cause of why I thought the indexes are not doing that well #3050

PR #3046, opened 2026-08-18
Add a JIT expression compiler

fulmicoton-dd:
## What does this change do?

- adds the `jitexpr` crate, including expression parsing, type inference, and Cranelift lowering
- integrates compiled expressions with calculated predicate queries
- implements the supported calculated-field functions and tracks coverage in `jitexpr/docs/FUNCTIONS.md`
- adds borrowed string results and a fixed-capacity call arena for functions that construct strings

`LEFT`, `RIGHT`, and `SUBSTRING` use UTF-8 byte offsets. Slices that split a code point return null.

## Testing

- `cargo test -p jitexpr`


fulmicoton-dd:
Superseded by the two draft PRs #3047 and #3048.

PR #3047, opened 2026-08-19
Calculed fields: Add standalone JIT expression compiler

fulmicoton-dd:
JitExpr is an expression JIT compiler for tantivy.
The goal here is to be able to evaluate dynamically defined predicate over columns.

The compilation goes through different representations:

```
  UntypedExpr
      ↓ injecting variable types, and type checking
  TypedExpr
      ↓ lowering
  Cranelift IR
      ↓ Cranelift code generation
  Machine code (or assembly)
  ```

Function specific logic is isolated in a FnCall trait that offers rails for AI / humans to extend the system
safely and in a uniform manner.

Input and output variable are 16 bytes VariableValue object that are Copy.
Primitive types are using 8 bytes for the value and 8bytes for the bool (alignement)

String are also supposed as Option<&str> and also take 16 bytes.
strings are not manipulated by the cranelift program itself. Instead the program calls native function and
treats them as opaque objects.

Function that return reference to an input string or a literal just return a reference over one or the other.
Function that build string use a string arena as their scratch pad to write their bytes and return a &str
pointing to it.

As a result the return type has a lifetime limited to &self (literal + string arena) and the input lifetime (regex extract for instance).

# Use of AI

Most of the function themselves (all but ADD, REGEXP_EXTRACT, LOWER), the unit tests, the serialization format code is isolated and has been coded by LLM with very little review.

### DIFF
--- a/jitexpr/examples/basic.rs
+++ b/jitexpr/examples/basic.rs
@@ -1,0 +1,43 @@
+use std::collections::HashMap;
+use std::error::Error;
+use std::sync::Arc;
+
+use jitexpr::ast::{Function, InferredTypeSet, UntypedExpr, infer_types};
+use jitexpr::compile::{CompiledFn, CompiledFnCtx, compile};
+use jitexpr::types::{VarType, VariableValue};
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // A simple expression that goes:
+    // my_col + 1
+    let untyped_expr = Function::Add.call_untyped_expr(vec![
+        UntypedExpr::variable("my_col"),
+        UntypedExpr::literal(1.0f64),
+    ]);
+
+    // Infer types does not return specific types, but instead a set of acceptable
+    // types for each variables.
+    let inferred_types: HashMap<&str, InferredTypeSet> = infer_types(&untyped_expr)?;
+    assert_eq!(
+        inferred_types.get("my_col").unwrap(),
+        &InferredTypeSet::NUMERICAL
+    );
+
+    // This is then up to us to decide the actual type for each variable.
+    // In tantivy, this means picking the first column with a type in inferred_types.
+    //
+    // If none match then we should use the VarType::None.
+    let variable_types: HashMap<&str, VarType> =
+        std::iter::once(("my_col", VarType::F64)).collect();
+
+    let compiled_fn: Arc<CompiledFn> = compile(&untyped_expr, &variable_types)?;
+    let mut compiled_fn_ctx = CompiledFnCtx::new(compiled_fn);
+
+    // We use a nullable wrapper around the value union to pass typed variables.
+    // For present values, it is up to us to populate the correct union member.
+    // Not doing so is UB.
+    let input: Box<[VariableValue]> = vec![VariableValue::from(1.2f64)].into_boxed_slice();
+    let output = unsafe { compiled_fn_ctx.call(&input[..]) };
+    assert_eq!(unsafe { output.as_f64() }, Some(1.2f64 + 1.0f64));
+
+    Ok(())
+}
--- a/jitexpr/src/ast/infer_types.rs
+++ b/jitexpr/src/ast/infer_types.rs
@@ -1,0 +1,296 @@
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
+
+use crate::ast::{Function, Literal, UntypedExpr};
+use crate::functions::InvalidFunctionCall;
+use crate::types::VarType;
+
+#[derive(Default, Copy, Clone, Debug, Eq, PartialEq)]
+pub struct InferredTypeSet {
+    pub string: bool,
+    pub i64: bool,
+    pub u64: bool,
+    pub f64: bool,
+    pub boolean: bool,
+}
+
+impl InferredTypeSet {
+    pub const NONE: InferredTypeSet = InferredTypeSet {
+        string: false,
+        i64: false,
+        u64: false,
+        f64: false,
+        boolean: false,
+    };
+
+    pub const ALL: InferredTypeSet = InferredTypeSet {
+        string: true,
+        i64: true,
+        u64: true,
+        f64: true,
+        boolean: true,
+    };
+
+    pub const NUMERICAL: InferredTypeSet = InferredTypeSet {
+        i64: true,
+        u64: true,
+        f64: true,
+        boolean: false,
+        string: false,
+    };
+
+    pub const I64: InferredTypeSet = InferredTypeSet {
+        i64: true,
+        ..Self::NONE
+    };
+
+    pub const U64: InferredTypeSet = InferredTypeSet {
+        u64: true,
+        ..Self::NONE
+    };
+
+    pub const F64: InferredTypeSet = InferredTypeSet {
+        f64: true,
+        ..Self::NONE
+    };
+
+    pub const STRING: InferredTypeSet = InferredTypeSet {
+        string: true,
+        ..Self::NONE
+    };
+
+    pub const BOOLEAN: InferredTypeSet = InferredTypeSet {
+        boolean: true,
+        ..Self::NONE
+    };
+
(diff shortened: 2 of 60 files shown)

### CODE REVIEW

File: jitexpr/src/functions/lower.rs [hunk 1]
```diff
@@ -0,0 +1,333 @@
...
+    // LOWER calls whose input points into an earlier, disjoint arena allocation.
+    // SAFETY: CompiledFn exclusively owns and passes this arena for the call.
+    let Some(output_ptr) = (unsafe { &mut *string_arena }).allocate(output_len) else {
+        return RawStr::none();
+    };
+
+    // SAFETY: Same input contract as above. The fixed arena never reallocates,
+    // and its new output range starts after any arena-backed input range.
+    let input =
+        unsafe { std::str::from_utf8_unchecked(std::slice::from_raw_parts(input_ptr, input_len)) };
```
fulmicoton, line 184: this code is really dumb...

jitexpr/examples/basic.rs
Mallets, line 1: Very nice, I like this kind of examples.

jitexpr/src/compile/mod.rs
```diff
@@ -0,0 +1,362 @@
...
+
+use crate::ast::UntypedExpr;
+use crate::functions::NativeFunctions;
+use crate::types::{VarType, VariablePrimitiveOpt, VariableValue};
+
+/// Compiles an expression into an immutable, shareable native function.
+pub fn compile(
+    untyped_expr: &UntypedExpr,
+    var_types: &HashMap<&str, VarType>,
+) -> Result<Arc<CompiledFn>, CompileError> {
```
Mallets, line 33: nit: compile could return a `CompiledFn` that is something like `CompiledFn(Arc<CompiledFnInner>)` and remove the `Arc` exposure to every caller.

jitexpr/examples/basic.rs
```diff
@@ -0,0 +1,43 @@
...
+    );
+
+    // This is then up to us to decide the actual type for each variable.
+    // In tantivy, this means picking the first column with a type in inferred_types.
+    //
+    // If none match then we should use the VarType::None.
+    let variable_types: HashMap<&str, VarType> =
+        std::iter::once(("my_col", VarType::F64)).collect();
+
+    let compiled_fn: Arc<CompiledFn> = compile(&untyped_expr, &variable_types)?;
```
Mallets, line 32: I'm missing the point of returning separately a `CompiledFnCtx` and `CompiledFn`. In this example the actual `.call` happens on `compiled_fn_ctx` and not on `compiled_fn`. Can `CompiledFn` be used without being passed to  `CompiledFnCtx`?

jitexpr/src/ast/literal.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    pub fn types(&self) -> InferredTypeSet {
+        match self {
+            Literal::None => InferredTypeSet::ALL,
+            Literal::Bool(_) => InferredTypeSet::BOOLEAN,
+            Literal::I64(value) => InferredTypeSet {
+                i64: true,
+                u64: *value >= 0,
+                f64: (*value as f64) as i128 == *value as i128,
```
Mallets, line 29: I don't really understand this magic of casting a `f64` to `i128` :)

jitexpr/src/ast/infer_types.rs
```diff
@@ -0,0 +1,293 @@
...
+        i64: true,
+        u64: true,
+        f64: true,
+        boolean: false,
+        string: false,
+    };
+
+    pub const I64: InferredTypeSet = InferredTypeSet {
+        i64: true,
+        ..Self::NONE
```
Mallets, line 43: TIL defaults on `const` definitions could be inherited from other `const`.

jitexpr/src/ast/serialize.rs
Mallets, line 1: This file is used only in `bin/jitexpr-asm`, it's not used anywhere else to parse user input.

jitexpr/src/compile/compiled_fn.rs
```diff
@@ -0,0 +1,152 @@
...
+use super::{StringArena, TypedExpr, TypedVariable};
+use crate::types::{VarType, VariableValue};
+
+#[cfg(not(any(
+    all(target_arch = "x86_64", not(target_os = "windows")),
+    target_arch = "aarch64"
+)))]
+compile_error!(
+    "the direct VariableValue JIT return ABI is only implemented for x86-64 System V and AArch64"
+);
```
Mallets, line 15: This `#[cfg]` does not exclude `aarch64-pc-windows-*`, I didn't test but I'm expecting it to fail as well.

jitexpr/src/compile/mod.rs
fulmicoton: This is voluntary. I actually like to expose the arc explicitly except for object that are used a lot (e.g. Schema).
I actually want users to know this is an object that costs nothing to clone.

jitexpr/examples/basic.rs
fulmicoton: Make function compilation cachable.
It takes a few ms to compile but generally the compilation applies to all splits. (rather we can generate/cache one version per set of column types)

jitexpr/src/ast/literal.rs
fulmicoton: I will try to add comments and wrangel the code into something that makes sense.

fulmicoton: with furhter thoughts we don't want that behavior and we want to always accept f64.

Explaining the presence of i128 though.

The question this was trying to answer is given a u64, when can it be translated into f64 in a loss less manner.

By lossless, we do not just mean:
u64 -> f64 -> u64 brings us back to the same value, 
we also want to make sure that the f64 value was indeed representing the original u64 value so that 
we do not have strange side effect if we did arithmetic in the f64 world.

u64::MAX does NOT map to its value in f64 world, but remaps correctly to u64::MAX due to saturation.
Using u128 make it possible to rule out u64::MAX as having a f64 representation.

jitexpr/src/ast/serialize.rs
fulmicoton: good point i will try to isolate it.

jitexpr/src/ast/literal.rs
Mallets: Thanks for the explanation!

jitexpr/src/compile/compiled_fn.rs
fulmicoton: yes it is not supported because the ABI does not support returning more than a single word via register. I commented along these lines. If we want to support windows the natural fix would be to change that to passing a destination value pointer as arg.

jitexpr/src/ast/serialize.rs
fulmicoton: After investigation, the benefit might not be as large as the gain.
(either via feature flag or via creation of a project).

Let's do the separate project if we end up adding some binary specific dependency. Right now this is not the case.

fulmicoton: (right now it has the benefit of offering a good looking Debug/Display. Deserialization is pretty useless though)

jitexpr/src/functions/add.rs
```diff
@@ -0,0 +1,482 @@
...
+use crate::functions::{FnCall, FnCallEnum};
+use crate::types::VarType;
+
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct AddFnCall {
+    pub(crate) args: Box<[TypedExpr]>,
+}
+
+impl FnCall for AddFnCall {
+    const ARG_COUNT: super::ArgumentCount = super::ArgumentCount::Any;
```
Mallets, line 33: `Add` accept any number of arguments while `Multiple`, `Subtract`, `Divide`, etc. accept exactly `2` arguments. Should we align the semantics of `Add` to accept exactly `2` arguments?

PR #3056, opened 2026-08-26
Fix RamDirectory::atomic_write not waiting for reload watchers

### LINKED ISSUE #1824: Missing search results with in memory index

Divyesh-k:
commit() on a RamDirectory broadcast the meta.json watch callbacks (including IndexReader::reload() under ReloadPolicy::OnCommitWithDelay) on a detached thread and dropped the result without waiting, so a search performed immediately after commit() returned could still miss documents from that commit. Wait for the broadcast to finish before atomic_write() returns.

Fixes #1824

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Tantivy 0.26 (Unreleased)
 ================================
 
 ## Bugfixes
+- Fix `RamDirectory::atomic_write` not waiting for `OnCommitWithDelay` watchers (e.g. `IndexReader::reload()`), so a search right after `commit()` on an in-memory index could miss documents from that commit [#1824](https://github.com/quickwit-oss/tantivy/issues/1824)(@Divyesh-k)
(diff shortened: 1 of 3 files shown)


fulmicoton:
I don't think we want to merge that.

The behavior expected here is:

something we will never be able to offer with the MMapDirectory
the Ram directory main purpose (granted it can be used differently) for unit test, so we should not have it behave "better" than the mmap directory
the impact on write performance could be massive.

PR #3058, opened 2026-08-26
Add read-only directory support

shashankbrgowda:
## Summary

Adds `ReadOnlyDirectory` for opening an immutable index without requiring write access to its directory.

- Passes read operations to the underlying directory.
- Rejects write and delete operations.
- Skips `META_LOCK`.
- Rejects other locks, including the writer lock.
- Disables filesystem watching.

Two convenience methods are included:

```rust
Index::open_read_only(directory)
Index::open_read_only_in_dir(path)
```

## Motivation

`IndexReader` acquires `META_LOCK` while loading segment readers. This requires write access to the index directory, even when the index is immutable and used only for searching.

`ReadOnlyDirectory` skips this lock so an index can be used from a read-only filesystem.

A custom `Directory` wrapper can provide similar behaviour in Rust, but this workaround is not available to users of bindings such as `tantivy-py`, because the `Directory` interface is not exposed to Python. Providing `ReadOnlyDirectory` in Tantivy allows bindings to expose the same behaviour without maintaining their own implementation.

The underlying index must remain unchanged while it is open. This mode must not be used while another process is writing to or garbage-collecting the same index.

Addresses #557 https://github.com/quickwit-oss/tantivy/issues/557#issuecomment-5430243628

## Testing

Tests cover:

- Reading from the wrapped directory.
- Rejection of writes and deletes.
- Metadata and writer lock behaviour.
- Searching and manually reloading an index.
- Opening a real read-only filesystem directory.
- Opening the same index with multiple readers.
- Verifying that `.tantivy-meta.lock` is not created.

Built with GPT-5.6 Sol in the Codex desktop harness.

### DIFF
--- a/src/directory/directory.rs
+++ b/src/directory/directory.rs
@@ -100,10 +100,12 @@ fn retry_policy(is_blocking: bool) -> RetryPolicy {
 /// Write-once read many (WORM) abstraction for where
 /// tantivy's data should be stored.
 ///
-/// There are currently two implementations of `Directory`
+/// There are currently three implementations of `Directory`
 ///
 /// - The [`MMapDirectory`][crate::directory::MmapDirectory], this should be your default choice.
 /// - The [`RamDirectory`][crate::directory::RamDirectory], which should be used mostly for tests.
+/// - The [`ReadOnlyDirectory`][crate::directory::ReadOnlyDirectory], which wraps another directory
+///   containing an immutable index.
 pub trait Directory: DirectoryClone + fmt::Debug + Send + Sync + 'static {
     /// Opens a file and returns a boxed `FileHandle`.
     ///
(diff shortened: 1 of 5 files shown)


fulmicoton:
This is a just a read only directory, and that functionality can leave outside of tantivy, so this is just about including batteries, correct? 

I would call it immutable more than read only maybe.
Read only could mean "I do not write but there is a writer in a different process". 
This is not compatible with this.

shashankbrgowda:
Thanks for the feedback @fulmicoton . 

Correct, this is mainly about including the batteries. Rust users can implement a custom directory wrapper outside tantivy, but tantivy-py users cannot provide their own directory implementation. Including this in tantivy would allow the python bindings to open a static index from a read only filesystem.

I agree Immutable communicates the api better than read only. Shall I rename the directory and methods accordingly ?

> This is a just a read only directory, and that functionality can leave outside of tantivy, so this is just about including batteries, correct?
> 
> I would call it immutable more than read only maybe. Read only could mean "I do not write but there is a writer in a different process". This is not compatible with this.

shashankbrgowda:
@fulmicoton I’ve renamed the API from read-only to immutable. Please let me know if you had different naming in mind. Thanks

PR #3070, opened 2026-09-02
Added the capacity to build scorer and seek_danger in a joint manner.

fulmicoton:
This PR introduces an optimization follow up to the seek_danger logic.

seek_danger has been introduced to allow Scorer for which seeking is someone slow to lazily just answer "that document don't match" instead of seeking for the actual next document they match.
e.g. A phrase scorer would have to find the next doc.

That optimization was NOT usable however for scorer initialization. All scorer have to be initialized and positioned on their first doc.

The main motivation is the future addition of predicate queries that are based on a calculated field.
Asking for the first doc of such a Score would mean scanning and evaluating that predicate for every doc until one matches.

This PR addresses that issue by adding a new method to Weight called scorer_danger. It is the equivalent of seek_danger but for Scorer initialization.

The only caller is BooleanWeight, which will use it to create the scorers associated to its must clause.
All of the code that manipulates scorer in a "dangerous", almost undefined state is contained in create_aligned_scorers_for_intersection.

### DIFF
--- a/src/query/boolean_query/boolean_weight.rs
+++ b/src/query/boolean_query/boolean_weight.rs
@@ -359,7 +359,7 @@ impl<TScoreCombiner: ScoreCombiner> BooleanWeight<TScoreCombiner> {
             remove_and_count_all_and_empty_scorers(&mut must_scorers);
 
         if must_special_scorer_counts.num_empty_scorers > 0 {
-            return Ok(SpecializedScorer::Other(Box::new(EmptyScorer)));
+            return Ok(SpecializedScorer::empty());
         }
 
         let mut should_scorers = per_occur_scorers.remove(&Occur::Should).unwrap_or_default();
@@ -374,7 +374,7 @@ impl<TScoreCombiner: ScoreCombiner> BooleanWeight<TScoreCombiner> {
 
         if exclude_special_scorer_counts.num_all_scorers > 0 {
             // We exclude all documents at one point.
-            return Ok(SpecializedScorer::Other(Box::new(EmptyScorer)));
+            return Ok(SpecializedScorer::empty());
         }
 
         let effective_minimum_number_should_match = self
@@ -386,7 +386,7 @@ impl<TScoreCombiner: ScoreCombiner> BooleanWeight<TScoreCombiner> {
             if effective_minimum_number_should_match > num_of_should_scorers {
                 // We don't have enough scorers to satisfy the minimum number of should matches.
                 // The request will match no documents.
-                return Ok(SpecializedScorer::Other(Box::new(EmptyScorer)));
+                return Ok(SpecializedScorer::empty());
             }
             match effective_minimum_number_should_match {
                 0 if num_of_should_scorers == 0 => ShouldScorersCombinationMethod::Ignored,
--- a/src/query/boolean_query/boolean_weight.rs
+++ b/src/query/boolean_query/boolean_weight.rs
@@ -1,6 +1,7 @@
+use std::cmp::Ordering;
 use std::collections::HashMap;
 
-use crate::docset::COLLECT_BLOCK_BUFFER_LEN;
+use crate::docset::{SeekDangerResult, COLLECT_BLOCK_BUFFER_LEN};
 use crate::index::SegmentReader;
 use crate::postings::FreqReadingOption;
 use crate::query::disjunction::Disjunction;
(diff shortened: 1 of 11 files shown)


src/query/boolean_query/boolean_weight.rs
```diff
@@ -170,6 +179,68 @@ fn effective_should_scorer_for_union<TScoreCombiner: ScoreCombiner>(
...
+        }
+    }
+
+    // This eventually terminates because, at each iteration, the pair
+    // (candidate, num_scorer_aligned) increases according to lexicographic order.
+    //
+    // Eventually, either candidate will reach TERMINATED or
+    // num_scorer_aligned will reach scorers.len().
+    let mut num_scorer_aligned = 0;
+    for scorer_id in (0..scorers.len()).cycle() {
```
PSeitz-dd:
```suggestion
    for scorer_ord in (0..scorers.len()).cycle() {
```

src/query/boolean_query/boolean_weight.rs
```diff
@@ -170,6 +179,69 @@ fn effective_should_scorer_for_union<TScoreCombiner: ScoreCombiner>(
...
+    // We now this eventually terminates because at each iteration because
+    // at each iteration the couple (candidate, num_scorer_aligned) gets ever greater
+    // according to the lexicographic order.
+    //
+    // Eventually, either candidate will reach TERMINATED or
+    // num_scorer_aligned will reach scorers.len().
+    let mut num_scorer_aligned = 0;
+    for scorer_id in (0..scorers.len()).cycle() {
+        if let SeekDangerResult::SeekLowerBound(seek_lower_bound) =
+            scorers[scorer_id].seek_danger(candidate)
```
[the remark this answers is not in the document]
PSeitz-dd: I think this is correct. not sure if it's a problem though in any implementation currently

src/query/boolean_query/boolean_weight.rs
```diff
@@ -181,7 +252,7 @@ enum ShouldScorersCombinationMethod {
 
 /// Weight associated to the `BoolQuery`.
 pub struct BooleanWeight<TScoreCombiner: ScoreCombiner> {
-    weights: Vec<(Occur, Box<dyn Weight>)>,
+    per_occur_weights: HashMap<Occur, Vec<Box<dyn Weight>>>,
```
PSeitz-dd: I think it would make sense to move this to its own struct with some methods, e.g.
```rust
struct OccurWeights{
    should: Vec<Box<dyn Weight>>>>,
    must: Vec<Box<dyn Weight>>>>,
    must_not: Vec<Box<dyn Weight>>>>,
}
impl OccurWeights{
   fn get(&self, occur: Occur) -> impl Iterator<Item = &dyn Weight>                                                                                                                                                                                            
   fn count(&self, occur: Occur) -> usize                                                                                                                                                                                                                      
   fn promote_should_to_must(&mut self)
}
```

src/query/weight.rs
```diff
@@ -71,6 +71,36 @@ pub trait Weight: Send + Sync + 'static {
...
+    fn scorer_danger(
+        &self,
+        reader: &SegmentReader,
+        target: DocId,
+        boost: Score,
+    ) -> crate::Result<(SeekDangerResult, Box<dyn Scorer>)> {
+        let mut scorer = self.scorer(reader, boost)?;
+        let scorer_doc = scorer.doc();
+        let seek_danger_result = match scorer_doc.cmp(&target) {
+            std::cmp::Ordering::Less => scorer.seek_danger(target),
```
PSeitz:
```suggestion
            Ordering::Less => scorer.seek_danger(target),
```

src/query/boolean_query/boolean_weight.rs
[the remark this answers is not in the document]
fulmicoton: ah yeah that's a good point.

src/query/boolean_query/boolean_weight.rs
```diff
@@ -170,6 +179,70 @@ fn effective_should_scorer_for_union<TScoreCombiner: ScoreCombiner>(
...
+        }
+    }
+
+    // This eventually terminates because, at each iteration, the pair
+    // (candidate, num_scorer_aligned) increases according to lexicographic order.
+    //
+    // Eventually, either candidate will reach TERMINATED or
+    // num_scorer_aligned will reach scorers.len().
+    let mut scorer_ord = 0;
+    while num_scorer_aligned < scorers.len() {
```
fulmicoton, line 228: If you are wondering the change of the loop condition is not AI going crazy.
I just like making sure the exit condition is expressed in the loop statement.

src/query/boolean_query/boolean_weight.rs
fulmicoton: done

src/query/boolean_query/boolean_weight.rs
```diff
@@ -226,30 +317,47 @@ impl<TScoreCombiner: ScoreCombiner> BooleanWeight<TScoreCombiner> {
...
 
-        // Indicate how should clauses are combined with must clauses.
+        let intersection_weights: &[Box<dyn Weight>] = self.per_occur_weights.get(Occur::Must);
+
         let mut must_scorers: Vec<Box<dyn Scorer>> =
-            per_occur_scorers.remove(&Occur::Must).unwrap_or_default();
-        let must_special_scorer_counts = remove_and_count_all_and_empty_scorers(&mut must_scorers);
+            create_aligned_scorers_for_intersection(intersection_weights, reader, boost)?;
+
+        if !must_scorers.is_empty() {
```
PSeitz, line 326:
```suggestion
        // TODO: pass the current doc to the scorer creation below (via `scorer_danger`) and remove the early exit.
        if !must_scorers.is_empty() {
```

PSeitz (approved):

PR #3072, opened 2026-09-03
Add DocPredicateQuery and a generic FunctionPredicate implementation

fulmicoton-dd:
## Summary
- Adds `DocPredicateQuery`, a query that evaluates a per-document predicate against each document of a segment, with support for `Weight::scorer_danger` to allow fast seeking within intersections.
- Adds `FunctionPredicate`, a generic `DocPredicate` implementation built from a plain per-segment factory closure that returns a per-document closure. Has no dependency on fast fields or any other segment data structure.

## Test plan
- `cargo test query::doc_predicate_query`

### DIFF
--- a/src/query/doc_predicate_query/function_predicate.rs
+++ b/src/query/doc_predicate_query/function_predicate.rs
@@ -1,0 +1,99 @@
+use super::{DocPredicate, SegmentDocPredicate};
+use crate::index::SegmentReader;
+use crate::DocId;
+
+/// Blanket [`SegmentDocPredicate`] implementation for any per-document
+/// closure.
+///
+/// Hidden contract: [`SegmentDocPredicate::eval`] takes `&self`, not
+/// `&mut self`, so the closure must be callable through an immutable
+/// reference (`Fn`, not `FnMut`). A caller that needs per-document mutable
+/// state must capture it behind its own synchronization primitive (for
+/// example `Arc<Mutex<_>>` or `Arc<AtomicUsize>`) rather than relying on
+/// interior mutability provided by this blanket impl.
+impl<F> SegmentDocPredicate for F
+where F: Fn(DocId) -> bool + Send + Sync + 'static
+{
+    fn eval(&self, doc_id: DocId) -> bool {
+        (self)(doc_id)
+    }
+}
+
+/// A [`DocPredicate`] built from a plain factory function.
+///
+/// `FunctionPredicate` wraps a factory closure that is called once per
+/// segment (via [`DocPredicate::doc_predicate`]) and returns a per-document
+/// closure evaluated once per candidate document. This predicate has no
+/// dependency on fast fields or any other segment data structure: the
+/// factory receives the [`SegmentReader`] and is free to ignore it entirely,
+/// for example to build a predicate over `doc_id` alone.
+///
+/// # Example
+///
+/// ```
+/// use tantivy::query::doc_predicate_query::FunctionPredicate;
+/// use tantivy::query::doc_predicate_query::DocPredicateQuery;
+/// use tantivy::index::SegmentReader;
+/// use tantivy::DocId;
+///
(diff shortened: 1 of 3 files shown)


src/query/doc_predicate_query/function_predicate.rs
```diff
@@ -0,0 +1,99 @@
...
+
+/// Blanket [`SegmentDocPredicate`] implementation for any per-document
+/// closure.
+///
+/// Hidden contract: [`SegmentDocPredicate::eval`] takes `&self`, not
+/// `&mut self`, so the closure must be callable through an immutable
+/// reference (`Fn`, not `FnMut`). A caller that needs per-document mutable
+/// state must capture it behind its own synchronization primitive (for
+/// example `Arc<Mutex<_>>` or `Arc<AtomicUsize>`) rather than relying on
+/// interior mutability provided by this blanket impl.
```
PSeitz: I don't think this blanket provides interior mutability?

src/query/doc_predicate_query/function_predicate.rs
```diff
@@ -0,0 +1,99 @@
...
+/// let _query: DocPredicateQuery = predicate.into();
+/// ```
+pub struct FunctionPredicate<F> {
+    factory: F,
+}
+
+impl<F> FunctionPredicate<F> {
+    /// Creates a new predicate from a per-segment factory function.
+    ///
+    /// `factory` is called once per segment and must return a per-document
```
PSeitz: this could be in the type instead a comment

src/query/doc_predicate_query/function_predicate.rs
```diff
@@ -0,0 +1,99 @@
...
+/// use tantivy::DocId;
+///
+/// // Matches every even doc id, in every segment.
+/// let predicate = FunctionPredicate::new(|_segment_reader: &SegmentReader| {
+///     Ok(move |doc_id: DocId| doc_id % 2 == 0)
+/// });
+/// let _query: DocPredicateQuery = predicate.into();
+/// ```
+pub struct FunctionPredicate<F> {
+    factory: F,
```
PSeitz:
```suggestion
    segment_predicate_factory: F,
```

src/query/doc_predicate_query/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+    }
+}
+
+impl<TSegmentDocPredicate: SegmentDocPredicate> DocPredicateDocSet<TSegmentDocPredicate> {
+    fn find_match(&mut self, mut target: DocId) -> DocId {
+        loop {
+            match self.seek_danger(target) {
+                SeekDangerResult::Found => return target,
+                SeekDangerResult::SeekLowerBound(next_target) => {
+                    if next_target >= self.max_doc {
```
PSeitz: nitpick: we have this check twice per loop, in and after seek_danger

src/query/doc_predicate_query/function_predicate.rs
fulmicoton: apologies, i used AI in a sloppy manner and asked it to fix warnings so it added docs and I didn't read the result. shame on me

src/query/doc_predicate_query/function_predicate.rs
fulmicoton: good point

src/query/doc_predicate_query/function_predicate.rs
fulmicoton: agreed

src/query/doc_predicate_query/mod.rs
fulmicoton: it is difficult to do without any extra conditions in a clear way. I changed the check to compare with TERMINATED instead, and did not assign self.doc though

PSeitz (approved):

PR #3081, opened 2026-09-07
(Calculated fields)  Add feature-gated JIT expression document predicate

fulmicoton:
## Summary

A predicate for jitexpr expression.
The point is to be able to run queries where the predicate includes calculated fields.

This is a first PR with horrible performance.

# Follow up needed

- a compilation cache: right now we compile for every segment
- a term ord -> bool cache or at least a term ord -> str cache for dictionary encoded string fields
- support for a plain codec (= not dictionary encoded)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,6 +76,9 @@ measure_time = "0.9.0"
 arc-swap = "1.5.0"
 bon = "3.3.1"
 
+# EXPERIMENTAL. The API is likely to change in the near future.
+jitexpr = { version = "0.1", path = "./jitexpr", optional = true }
+
 columnar = { version = "0.7", path = "./columnar", package = "tantivy-columnar" }
 sstable = { version = "0.7", path = "./sstable", package = "tantivy-sstable", optional = true }
 stacker = { version = "0.7", path = "./stacker", package = "tantivy-stacker" }
@@ -149,6 +152,7 @@ failpoints = ["fail", "fail/failpoints"]
 unstable = []                            # useful for benches.
 
 quickwit = ["sstable", "futures-util", "futures-channel"]
+jitexpr = ["dep:jitexpr"]
 
 # Compares only the hash of a string when indexing data.
 # Increases indexing speed, but may lead to extremely rare missing terms, when there's a hash collision.
--- a/src/query/doc_predicate_query/jitexpr_predicate.rs
+++ b/src/query/doc_predicate_query/jitexpr_predicate.rs
@@ -1,0 +1,535 @@
+use std::collections::HashMap;
+
+use columnar::{ColumnType, DynamicColumn, StrColumn};
+use jitexpr::ast::{infer_types_with_target, InferredTypeSet, TypeError, UntypedExpr};
+use jitexpr::compile::{compile, CompiledFnCtx};
+use jitexpr::types::{VarType, VariableValue};
+
+use super::{DocPredicate, SegmentDocPredicate};
+use crate::index::SegmentReader;
+use crate::{DocId, TantivyError};
+
+/// A [`DocPredicate`] that evaluates a boolean JIT expression against fast fields.
+///
+/// Requires the `jitexpr` feature. Variable names are resolved as fast-field names
+/// for each segment, supporting boolean, numeric, and string columns. Missing or
+/// incompatible columns are left unbound, so the compiler treats them as `None`.
+/// For bound columns, multivalued documents contribute their first value, and a
+/// document missing any input is skipped before evaluating the expression.
+/// Only a present `true` result matches.
+///
+/// ```
+/// use tantivy::jitexpr::ast::deserialize;
+/// use tantivy::query::doc_predicate_query::{DocPredicateQuery, JitExprPredicate};
+///
+/// let expression = deserialize("(EQ (ADD price 1u64) 10u64)").unwrap();
+/// let query: DocPredicateQuery = JitExprPredicate::new(expression).unwrap().into();
+/// ```
+#[derive(Clone, Debug)]
+pub struct JitExprPredicate {
+    expression: UntypedExpr,
+    inferred_inputs: Vec<(String, InferredTypeSet)>,
+}
+
+impl JitExprPredicate {
+    /// Creates a predicate after inferring its inputs and requiring a boolean result.
+    pub fn new(expression: UntypedExpr) -> Result<Self, TypeError> {
+        let inferred_types: HashMap<&str, InferredTypeSet> =
+            infer_types_with_target(&expression, InferredTypeSet::BOOLEAN)?;
+        let mut inferred_inputs: Vec<(String, InferredTypeSet)> = inferred_types
+            .into_iter()
+            .map(|(name, types)| (name.to_string(), types))
+            .collect();
+        inferred_inputs.sort_unstable_by(|left, right| left.0.cmp(&right.0));
+        Ok(Self {
+            expression,
+            inferred_inputs,
+        })
+    }
+
+    /// Returns the expression evaluated by this predicate.
+    pub fn expression(&self) -> &UntypedExpr {
+        &self.expression
+    }
+}
+
+impl DocPredicate for JitExprPredicate {
+    type SegmentDocPredicate = JitExprSegmentPredicate;
+
+    fn doc_predicate(
+        &self,
+        segment_reader: &SegmentReader,
+    ) -> crate::Result<JitExprSegmentPredicate> {
+        let mut variable_types = HashMap::with_capacity(self.inferred_inputs.len());
+        let mut opened_columns: HashMap<&str, DynamicColumn> =
+            HashMap::with_capacity(self.inferred_inputs.len());
+
+        // Let's see which column (we can have several column type associated
+        // to the same column name in tantivy!) we will pick for each variable name.
+        for (name, accepted_types) in &self.inferred_inputs {
+            let Some(column) = open_input_column(segment_reader, name, *accepted_types)? else {
+                // If we do not have a valid column for that expression, we do not
+                // fill the HashMap at all.
+                //
+                // The compiler will replace the expression and make it behave like the null
+                // literal.
+                continue;
+            };
+            let Some(var_type) = var_type_for_column_type(column.column_type()) else {
+                continue;
+            };
+            variable_types.insert(name.as_str(), var_type);
+            opened_columns.insert(name.as_str(), column);
+        }
+
+        let Ok(compiled) = compile(&self.expression, &variable_types) else {
+            return Ok(ALWAYS_FALSE_SEGMENT_PREDICATE);
+        };
+        if compiled.result_type() != VarType::Bool {
+            return Ok(ALWAYS_FALSE_SEGMENT_PREDICATE);
+        }
+
+        // The compiler owns the definitive ABI order. Do not rely on inference
+        // or HashMap iteration order when building the argument slots.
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -76,6 +76,9 @@ measure_time = "0.9.0"
 arc-swap = "1.5.0"
 bon = "3.3.1"
 
+# EXPERIMENTAL. The API is likely to change in the near future.
+jitexpr = { version = "0.1", path = "./jitexpr", optional = true }
```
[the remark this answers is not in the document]
trinity-1686a: this isn't enabled by default , it's non-breaking for people not opting in the experimental feature

src/query/doc_predicate_query/jitexpr_predicate.rs
```diff
@@ -0,0 +1,508 @@
...
+use crate::index::SegmentReader;
+use crate::{DocId, TantivyError};
+
+/// A [`DocPredicate`] that evaluates a boolean JIT expression against fast fields.
+///
+/// Requires the `jitexpr` feature. Variable names are resolved as fast-field names
+/// for each segment, supporting boolean, numeric, and string columns. Missing or
+/// incompatible columns are left unbound, so the compiler treats them as `None`.
+/// For bound columns, multivalued documents contribute their first value, and a
+/// document missing any input is skipped before evaluating the expression.
```
trinity-1686a: that seems wrong based on unit tests and the existence of a IS_NULL predicate?

src/query/doc_predicate_query/jitexpr_predicate.rs
```diff
@@ -0,0 +1,508 @@
...
+impl JitExprPredicate {
+    /// Creates a predicate after inferring its inputs and requiring a boolean result.
+    pub fn new(expression: UntypedExpr) -> Result<Self, TypeError> {
+        let inferred_types: HashMap<&str, InferredTypeSet> =
+            infer_types_with_target(&expression, InferredTypeSet::BOOLEAN)?;
+        let mut inferred_inputs: Vec<(String, InferredTypeSet)> = inferred_types
+            .into_iter()
+            .map(|(name, types)| (name.to_string(), types))
+            .collect();
+        inferred_inputs.sort_unstable_by(|left, right| left.0.cmp(&right.0));
```
trinity-1686a: i don't understand why this needs sorting, afaict it's only used to fill hashmaps

src/query/doc_predicate_query/jitexpr_predicate.rs
```diff
@@ -0,0 +1,508 @@
...
+                self.expression
+            )));
+        }
+
+        // The compiler owns the definitive ABI order. Do not rely on inference
+        // or HashMap iteration order when building the argument slots.
+        let mut columns = Vec::with_capacity(compiled_fn.inputs().len());
+        for input in compiled_fn.inputs() {
+            let column = opened_columns
+                .remove(input.variable_name.as_ref())
```
trinity-1686a: if the same expression asks for the same column with different types, the result doesn't match as many things as expected. Note, this likely come from the deeper engine, but i wondered about this based on this HashMap::remove pattern not being able to handle a column with multiple types
```rs
    #[test]
    fn test_multi_typed_field_matches_either_type() {
        let json = schema_builder.add_json_field("json", FAST);
        let mut writer = index.writer_for_tests().unwrap();
        writer
            .add_document(doc!(json => serde_json::json!({"myfield": 2u64})))
            .unwrap();
        writer
            .add_document(doc!(json => serde_json::json!({"myfield": "b"})))
            .unwrap();
        writer
            .add_document(doc!(json => serde_json::json!({"myfield": 3u64})))
            .unwrap();
        writer
            .add_document(doc!(json => serde_json::json!({"myfield": "c"})))
            .unwrap();
        writer.commit().unwrap();
        let searcher = index.reader().unwrap().searcher();
        assert_eq!(
            searcher
                .search(
                    &query(r#"(OR (EQ json.myfield 2u64) (EQ json.myfield "b"))"#),
                    &Count
                )
                .unwrap(),
            2 // assertion fails, expected 2 got 1
        );
    }
```

in fact if querying only for string on that multi-typed column, nothing matches
```rs
    #[test]
    fn test_multi_typed_field_matches_second_type() {
        // omited
        assert_eq!(
            searcher
                .search(&query(r#"(EQ json.myfield "b")"#), &Count)
                .unwrap(),
            1 // assertion fails, expected 1 got 0
        );
    }
```
(that second one might also mean type extraction could be improved?)

src/query/doc_predicate_query/jitexpr_predicate.rs
```diff
@@ -0,0 +1,535 @@
...
+        };
+        state.eval(doc_id)
+    }
+}
+
+impl JitExprEvalState {
+    fn eval(&mut self, doc_id: DocId) -> bool {
+        // Input_values is just a buffer we share to avoid allocations
+        let input_values: &mut Vec<VariableValue<'_>> =
+            unsafe { std::mem::transmute(&mut self.input_values) };
```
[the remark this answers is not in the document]
trinity-1686a: codex is right. You can add something like a `struct ClearOnDrop<T>(&mut Vec<T>);` which implements DerefMut+Drop to have the code look almost identical, while being safe when facing panic/early return

Cargo.toml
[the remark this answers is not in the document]
trinity-1686a: this isn't enabled by default , it's non-breaking for people not opting in to the experimental feature

src/query/doc_predicate_query/jitexpr_predicate.rs
fulmicoton: you are right!

src/query/doc_predicate_query/jitexpr_predicate.rs
fulmicoton: yes that's one of the corner I decide to take. We pick only one column type and try to pick the right one when it makes sense.

We might have to revisit that later.
I can try to put a comment somewhere if it is not here yet.

src/query/doc_predicate_query/jitexpr_predicate.rs
fulmicoton: yes you are right!

I think I fixed it in the last commit.
```
 We want missing columns to behave like all values are Null.
    Before the bugfix we assumed that a missing column would result in a returned value of false,
    as an optimization.

    However through negation, it is perfectly possible to get an expression that returns true
    on a missing column.

    This PR changes things so that
    1- when calculating an expression, we just fill missing variables with null and leave the evaluation
    logic to jitexpr.
    2- when we can trivially spot that an expression is constant (all variables are missing, or there are no variables to begin with)
    then we emit a const predicate which optimizes into an AllScorer or an EmptyScorer.
```

src/query/doc_predicate_query/jitexpr_predicate.rs
fulmicoton: I kept your two unit tests (slightly modified) and left them commented

src/query/doc_predicate_query/jitexpr_predicate.rs
```diff
@@ -0,0 +1,415 @@
...
+    reader: &SegmentReader,
+    name: &str,
+    accepted_types: InferredTypeSet,
+) -> crate::Result<Option<DynamicColumn>> {
+    for handle in reader.fast_fields().dynamic_column_handles(name)? {
+        let Some(var_type) = var_type_for_column_type(handle.column_type()) else {
+            continue;
+        };
+        if accepted_types.contains(var_type) {
+            return Ok(Some(handle.open()?));
```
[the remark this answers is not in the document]
fulmicoton: this is a valid point and a known problem we want to revisit in the future.

src/query/doc_predicate_query/jitexpr_predicate.rs
```diff
@@ -0,0 +1,527 @@
...
+
+impl JitExprPredicate {
+    /// Creates a predicate after inferring its inputs and requiring a boolean result.
+    pub fn new(expression: UntypedExpr) -> Result<Self, TypeError> {
+        let inferred_types: HashMap<&str, InferredTypeSet> =
+            infer_types_with_target(&expression, InferredTypeSet::BOOLEAN)?;
+        let inferred_inputs: Vec<(String, InferredTypeSet)> = inferred_types
+            .into_iter()
+            .map(|(name, types)| (name.to_string(), types))
+            .collect();
```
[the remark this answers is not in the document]
trinity-1686a: two opinions on this:
- if we assume X must be singly typed (we only take the 1st value of "the first column", we don't check multiple columns for multi-valued fields),`open_input_column` collapses x to being all null, so that IS_NULL(ADD ..) matches everything
- if we allow X to be multi-typed, inference should say so, but we don't support that yet

i don't know if we want that to be an error, or handled in some other way. For instance do we want `(OR (EQ (ADD x 1u64) 2u64) (NOT x))` to be legal and match 1 or false (today only one of those because of our limitations), or do we want to reject this? I think rejecting is okay if we say we don't support multi-typed columns in calculated fields at the moment

src/query/doc_predicate_query/jitexpr_predicate.rs
```diff
@@ -0,0 +1,527 @@
...
+    Ok(None)
+}
+
+fn var_type_for_column_type(column_type: ColumnType) -> Option<VarType> {
+    match column_type {
+        ColumnType::Bool => Some(VarType::Bool),
+        ColumnType::I64 => Some(VarType::I64),
+        ColumnType::U64 => Some(VarType::U64),
+        ColumnType::F64 => Some(VarType::F64),
+        ColumnType::Str => Some(VarType::Str),
```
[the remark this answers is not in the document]
trinity-1686a: at best this needs documentation, facet handling can wait for later

src/query/doc_predicate_query/jitexpr_predicate.rs
```diff
@@ -0,0 +1,634 @@
...
+/// A wrapper to make sure the variable value buffer is cleared even if the evaluation
+/// panicked.
+struct ClearOnDrop<'a>(&'a mut Vec<VariableValue<'a>>);
+
+impl<'a> ClearOnDrop<'a> {
+    fn wrap(input_values: &'a mut Vec<VariableValue<'static>>) -> Self {
+        debug_assert!(input_values.is_empty());
+        // Input_values is just a buffer we share to avoid allocations
+        let lower_lifetime_input_values: &mut Vec<VariableValue<'_>> =
+            unsafe { std::mem::transmute(input_values) };
```
trinity-1686a, line 225: debug_assert doesn't protect in production build, ignoring it for safety analysis

safety assessment:
- calling ClearOnDrop::wrap on an non-empty Vec only transmute to reduce lifetime, VariableValue lifetime is covariant so this is fine (the entire Vec is invariant, why we need transmute and make sure to clear before making the 'static things available again)
- the 'static value is not usable while ClearOnDrop due to borrowing. Vec is cleared on Drop, so values with a shorter lifetime than 'static are never directly observable, fine too (that's the goal of this abstraction, making sure no panic happen is too hard to check otherwise)
- mem::forget (or any other way to leak values) can end a borrow without calling Drop, causing effective lifetime transmutation. Defending against that is kinda annoying (would need a design like `fn call_with_reduced_lifetime(&'a mut Vec<VariableValue<'static>>, FnOnce(&'a mut Vec<VariableValue<'a>>) -> T) -> T` to enforce properly), making `wrap` unsafe with the contract that ClearOnDrop mustn't be leaked is probably enough given the only usage trivially doesn't leak ClearOndrop

trinity-1686a (approved):

PR #3088, opened 2026-09-08
fix: disable buffered union seek_danger override

### LINKED ISSUE #3086: BufferedUnionScorer::seek keeps a child that seek_danger left invalid
`BufferedUnionScorer::seek` keeps a child that `seek_danger` left invalid, so a document holding one term of an AND child is returned by the union.

## Describe the bug

**What did you do?**

I ran a `BooleanQuery` with two `Must` clauses on a one-segment index: a `TermQuery` filter with few postings (a "bucket" of documents), and a `DisjunctionMaxQuery` whose branches are per-field ANDs, each a `BooleanQuery` of two `Must` `TermQuery`s (`content: service AND order`, `title: service AND order`). The filter has the lowest cost, so the outer `Intersection` drives with it and calls `seek_danger` on the dis_max's `BufferedUnionScorer`.

**What happened?**

The result contains a document that holds only one of the two required terms of one branch: document 11010 in the program below has `content: "purchase order"` and no `service` anywhere, and neither branch matches it. The union returned it as a match.

Mechanism, as I read `src/query/union/buffered_union.rs` and `src/query/intersection.rs` in 0.26.1:

1. The outer intersection calls `BufferedUnionScorer::seek_danger(target)` with a target outside the union's buffered horizon (`HORIZON` = 4,096), so the union forwards `seek_danger(target)` to each child.
2. The first child is `Intersection<TermScorer, TermScorer>` (`content: service AND order`). Its `seek_danger` seeks `left` (`order`, the cheaper term) to a document past the target that holds `order` alone, returns `SeekLowerBound(that_doc)`, and leaves `right` behind. The child is now in the invalid state the `seek_danger` contract allows, and its `doc()` names a document that is not in the intersection.
3. The second child (`title: service AND order`) returns `Found`, so the union sets `is_hit` and calls `self.seek(target)` to get back to a valid state.
4. In `seek`, outside-horizon branch, each child is re-seeked only `if docset.doc() < target`. The invalid child's `doc()` is past the target, so it is left as it is. `refill` then reads `doc()` of every child and inserts the one-term document into the union's bitset. Once the outer intersection's driver lands on that document (it is in the bucket), the union answers `Found` from its buffer.

The conditions are: the target is outside the union's horizon, a child returns `SeekLowerBound` before another child returns `Found` (the loop breaks at the first hit), the lower-bound document lies inside the new horizon, and the outer driver also matches it.

This came from a real data set: on a 16.35 million document sample, the query `service order` under a filter of 942 values matched 338 documents through this path and 331 without it. Lucene (OpenSearch 3.x) on the same data answers 331. Seven of the 338 documents hold one word only.

A fix that seems sufficient: in `seek`'s outside-horizon branch, seek every child to `target.max(docset.doc())`, so a child left invalid by `seek_danger` is realigned by its own `seek` (for `Intersection`, `left.seek(doc)` is a no-op and `go_to_first_doc` aligns the other terms). Or remember which children returned `SeekLowerBound` in `seek_danger` and re-seek those before `refill` reads their `doc()`. The in-horizon `seek` path is not affected, because consecutive `seek_danger` targets increase, so a union seeked outside its horizon stays outside until a refill.

My workaround: a wrapper `Query`/`Weight`/`Scorer` around the `DisjunctionMaxQuery` that does not override `seek_danger`, so the outer intersection reaches the union through the trait default (`seek`, then compare) and the union is never seeked into the danger zone. The wrapper costs some skipping on the union but returns the correct 331.

**What was expected?**

Only documents that match at least one dis_max branch: `[6000, 11000]` in the program below.

## Which version of tantivy are you using?

`tantivy = "0.26.1"` from crates.io (`Cargo.lock` resolves `tantivy 0.26.1`, `registry+https://github.com/rust-lang/crates.io-index`). Rust 1.98, Linux x86_64. Reproduced in a debug test build and in a release build.

## To Reproduce

Deterministic. One segment written by a single indexing thread, so document ids follow insertion order. The numbers in the comments are the document ids.

`Cargo.toml`:

```toml
[dependencies]
tantivy = "=0.26.1"
```

`src/main.rs`:

```rust
use tantivy::collector::DocSetCollector;
use tantivy::query::{BooleanQuery, DisjunctionMaxQuery, Occur, Query, TermQuery};
use tantivy::schema::{Field, IndexRecordOption, Schema, INDEXED, TEXT};
use tantivy::{doc, DocId, Index, IndexWriter, Term};

    let mut builder = Schema::builder();
    let project = builder.add_u64_field("project", INDEXED);
    let title = builder.add_text_field("title", TEXT);
    let content = builder.add_text_field("content", TEXT);
    let index = Index::create_in_ram(builder.build());
    let mut w: IndexWriter = index.writer_with_num_threads(1, 50_000_000)?;

    // One segment, doc ids in insertion order. "service" is on every filler document, so
    // "order" is the cheaper content term and drives the content AND.
    let filler = |w: &mut IndexWriter, n: usize| -> tantivy::Result<()> {
        for _ in 0..n {
            w.add_document(doc!(project => 999u64, content => "service team meeting"))?;
        }
        Ok(())
    };
    let both = |w: &mut IndexWriter| w.add_document(doc!(project => 999u64, content => "service order"));
    let bucket_title = |w: &mut IndexWriter| {
        w.add_document(doc!(project => 7u64, title => "service order", content => "nothing here"))
    };

    both(&mut w)?; //            0: the content AND's first match
    filler(&mut w, 5_999)?; //   1..=5999
    bucket_title(&mut w)?; //    6000: in the bucket, both words in the title
    filler(&mut w, 4_199)?; //   6001..=10199
    both(&mut w)?; //            10200: a content match past 6000's horizon, before 11000
    filler(&mut w, 799)?; //     10201..=10999
    bucket_title(&mut w)?; //    11000: in the bucket, both words in the title
    filler(&mut w, 9)?; //       11001..=11009
    w.add_document(doc!(project => 7u64, content => "purchase order"))?; // 11010: "order" alone
    filler(&mut w, 90)?; //      11011..=11100
    both(&mut w)?; //            11101: keeps the content AND alive past 11010
    w.commit()?;

    let searcher = index.reader()?.searcher();
    assert_eq!(searcher.segment_readers().len(), 1);

    let and = |field: Field| -> Box<dyn Query> {
        Box::new(BooleanQuery::new(
            ["service", "order"]
                .iter()
                .map(|word| {
                    let term = Term::from_field_text(field, word);
                    let q: Box<dyn Query> = Box::new(TermQuery::new(term, IndexRecordOption::Basic));
                    (Occur::Must, q)
                })
                .collect(),
        ))
    };
    // The faulting child (content) is listed before the child that finds the target (title).
    let dis_max: Box<dyn Query> =
        Box::new(DisjunctionMaxQuery::with_tie_breaker(vec![and(content), and(title)], 0.0));
    let bucket: Box<dyn Query> = Box::new(TermQuery::new(
        Term::from_field_u64(project, 7),
        IndexRecordOption::Basic,
    ));
    let query = BooleanQuery::new(vec![(Occur::Must, bucket), (Occur::Must, dis_max)]);

    let mut docs: Vec<DocId> = searcher
        .search(&query, &DocSetCollector)?
        .into_iter()
        .map(|address| address.doc_id)
        .collect();
    docs.sort_unstable();
    println!("matched: {docs:?}");
    // Expected [6000, 11000]. Document 11010 holds "order" and not "service".
    assert_eq!(docs, vec![6000, 11000]);
    Ok(())
}
```

Output on 0.26.1:

```
matched: [6000, 11000, 11010]
thread 'main' panicked at src/main.rs:
assertion `left == right` failed
  left: [6000, 11000, 11010]
 right: [6000, 11000]
```

Two details of the layout matter. The content match at 10200 keeps the content AND positioned before 11010 when the union is seeked to 11000, and it sits past 6000's horizon so the refill at 6000 does not advance the AND past 11010. The two bucket matches are more than 4,096 documents apart, so the seek to 11000 takes the danger path. Swapping the two dis_max branches hides the fault, because the union breaks out of its loop at the first `Found` before the content AND is asked.

PSeitz:
BufferedUnion is violating the seek_danger contract. 
We need to call seek_danger until we return Found, and are not allowed to call doc before that. `BufferedUnion::seek_danger` is calling `seek`, which is calling `doc` on childs which may be in an invalid state.

I checked and there's no easy fix, so I'll remove this for now. We'll add it back later probably on SimpleUnion and use that in intersections.

closes #3086

### DIFF
--- a/src/query/union/buffered_union.rs
+++ b/src/query/union/buffered_union.rs
@@ -1,6 +1,6 @@
 use common::TinySet;
 
-use crate::docset::{DocSet, SeekDangerResult, COLLECT_BLOCK_BUFFER_LEN, TERMINATED};
+use crate::docset::{DocSet, COLLECT_BLOCK_BUFFER_LEN, TERMINATED};
 use crate::query::score_combiner::{DoNothingCombiner, ScoreCombiner};
 use crate::query::size_hint::estimate_union;
 use crate::query::Scorer;
@@ -131,12 +131,6 @@ impl<TScorer: Scorer, TScoreCombiner: ScoreCombiner> BufferedUnionScorer<TScorer
         }
         false
     }
-
-    fn is_in_horizon(&self, target: DocId) -> bool {
-        // wrapping_sub, because target may be < window_start_doc
-        let gap = target.wrapping_sub(self.window_start_doc);
-        gap < HORIZON
-    }
 }
 
 impl<TScorer, TScoreCombiner> DocSet for BufferedUnionScorer<TScorer, TScoreCombiner>
@@ -251,49 +245,8 @@ where
         }
     }
 
-    fn seek_danger(&mut self, target: DocId) -> SeekDangerResult {
-        if target >= TERMINATED {
-            return SeekDangerResult::SeekLowerBound(TERMINATED);
-        }
-        if self.is_in_horizon(target) {
-            // Our value is within the buffered horizon and the docset may already have been
-            // processed and removed, so we need to use seek, which uses the regular advance.
-            let seek_doc = self.seek(target);
-            if seek_doc == target {
-                return SeekDangerResult::Found;
-            } else {
-                return SeekDangerResult::SeekLowerBound(seek_doc);
-            };
-        }
-
-        // The docsets are not in the buffered range, so we can use seek_into_the_danger_zone
-        // of the underlying docsets
-        let mut is_hit = false;
-        let mut min_new_target = TERMINATED;
-
-        for docset in self.docsets.iter_mut() {
-            match docset.seek_danger(target) {
-                SeekDangerResult::Found => {
-                    is_hit = true;
-                    break;
-                }
-                SeekDangerResult::SeekLowerBound(new_target) => {
-                    min_new_target = min_new_target.min(new_target);
-                }
-            }
-        }
-
-        // The API requires the DocSet to be in a valid state when `seek_into_the_danger_zone`
-        // returns Found.
-        if is_hit {
-            // The doc is found. Let's make sure we position the union on the target
-            // to bring it back to a valid state.
-            self.seek(target);
-            SeekDangerResult::Found
-        } else {
-            SeekDangerResult::SeekLowerBound(min_new_target)
-        }
-    }
+    // Use the default seek_danger: forwarding to children can leave them invalid
+    // when another child matches, making an ordinary seek/refill unsafe (#3086).
 
     #[inline]
     fn doc(&self) -> DocId {
--- a/src/query/disjunction.rs
+++ b/src/query/disjunction.rs
@@ -27,6 +27,11 @@ struct ScorerWrapper<T> {
     current_doc: DocId,
 }
 
+#[cfg(test)]
+pub(super) fn seek_danger_test_wrapper(scorer: Box<dyn Scorer>) -> impl DocSet {
+    ScorerWrapper::new(scorer)
+}
+
 impl<T: Scorer> ScorerWrapper<T> {
     fn new(scorer: T) -> Self {
         let current_doc = scorer.doc();
--- a/src/query/mod.rs
+++ b/src/query/mod.rs
@@ -32,6 +32,8 @@ mod union;
 mod weight;
 
 #[cfg(test)]
+mod seek_danger_tests;
+#[cfg(test)]
 mod vec_docset;
 
 pub(crate) mod score_combiner;
--- a/src/query/seek_danger_tests.rs
+++ b/src/query/seek_danger_tests.rs
@@ -1,0 +1,328 @@
+//! Shared contract tests for query-side `seek_danger` implementations.
+//!
+//! When adding an override, add its query/scorer to `Kind`, `IMPLEMENTATIONS`, and
+//! `Expr::query` (or `scorer`). Test it both alone and nested: a miss may leave a
+//! child invalid even when its parent returns Found.
+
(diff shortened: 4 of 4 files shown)


src/query/union/buffered_union.rs
```diff
@@ -251,49 +245,8 @@ where
...
-        if is_hit {
-            // The doc is found. Let's make sure we position the union on the target
-            // to bring it back to a valid state.
-            self.seek(target);
-            SeekDangerResult::Found
-        } else {
-            SeekDangerResult::SeekLowerBound(min_new_target)
-        }
-    }
+    // Use the default seek_danger: forwarding to children can leave them invalid
```
trinity-1686a, line 248: couldn't we track which child we left in an _dangered_ state, and handle those differently? (or assume we always left them in a _dangered_ state, even in BufferedUnionScorer::seek())

trinity-1686a (approved):

src/query/union/buffered_union.rs
PSeitz-dd: yes, but it may introduce other new subtleties, which I don't like for a bugfix

BufferedUnion is unlikely the right implementation to use anyway here, so I rather remove it. (I have SimpleUnion chosen for most intersections on another branch, which I plan to open a PR for eventually)

trinity-1686a: >yes, but it may introduce other new subtleties, which I don't like for a bugfix

agreed

PR #3089, opened 2026-09-08
CHANGELOG for 0.26.2 bugfixes

PSeitz:


trinity-1686a (approved):
can you rebase before merge, otherwise #3086 won't be in there (assuming this is the exact commit that gets tagged)

PSeitz:
the changes are cherry-picked on the `release_0.26` branch

PR #3095, opened 2026-09-11
RamDirectory reduce memcopy / Remove flush from Directory

PSeitz:
Remove intermediate flushes. 
The core idea is that intermediate flushes in tantivy while writing the index are unnecessary, since partial written index is not readable anyways. This changes the API contract, so the implementation can decide when to flush. We only guarantee written on disk on the final `finish` call on the writer.

## Summary

- Include active `VecWriter` allocations in `RamDirectory::total_mem_usage`, reporting shared usage in 10 KB increments.
- Remove intermediate `flush` behavior from the `Directory` contract. Intermediate writes are an implementation detail; Tantivy only decides when a file is finished.
- Move the completed `Vec<u8>` directly into `RamDirectory` instead of copying it on flush.
- Remove unnecessary per-field flushes from postings, positions, and field norms.
- Rename `TerminatingWrite` to `FinishableWrite` to reflect the finalization semantics.
- Remove the unused `RamDirectory::deep_clone` (it's unused on github).

## Testing

- `cargo test --workspace`

### DIFF
--- a/benches/merge_segments.rs
+++ b/benches/merge_segments.rs
@@ -16,7 +16,7 @@ use rand::rngs::StdRng;
 use rand::SeedableRng;
 use tantivy::directory::error::{DeleteError, OpenReadError, OpenWriteError};
 use tantivy::directory::{
-    AntiCallToken, Directory, FileHandle, OwnedBytes, TerminatingWrite, WatchCallback, WatchHandle,
+    AntiCallToken, Directory, FileHandle, FinishableWrite, OwnedBytes, WatchCallback, WatchHandle,
     WritePtr,
 };
 use tantivy::indexer::{merge_filtered_segments, NoMergePolicy};
@@ -40,8 +40,8 @@ impl Write for NullWriter {
     }
 }
 
-impl TerminatingWrite for NullWriter {
-    fn terminate_ref(&mut self, _token: AntiCallToken) -> io::Result<()> {
+impl FinishableWrite for NullWriter {
+    fn finish_ref(&mut self, _token: AntiCallToken) -> io::Result<()> {
         Ok(())
     }
 }
@@ -63,8 +63,8 @@ impl Write for InMemoryWriter {
     }
 }
 
-impl TerminatingWrite for InMemoryWriter {
-    fn terminate_ref(&mut self, _token: AntiCallToken) -> io::Result<()> {
+impl FinishableWrite for InMemoryWriter {
+    fn finish_ref(&mut self, _token: AntiCallToken) -> io::Result<()> {
         let bytes = OwnedBytes::new(std::mem::take(&mut self.buffer));
         self.blobs.write().unwrap().insert(self.path.clone(), bytes);
         Ok(())
--- a/src/directory/ram_directory.rs
+++ b/src/directory/ram_directory.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::io::{self, BufWriter, Cursor, Write};
 use std::path::{Path, PathBuf};
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, RwLock};
 use std::{fmt, result};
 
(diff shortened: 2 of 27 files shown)


benches/merge_segments.rs
```diff
@@ -63,8 +63,8 @@ impl Write for InMemoryWriter {
     }
 }
 
-impl TerminatingWrite for InMemoryWriter {
-    fn terminate_ref(&mut self, _token: AntiCallToken) -> io::Result<()> {
+impl FinishableWrite for InMemoryWriter {
+    fn finish_ref(&mut self, _token: AntiCallToken) -> io::Result<()> {
```
fulmicoton: The rename is just painful to review and will break backward compat. was it necessary?

src/directory/ram_directory.rs
```diff
@@ -232,18 +245,14 @@ impl Directory for RamDirectory {
...
-            // The writer's allocation is tracked by `RamDirectory`; an additional buffer would not
-            // be included in `total_mem_usage()`.
-            Ok(BufWriter::with_capacity(0, Box::new(vec_writer)))
+        if self.fs.read().unwrap().exists(path) {
+            return Err(OpenWriteError::FileAlreadyExists(path_buf));
         }
+        let vec_writer = VecWriter::new(path_buf, self.clone());
+        // The writer's allocation is tracked by `RamDirectory`; an additional buffer would not
+        // be included in `total_mem_usage()`.
+        Ok(BufWriter::with_capacity(0, Box::new(vec_writer)))
```
fulmicoton: One problem with this, is that if we do a lot of small writes, the BufWriter will not amortize the dynamic call.

Are you sure there is no backward regression in that case?

benches/merge_segments.rs
PSeitz: Not, necessary. I thought it's a better name: terminating, unlike finish, does not imply orderly flush.
I can revert if you prefer terminating.

src/directory/ram_directory.rs
PSeitz: ah good catch, pretty sure we want `BufWriter::new`

PSeitz: changed to BufWriter::new

benches/merge_segments.rs
PSeitz: removed the rename, to make review easier

src/directory/tests.rs
```diff
@@ -144,21 +145,23 @@ fn test_rewrite_forbidden(directory: &dyn Directory) -> crate::Result<()> {
...
     let test_path: &'static Path = Path::new("some_path_for_test");
-    {
-        assert!(directory.open_read(test_path).is_err());
-        let _w = directory.open_write(test_path).unwrap();
-        assert!(directory.exists(test_path).unwrap());
-        assert!(directory.open_read(test_path).is_ok());
-        assert!(directory.delete(test_path).is_ok());
-    }
+    assert!(directory.open_read(test_path).is_err());
+    directory
```
fulmicoton: this is not the same test at all.

before this was testing if a file with a write object opened was seen as existing.

src/directory/ram_directory.rs
```diff
@@ -14,71 +15,132 @@ use crate::directory::{
...
+        Self {
+            shared_usage,
+            reported_bytes: 0,
+            unreported_bytes: 0,
+        }
+    }
+
+    fn add(&mut self, num_bytes: usize) {
+        self.unreported_bytes += num_bytes;
+        if self.unreported_bytes >= MEMORY_USAGE_UPDATE_THRESHOLD {
```
fulmicoton, line 37: ah nice trick, you do that to avoid contention on the Atomic?

src/directory/ram_directory.rs
```diff
@@ -14,71 +15,132 @@ use crate::directory::{
...
+            self.shared_usage
+                .fetch_add(self.unreported_bytes, Ordering::Relaxed);
+            self.reported_bytes += self.unreported_bytes;
+            self.unreported_bytes = 0;
+        }
+    }
+
+    fn finish(&mut self) {
```
fulmicoton: should we call that release ?

src/directory/ram_directory.rs
```diff
@@ -14,71 +15,132 @@ use crate::directory::{
...
 }
 
 impl TerminatingWrite for VecWriter {
     fn terminate_ref(&mut self, _: AntiCallToken) -> io::Result<()> {
-        self.flush()
+        let mut data = std::mem::take(self.data.get_mut());
+        data.shrink_to_fit();
+        let mut fs = self.shared_directory.fs.write().unwrap();
+        fs.active_writers.remove(&self.path);
+        fs.write_owned(self.path.clone(), data);
```
fulmicoton, line 123: when the data is in the Ram directory we don't track its memory at all?

src/directory/ram_directory.rs
PSeitz: yep, probably not necessary here, but better to be on the safe side when it comes to lock contention

src/directory/ram_directory.rs
PSeitz: yes, renamed

src/directory/ram_directory.rs
PSeitz: We track it via `FileSlice`. We have a short double accounting window, when it's moved to `RamDirectory`

src/directory/tests.rs
PSeitz: fixed the test, and there was also the issue that the `exists` implementation did not check active writers.

fulmicoton-dd (approved):

PR #3098, opened 2026-09-13
Add cargo-fuzz harness and ClusterFuzzLite CI

osyniakov:
Addresses the OpenSSF Scorecard [Fuzzing check](https://scorecard.dev/viewer/?uri=github.com/quickwit-oss/tantivy), which scores 0 for tantivy, because the repository has no fuzzing harness. This adds one — **no source changes**, so it can be reviewed on its own.

## Why

Scorecard's implementation ([`checks/raw/fuzzing.go`](https://github.com/ossf/scorecard/blob/main/checks/raw/fuzzing.go); the published docs don't mention Rust, the code does) detects two signals:

- **cargo-fuzz** — any `*.rs` file in the repository containing `libfuzzer_sys`.
- **ClusterFuzzLite** — the presence of `.clusterfuzzlite/Dockerfile`.

Both are satisfied here, and both are backed by targets that actually build and run rather than files that merely satisfy a scanner.

## What's here

- `fuzz/` — a cargo-fuzz crate with six targets. It declares its own `[workspace]` and is not a member of the root workspace, so cargo-fuzz's sanitizer flags cannot leak into the normal build. It depends on `tantivy` with `default-features = false` to keep the fuzz binaries lean.
- `fuzz/seeds/` — a small checked-in seed corpus for the two targets that parse a container format, and the generator that produces it.
- `.clusterfuzzlite/` — the OSS-Fuzz `base-builder-rust` container, plus a build script that discovers targets by globbing `fuzz/fuzz_targets/`, so adding a target needs no CI change. It also packs `fuzz/seeds/<target>/` into `$OUT/<target>_seed_corpus.zip`.
- `.github/workflows/cflite_pr.yml` — PR fuzzing scoped to the code a change touches (`mode: code-change`), 300s, ASan.
- `.github/workflows/cflite_batch.yml` — a longer run of every target daily.
- `.dockerignore` — the build context otherwise includes ~5 GB of `target/`.
- `fuzz/README.md` — how to run, reproduce and minimise, how the seeds work, and how to pick targets.

Actions are pinned by commit SHA, matching the existing workflows.

## Targets

| Target | What it fuzzes |
|---|---|
| `query_grammar` | `parse_query` / `parse_query_lenient` — the raw query-string grammar |
| `query_parser` | `QueryParser` against a real schema: field resolution, typed values, ranges |
| `tokenizer` | the simple, whitespace, raw and ngram tokenizers, and the lowercase, remove-long, ASCII-folding and compound-splitting filters |
| `sstable_dictionary` | `Dictionary::from_bytes` for the SSTable term dictionary format |
| `columnar_reader` | `ColumnarReader::open` for the columnar (fast field) format, then opening every column handle it lists |
| `common_vint` | `VInt` / `VIntU128` decoding, plus a serialize/deserialize round-trip property |

Targets were chosen for APIs whose input is genuinely attacker-controlled — query strings, document text, serialized bytes — and whose contract is to return an error rather than panic. `tokenizer` goes past "does not crash" and asserts the invariants `Token` documents: offsets in bounds, and on UTF-8 character boundaries, since downstream code slices `&text[offset_from..offset_to]`.

`FacetTokenizer` is deliberately excluded (it expects facet-encoded input, not free text) and so are the low-level `read_u32_vint` helpers, which have documented preconditions. Fuzzing those reports contract violations as crashes and drowns out real findings; `fuzz/README.md` says so, so the exclusions are not quietly re-added later.

## Why the byte targets ship a seed corpus

`columnar_reader` and `sstable_dictionary` parse formats that carry a footer, a declared length and offsets. Blind mutation does not synthesise one, so without seeds those targets spend their whole budget being rejected at the first length check.

Measured on `columnar_reader` under ASan, from an empty corpus, for the 300s `cflite_pr.yml` gives a PR: **42.8M executions, 109 edges, no crash** — all of it in the guards at the top of `ColumnarReader::open`, none of it in the parser below. With the seeds, the same target reaches **1222 edges** and finds real bugs in seconds.

The seeds are generated rather than hand-written, by `fuzz/seeds/generator` — a crate with its own `[workspace]` that is not a fuzz target, so `cargo fuzz build` never builds it and CI never pays for it. Regenerating after a format change is one command.

## Verification

The ClusterFuzzLite path was exercised end to end with the real OSS-Fuzz toolchain, not assumed:

```
docker build -t tantivy-cflite -f .clusterfuzzlite/Dockerfile .
docker run --rm -e FUZZING_LANGUAGE=rust -e SANITIZER=address \
  -e FUZZING_ENGINE=libfuzzer -e ARCHITECTURE=x86_64 \
  -v /tmp/out:/out tantivy-cflite compile
```

- The image builds on `gcr.io/oss-fuzz-base/base-builder-rust`.
- `compile` runs `.clusterfuzzlite/build.sh` and exits 0.
- All six binaries and both `<target>_seed_corpus.zip` files are installed into `$OUT`.
- `run_fuzzer` picks the seeds up: `Using seed corpus: columnar_reader_seed_corpus.zip`.
- `cargo fuzz list` shows all six targets; `cargo +nightly fmt --check` is clean for the new crates.

## Heads-up: this finds real bugs today

Running the targets reproduces several pre-existing bugs, so `cflite_pr` will report failures on changes touching those areas until fixes land. The fixes will be applied as a follow-up, see https://github.com/osyniakov/tantivy/pull/4 for details


PSeitz:
Hi, thanks for the PR. I'd suggest OpenSSF add `proptest` to its scorecard, instead of adding additional fuzzing to the repo. 

Also it doesn't make sense to test internal APIs(`tantivy-` prefixed subcrates). 

Two issues on the query parser seem to be real though.

osyniakov:
Hi @PSeitz, thanks for reviewing.

I agree that recognizing proptest would make Scorecard more consistent with its support for property-testing frameworks in other languages.

My motivation for coverage-guided fuzzing was also the additional bug discovery: this run found both the " *" panic and the non-progressing IN [...] parsing loop. Proptest could potentially discover these with suitable strategies, but its standard generator doesn’t use execution coverage to guide subsequent inputs, so I see the approaches as complementary.

I understand the decision to avoid additional infrastructure.

PR #3107, opened 2026-09-16
Added jitexpr compilation cache

fulmicoton:


jitexpr/src/compile/cache.rs
```diff
@@ -0,0 +1,309 @@
...
+    fn slot_opt(
+        &self,
+        untyped_expr: &UntypedExpr,
+        var_types: &HashMap<&str, VarType>,
+    ) -> Option<CompilationSlot> {
+        // That function does take the lock but only does trivial things that
+        // cannot panick before releasing it.
+        let mut inner_guard = self.inner.lock().unwrap();
+        let entries = inner_guard.entries()?;
+        let key = ExprCacheKey::new(untyped_expr, var_types);
```
trinity-1686a: could technically be put outside the critical section (likely changes nothing in practice)

jitexpr/src/compile/cache.rs
```diff
@@ -0,0 +1,309 @@
...
+
+impl ExprCacheKey {
+    fn new(untyped_expr: &UntypedExpr, var_types: &HashMap<&str, VarType>) -> ExprCacheKey {
+        let mut sorted_var_types: Vec<(String, VarType)> = Vec::with_capacity(var_types.len());
+        for (variable_name, var_type) in var_types {
+            sorted_var_types.push((variable_name.to_string(), *var_type));
+        }
+        sorted_var_types.sort_unstable();
+        ExprCacheKey {
+            expr: untyped_expr.to_string(),
```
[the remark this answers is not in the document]
trinity-1686a: searching for a NaNs is asking for troubles, beside NaN ∉ [-∞, +∞] i don't think we should have much guarantees

jitexpr/src/compile/cache.rs
```diff
@@ -0,0 +1,309 @@
...
+        let second = cache
+            .compile(
+                &untyped_expr,
+                &HashMap::from([("right", VarType::U64), ("left", VarType::U64)]),
+            )
+            .unwrap();
+
+        assert!(Arc::ptr_eq(&first, &second));
+        assert_eq!(cache.len(), 1);
+    }
```
trinity-1686a, line 199: this tests is a bit strange, HashMap isn't an ordered map, it doesn't promise iteration in insertion order, comparing calls with `hashmap[left right]` with `hashmap[right left]` is eye-candy
it would make more sense to pass a few var_types and make sure they are in the right order inside the resulting key

trinity-1686a (approved):

jitexpr/src/compile/cache.rs
fulmicoton: you are right. I think I wanted to make sure we did not build the key if the cache is disabled, but the cache being disabled is not really a use case we care about. It would also be possible to do it by adding a enabled flag outside of the mutex.

jitexpr/src/compile/cache.rs
fulmicoton: you are right.

jitexpr/src/compile/cache.rs
trinity-1686a: oh yeah, that tracks

jitexpr/src/ast/literal.rs
```diff
@@ -43,10 +44,11 @@ impl Literal {
                 ..InferredTypeSet::NONE
             },
             Literal::F64(value) => {
+                let value: f64 = value.get();
                 let is_integral: bool = value.fract() == 0.0;
                 InferredTypeSet {
-                    i64: is_integral && *value >= i64::MIN as f64 && *value < -(i64::MIN as f64),
-                    u64: is_integral && *value >= 0.0 && *value < u64::MAX as f64,
+                    i64: is_integral && value >= i64::MIN as f64 && value < -(i64::MIN as f64),
```
trinity-1686a, line 50: i'm confused as to why not <= i64::MAX as f64

jitexpr/src/ast/literal.rs
```diff
@@ -43,10 +44,11 @@ impl Literal {
...
+                    u64: is_integral && value >= 0.0 && value < u64::MAX as f64,
```
trinity-1686a, line 51: doesn't need to be strict LT, LE is enough (by definition a u64 can encode u64::MAX)

jitexpr/src/compile/cache.rs
fulmicoton: using hasher with different seeds could have been working too (went for your solution because the DefaultHasher is hiding too much)

jitexpr/src/compile/typed_expr.rs
```diff
@@ -34,29 +34,27 @@ impl TypedExpr {
...
+                    TypedExprAst::Literal(TypedLiteral::F64(SafeF64::from_integer(value)))
                 }
                 (TypedExprAst::Literal(TypedLiteral::F64(value)), VarType::U64)
-                    if value.is_finite()
-                        && value.fract() == 0.0
-                        && value >= 0.0
-                        && value < u64::MAX as f64 =>
+                    if value.get().fract() == 0.0
+                        && value.get() >= 0.0
+                        && value.get() < u64::MAX as f64 =>
```
trinity-1686a, line 48: LT vs LE

jitexpr/src/compile/typed_expr.rs
```diff
@@ -34,29 +34,27 @@ impl TypedExpr {
...
+                    TypedExprAst::Literal(TypedLiteral::U64(value.get() as u64))
                 }
                 (TypedExprAst::Literal(TypedLiteral::F64(value)), VarType::I64)
-                    if value.is_finite()
-                        && value.fract() == 0.0
-                        && value >= i64::MIN as f64
-                        && value < -(i64::MIN as f64) =>
+                    if value.get().fract() == 0.0
+                        && value.get() >= i64::MIN as f64
+                        && value.get() < -(i64::MIN as f64) =>
```
trinity-1686a, line 55: could use i64::MAX

jitexpr/src/functions/round.rs
```diff
@@ -57,12 +57,11 @@ fn constant_precision(
         Literal::I64(value) => Some(*value),
         Literal::U64(value) => i64::try_from(*value).ok(),
         Literal::F64(value)
-            if value.is_finite()
-                && value.fract() == 0.0
-                && *value >= i64::MIN as f64
-                && *value < -(i64::MIN as f64) =>
+            if value.get().fract() == 0.0
+                && value.get() >= i64::MIN as f64
+                && value.get() < -(i64::MIN as f64) =>
```
trinity-1686a, line 62: given how many times this logic exists, there could be an helper for it

jitexpr/src/types.rs
```diff
@@ -11,6 +15,83 @@ pub enum VarType {
...
+}
+
+// Sound because NaN is excluded, so equality is reflexive.
+impl Eq for SafeF64 {}
+
+impl Ord for SafeF64 {
+    fn cmp(&self, other: &SafeF64) -> Ordering {
+        // `total_cmp` returns `Equal` exactly when the bit patterns match, so
+        // this order is consistent with `PartialEq` above.
+        self.0.total_cmp(&other.0)
```
trinity-1686a, line 77: note that it means we consider -0.0 and +0.0 to be distinct. It's also what we do in other places (range query from +0.0 to +1.0 doesn't return docs matching -0.0 iirc), so not a change of status quo, but worth mentioning as it's surprising

jitexpr/src/ast/literal.rs
trinity-1686a: hum, actually, with f64 being 53b precision that might be wrong

jitexpr/src/ast/literal.rs
fulmicoton: yes

jitexpr/src/compile/typed_expr.rs
fulmicoton: not really. It is tricky but u64::MAX as f64 is numerically greater than u64::MAX.

The test described here is the right one to get the domain where

f64 ---- coerce ---> u64 --- coerce ---->  f64

is injective (i am lying because of neg 0 of course)

PR #3112, opened 2026-09-18
test block value source

fulmicoton:
Changing the way aggregation access their value.

They now get values via a ValueSource abstraction.
The aggregation collector also gets the possibility to
register ValueSourceProvider describing value columns that are computed
on the fly.

Finally, segment aggregation that require a full column
now manipulates a Arc<dyn ColumnValue> directly.


src/aggregation/bucket/histogram/histogram.rs
```diff
@@ -484,18 +483,19 @@ impl<B: BucketIdSlot> SegmentAggregationCollector for SegmentHistogramCollector<
         store.maybe_densify(dense_range);
 
         let req = &self.req_data;
+        let field_type = req.accessor.column_type();
```
PSeitz, line 486: this adds a dynamic dispatch call vs before value on the stack

PSeitz: we don't dispatch over many parents yet, so in a nested aggregation with many buckets, this may be many calls

src/aggregation/bucket/term_agg/flattened_term_histogram.rs
```diff
@@ -223,32 +215,30 @@ impl BucketResolver for ComputedBucketResolver {
...
     boundaries: [u64; NUM_BUCKETS],
     num_buckets: usize,
 }
 
 impl<const NUM_BUCKETS: usize> LinearBucketResolver<NUM_BUCKETS> {
+    /// Returns `None` only when no padding sentinel exists (see below), in which case the caller
+    /// falls back to [`ComputedBucketResolver`].
     fn new(
         hist_req_data: &HistogramAggReqData,
+        hist_values: &Arc<dyn ColumnValues>,
```
PSeitz:
```suggestion
        column_values: &Arc<dyn ColumnValues>,
```
hist_values could be confused with the histogram values

src/aggregation/bucket/term_agg/flattened_term_histogram.rs
```diff
@@ -375,7 +365,8 @@ struct FlattenedTermHistogramCollector<R: BucketResolver, const LANES: usize> {
     /// `bucket_pos` mapped to time-bucket index 0.
     base_pos: i64,
     terms_req_data: TermsAggReqData,
-    /// The (cloned, normalized) histogram request: its column + interval/offset/bounds.
+    /// The full terms full column's values
```
PSeitz:
```suggestion
    /// The terms full column's values
```

PSeitz: one full is enough :)

src/aggregation/value_source/block_accessor.rs
```diff
@@ -37,54 +23,41 @@ pub(crate) struct ColumnBlockAccessor {
     missing_docids_cache: Vec<DocId>,
     /// Scratch buffer available to sources for translating document IDs into value row IDs.
     row_id_cache: Vec<RowId>,
-    /// Cardinality reported by the source that loaded the latest block.
-    /// For the moment this is reporting the cardinality of the full column, not
-    /// something specific to the block.
+    /// Cardinality here is describes the relationship with the loaded doc_id_cache and val_cache.
+    ///
+    /// For physical columns, we typically set the full column cardinality,
```
PSeitz:
```suggestion
    /// For physical columns, we typically set the complete column cardinality,
```
typically is a bit of a weak API contract, can we make it always or clarify?

PR #3114, opened 2026-09-21
Fix Quadratic Complexity Issue in Aggregation Merge

PSeitz:
Up to 50x faster multi terms (1000 segments case, higher diff for more segments).

# Quadratic Complexity

Fixes an quadratic performance issue in `merge_maps` (merge aggregations). Affected aggregations:
`multi_terms`,`composite`, `range`

Before, on every merge we iterated over `left` and checked if the value exists on `right`. `left` accumulated the entries (truncating happens after all segments have been merged). Left may grow after ever merge step, so we iterate more and more data. The fix is to iterate over `right` and add them to left. 

### Multiterms Optimizations
 - Repeated dictionary decoding: Individual string lookups repeatedly decoded the same blocks. Batched, sorted ordinal lookups improved full-query performance 2.3×–13×, depending on the workload.
 - Wasted finalization: We formatted keys and finalized sub-aggregations before discarding excess buckets. Now we apply the final cutoff first and materialize only survivors. 

# Benchmarks
```
Before
1000_segments
nested_terms_status_and_zipf_1000         Memory: 29.0 MB               Avg: 50.5496ms
multi_terms_status_and_zipf_1000          Memory: 36.5 MB               Avg: 211.6050ms
nested_terms_many_and_zipf_1000           Memory: 175.9 MB              Avg: 221.2388ms
multi_terms_many_and_zipf_1000            Memory: 59.5 MB               Avg: 2.51s
nested_terms_many_and_zipf_1000_top500    Memory: 1.7 GB                Avg: 2.13s
multi_terms_many_and_zipf_1000_top500     Memory: 251.9 MB              Avg: 18.70s

After
1000_segments
nested_terms_status_and_zipf_1000         Memory: 29.0 MB               Avg: 50.9583ms (+0.81%)
multi_terms_status_and_zipf_1000          Memory: 27.8 MB (-23.77%)     Avg: 36.2329ms (-82.88%)
nested_terms_many_and_zipf_1000           Memory: 175.9 MB              Avg: 206.8481ms (-6.50%)
multi_terms_many_and_zipf_1000            Memory: 39.2 MB (-34.11%)     Avg: 111.696ms (-95.56%)
nested_terms_many_and_zipf_1000_top500    Memory: 1.7 GB                Avg: 2.03 (-4.82%)
multi_terms_many_and_zipf_1000_top500     Memory: 172.8 MB (-31.39%)    Avg: 292.654ms (-98.44%)
```


src/aggregation/bucket/multi_terms/mod.rs
```diff
@@ -1095,6 +1073,59 @@ where
...
+        .fields
+        .iter()
+        .zip(&req_data.missing_accessors)
+        .enumerate()
+    {
+        let mut ords_and_positions = Vec::new();
+        for (position, entry) in entries.iter().enumerate() {
+            let value = packing.unpack_value(&entry.key, field_idx);
+            if field.column_type == ColumnType::Str
+                && !missing.as_ref().is_some_and(|m| m.missing_value == value)
```
trinity-1686a: imo more readable as
```suggestion
                && missing.as_ref() != Some(m.missing_value)
```
or
```suggestion
                && missing.as_ref().is_none_or(|m| m.missing_value != value)
```

trinity-1686a (approved):

src/aggregation/bucket/multi_terms/mod.rs
PSeitz: actually I prefer `.map(|m| m.missing_value != value).unwrap_or(false)` if that's okay with you

trinity-1686a: why not, i mostly think inverting the result of an is_some_and is particularly hard to think about

